PR #59, opened 2025-02-18
fix(aws-cdk): CLI doesn't work on Node 16

rix0rrr:
During the migration, we updated `cdk-from-cfn` to the latest version. Between version `0.162.1` and `0.189.0` a change to that package made it incompatible with Node 16 (requires Node 18+).

Even though both Node 16 and Node 18 are EOL right now and everyone should be on Node 20+, just to be safe we'll roll back to an old version.

This does mean that we have an old version of the spec used by `cdk migrate`.

---

### DIFF
--- a/packages/@aws-cdk/cli-lib-alpha/THIRD_PARTY_LICENSES
+++ b/packages/@aws-cdk/cli-lib-alpha/THIRD_PARTY_LICENSES
@@ -23884,7 +23884,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 ----------------
 
-** cdk-from-cfn@0.189.0 - https://www.npmjs.com/package/cdk-from-cfn/v/0.189.0 | MIT OR Apache-2.0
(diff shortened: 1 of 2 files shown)


ahammond:
This will unblock a bunch of our failing cdk pipelines.

mrgrain:
> This will unblock a bunch of our failing cdk pipelines.

@ahammond curious to understand why you are still on node 16. Feel free to reach out in DM if you cannot share publicly.

ahammond:
@mrgrain back in the day, we hardcoded an override of the build image to use the new standard:6 image (we were moving to node18, IIRC). Forgot all about it. We added support for using node20 (or whatever is in the .nvmrc) and it worked fine for the Build stage. Doesn’t do anything for the UpdatePipeline stage though. Then node16 stopped working and… ouch. 

So, we are overriding and unsetting the N_PRESERVE_COREPACK to get the modern, working corepack and blasting forward with node22. 

We will probably look at projen pipelines going forward. If we can get that to play nicely in the context of an nx monorepo, we might be able to keep cdk around.

ahammond:
I just looked up the PR where I did this. It was Apr 2023 to get support for node16. So we must have been moving from 14 at the time.

PR #81, opened 2025-02-20
chore: skip jsii packaging on regular build

rix0rrr:
jsii packaging is done as part of the 'release' workflow, so don't run the packaging steps as part of the normal build. This saves ~17% of build time on my machine.

---


mrgrain (changes requested):
I'm not sure about this. There is a class of errors that only occurs when you pack for jsii. Are we really okay only noticing this at release time?

rix0rrr:
I'm not saying it never happens, but I think I can recall this happening... oh... once ever year on some complex deep inheritance thing in the construct library?

Do we want to pay the wait time for every local and PR build for those rare cases?

mrgrain (approved):

PR #82, opened 2025-02-20
refactor(cli): re-organize api and utils

mrgrain:
This is in preparation of splitting out certain parts of the code into a separate library that can than be shared between `aws-cdk` (the CLI) and `toolkit-lib` (the Toolkit lib).

**No functional changes!** Only moving files and updating imports.

Specifically prepares for:

- using `ioHost` in `work-graph`
- using `ioHost` in `stack-events`
- splitting out `utils` into a the shared library

---


.projenrc.ts
```diff
@@ -863,7 +863,7 @@ cli.gitignore.addPatterns('build-info.json');
 const cliPackageJson = `${cli.workspaceDirectory}/package.json`;
 
 cli.preCompileTask.prependExec('./generate.sh');
-cli.preCompileTask.prependExec('ts-node scripts/user-input-gen.ts');
+cli.preCompileTask.prependExec('ts-node --prefer-ts-exts scripts/user-input-gen.ts');
```
mrgrain, line 866: This was need to make the build more manageable. Otherwise the script would always use the js file and you would need to manually delete it.

packages/aws-cdk/test/util/serialize.test.ts
mrgrain, line 1: There were 2 separate test files in different directories for this. I have consolidated into a single file.

iliapolo (approved):

mrgrain:
Overriding merge.

PR #97, opened 2025-02-21
chore: fix codecov workflow

kaizencc:
runs are currently failing because `runs-on` is required: https://github.com/aws/aws-cdk-cli/actions/runs/13461282048

---


projenrc/codecov.ts
```diff
@@ -24,6 +24,7 @@ export class CodeCovWorkflow extends Component {
     });
 
     this.workflow.addJob('collect', {
+      runsOn: ['ubuntu-latest'],
```
kaizencc, line 27: @mrgrain we should make this either required in projen or default to ubuntu-latest right

mrgrain: Huh. That might have changed tbh.

mrgrain, line 27: Use this like everywhere else?
```suggestion
      runsOn: ['aws-cdk_ubuntu-latest_4-core'],
```

kaizencc: sure, but quite a few of our workflows are using ubuntu-latest.

iankhou (approved):

PR #99, opened 2025-02-21
refactor(cli): optimize appsync function hotswap performance

sundersc:
**Optimize AppSync Function Hotswap Performance**

This commit addresses a performance bottleneck in the AppSync function hotswap mechanism by reducing redundant API calls.

*Previous Behavior:*
The hotswap feature was invoking `listFunctions` separately for each AppSync function requiring an update. This approach led to excessive API requests and potential AWS AppSync throttling.

*Improvements:*
- Eliminate repetitive API interactions
- Enhance overall deployment efficiency
- Reduce potential throttling risks

Resolves one of the root cause of https://github.com/aws-amplify/amplify-category-api/issues/3159

---


packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -116,8 +120,10 @@ export async function isHotswappableAppSyncChange(
             delete sdkRequestObject.runtime;
           }
 
-          const functions = await sdk.appsync().listFunctions({ apiId: sdkRequestObject.apiId });
-          const { functionId } = functions.find((fn) => fn.name === physicalName) ?? {};
+          if (!functions.hasOwnProperty(sdkRequestObject.apiId)) {
+            functions[sdkRequestObject.apiId] = await sdk.appsync().listFunctions({ apiId: sdkRequestObject.apiId });
+          }
+          const { functionId } = functions[sdkRequestObject.apiId].find((fn) => fn.name === physicalName) ?? {};
```
iankhou, line 126: Can we include a unit test that validates that:
1. `listFunctions` is called with proper params when `sdkRequestObject.apiId` is not present
2. `listFunctions` is NOT called when `sdkRequestObject.apiId` is present
?

iankhou (changes requested):

sundersc:
I'm closing the PR. Based on more testing, this approach is not going to improve the performance. If there are multiple functions in a stack, all those functions are being hotswapped at the same time, so caching is not helping here.

I'm working with AppSync team to check if it is possible to support server side filtering for `list-functions` based on function name.

PR #102, opened 2025-02-23
chore(cli): show warning when using deprecated version

tttol:
Fixes https://github.com/aws/aws-cdk/issues/298

## What I changed
Show the warning message when Toolkit uses deprecated version of aws-cdk. The message is like below.
```bash
You are using deprecated version(aws-cdk@X.X.X): This version has been deprecated
```

## Additional information
- I'm new to aws-cdk.
- I didn't submit PR to [aws-cdk-cli-testing](https://github.com/aws/aws-cdk-cli-testing) because I think that integration test is not needed in my change. But I am not confident in my judgment. Please let me know if additional integration tests are needed.

---


packages/aws-cdk/lib/cli/util/npm.ts
```diff
@@ -19,3 +19,17 @@ export async function getLatestVersionFromNpm(): Promise<string> {
 
   return latestVersion;
 }
+
+export async function checkIfDeprecated(version: string): Promise<string | null> {
+  try {
+    const { stdout, stderr } = await exec(`npm view aws-cdk@${version} deprecated --silent`, { timeout: 3000 });
```
rix0rrr, line 25: Can you work this into the existing version check in some way? (it may be that you will need to change that existing check to return information about more than 1 version at a time).

I'd like to avoid doing unnecessary calls to NPM.

rix0rrr (changes requested):

packages/aws-cdk/lib/cli/util/npm.ts
rix0rrr: We're doing upgrade notice work here already: https://github.com/aws/aws-cdk-cli/blob/huijbers/install-deps/packages/aws-cdk/lib/cli/version.ts/#L86

tttol: @rix0rrr 
Do you mean that calling `checkIfDeprecated` out of `if (laterVersion) {` block is better?
Ref: https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/cli/version.ts#L125
```diff
    try {
      const laterVersion = await latestVersionIfHigher(currentVersion, versionCheckCache ?? new VersionCheckTTL());
+    const deprecationMessage = await checkIfDeprecated(currentVersion);
+    if (deprecationMessage) {
+      info(`You are using deprecated version(aws-cdk@${currentVersion}): ${deprecationMessage}`);
+    }
      if (laterVersion) {
        const versionMessage = await getVersionMessage(currentVersion, laterVersion);
```

rix0rrr: I mean I don't want 2 calls to `npm view`. I would like a single call to `npm view` to return both bits of information:

- What is the latest version?
- Is the current version deprecated?

It might be that that is not possible, but I would like that to be researched and confirmed first.

mrgrain: It's somewhat possible to get this with a single request:
- https://registry.npmjs.com/aws-cdk/

The downside is that this is a very large response object. Individual versions or latest can be fetched as well:

- https://registry.npmjs.com/aws-cdk/2.1003.0
- https://registry.npmjs.com/aws-cdk/latest

---

It might be prudent to change this feature over to use https calls anyway, to be less reliant on the `npm` binary. If we download the huge file, we definitely need to cache the result. Making two smaller https requests in parallel migh be faster, not sure. Probably still want to have a cache of sort or at least ensure this isn't checked all the time.

tttol: @rix0rrr @mrgrain 
Thank you for your advice. I will try to retrieve the latest version and deprecation message in a single `npm view`call. 
```sh
$ npm view aws-cdk@1.204.0 deprecated name --json        
{
  "deprecated": "AWS CDK v1 has reached End-of-Support on 2023-06-01.\nThis package is no longer being updated, and users should migrate to AWS CDK v2.\n\nFor more information on how to migrate, see https://docs.aws.amazon.com/cdk/v2/guide/migrating-v2.html",
  "name": "aws-cdk"
}
```

rix0rrr: Hm. I don't think this single `npm view` call will be able to retrieve both bits of information. 

I guess the best solution would be then to do 2 parallel calls to `npm view` for both bits of information, and run them in parallel with `await Promise.all()`.

rix0rrr: Heyyy that's what you did 😆

packages/aws-cdk/lib/cli/util/npm.ts
```diff
@@ -1,21 +1,30 @@
...
-export async function getLatestVersionFromNpm(): Promise<string> {
-  const { stdout, stderr } = await exec('npm view aws-cdk version', { timeout: 3000 });
-  if (stderr && stderr.trim().length > 0) {
-    debug(`The 'npm view' command generated an error stream with content [${stderr.trim()}]`);
+export async function execNpmView(currentVersion: string) {
+  const [latestResult, currentResult] = await Promise.all([
+    exec('npm view aws-cdk@latest version', { timeout: 3000 })
+      .catch(err => {
+        throw new ToolkitError(`Failed to fetch latest version info: ${err.message}`);
+      }),
```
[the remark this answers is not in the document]
tttol: >  Is catch with a throw inside the right way to do this?
It’s not the wrong way in TypeScript, but this implementation is a little confusing. I used a single catch to handle all thrown exceptions.

packages/aws-cdk/lib/cli/version.ts
```diff
@@ -84,20 +84,27 @@ export class VersionCheckTTL {
...
   if (!(await cacheFile.hasExpired())) {
-    return null;
+    return [];
   }
 
-  const latestVersion = await getLatestVersionFromNpm();
-  const isNewer = semver.gt(latestVersion, currentVersion);
+  const packageInfo = await execNpmView(currentVersion);
+  const latestVersion = packageInfo.latestVersion;
   await cacheFile.update(latestVersion);
```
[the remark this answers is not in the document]
tttol: I changed the logic to cache the entire `packageInfo`.

packages/aws-cdk/lib/legacy-exports-source.ts
```diff
@@ -21,9 +21,10 @@ export { setSdkTracing as enableTracing } from './api/aws-auth/tracing';
 export { aliases, command, describe } from './commands/docs';
 export { Deployments } from './api/deployments';
 export { cliRootDir as rootDir } from './cli/root-dir';
-export { latestVersionIfHigher, versionNumber } from './cli/version';
+export { getVersionMessages, versionNumber } from './cli/version';
```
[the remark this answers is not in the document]
tttol: I removed this from `legacy-exports-source.ts` and `legacy-exports.ts`

packages/aws-cdk/lib/legacy-exports-source.ts
mrgrain, line 1: please make sure no other changes than the removal are in this file

tttol: I removed unnecessary changes.

rix0rrr (approved):

mrgrain:
@tttol We are good now. Please stop updating the branch as this will reset tests and force a re-run. Will let you know if you need any more changes.

mrgrain:
@tttol The build is failing with a linter error, I think once that's fixed we are good.

tttol:
> The build is failing with a linter error, I think once that's fixed we are good.

I fixed it.

> Please stop updating the branch as this will reset tests and force a re-run. 

Sorry, I made a mistake. I update the branch after resolving the conflict.

PR #105, opened 2025-02-24
chore(toolkit): cleanup public interfaces

mrgrain:
Clean-up public interface and docs generation in preparation of the release.

This mostly moves some previously public interfaces to be private, renames interface and removes unused options.

---


packages/aws-cdk/lib/cli/cdk-toolkit.ts
mrgrain, line 1: Enum values need to be the same in CLI and Toolkit Lib. This option is not publicly exposed as a CLI option, so changing the value is okay

rix0rrr: I would advise against load-bearing enum values to begin with. Just because TypeScript allows you to do it doesn't mean it's a good idea.

rix0rrr (approved):

PR #112, opened 2025-02-24
feat(bootstrap): Add ECR resource policy for emr-serverless containers

graydenshand:
This PR modifies the default bootstrap template to support EMR Serverless containers.

Without this policy, containers created using the `DockerImageAsset` construct cannot be used with EMR Serverless applications.

There is precedence for this pattern to support Lambda function containers, so I don't think this should be too controversial of a change.

Relevant documentation on using custom images in EMR Serverless:
- https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/application-custom-image.html


graydenshand:
@mrgrain @kaizencc @rix0rrr Anyone able to take a look at this?

rix0rrr (changes requested):
Thanks! I think I can approve this, but it needs a bump of the bootstrap template version as well.

rix0rrr (approved):

mrgrain:
@graydenshand You can stop updating from main now. We take care of it from here. Updating will just invalidate the tests and make everything take longer.

PR #118, opened 2025-02-24
chore: issue labeler workflow

kaizencc:
Attempting to copy and projenify this workflow: https://github.com/aws/aws-cdk/blob/main/.github/workflows/issue-label-assign.yml

---


projenrc/issue-labeler.ts
```diff
@@ -0,0 +1,87 @@
...
+  target: 'pull-requests' | 'issues' | 'both';
+  excludedExpressions?: string[];
+  includedExpressions?: string[];
+  includedLabels?: string[];
+  excludedLabels?: string[];
+  defaultArea?: string;
+  parameters?: string;
+  affixes?: string;
+  areaIsKeyword?: boolean;
+  needEnvs?: boolean;
```
iliapolo, line 29: How come one jobs needs it and one doesn't? I see its [configured on the workflow level](https://github.com/aws/aws-cdk/blob/main/.github/workflows/issue-label-assign.yml#L58) in the aws-cdk repo.

projenrc/issue-labeler.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    this.workflow = repo.github.addWorkflow('issue-label-assign');
+    this.workflow.on({
+      pullRequestTarget: { types: ['opened'] },
+      issues: { types: ['opened', 'edited'] },
+    });
+
+    this.workflow.addJob('Triage Issues', triageManagerJob({
+      target: 'issues',
+      excludedExpressions: ['CDK CLI Version', 'TypeScript', 'Java', 'Python', 'Go'],
```
iliapolo, line 72: Are you sure this works? In the aws-cdk repo this is [configured as a string](https://github.com/aws/aws-cdk/blob/main/.github/workflows/issue-label-assign.yml#L35).

projenrc/issue-labeler.ts
```diff
@@ -0,0 +1,87 @@
...
+      includedLabels: ['needs-triage'],
+      excludedLabels: ['p1', 'p2', 'p0', 'effort-small', 'effort-medium', 'effort-large', 'guidance'],
```
iliapolo, line 74: same here

.github/workflows/issue-label-assign.yml
```diff
@@ -0,0 +1,61 @@
...
+    runs-on: aws-cdk_ubuntu-latest_4-core
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Triage Manager
+        uses: aws-github-ops/aws-issue-triage-manager@main
+        with:
+          target: pull-requests
+          areaIsKeyword: true
```
iliapolo, line 59: Casing doesn't bother the action? See https://github.com/aws/aws-cdk/blob/main/.github/workflows/issue-label-assign.yml#L36C11-L36C26

iliapolo (changes requested):
There are a lot of difference between the resulting workflow and the one in the aws-cdk repo. If those are expected, please add a note. If not, please fix :)

projenrc/issue-labeler.ts
kaizencc: its configured on the workflow level, yes. but it doesn't look like the other job uses those env variables, even if it might have access to it.

projenrc/issue-labeler.ts
kaizencc: missed a cnversion step :(

projenrc/issue-labeler.ts
kaizencc: i didn't add the guidance step, which would utilize the workflow level env vars (probs why they put it on the workflow level).

however, projen doesn't currently support workflow-level env variables, so if we were to include the guidance job i'd just add `needEnvs: true` to that step for now (until we have projen support).

.github/workflows/issue-label-assign.yml
kaizencc: it definitely does, i just missed doing a conversion step that i had planned

projenrc/issue-labeler.ts
iliapolo: Followup then - do we not need the guidance step?

.github/workflows/issue-label-assign.yml
```diff
@@ -0,0 +1,49 @@
...
+    name: Triage Manager
+    runs-on: aws-cdk_ubuntu-latest_4-core
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Triage Manager
+        uses: aws-github-ops/aws-issue-triage-manager@main
+        with:
+          github-token: ${{ secrets.PROJEN_GITHUB_TOKEN }}
```
kaizencc, line 45: https://github.com/aws/aws-cdk/pull/19727/files this doesn't tell me why but it seems it was intentional

kaizencc (commented):
i believe the issue-label-assign workflow has the equivalent job to what exists on aws-cdk now

iliapolo (approved):

PR #122, opened 2025-02-24
chore: add git-secrets-scan to repository

kaizencc:
Will also need to be added to aws-cdk-cli-testing i believe. `git-secrets-scan.sh` is copied directly from the old repo; i didn't want to also change it to a typescript file. it runs now before `npx projen build`.

---


.projenrc.ts
```diff
@@ -237,11 +238,23 @@ repoProject.eslint = new pj.javascript.Eslint(repoProject, {
...
+// always scan for git secrets before building
+const gitSecretsScan = repoProject.addTask('git-secrets-scan', {
+  steps: [
+    {
+      exec: '/bin/bash ./projenrc/git-secrets-scan.sh',
+    },
+  ],
+});
+
+repoProject.tasks.tryFind('build')?.prependSpawn(gitSecretsScan);
```
iliapolo, line 256:
```suggestion
repoProject.tasks.tryFind('build')!.prependSpawn(gitSecretsScan);
```

So that we fail in case something here changes unexpectedly. Also, probably best to run this AFTER the projen default task because that step can generate code and files.

kaizencc: now its after default but also after build and eslint, but i think thats fine

rix0rrr (approved):

PR #128, opened 2025-02-25
feat(cli): can match notices against Node version

rix0rrr:
This allows publishing notices that match against the version of Node we're executing on.

Implemented by generalizing the code that matches against CLI version and bootstrap versions: it can now match against arbitrary named components.

Also implement more complex matching rules: currently we match the pattern if one of them matches:

```ts
// Matches if one of A, B or C matches
components: [A, B, C]
```

Instead, generalize to Disjunctive Normal Form and treat the current case as a special case of DNF where every conjunction has one element: 

```ts
// The above gets interpreted as
components: [[A], [B], [C]]

// More complex rules: A and B together, or C
components: [[A, B], [C]]
```

This way we can write rules to say that "component X on Node version Y" should get a notice.

---

### DIFF
--- a/packages/aws-cdk/lib/notices.ts
+++ b/packages/aws-cdk/lib/notices.ts
@@ -421,7 +421,7 @@ export interface Notice {
    * If the `components` is a single-level array, it is evaluated as an OR; it
    * matches if any of the components matches.
    */
-  components: Component[] | Component[][];
+  components: Array<Component | Component[]>;
   schemaVersion: string;
   severity?: string;
 }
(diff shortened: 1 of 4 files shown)


packages/aws-cdk/lib/notices.ts
```diff
@@ -84,32 +84,105 @@ export interface NoticesFilterFilterOptions {
   readonly bootstrappedEnvironments: BootstrappedEnvironment[];
 }
 
-export class NoticesFilter {
+export abstract class NoticesFilter {
```
rix0rrr, line 87: To make it very clear that this class doesn't have any instance methods on purpose.

otaviomacedo (approved):

PR #131, opened 2025-02-25
docs(toolkit-lib): include `VERSION` file

mrgrain:
The docs build requires this file to contain the current version number.

---

### DIFF
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -1204,18 +1204,19 @@ for (const tsconfig of [toolkitLib.tsconfigDev]) {
   }
 }
 
+// Ad a command for the docs
 const toolkitLibDocs = toolkitLib.addTask('docs', {
   exec: 'typedoc lib/index.ts',
   receiveArgs: true,
 });
-toolkitLib.packageTask.spawn(toolkitLibDocs, {
-  // the nested directory is important
-  // the zip file needs to have this structure when created
-  args: ['--out dist/docs/cdk/api/toolkit-lib'],
-});
-toolkitLib.packageTask.exec('zip -r ../docs.zip cdk ', {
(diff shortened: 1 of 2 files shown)


.projenrc.ts
```diff
@@ -1204,18 +1204,19 @@ for (const tsconfig of [toolkitLib.tsconfigDev]) {
...
-});
-toolkitLib.packageTask.exec('zip -r ../docs.zip cdk ', {
-  cwd: 'dist/docs',
-});
+
+// When packaging, output the docs into a specific nested directory
+// This is required because the zip file needs to have this structure when created
+toolkitLib.packageTask.spawn(toolkitLibDocs, { args: ['--out dist/docs/cdk/api/toolkit-lib'] });
+// The docs build needs the version in a specific file at the nested root
+toolkitLib.packageTask.exec('(cat dist/version.txt || echo "latest") > dist/docs/cdk/api/toolkit-lib/VERSION');
```
mrgrain, line 1217: `dist/version.txt` is only available during release. Otherwise just default to the generic `latest`.

iliapolo, line 1217: Not that it really matters but wouldn't it make more sense to default to 0.0.0 - which is the version number during development?

iliapolo: Not that it really matters but wouldn't it make more sense to default to `0.0.0` - which is the version number during development?

iliapolo (approved):

iliapolo:
Removed from queue if you want to address the comment

.projenrc.ts
mrgrain: My thinking was that worst case scenario we will show "latest" on the docs website. That feels better than accidentally showing "0.0.0".

PR #149, opened 2025-02-25
feat(cli): control library init version with a command-line parameter

rix0rrr:
Allows controlling of the `cdk init` version of `aws-cdk-lib`:

```
cdk init --lib-version=1.2.3 
```

Mainly intended to be used by tests (to smoke-test that upcoming library versions still `init` properly), but it can be used by users as well.

---


mrgrain (commented):
Should we just make this a cli feature?

rix0rrr:
> Should we just make this a cli feature

Fair! Yeah, why not

mrgrain (approved):

PR #150, opened 2025-02-25
refactor(cli): stack monitor uses io-host

mrgrain:
Provides stack events to the IoHost in a structured way.

The main change here is that `StackActivityMonitor` does not directly talk to the `ActivityPrinter` anymore. Instead messages are emitted to the `IoHost`. This caused two related changes:
- Progress tracking has been split into a more generic `StackProgressMonitor` and progress information is now provided directly by the `StackActivityMonitor`
- Originally we only had a single timer that covered polling events and printing them. With the split, we now have two separate timers: one for polling events and a separate one for updating the view

Removed low-level options:

- `progress` -> now a `CliIoHost` property
- `ci` -> use value from `CliIoHost` directly instead of passing through all the layers
- `quiet` -> surprisingly this was unused by CLI. We should have something like this again, but can be implemented in a follow up PR.

Manually tested the changes in addition to integ tests.

---


packages/aws-cdk/lib/commands/deploy.ts
mrgrain, line 1: This is now only a CLI concern (not toolkit anymore)

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -282,520 +325,3 @@ export class StackActivityMonitor {
     return path;
   }
 }
-
-function padRight(n: number, x: string): string {
-  return x + ' '.repeat(Math.max(0, n - x.length));
-}
```
mrgrain, line 288: moved to `util/string-manipulation.ts`

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -282,520 +325,3 @@ export class StackActivityMonitor {
...
-
-/**
- * Infamous padLeft()
- */
-function padLeft(n: number, x: string): string {
-  return ' '.repeat(Math.max(0, n - x.length)) + x;
-}
```
mrgrain, line 295: moved to `util/string-manipulation.ts`

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -282,520 +325,3 @@ export class StackActivityMonitor {
...
-  const resources = (template && template.Resources) || {};
-  let maxWidth = 0;
-  for (const id of Object.keys(resources)) {
-    const type = resources[id].Type || '';
-    if (type.length > maxWidth) {
-      maxWidth = type.length;
-    }
-  }
-  return maxWidth;
-}
```
mrgrain, line 307: moved to `util/cloudformation.ts`

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -282,520 +325,3 @@ export class StackActivityMonitor {
...
-
-interface PrinterProps {
```
mrgrain, line 309: Rest of this file moved to various files in `cli/activity-printer` but also got refactored a lot

packages/@aws-cdk/toolkit-lib/lib/api/io/private/codes.ts
```diff
@@ -109,6 +109,24 @@ export const CODES = {
     description: 'Confirm deploy security sensitive changes',
     level: 'info',
   }),
+
+  CDK_TOOLKIT_I5501: codeInfo({
+    code: 'CDK_TOOLKIT_I5501',
+    description: 'Stack Monitoring Start',
+    level: 'debug',
```
rix0rrr, line 116: Does `level` even make sense here? And does it make sense for it to be different from the "event" level?

rix0rrr: Does this imply the starting of a single stack, of the starting of an entire monitoring operation potentially containing multiple stacks?

If this relates to a single stack, don't we need context information to relate the start, stop and update events to each other?

packages/@aws-cdk/toolkit-lib/lib/api/io/private/codes.ts
```diff
@@ -121,6 +139,11 @@ export const CODES = {
     description: 'No stacks found',
     level: 'error',
   }),
+  CDK_TOOLKIT_E5500: codeInfo({
```
rix0rrr, line 142: Shouldn't this have a payload?

packages/@aws-cdk/toolkit-lib/lib/api/io/private/codes.ts
rix0rrr: Don't we need `resourcesTotal` in the payload as well?

packages/@aws-cdk/toolkit-lib/lib/api/io/private/codes.ts
```diff
@@ -109,6 +109,24 @@ export const CODES = {
...
+  CDK_TOOLKIT_I5501: codeInfo({
+    code: 'CDK_TOOLKIT_I5501',
+    description: 'Stack Monitoring Start',
+    level: 'debug',
+  }),
+  CDK_TOOLKIT_I5502: codeInfo({
+    code: 'CDK_TOOLKIT_I5502',
+    description: 'Stack Monitoring Activity event',
+    level: 'info',
+    interface: 'StackActivity',
```
rix0rrr, line 122: Do we have our own copy of this type? I'm not comfortable exposing other people's types as part of our API.

Also if we make our own copy of the type, we'll have the ability to add fields if we feel that's convenient, and we can simplify and omit fields that aren't necessary for our current implementation.

packages/aws-cdk/lib/cli/activity-printer/base.ts
```diff
@@ -0,0 +1,194 @@
...
+      clearTimeout(this.tickTimer);
+    }
+  }
+
+  private scheduleNextTick() {
+    if (!this.active) {
+      return;
+    }
+
+    this.tickTimer = setTimeout(() => void this.tick(), this.updateSleep);
```
rix0rrr, line 170: It seems like we can respond to events immediately as they come in, in `addActivity`. I don't see why we need another timer here, and it leads to compounding latency of 2 polling timers.

packages/aws-cdk/lib/cli/activity-printer/current.ts
```diff
@@ -0,0 +1,170 @@
...
+ * along with their stack trace when the monitoring ends.
+ *
+ * Resources that failed deployment because they have been cancelled are
+ * not included.
+ */
+export class CurrentActivityPrinter extends ActivityPrinterBase {
+  /**
+   * This looks very disorienting sleeping for 5 seconds. Update quicker.
+   */
+  public readonly updateSleep: number = 2_000;
```
rix0rrr, line 25: As discussed in the call, the producer decides the polling interval, and we're not allowing it to  vary based on the printer just yet. This can go away.

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -137,72 +124,100 @@ export class StackActivityMonitor {
...
+
+  constructor({
+    cfn,
+    ioHost,
+    action,
+    stack,
+    stackName,
+    resourcesTotal,
+    changeSetCreationTime,
+    pollingInterval = 1000,
```
rix0rrr, line 140: 2000?

mrgrain: fixed

packages/aws-cdk/lib/cli/activity-printer/current.ts
mrgrain: done

packages/@aws-cdk/toolkit-lib/lib/api/io/private/codes.ts
mrgrain: done

packages/aws-cdk/lib/cli/activity-printer/base.ts
mrgrain: Done!

packages/@aws-cdk/toolkit-lib/lib/api/io/private/codes.ts
mrgrain: > Does level even make sense here? And does it make sense for it to be different from the "event" level?

Currently every message has a level. It's defined as the "recommended printable output for a given verbosity". Acknowledging that different implementors might have a different opinion. But you are right, change it to `info`.

> If this relates to a single stack, don't we need context information to relate the start, stop and update events to each other?

Relates to a single stack. I just forgot documenting the interface. Nice catch!

mrgrain: > Don't we need resourcesTotal in the payload as well?

Sure, why not.

mrgrain: done

PR #151, opened 2025-02-25
refactor(cli): require approval is a part of the CliIoHost

kaizencc:
`CliIoHost` now governs when to ask for approval or not. This was previously passed into deploy via deployOptions. 

- `deployOptions.requireApproval` was deprecated before, and now it is no longer being used (this may have to be pulled into a different change that removes the `requireApproval` option entirely in a minor version bump of toolkit-lib.
- `CliIoHost` has a property, `requireApproval` that can be set. When the `requestResponse` API is called, we check to see if the message code is a known `requireApproval` code and if so, see if the request is necessary or if the default can be used.

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -246,8 +246,6 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
 
     await migrator.tryMigrateResources(stackCollection, options);
 
-    const requireApproval = options.requireApproval ?? RequireApproval.NEVER;
```
kaizencc, line 249: this breaks the behavior of the `requireApproval` option but i don't have a good way or hooking it up to the `requireApproval` governed by the `CliIoHost` -- `toolkit.ts` only has a concept of an `IIoHost`

packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
```diff
@@ -46,20 +46,28 @@ export const confirm = (
...
 }, boolean> => {
-  return prompt(code, `${chalk.cyan(question)} (y/n)?`, defaultResponse, {
-    motivation,
-    concurrency,
-  });
+  return prompt(
+    code,
+    `${chalk.cyan(question)} (y/n)?`,
+    defaultResponse,
+    { motivation, concurrency, approvalLevel: permissionChangeType },
```
mrgrain, line 58: `approvalLevel` needs to be unique just for that message, not part of prompt

packages/@aws-cdk/toolkit-lib/test/actions/deploy.test.ts
mrgrain, line 1: We probably want a test that make sure we are getting the approval prompt and can respond to it

packages/aws-cdk/lib/toolkit/cli-io-host.ts
```diff
@@ -163,6 +164,13 @@ export interface CliIoHostProps {
   readonly isCI?: boolean;
 
   /**
+   * In what scenarios should the CliIoHost ask for approval
+   *
+   * @default RequireApproval.BROADENING
+   */
+  readonly requireApproval?: RequireApproval;
```
mrgrain, line 171: Add `deploy` somewhere in here? Cause it's different than e.g. `destroy`.


```suggestion
  readonly requireDeployApproval?: RequireApproval;
```

packages/aws-cdk/lib/toolkit/cli-io-host.ts
```diff
@@ -454,6 +502,14 @@ export class CliIoHost implements IIoHost {
         throw new ToolkitError(`${motivation}, but concurrency is greater than 1 so we are unable to get a confirmation from the user`);
       }
 
+      // Special approval prompt
+      // Determine if the message needs approval. If it does, continue (it is a basic confirmation prompt)
+      // If it does not, return success (true). We only check messages with codes that we are aware
+      // are requires approval codes.
+      if (this.requiresApproval(msg) && this.skipApprovalStep(msg)) {
```
mrgrain, line 509: nit: would that not just be one helper?

packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
mrgrain: This is blocking.

mrgrain (changes requested):
PR looks good, got a couple of nits.

Blocking is only that we should not add additional datas to a generic helper.

packages/@aws-cdk/toolkit-lib/test/actions/deploy.test.ts
```diff
@@ -53,10 +54,9 @@ describe('deploy', () => {
 
   test('request response when require approval is set', async () => {
```
kaizencc, line 55: this test tests that we are getting the approval prompt... do you also want me to test that we can respond to it? or is that self-explanitory

packages/@aws-cdk/toolkit-lib/lib/toolkit/types.ts
```diff
@@ -107,3 +108,14 @@ export interface ConfirmationRequest {
    */
   readonly concurrency?: number;
 }
+
+/**
+ * Payload for a yes/no confirmation in deploy. Includes information on
+ * what kind of change is being made.
+ */
+export interface DeployConfirmationRequest extends ConfirmationRequest {
```
mrgrain, line 116: Nit: I'd move this into `actions/deploy/index.ts`

packages/@aws-cdk/toolkit-lib/test/actions/deploy.test.ts
```diff
@@ -53,10 +54,9 @@ describe('deploy', () => {
...
     // WHEN
+    ioHost.requireDeployApproval = RequireApproval.ANY_CHANGE;
```
mrgrain, line 57: I don't understand that effect this has on the test?

packages/@aws-cdk/toolkit-lib/test/actions/deploy.test.ts
mrgrain: I think that's enough. Otherwise we're just testing that the `TestIoHost` works.

I'd add a `data` prop to the expected message though.

packages/@aws-cdk/toolkit-lib/test/actions/deploy.test.ts
kaizencc: this is how we set require approval now. i can add a comment in the code

mrgrain (approved):

PR #152, opened 2025-02-25
chore: correct require approval default

kaizencc:
---


packages/@aws-cdk/cloud-assembly-schema/schema/version.json
```diff
@@ -1,4 +1,4 @@
 {
-  "schemaHash": "1063bafc562a50e6dafeb8664275f5d61ff631542d47992963bfaf5cf277ab6e",
-  "revision": 40
+  "schemaHash": "4244f1ed6fcece9abcfb319c637fd2eb863a5deef9cc36f05f7d52377ce60012",
+  "revision": 41
```
iliapolo, line 3: Hmm. Is there a way to avoid this...? @rix0rrr ?

mrgrain: +1 seems unnecessary for a doc change

kaizencc: this was an oversight from a git cherry pick :(

mrgrain: 💩 Maybe we can wait and group it with the text change? We could push directly to main, but I'm not sure about the consequences :(

rix0rrr: You can hax0r around it by copying the new hash but moving the old version number back.

mrgrain: Like this?

```suggestion
  "revision": 40
```

kaizencc: i guess that should work?

packages/@aws-cdk/cloud-assembly-schema/schema/version.json
```diff
@@ -1,4 +1,4 @@
 {
-  "schemaHash": "1063bafc562a50e6dafeb8664275f5d61ff631542d47992963bfaf5cf277ab6e",
+  "schemaHash": "4244f1ed6fcece9abcfb319c637fd2eb863a5deef9cc36f05f7d52377ce60012",
   "revision": 40
 }
```
mrgrain, line 4:
```suggestion
}

```

mrgrain (approved):

PR #163, opened 2025-02-27
fix(cli): notices for ranged alpha modules are not showing up correctly

GavinZZ:
Fixes an issue where notices don't show up when we define a range of pre-releases semver for alpha modules. 

---


rix0rrr (approved):

packages/aws-cdk/test/cloud-assembly-trees/experimental-module-pre-release-semver/tree.json
mrgrain, line 1: Can you please add a test fixture that's more focused on the feature you are testing. Something like the existing ones that have around 100 LOC.

mrgrain (changes requested):

mrgrain (approved):

PR #181, opened 2025-03-03
chore(adc): install dependencies before publishing

rix0rrr:
Otherwise we can't find the AWS SDK dependencies that we need.

---

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -22,10 +22,9 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
-          fetch-depth: 0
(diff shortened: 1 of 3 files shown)


.github/workflows/integ.yml
```diff
@@ -22,9 +22,10 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
+          fetch-depth: 0
       - name: Fetch tags from origin repo
         run: |-
-          git remote add upstream https://github.com/aws/aws-cdk-cli.git
+          git remote add upstream git@github.com:aws/aws-cdk-cli.git
```
mrgrain, line 28: self-mutation will catch this but ;(

mrgrain (approved):

PR #184, opened 2025-03-03
feat: record CLI version in manifest package

rix0rrr:
Record the current CLI version in the manifest package when we release the manifest.

This by itself will not trigger a new release of the `cloud-assembly-schema` package, so the version only gets recorded (and published!) when there is a new version of `cloud-assembly-schema` to publish anyway, and the version will only be displayed back to the user if there is an incompatibility of the major version number.

That means we won't necessarily show the *lowest* possible CLI version number when the user needs to upgrade, but we'll show a valid CLI version number that will definitely support the given manifest.

---


projenrc/insert-task-step.ts
```diff
@@ -0,0 +1,29 @@
+import { Component, Project, TaskStep } from 'projen';
+
+export interface InsertTaskStepProps {
+  readonly taskName: string;
+  readonly insertSteps: TaskStep[];
+  readonly beforeExec: string;
```
iliapolo, line 6: Confusing name...to me this reads like "execute the provided task before this exec". When actually its "this is the exec right after the provided task"?

.projenrc.ts
```diff
@@ -374,11 +375,24 @@ new JsiiBuild(cloudAssemblySchema, {
...
   cloudAssemblySchema.addPackageIgnore('!*.d.ts');
   cloudAssemblySchema.addPackageIgnore('** /scripts');
 })();
 
+new InsertTaskStep(repoProject, {
+  taskName: 'release',
+  insertSteps: [
+    { exec: 'ts-node projenrc/copy-cli-version-to-assembly.task.ts' },
+  ],
+  beforeExec: 'yarn workspaces run build',
```
iliapolo, line 393: I suggest making the necessary changes to `InsertTaskStep` so that invoking it eventually looks like this:

```suggestion
  afterExec: 'yarn workspaces run bump',
```

Since the `bump` tasks is actually what is important here.

.projenrc.ts
```diff
@@ -374,11 +375,24 @@ new JsiiBuild(cloudAssemblySchema, {
 (() => {
   cloudAssemblySchema.preCompileTask.exec('tsx projenrc/update.ts');
 
+  // This file will be generated at release time. It needs to be gitignored or it will
+  // fail projen's "no tamper" check, which means it must also be generated every build time.
+  cloudAssemblySchema.preCompileTask.exec('[[ -f cli-version.json ]] || echo \'{ "version": "" }\' > cli-version.json');
```
iliapolo, line 380: Can we not call the same `copy-cli-version-to-assembly.task.ts` script and have it create the empty file in case the version is 0.0.0? Just trying to minimize bash.

.projenrc.ts
iliapolo: > I had that first but I find a "before" much more natural for an "insert" operation.

hh so call it `AppendTaskStep` :)

Its fine. just a personal preference.

projenrc/insert-task-step.ts
```diff
@@ -0,0 +1,29 @@
...
+    const task = this.project.tasks.tryFind(this.props.taskName);
+    if (!task) {
+      throw new Error(`Did not find task ${this.props.taskName}`);
+    }
+
+    // Find the bump task, and do the CLI version copy straight after
+    const bumpIx = task.steps.findIndex(s => s.exec === this.props.beforeExec);
+    if (bumpIx === -1) {
+      throw new Error(`Did not find step: ${this.props.beforeExec}`);
+    }
```
iliapolo, line 24: Missed some ;)

projenrc/copy-cli-version-to-assembly.task.ts
```diff
@@ -0,0 +1,24 @@
+import { promises as fs } from 'fs';
```
mrgrain, line 1: just FYI, not sure which is better
```suggestion
import * fs from 'fs/promises';
```

mrgrain (approved):

PR #185, opened 2025-03-03
chore: change SSM timestamp name to `cli-npm`

rix0rrr:
We used to publish under the key name `cli`, but the library publishing process is writing timestamps under `npm`, `pypi`, etc.

Having the keys

* `npm`
* `pypi`
* `nuget`
* ...
* `cli`

Was confusing, so we're changing it to `cli-npm` instead.

---

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -143,7 +143,9 @@ jobs:
           - init-typescript-lib
           - tool-integrations
   integ:
-    needs: integ_matrix
+    needs:
+      - prepare
+      - integ_matrix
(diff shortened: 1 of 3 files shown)


.github/workflows/integ.yml
```diff
@@ -143,15 +143,13 @@ jobs:
           - init-typescript-lib
           - tool-integrations
   integ:
-    needs:
```
iliapolo, line 146: Got some locally linked projen stuff? Should be fine with self-mutate I guess.

iliapolo (approved):

PR #188, opened 2025-03-04
fix(cli): write notices to stderr or don't write them at all

### LINKED ISSUE #33589: aws-cdk-notices: CDK Notice being parsed incorrectly causing YAMLSemanticError
### Describe the bug

When running a CDK Deploy, due to the CDK Notice, it seems to be picked up and parsed by YAML? 

This causes a YAMLSemanticError to be produced within one of aws-cdk yaml dependencies. 

### Regression Issue

2.177.0

### Expected Behavior

CDK Deploy should not fail on YAMLSemanticError.

### Current Behavior

Error Log Output after CDK Deploy:
```
/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/resolveSeq-d03cb037.js:1856
                doc.errors.push(new PlainValue.YAMLSemanticError(item, msg));
                                ^
YAMLSemanticError: Implicit map keys need to be followed by map values
    at resolveBlockMapItems (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/resolveSeq-d03cb037.js:1856:33)
    at Object.resolveMap [as resolve] (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/resolveSeq-d03cb037.js:1658:79)
    at resolveByTagName (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/resolveSeq-d03cb037.js:1430:25)
    at resolveTag (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/resolveSeq-d03cb037.js:1460:17)
    at resolveNodeValue (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/resolveSeq-d03cb037.js:1581:23)
    at Object.resolveNode (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/resolveSeq-d03cb037.js:1629:15)
    at parseContents (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/Document-9b4560a1.js:357:30)
    at Document.parse (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/Document-9b4560a1.js:615:5)
    at parseDocument (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/index.js:44:37)
    at Object.parse (/app/node_modules/aws-cdk-lib/node_modules/yaml/dist/index.js:55:15) {
  source: PlainValue {
    error: null,
    range: Range { start: 21443, end: 21531 },
    valueRange: Range { start: 21443, end: 21531 },
    props: [],
    type: 'PLAIN',
    value: null,
    resolved: Scalar {
      value: "\x1B[37mNOTICES         (What's this? https://github.com/aws/aws-cdk/wiki/CLI-Notices)\x1B[39m",
      range: [Array],
      type: 'PLAIN'
    }
  }
}
[06:22:00] failed command: npx ts-node ./src/index.ts
```

### Reproduction Steps

npx aws-cdk deploy

### Possible Solution

Stop CDK notices from being picked up by aws-cdk-lib yaml for parsing. 

Have tried using environment variable: `CDK_NOTICES=false` as suggested, but it doesn't ignore it. 

Ref: https://github.com/aws/aws-cdk/wiki/CLI-Notices#what-if-i-dont-want-notices

### CDK CLI Version

2.1000.3

### Framework Version

_No response_

### Node.js Version

22.14.0

### OS

Amazon Linux 2

### Language

TypeScript

### Language Version

TypeScript (5.3.3)

### Other information

Node 22.14.0 / NPM 10.9.2

rix0rrr:
On CI systems, the CDK CLI tries to avoid writing to `stderr` because there are a couple of CI systems that are commonly configured to fail if any output is written to `stderr`. That means all output, like notices, must go to `stdout`.

Some commands (like `cdk synth` or `cdk bootstrap --show-template`) produce usable output on `stdout`, and these are commonly scripted, like piping their output to a file.

However, because notices must go to `stdout`, these now interfere with the output of these commands.

This needs a more thorough reworking of the CLI output streams, but there is a risk of affecting users who are currently relying on the fact that all output goes to `stdout`.

In this PR, we are doing the first steps to solving this situation:

- Notices will always go to `stderr`, so that they will never interfere with `stdout` anymore.
- We try to detect what CI system we are running on, and we will completely suppress notices *unless* we determine that we are running on a CI system where it is "safe" to write to `sterr` (fail closed).

"Safe" in this case means that the CI system doesn't come with an easy to toggle checkbox that makes commands fail based on what they print, instead of their exit codes. The only systems I'm aware of that have this checkbox are "Azure DevOps", and "TeamCity running PowerShell scripts".

Even though we know the systems that are "unsafe", we will only show notices on systems known to be "safe".

To make the existing `Notices` code play nice with the `IoHost`, I needed to do a couple of things:

* Make a helper class, `IoHostEmitter`, to format messages that are sent to the IoHost. This converts strings or structured messages into an `IoMessage`.
* Thread instances of this helper class to everywhere we emit messages. This required making the static `NoticesFilter` class a proper class with an `ioHostEmitter` member.
* Add a `FakeIoHost` class for tests, to check for the messages we're emitting (as opposed to spying on `stdout` or spying on the `logging` members)

Fixes https://github.com/aws/aws-cdk/issues/33589.

---


packages/aws-cdk/lib/notices.ts
```diff
@@ -36,13 +36,6 @@ export interface NoticesProps {
    */
   readonly output?: string;
 
-  /**
-   * Global CLI option for whether we show notices
-   *
-   * @default true
-   */
-  readonly shouldDisplay?: boolean;
```
rix0rrr, line 44: Messages are now always emitted -- whether they are displayed or not is up to the IoHost.

PR #191, opened 2025-03-04
chore(toolkit-lib): fix code registry is disconnected from usage

mrgrain:
Previously the declared `level` and payload interfaces were only loosely coupled to actual messages being created. This caused a number of issues with incorrectly documented levels and payload interfaces.

Instead of the previous approach, the code registry is now providing a strongly typed "message builder" function specific to every registered code. This function uses generics, so it can correctly ensure at build time that payloads are as expected.

This is the first PR of this series to keep the diff small. Next steps will be to make the code registry available to the CLI package and to remove the old helpers for creating messages.

---


packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
```diff
@@ -29,7 +33,6 @@ export function defaultMessageCode(level: IoMessageLevel, category: IoMessageCod
   const code = `CDK_${category}_${levelIndicator}0000` as IoMessageCode;
   return {
     code,
-    description: `Generic ${level} message for CDK_${category}`,
```
mrgrain, line 32: this wasn't used and only existed to satisfy the interface

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/action-aware.ts
```diff
@@ -28,3 +14,24 @@ export interface ActionAwareIoHost extends IIoHost {
   notify<T>(msg: ActionLessMessage<T>): Promise<void>;
   requestResponse<T, U>(msg: ActionLessRequest<T, U>): Promise<U>;
 }
+
+/**
+ * An IoHost wrapper that adds the given action to an actionless message before
+ * sending the message to the given IoHost
+ */
+export function withAction(ioHost: IIoHost, action: ToolkitAction): ActionAwareIoHost {
```
iliapolo, line 22: Seems like we already have this function here:

https://github.com/aws/aws-cdk-cli/blob/01acccf590c67d10360ac46990958252a0663e44/packages/%40aws-cdk/toolkit-lib/lib/api/io/private/logger.ts#L14

Can we reuse?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/toolkit-action.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * The current action being performed by the CLI. 'none' represents the absence of an action.
+ */
+export type ToolkitAction =
```
iliapolo, line 4: Unrelated to this PR but it just caught my so putting it out here. This list is slightly different than then one in the CLI. Intentional? 

https://github.com/aws/aws-cdk-cli/blob/01acccf590c67d10360ac46990958252a0663e44/packages/aws-cdk/lib/toolkit/cli-io-host.ts#L91-L110

iliapolo (approved):
Dequeued to address minor comments.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/toolkit-action.ts
mrgrain: Yes, the CLI has more. To be unified soon.

PR #197, opened 2025-03-05
fix: `--force` flag does not applies to assets

### LINKED ISSUE #14474: cdk-assets: Remove asset from staging bucket on failed deployment
In #12536, it has been noted that part of the problem is that a corrupted zip file may be uploaded to the staging bucket. At this point, CDK will no-longer attempt to upload the asset, again, because it detects that an asset with the corresponding hash resides within the bucket. After reaching this state, it is necessary to manually remove the affected asset, or assets, from the staging bucket before a successful deployment can occur. In cases where the deployment of a given asset fails, the asset should be removed from the staging bucket to ensure that this "broken" state is not reached.

### Use Case

This change would help ensure that CDK does not attempt to use a corrupt pre-existing asset from the staging bucket during deployment. 

### Alternatives
Provide a CLI flag to ensure that assets are overwritten in the staging bucket on every deployment.

### Other

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

rix0rrr:
The `cdk deploy --force` flag is intended to disable all smartness around saving work. If set, it won't check whether assets already exist in the cloud, and remove the build and publishing steps from the work graph.

However, this by itself is not enough to make sure the asset truly gets published again, because the `publish()` action has its own version of short-circuiting again.

Rather than remove the short-circuiting behavior from `cdk-assets`, we add another `{ force }` flag there as well, which gets its value from the CLI's `--force` flag.

This will make it possible to recover from corrupted assets which were accidentally published, as fixed in https://github.com/aws/aws-cdk/pull/33692, by running `rm -rf cdk.out && cdk deploy --force`.

Fixes https://github.com/aws/aws-cdk/issues/14474.

---


iankhou (approved):

packages/aws-cdk/lib/api/deployments/deployments.ts
```diff
@@ -272,6 +272,13 @@ interface PublishStackAssetsOptions extends AssetOptions {
    * Stack name this asset is for
    */
   readonly stackName?: string;
+
+  /**
+   * Always publish, even if it already exists
+   *
+   * @default False
```
kaizencc, line 279:
```suggestion
   * @default false
```

mrgrain (changes requested):
Needs adding to toolkit-lib as well (as `forcePublishAssets` or something)

mrgrain (approved):

PR #199, opened 2025-03-05
docs(toolkit-lib): add CODE_REGISTRY directly to doc site

kaizencc:
Looks like this on the doc site:

Screenshot 2025-03-05 at 5 03 51 PM

---


kaizencc:
I have no reason besides I think this is the right place to put the code registry. imo it doesn't make sense to have it on the base level so under the `Toolkit`umbrella seems like the most logical place to put it.

mrgrain (approved):

PR #200, opened 2025-03-06
feat(toolkit-lib): wrap errors from assembly builder into AssemblyError

mrgrain:
Fixes errors thrown from a CloudAssembly produced by a builder function are unstructured. Wrapping errors will allow better handling by implementors.

Also introduced the `source` property on `ToolkitError`s as per the RFC, to denote the source of an error (user or toolkit).

---

### DIFF
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/api/toolkit-error.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/api/toolkit-error.ts
@@ -1,7 +1,9 @@
-const TOOLKIT_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.ToolkitError');
-const AUTHENTICATION_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.AuthenticationError');
-const ASSEMBLY_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.AssemblyError');
-const CONTEXT_PROVIDER_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.ContextProviderError');
+import type * as cxapi from '@aws-cdk/cx-api';
+
+const TOOLKIT_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.ToolkitError');
+const AUTHENTICATION_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.AuthenticationError');
+const ASSEMBLY_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.AssemblyError');
+const CONTEXT_PROVIDER_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.ContextProviderError');
 
 /**
  * Represents a general toolkit error in the AWS CDK Toolkit.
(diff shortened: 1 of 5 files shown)


packages/@aws-cdk/tmp-toolkit-helpers/src/api/toolkit-error.ts
```diff
@@ -1,7 +1,9 @@
-const TOOLKIT_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.ToolkitError');
-const AUTHENTICATION_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.AuthenticationError');
-const ASSEMBLY_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.AssemblyError');
-const CONTEXT_PROVIDER_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit.ContextProviderError');
+import type * as cxapi from '@aws-cdk/cx-api';
+
+const TOOLKIT_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.ToolkitError');
+const AUTHENTICATION_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.AuthenticationError');
+const ASSEMBLY_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.AssemblyError');
+const CONTEXT_PROVIDER_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.ContextProviderError');
```
mrgrain, line 6: Missed these in the rename. They are private symbols, so it's fine.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/toolkit-error.ts
```diff
@@ -61,20 +74,61 @@ export class AuthenticationError extends ToolkitError {
...
+ * Represents an error causes by cloud assembly synthesis
+ *
+ * This includes errors thrown during app execution, as well as failing annotations.
  */
 export class AssemblyError extends ToolkitError {
-  constructor(message: string) {
+  /**
+   * An AssemblyError with an original error as cause
+   */
+  public static fromCause(message: string, error: unknown, stacks?: cxapi.CloudFormationStackArtifact[]): AssemblyError {
```
iliapolo, line 85: Does this need to accept `stacks` as well? I don't see a usage of it and it creates some inconsistent naming with `fromStacks`.

packages/@aws-cdk/tmp-toolkit-helpers/test/api/toolkit-error.test.ts
```diff
@@ -3,35 +3,59 @@ import { AssemblyError, AuthenticationError, ContextProviderError, ToolkitError
...
     expect(toolkitError.type).toBe('toolkit');
     expect(authError.type).toBe('authentication');
     expect(assemblyError.type).toBe('assembly');
+    expect(assemblyCauseError.type).toBe('assembly');
     expect(contextProviderError.type).toBe('context-provider');
   });
 
   test('isToolkitError works', () => {
+    expect(toolkitError.source).toBe('toolkit');
+    toolkitError.cause;
```
iliapolo, line 20: Leftover?

packages/aws-cdk/lib/api/cxapp/cloud-assembly.ts
```diff
@@ -336,11 +336,11 @@ export class StackCollection {
     }
 
     if (errors && failAt != 'none') {
-      throw new AssemblyError('Found errors');
+      throw AssemblyError.fromStacks('Found errors', this.stackArtifacts);
```
iliapolo, line 339: Suggestion to rename to `withStacks` and `withCause` instead, but your call.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/toolkit-error.ts
mrgrain: Good catch!

packages/aws-cdk/lib/api/cxapp/cloud-assembly.ts
mrgrain: I had that before! Happy to change.

kaizencc (approved):
youve addressed eli's comments and its late for him so im just going to approve

PR #209, opened 2025-03-07
refactor(cli): use IoHelper in internal APIs

mrgrain:
This PR is a lot. There is no good way of reviewing it, so thanks for taking the time.

The main change here is the usage of the `IoHelper` in all internal APIs. Previously we would pass an `IoHost` and an `action` around as pair, but separate value. Instead we now use the `IoHelper` which knows how to send messages with an action.

I also decided to rename all properties and params respectively. This is partly what makes the PR so large. Note that all changes are to private APIs.

---

### DIFF
--- a/packages/aws-cdk/lib/cli/cli.ts
+++ b/packages/aws-cdk/lib/cli/cli.ts
@@ -5,6 +5,7 @@ import { parseCommandLineArguments } from './parse-command-line-arguments';
 import { checkForPlatformWarnings } from './platform-warnings';
 
 import * as version from './version';
+import { asIoHelper } from '../../../@aws-cdk/tmp-toolkit-helpers/src/api/io/private';
 import { SdkProvider } from '../api/aws-auth';
 import { SdkToCliLogger } from '../api/aws-auth/sdk-logger';
 import { setSdkTracing } from '../api/aws-auth/tracing';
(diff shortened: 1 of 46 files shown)


packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -107,7 +108,7 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
       proxyAddress: argv.proxy,
       caBundlePath: argv['ca-bundle-path'],
     },
-    logger: new SdkToCliLogger(ioHost),
+    logger: new SdkToCliLogger(asIoHelper(ioHost, ioHost.currentAction as any)),
```
mrgrain, line 111: `ioHost.currentAction` is extending `ToolkitAction` with additional values. This is fine.
Same again in L189

packages/aws-cdk/test/api/deployments/deploy-stack.test.ts
```diff
@@ -391,35 +393,6 @@ test('reuse previous parameters if requested', async () => {
   } as CreateChangeSetCommandInput);
 });
 
-describe('ci=true', () => {
```
mrgrain, line 394: deleted since this test was just testing that the CI dependent stream selection works.

kaizencc (approved):
seems like a big refactor, but there are no additional things you're sliding in and if the tests pass, i trust you...

mrgrain:
Added `pr/exempt-integ-test` because both failing integ tests are external issues.

PR #211, opened 2025-03-08
feat(cli): can ignore errors and return dummy value in CloudControl API context provider

go-to-k:
CDK app with CC API Context Provider fails if the resource we want to get doesn't exist.

```
[Error at /LookupStack] Encountered CC API error while getting resource MyLookupTestRole. Error: ResourceNotFoundException: AWS::IAM::Role Handler returned status FAILED: The role with name MyLookupTestRole cannot be found. (Service: Iam, Status Code: 404, Request ID: xxxx-xxx-xxxx-xxxx) (HandlerErrorCode: NotFound, RequestToken: xxxx-xxx-xxxx-xxxx)
```

Also CC API Context Provider (`CcApiContextProviderPlugin`) cannot ignore errors even if `ignoreErrorOnMissingContext` is passed in aws-cdk-lib because the current code in aws-cdk-**cli** doesn't handle the property.

Sample cdk-lib code (based on my [PR](https://github.com/aws/aws-cdk/pull/33603) for IAM Role fromLookup):

```ts
    const response: {[key: string]: any}[] = ContextProvider.getValue(scope, {
      provider: cxschema.ContextProvider.CC_API_PROVIDER,
      props: {
        typeName: 'AWS::IAM::Role',
        exactIdentifier: options.roleName,
        propertiesToReturn: [
          'Arn',
        ],
      } as cxschema.CcApiContextQuery,
      dummyValue: [
        {
          // eslint-disable-next-line @cdklabs/no-literal-partition
          Arn: 'arn:aws:iam::123456789012:role/DUMMY_ARN',
        },
      ],
      ignoreErrorOnMissingContext: options.returnDummyRoleOnMissing, // added
    }).value;
```

However it would be good if the provider can ignore errors and return any dummy value to cdk library. This allows all resources to be **used if available, or created if not.**

Actually, SSM and KMS provider can ignore errors:

KMS: https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/context-providers/keys.ts#L43-L48

SSM: https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/context-providers/ssm-parameters.ts#L27-L30

For example, in cdk-lib, we can specify [ignoreErrorOnMissingContext](https://github.com/aws/aws-cdk/blob/v2.182.0/packages/aws-cdk-lib/aws-kms/lib/key.ts#L741-L744) in the `fromLookup`. The `dummyValue` and `ignoreErrorOnMissingContext` properties can also be specified in [GetContextValueOptions](https://github.com/go-to-k/aws-cdk/blob/45a276fd0fc9b7b08f69f7faf3d0091796ab1663/packages/aws-cdk-lib/core/lib/context-provider.ts#L31-L45).

## cli-integ

https://github.com/aws/aws-cdk-cli-testing/pull/50

---


packages/aws-cdk/lib/context-providers/cc-api-provider.ts
```diff
@@ -49,26 +49,30 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
+        throw new ContextProviderError(`Could not get resource ${args.exactIdentifier}.`);
       }
     } catch (err) {
-      throw new ContextProviderError(`Encountered CC API error while getting resource ${exactIdentifier}. Error: ${err}`);
+      const dummyValue = this.getDummyValueIfErrorIgnored(args);
+      if (dummyValue) {
+        const propsObj = getResultObj(dummyValue, 'dummy-id', args.propertiesToReturn);
+        resultObjs.push(propsObj);
+        return resultObjs;
+      }
```
rix0rrr, line 74: Should this behavior be in a `try/catch` in `getValue()` ? It seems duplicated.

packages/aws-cdk/test/context-providers/cc-api-provider.test.ts
```diff
@@ -240,4 +241,190 @@ test('error by specifying neither exactIdentifier or propertyMatch', async () =>
...
+      exactIdentifier: 'bad-identifier',
+      propertiesToReturn: ['DBInstanceArn', 'StorageEncrypted'],
+      ignoreErrorOnMissingContext: true,
+      dummyValue: [
+        {
+          DBInstanceArn: 'arn:aws:rds:us-east-1:123456789012:db:dummy-instance',
+          StorageEncrypted: 'true',
+        },
+      ],
+    } as TestQueryWithDummy);
```
rix0rrr, line 264: Use `satisfies` instead of `as` everywhere.

packages/aws-cdk/test/context-providers/cc-api-provider.test.ts
```diff
@@ -240,4 +241,190 @@ test('error by specifying neither exactIdentifier or propertyMatch', async () =>
...
+      } as TestQueryWithDummy),
+    ).rejects.toThrow('Encountered CC API error while getting resource bad-identifier.');
+  });
+});
 /* eslint-enable */
+
+interface TestQueryWithDummy extends CcApiContextQuery {
+  ignoreErrorOnMissingContext?: boolean;
+  dummyValue?: any;
+}
```
rix0rrr, line 430: I'm confused. Shouldn't these fields be part of `CcApiContextQuery` itself? 

Why do we define them only for the tests, and access them with introspection in `getDummyValueIfErrorIgnored` ?

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
rix0rrr: Also: let's decide what a good error protocol would be here. I'm not convinced that return a single dummy object (which a caller has to supply a value for, and then has to test for that value... and it needs to be different from the *default* dummy object) is the best protocol.

Also; context provider return values are usually cached. Do we really want "not found" dummy values to be cached? (I.e., if they notice and create the resource after the first synth, we will still behave as if the resource doesn't exist). 

Is that behavior going to match expected user behavior?

rix0rrr (changes requested):

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
go-to-k: > Should this behavior be in a try/catch in getValue() ? It seems duplicated.

Certainly. Decided to use try/catch in findResources so that the following errors are not ignored

```ts
    if (args.exactIdentifier && args.propertyMatch) {
      throw new ContextProviderError(`Specify either exactIdentifier or propertyMatch, but not both. Failed to find resources using CC API for type ${args.typeName}.`);
    }
    if (!args.exactIdentifier && !args.propertyMatch) {
      throw new ContextProviderError(`Neither exactIdentifier nor propertyMatch is specified. Failed to find resources using CC API for type ${args.typeName}.`);
    }
```

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
```diff
@@ -82,22 +86,20 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
-        TypeName: typeName,
+        TypeName: args.typeName,
       });
       result.ResourceDescriptions?.forEach((resource) => {
         const id = resource.Identifier ?? '';
         if (id !== '') {
           const propsObject = JSON.parse(resource.Properties ?? '');
 
-          const filters = Object.entries(propertyMatch);
+          const filters = Object.entries(args.propertyMatch!);
```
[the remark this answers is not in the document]
go-to-k: I tried that pattern.

https://github.com/aws/aws-cdk-cli/pull/211/commits/2090861bb52cf745025b82ffb9618f3a3132f788

However, this was reverted as a result of calling `getDummyValueIfErrorIgnored` at the caller, which allowed the original code to be kept.

packages/aws-cdk/test/context-providers/cc-api-provider.test.ts
go-to-k: Sorry, I had lost track of the existence of the `CcApiContextQuery` interface.
Changed the interface then removed the `TestQueryWithDummy`.

https://github.com/aws/aws-cdk-cli/pull/211/commits/256059ac11a5c55ab0a3438d21fd0df4c99f3091

packages/aws-cdk/test/context-providers/cc-api-provider.test.ts
go-to-k: This was made unnecessary by the following.

https://github.com/aws/aws-cdk-cli/pull/211#discussion_r1990544632

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
go-to-k: > Also: let's decide what a good error protocol would be here. I'm not convinced that return a single dummy object (which a caller has to supply a value for, and then has to test for that value... and it needs to be different from the default dummy object) is the best protocol.

I see. However, throwing an error will cause the CLI to exit, so it must be returned with the normal return type (`{[key: string]: any} []`). The normal return value is also constructed and returned based on the following function.

Would it then be better to return an array of objects such as `[{ Identifier: ‘dummy-id’ }]` or `[{ ErrorIgnored: true }]` (or an empty object...)?

```ts
export function getResultObj(jsonObject: any, identifier: string, propertiesToReturn: string[]): {[key: string]: any} {
  const propsObj = {};
  propertiesToReturn.forEach((propName) => {
    Object.assign(propsObj, { [propName]: findJsonValue(jsonObject, propName) });
  });
  Object.assign(propsObj, { ['Identifier']: identifier });
  return propsObj;
}
```

> Also; context provider return values are usually cached. Do we really want "not found" dummy values to be cached? (I.e., if they notice and create the resource after the first synth, we will still behave as if the resource doesn't exist).

I believe this is a matter on the cdk-lib side. The command for clearing the context can avoid that behaviour, but if we are concerned about the hassle, we could add a process on the cdk-lib side that also reports a missingContext if a dummy value is stored.

https://github.com/aws/aws-cdk/blob/v2.183.0/packages/aws-cdk-lib/core/lib/context-provider.ts#L108-L131

Either way, as cdk-cli, if we don't create a mechanism to ignore the error, the CDK CLI will terminate with an error when the resource is not found.  This means that it will be difficult to fulfil the CDK use case of "use an existing resource if it is available, otherwise create it", so I think this PR should make sense. The existing SSM and KMS providers already support it, so it would be natural to do it in the CC API as well.

go-to-k: Alternatively, I came up with the idea of throwing an error with a message indicating something "ignored" in the CLI and not doing `Annotations.addError` ([here](https://github.com/aws/aws-cdk/blob/v2.183.0/packages/aws-cdk-lib/core/lib/context-provider.ts#L127-L129)) if the message is detected on the cdk-lib side. However, this may result in less clean code as it handles the error message and may be a bit confusing as it goes against the meaning of the property name `ignoreErrorOnMissingContext`...

go-to-k: > Alternatively, I came up with the idea of throwing an error with a message indicating something "ignored" in the CLI and not doing Annotations.addError ([here](https://github.com/aws/aws-cdk/blob/v2.183.0/packages/aws-cdk-lib/core/lib/context-provider.ts#L127-L129)) if the message is detected on the cdk-lib side. However, this may result in less clean code as it handles the error message and may be a bit confusing as it goes against the meaning of the property name ignoreErrorOnMissingContext...

I'm not sure, but if [this code](https://github.com/aws/aws-cdk-cli/blob/cdk%40v2.1003.0/packages/aws-cdk/lib/context-providers/index.ts#L77) is changed, perhaps it could be handled more naturally...

Either way, I am willing to act once I have your opinion.

go-to-k: Sorry for the volume of comments (I've given it a lot of thought).

Either way, I am willing to act once I have your opinion.

rix0rrr: I had sort of missed how this error ignoring protocol, where we record `dummyValue` as the "official response"  had already been implemented in https://github.com/aws/aws-cdk/pull/31415 before, and is being used in 2 other context providers already.

I still can't say I'm happy about it, but there's precedent so we might as well do the same thing as these other context providers until we decide that all of their behaviors should be changed!

I've documented my understanding of these flags here: https://github.com/aws/aws-cdk/pull/33875. If you think this is accurate, then we should just be doing what the documented protocol says.

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
```diff
@@ -124,4 +132,18 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
     return resultObjs;
   }
+
+  private getDummyValueIfErrorIgnored(args: CcApiContextQuery): Record<string, any> | undefined {
+    if (!args.ignoreErrorOnMissingContext) {
+      return undefined;
+    }
+    if (!Array.isArray(args.dummyValue) || args.dummyValue.length === 0) {
+      return undefined;
+    }
```
rix0rrr, line 142: If we're saying that `dummyValue` should be an array to be valid according to the protocol, we should error loudly if this isn't the case.

go-to-k: I could not find what we are saying that the `dummyValue` should be an array but could find the code that handles it as an array.

https://github.com/aws/aws-cdk/pull/33258/files#diff-8b2a963b0382d6266fde393a3491fcc8ae2a38e9bbc4a2d6803b088257dd54ecR155-R164

But in cdk-lib, `dummyValue` is defined as `any`. Should we still make an error in cdk-cli if it is not an array?

packages/@aws-cdk/cloud-assembly-schema/schema/version.json
```diff
@@ -1,4 +1,4 @@
 {
-  "schemaHash": "ba7d47a7a023c39293e99a374af293384eaf1ccd207e515dbdc59dfb5cae4ed6",
-  "revision": 41
+  "schemaHash": "4c069fb88b4e8d69ffb38c74a978955d21d2fb20022266d298f1d9855f45ba78",
+  "revision": 43
```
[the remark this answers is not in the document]
go-to-k: I see. I guess I was running `yarn build` every time I changed JSDoc.
I fixed it by manually reverting to 41 and running `yarn build` only once.

https://github.com/aws/aws-cdk-cli/pull/211/commits/0262d51d508cab4f90e9f46ddb6377104f3844a0

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
```diff
@@ -31,12 +31,20 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
+      } else {
+        // use listResource
+        return await this.listResources(cc, args.typeName, args.propertyMatch!, args.propertiesToReturn);
+      }
+    } catch (err) {
+      const dummyValue = this.getDummyValueIfErrorIgnored(args);
+      if (dummyValue) {
+        return [getResultObj(dummyValue, 'dummy-id', args.propertiesToReturn)];
+      }
+      throw err;
```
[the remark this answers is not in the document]
go-to-k: Okay. Changed to check whether it was not found with `err instanceof ResourceNotFoundException`.

https://github.com/aws/aws-cdk-cli/pull/211/commits/0262d51d508cab4f90e9f46ddb6377104f3844a0

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
go-to-k: > If you think this is accurate, then we should just be doing what the documented protocol says.

I read and think it is accurate. So I have changed to use `ignoreFailedLookup` instead of `ignoreErrorOnMissingContext`.

https://github.com/aws/aws-cdk-cli/pull/211/commits/0262d51d508cab4f90e9f46ddb6377104f3844a0

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
go-to-k: But yes, the format of `dummyValue` should be an array, since it must be in the same format as the normal return value.

It is also unkind not to tell the user when `dummyValue` is improperly formatted, even though `ignoreFailedLookup` is true.

Therefore, as you say, I have made an error in such a case.

https://github.com/aws/aws-cdk-cli/pull/211/commits/0262d51d508cab4f90e9f46ddb6377104f3844a0

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
```diff
@@ -128,22 +136,18 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
-    }
-    const dummyValue = args.dummyValue[0];
-    if (typeof dummyValue !== 'object' || dummyValue === null) {
-      return undefined;
+  private getDummyObject(args: CcApiContextQuery): Record<string, any> | undefined {
+    if (!Array.isArray(args.dummyValue) || args.dummyValue.length === 0 || typeof args.dummyValue[0] !== 'object' || args.dummyValue[0] === null) {
+      throw new ContextProviderError(`dummyValue must be an array with at least one object. Failed to get dummy object for type ${args.typeName}.`);
     }
-    return dummyValue;
+    return getResultObj(args.dummyValue[0], 'dummy-id', args.propertiesToReturn);
```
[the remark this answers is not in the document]
go-to-k: A few more changes to `getDummyObject`.

https://github.com/aws/aws-cdk-cli/pull/211/commits/c3154974a64ac6bb558c59284bfdf49e5fe604bf

packages/@aws-cdk/cloud-assembly-schema/schema/version.json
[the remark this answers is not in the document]
go-to-k: Changed it again: 
https://github.com/aws/aws-cdk-cli/pull/211/commits/407cdb2a8c435ea1d07c85efd22464de78b18021

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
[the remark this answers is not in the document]
go-to-k: A few more changes to `getDummyObject`.

https://github.com/aws/aws-cdk-cli/pull/211/commits/c3154974a64ac6bb558c59284bfdf49e5fe604bf

https://github.com/aws/aws-cdk-cli/pull/211/commits/71b0ec53f031d1104c170148305b952864bd4cad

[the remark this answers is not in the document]
go-to-k: A few more changes to `getDummyObject` (and also changed to `getDummyObjects`).

https://github.com/aws/aws-cdk-cli/pull/211/commits/c3154974a64ac6bb558c59284bfdf49e5fe604bf

https://github.com/aws/aws-cdk-cli/pull/211/commits/8d9d94cd2e29efce845ee53a525ddc6ee72417ce

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
go-to-k: I missed that the property called by cdk-lib is still `IgnoreErrorOnMissingContext`.

https://github.com/aws/aws-cdk/pull/33875/files#diff-0d73bed6d4cd92ebf3292b51185a26f6e85c3e9cf6eb37c28bfc48a457600876R169-R172

So I changed the property name back to `IgnoreErrorOnMissingContext`.

https://github.com/aws/aws-cdk-cli/pull/211/commits/4ea2594cc993342553293e98ea5bb0edf0ae810b

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
```diff
@@ -19,8 +20,18 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
+      return result;
+    } catch (err) {
+      if (err instanceof ResourceNotFoundException) {
+        const dummyObjects = this.getDummyObjects(args);
+        if (dummyObjects) {
+          return dummyObjects;
+        }
+      }
+      throw err;
+    }
```
go-to-k, line 34: I moved the try/catch in `getValue` because I have made it so that it is determined by the type of error instance, so there is no longer a problem doing it with `getValue` instead of `findResources`.

packages/aws-cdk/lib/context-providers/cc-api-provider.ts
```diff
@@ -120,8 +136,21 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
         }
       });
     } catch (err) {
+      if (err instanceof ResourceNotFoundException && ignoreErrorOnMissingContext) {
+        throw err;
+      }
```
go-to-k, line 141: I thought that if the error is `ResourceNotFoundException` but `ignoreFailedLookup` is false, it should return the `ContextProviderError` that wraps this error as before, so I added `&& ignoreFailedLookup` to the condition.

rix0rrr (approved):

rix0rrr:
I've updated the PR a bit to simplify

rix0rrr:
Integ tests passed, try again on flaky build step

PR #215, opened 2025-03-10
refactor(cli): replace getters and setters with a public field

rix0rrr:
The `CliIoHost` had a number of fields that had the following pattern:

```ts
class CliIoHost {
  private _someField: string;

  /** Get someField */
  public get someField() {
    return this._someField;
  }

  /** Set someField */
  public set someField(value: string) {
    this._someField = value;
  }
}
```

There is no additional code in the getters and setters other than directly forwarding every access to a private field, so the above is equivalent to the following:

```ts
class CliIoHost {
  public someField: string;
}
```

Since the above is simpler, this PR proposed to remove the getters and setters and just expose the field mutably.

If we ever want to run additional code in a setter in some of these fields, we can refactor back to a getter and setter pair without impacting source or binary compatibility (forwards compatibility was the argument in the olden C++ and Java days for pre-emptively wrapping fields in getters and setters, but doesn't apply in JS land).

(🤖 This PR was generated using AI)

---

### DIFF
--- a/packages/aws-cdk/lib/toolkit/cli-io-host.ts
+++ b/packages/aws-cdk/lib/toolkit/cli-io-host.ts
@@ -90,13 +90,36 @@ export class CliIoHost implements IIoHost {
    */
   private static _instance: CliIoHost | undefined;
 
-  // internal state for getters/setter
-  private _currentAction: CliAction;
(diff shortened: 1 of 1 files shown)


mrgrain (approved):
thanks!

PR #217, opened 2025-03-10
refactor(cli): remove generic argument from `notify<T>()`

rix0rrr:
The `IIoHost` interface has the following method:

```ts
interface IIoHost {
  notify<T>(msg: IoMessage<T>): Promise<void>;
}
```

The generic parameter `T` is only used once, without bounds, for a singleton argument in input position. This means it is equivalent to the following:

```ts
interface IIoHost {
  notify(msg: IoMessage<unknown>): Promise<void>;
}
```

(Preferring `unknown` over `any` so that implementors are forced to test for the type of the `data` field, just like they would need to do with an argument of type `T`)

In the words of the [Java generics tutorial](https://docs.oracle.com/javase/tutorial/extra/generics/methods.html):

> Generic methods allow type parameters to be used to express
> dependencies among the types of one or more arguments to a method and/or
> its return type. If there isn't such a dependency, a generic method
> should not be used.

Or [similar advice for TypeScript](https://effectivetypescript.com/2020/08/12/generics-golden-rule/):

> Type Parameters Should Appear Twice
>
> Type parameters are for relating the types of multiple values. If a
> type parameter is only used once in the function signature, it's not
> relating anything.

---

### DIFF
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/io-helper.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/io-helper.ts
@@ -13,7 +13,7 @@ export type ActionLessRequest<T, U> = Omit<IoRequest<T, U>, 'action'>;
  * Wraps a client provided IoHost and provides additional features & services to toolkit internal classes.
  */
 export interface IoHelper extends IIoHost {
-  notify(msg: ActionLessMessage<any>): Promise<void>;
+  notify(msg: ActionLessMessage<unkown>): Promise<void>;
   requestResponse<T, U>(msg: ActionLessRequest<T, U>): Promise<U>;
 }
 
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/io-helper.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/io-helper.ts
@@ -13,7 +13,7 @@ export type ActionLessRequest<T, U> = Omit<IoRequest<T, U>, 'action'>;
  * Wraps a client provided IoHost and provides additional features & services to toolkit internal classes.
  */
 export interface IoHelper extends IIoHost {
-  notify<T>(msg: ActionLessMessage<T>): Promise<void>;
+  notify(msg: ActionLessMessage<any>): Promise<void>;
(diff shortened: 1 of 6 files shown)


packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/io-helper.ts
```diff
@@ -13,7 +13,7 @@ export type ActionLessRequest<T, U> = Omit<IoRequest<T, U>, 'action'>;
  * Wraps a client provided IoHost and provides additional features & services to toolkit internal classes.
  */
 export interface IoHelper extends IIoHost {
-  notify<T>(msg: ActionLessMessage<T>): Promise<void>;
+  notify(msg: ActionLessMessage<any>): Promise<void>;
```
mrgrain, line 16: This feels bad. Why are you are going against your own advice from the PR description?
```suggestion
  notify(msg: ActionLessMessage<unkown>): Promise<void>;
```

rix0rrr: Hah. Because of this:

image

Ideally we'd do something like:

```ts
IO.CDK_TOOLKIT_I5501.getData(msg)
```

Or something

rix0rrr: I had thought that TypeScript would treat an unbound `<T>` as unknown and force you to test. But looks like it just treats it like `any` and says "you know what you're doing". It's not like we used to have type safety that we've now lost

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/io-helper.ts
```diff
@@ -13,7 +13,7 @@ export type ActionLessRequest<T, U> = Omit<IoRequest<T, U>, 'action'>;
  * Wraps a client provided IoHost and provides additional features & services to toolkit internal classes.
  */
 export interface IoHelper extends IIoHost {
-  notify<T>(msg: ActionLessMessage<T>): Promise<void>;
+  notify(msg: ActionLessMessage<unkown>): Promise<void>;
```
mrgrain, line 16:
```suggestion
  notify(msg: ActionLessMessage<unknown>): Promise<void>;
```

mrgrain (approved):

PR #219, opened 2025-03-10
feat(toolkit-lib): emit marker messages during actions

### LINKED ISSUE #33286: Emit markers
Emit markers (start/complete) during all actions as trace:

- [x] synth
- [x] deploy
- [x] destroy
- [x] asset publishing
- [x] rollback
- [x] bootstrap 

These events should have unique codes

mrgrain:
Introduces the new concept of "Message Spans" to connect multiple messages for toolkit sub-tasks like "synth" or "build-assets" together. A span consists of a number of messages that have the same `span` value. 

At a minimum, a span is a pair of a `start` and an `end` message. Every end message also reports the elapsed time since the start of the span.

Message spans will allow integrators to more easily track the flow of "tasks" completed in the Toolkit.

## Changes

- Refactors the previous `Timer` class to "Message Spans"
- Introduces a new registry for spans at the bottom of the existing messages registry. A span consists of a start and an end message maker.
- This PR makes use of the new `IoHelper` to provide a nicer API
- There is some more of type magic to ensure correct payloads, I don't think this can be tested further
- I also changed the message registry to be private again, we are not ready to have it as part of the public interface (this doesn't affect our packages)
- Made some improvements to the bundling of declaration files. It's currently really slow because it ends up scanning through A LOT of SDK types. Will look into this more in a follow-up PR.

Fixes https://github.com/aws/aws-cdk/issues/33286

---


packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
mrgrain, line 1: All added messages are traces and won't show up in regular logging configurations (like our cli).

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/payloads/bootstrap-environment-progress.ts
```diff
@@ -1,6 +1,12 @@
 import type * as cxapi from '@aws-cdk/cx-api';
 
 export interface BootstrapEnvironmentProgress {
+  /**
+   * Uniquely identifies this marker amongst concurrent messages
+   *
+   * This is an otherwise meaningless identifier.
+   */
+  readonly marker: string;
```
rix0rrr, line 9: It would be nicer if `marker` was part of the message structure itself (like `data` is or isn't), rather than part of the payload.

packages/@aws-cdk/toolkit-lib/build-tools/bundle.mjs
mrgrain, line 1: Bundling currently takes a while because the declaration bundling started to scan through sdk packages. I plan to fix/work-around/whatever, but in the meantime make this slightly faster by doing all work in parallel.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/stack-selector.ts
mrgrain, line 1: Needs to move to shared because it is used in payloads.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/payloads/deploy.ts
```diff
@@ -55,3 +56,64 @@ export interface NeedRollbackFirstDeployStackResult {
 export interface ReplacementRequiresRollbackStackResult {
   readonly type: 'replacement-requires-rollback';
 }
+
+export interface StackDeployProgress {
+  /**
+   * Uniquely identifies this marker amongst concurrent messages
```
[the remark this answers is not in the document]
mrgrain: Yes, we need to identify the "Marker" or "Span" the message belongs to.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/cloud-assembly/stack-selector.d.ts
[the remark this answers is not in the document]
mrgrain: Yikes, weird.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/io-helper.ts
```diff
@@ -1,39 +1,59 @@
...
+    return this.ioHost.requestResponse({
+      ...msg,
+      action: this.action,
+    });
+  }
+
+  /**
+   * Create a new marker from a given registry entry
+   */
+  public marker<S extends MarkerStart, E extends MarkerEnd>(type: MarkerDefinition<S, E>): Marker<S, E> {
```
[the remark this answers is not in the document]
mrgrain: I kept the `ioHelper.span(...).begin(...)` approach. API looks nicer to me that way.

rix0rrr (approved):

PR #221, opened 2025-03-11
fix(cli): write notices to stderr or don't write them at all

### LINKED ISSUE #33589: aws-cdk-notices: CDK Notice being parsed incorrectly causing YAMLSemanticError

rix0rrr:
(Replaces #188).

On CI systems, the CDK CLI tries to avoid writing to `stderr` because there are a couple of CI systems that are commonly configured to fail if any output is written to `stderr`. That means all output, like notices, must go to `stdout`.

Some commands (like `cdk synth` or `cdk bootstrap --show-template`) produce usable output on `stdout`, and these are commonly scripted, like piping their output to a file.

However, because notices must go to `stdout`, these now interfere with the output of these commands.

This needs a more thorough reworking of the CLI output streams, but there is a risk of affecting users who are currently relying on the fact that all output goes to `stdout`.

In this PR, we are doing the first steps to solving this situation:

- Notices will always go to `stderr`, so that they will never interfere with `stdout` anymore.
- We try to detect what CI system we are running on, and we will completely suppress notices *unless* we determine that we are running on a CI system where it is "safe" to write to `sterr` (fail closed).

"Safe" in this case means that the CI system doesn't come with an easy to toggle checkbox that makes commands fail based on what they print, instead of their exit codes. The only systems I'm aware of that have this checkbox are "Azure DevOps", and "TeamCity running PowerShell scripts".

Even though we know the systems that are "unsafe", we will only show notices on systems known to be "safe".

Fixes https://github.com/aws/aws-cdk/issues/33589.

Also in this PR, because this grew.

* Introduce `IoDefaultMessages` in the CLI package, which helps migrate "legacy" logging code to just emit default warning/info/etc messages to the IoHost.
* Removed the ability to log with a `{ message: 'asdf' }` object to the global logger functions. This wasn't being used anywhere other than tests, and it's sort of pointless: if you know the code you should be using the `MessageMaker` to make a message object; if you don't know the code you can emit a string. There is no need to look up the right code given a level and a message object.
* Make it possible for result types to be any type, not just object types. This is necessary to cover the "result" from legacy logging, where the result is just a string.
* Updated many tests in a test file (`cli-io-host.test.ts`) that failed type checking, but succeeded running, and therefore didn't fail the build of #220.
* Centralized `TestIoHost` into the helper package, and renamed it to `MockIoHost`.
* Introducing a `FakeIoHost` in the CLI package to assert on messages emitted to an `IoHost`. 

---


packages/aws-cdk/lib/logging.ts
```diff
@@ -185,3 +170,60 @@ export const success = (input: LogInput<'I'>, ...args: unknown[]) => {
...
 };
+
+/**
+ * Helper class to emit standard log messages to an IoHost
+ *
+ * It wraps an `IoHelper`, and adds convenience methods for the various log
+ * levels, same as the global ones but scoped to a particular `IoHost` instead
+ * of the global singleton one.
+ */
+export class IoLogger {
```
mrgrain, line 181: So now we have two sets of deprecated logging functions?

mrgrain: I don't like anything that's permanent to be called logging. It gives a false impression to what it's doing. So maybe this?

```suggestion
export class IoMessageEmitter {
```

packages/aws-cdk/lib/notices.ts
```diff
@@ -298,10 +301,10 @@ export class Notices {
 
   private readonly context: Context;
   private readonly output: string;
-  private readonly shouldDisplay: boolean;
   private readonly acknowledgedIssueNumbers: Set<Number>;
   private readonly includeAcknowlegded: boolean;
   private readonly httpOptions: SdkHttpOptions;
+  private readonly ioHostEmitter: IoLogger;
```
mrgrain, line 307: Could we use the IoHelper here?

packages/aws-cdk/test/util/fake-io-host.ts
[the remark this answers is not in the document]
mrgrain: This is on me for starting a new directory without moving the old code.

FWIW I try to avoid `util` in the test directory because there also is a module called `util` and it just confuses people as to whether this directory is used for test utilities or for tests testing the `util` module.

packages/@aws-cdk/toolkit-lib/docs/message-registry.md
```diff
@@ -51,6 +51,10 @@ group: Documents
 | `CDK_TOOLKIT_I9100` | Bootstrap progress | `info` | {@link BootstrapEnvironmentProgress} |
 | `CDK_TOOLKIT_I9900` | Bootstrap results on success | `result` | [cxapi.Environment](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_cx-api.Environment.html) |
 | `CDK_TOOLKIT_E9900` | Bootstrap failed | `error` | {@link ErrorPayload} |
+| `CDK_TOOLKIT_I0100` | Notices decoration (the header or footer of a list of notices) | `info` | n/a |
+| `CDK_TOOLKIT_W0101` | A notice that is marked as a warning | `warn` | n/a |
+| `CDK_TOOLKIT_E0101` | A notice that is marked as an error | `error` | n/a |
+| `CDK_TOOLKIT_I0101` | A notice that is marked as informational | `info` | n/a |
```
mrgrain, line 57: This is probably out-of-scope for this PR, but I wonder if we could give these messages a useful payload like the raw notices object.

packages/aws-cdk/lib/logging.ts
rix0rrr: Ha, 100% what I had first and then I changed it :D

mrgrain (approved):

mrgrain:
@rix0rrr FYI this has a genuine integ test failure

https://github.com/aws/aws-cdk-cli/actions/runs/13791337672/job/38573116246?pr=221#step:13:667

rix0rrr:
> FYI this has a genuine integ test failure

Good point, and it can't be helped, really. 

https://github.com/aws/aws-cdk-cli-testing/pull/57

PR #238, opened 2025-03-17
fix(toolkit-lib): deployment total time includes stray output

mrgrain:
Fixes an extra `Deployment <number>` at the end of the "Total Time" message reporting the total deployment time.

Adds missing test cases for the `MessageSpan` feature (GenAI supported).

---


packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/span.ts
```diff
@@ -172,6 +169,22 @@ export class SpanMaker<S extends object, E extends SpanEnd> {
...
 
+function parseArgs<S extends object>(first: any, second?: S): { message: string | undefined; payload: S } {
+  const firstIsMessage = typeof first === 'string';
+
+  // When the first argument is a string or we have a second argument, then the first arg is the message
+  const message = (firstIsMessage || second) ? first : undefined;
+
+  // When the first argument is a string or we have a second argument,
+  // then the second arg is the payload, otherwise the first arg is the payload
+  const payload = firstIsMessage ? second : (second ?? first);
```
kaizencc, line 180: the comment here doesn't align with the code -- if the code is right, it should say "when the first argument is a string' only. based on the code, it should also say that the additional case is `second ?? first`, not just `first`.

packages/@aws-cdk/tmp-toolkit-helpers/test/api/io/span.test.ts
kaizencc, line 1: meta question for my benefit: what did you ask AI for, and how many iterations did it take for it to get it right?

kaizencc (commented):
small comment

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/span.ts
mrgrain:
```suggestion
  // When the first argument is a string or we have a second argument,
  // then the second arg is the payload, otherwise the first arg is the payload
  const payload = (firstIsMessage || second) ? second : first;
```

kaizencc (approved):

PR #239, opened 2025-03-17
docs: compatibility table for CLI/library versions

rix0rrr:
Describe the compatibility model, and add a table linking the CLI and library versions.

---

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,0 +1,93 @@
+# CDK CLI and AWS Construct Library compatibility table
+
+The table below show what version of the CDK CLI is required for a particular
+version of the AWS Construct Library:
+
+The table below contains the specific versions of the schema, along with the
+CLI that supports them and the Construct Library that uses them.
+
+| Library version | Minimum CLI Version | Schema Version |
+|-----------------|---------------------|----------------|
+| 2.174.0         | 2.174.0             |  39            |
+| 2.182.0         | 2.1000.0            |  40            |
+| (unreleased)    | 2.1003.0            |  41            |
+
+How to use this table: find your library version in the leftmost column.
+If it's not in there, find the highest version lower than your library version.
+The middle column indicates the minimum version of the CDK CLI you need to use.
+
+## How it works
+
+(This section describes how the protocol between App and CLI is versioned and
+is purely for reference; the table above contains the most important information.)
+
+The compatibility between the construct library and the CLI is based on the
+version of the [Cloud Assembly Schema](./packages/@aws-cdk/cloud-assembly-schema)
+used by the construct library and supported by the CLI app. The app writes
+a particular version `X` of the Cloud Assembly, and the CLI can read all versions
+up to and including the version `Y` that was current when it was released. We
+need `X <= Y`.
+
+```
+┌────────┐          ┌─────────┐           ┌──────────┐
+│        │ (writes) │         │  (reads)  │          │
+│  App   │─────────▶│ cdk.out │◀──────────│   CLI    │
+│        │          │         │           │          │
+└────────┘  schema  └─────────┘  schemas  └──────────┘
+             vX                  <= vY
+```
+
+### Compatibility rule for CLI < 2.1000.0
+
+Before CDK CLI `2.1000.0`, the CLI and the library were released in lockstep
+with the same version number, and so the versions `X` and `Y` above are the
+same for a given version number.
+
+An easy compatibility rule therefore is: the CLI supports apps written against
+the AWS Construct Library with the same version or lower. It might support more
+versions as well, but it *at least* supports those.
+
+This works because `v(App) <= v(CLI)` guarantees that `X <= Y`.
+
+### Compatibility rule for CLI >= 2.1000.0
+
+Since CDK CLI `2.1000.0`, the CLI and construct library have two different
+versions. The only rule that relates `X` and `Y` involves the *release date* of
+the respective packages. The CLI version that supports a particular schema version
+is always released before a CDK app starts emitting cloud assemblies using
+that schema.
+
+A general rule therefore is: a particular library version's Cloud Assemblies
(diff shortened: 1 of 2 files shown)


COMPATIBILITY.md
```diff
@@ -0,0 +1,45 @@
...
+| Schema Version | CLI Version | Library version |
+|----------------|-------------|-----------------|
+|  39            | 2.174.0     | 2.174.0         |
+|  40            | 2.1000.0    | 2.182.0         |
+|  41            | 2.1003.0    | (not used yet)  |
+
+How to use the table above: to find out what version of the CDK CLI you need
+to deploy your CDK App written against a particular library version, find
+your library version in the right-hand column and then see the minimum version
+of the CDK CLI you need in the middle column.
```
iliapolo, line 45:
```suggestion
| Schema Version | CLI Version | Library version |
|----------------|-------------|-----------------|
|  39            | 2.174.0     | >=2.174.0 <2.182.0         |
|  40            | 2.1000.0    | >=2.182.0         |
|  41            | 2.1003.0    | (not used yet)  |

How to use the table above: to find out what version of the CDK CLI you need, locate 
your library version range in the right-hand column, the CDK CLI version corresponding to it, is the minimum version you should use.
```

How do we keep this up to date?

rix0rrr: > How do we keep this up to date?

Currently manually. Of course we should try and automate its generation. I don't have good ideas for doing that just yet and scripting it will be a significant bit of work. I don't want to gate publishing this information on having an answer to that question.

As for the table, if you want to be even more technically correct we should be adding range operators to the middle column as well:

```md
| Schema Version | CLI Version | Library version |
|----------------|-------------|-----------------|
|  39            | >= 2.174.0     | >=2.174.0 <2.182.0         |
|  40            | >= 2.1000.0    | >=2.182.0         |
|  41            | >= 2.1003.0    | (not used yet)  |
```

But this adds to the visual noise. I think people will be able to infer those bounds pretty naturally, don't you think?

rix0rrr: > How do we keep this up to date?

Currently manually. Of course we should try and automate its generation. I don't have good ideas for doing that just yet and scripting it will be a significant bit of work. I don't want to gate publishing this information on having an answer to that question.

As for the table, if you want to be even more technically correct we should be adding range operators to the middle column as well:

```md
| Schema Version | CLI Version | Library version |
|----------------|-------------|-----------------|
|  39            | >= 2.174.0     | >=2.174.0 <2.182.0         |
|  40            | >= 2.1000.0    | >=2.182.0         |
|  41            | >= 2.1003.0    | (not used yet)  |
```

But this adds to the visual noise. I think people will be able to infer those bounds pretty naturally, don't you think?

iliapolo: > But this adds to the visual noise. I think people will be able to infer those bounds pretty naturally, don't you think? I don't think we need them.

iliapolo: > But this adds to the visual noise. I think people will be able to infer those bounds pretty naturally, don't you think? I don't think we need them.

It does add _some_ clutter, but its also familiar noise...I think it will help. Maybe just add a "all versions below should be read as minimum version numbers", or something?

COMPATIBILITY.md
```diff
@@ -0,0 +1,45 @@
...
+```
+                             this CLI version...
+                                    ⋎
+                                    o
+CLI           │     │     │         │       │      │     │      │
+        ─────┬┴─────┴┬────┴──┬────┬─┴────┬──┴──────┴┬────┴───┬──┴─────→
+Library      │       │       │    │      │          │        │
+             ⋏       ⋏       ⋏    ⋏      ⋏
+   ...supports at least these library versions
+```
```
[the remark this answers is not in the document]
rix0rrr: > It gets users thinking about **when** versions are released and that can be tricky

Can you tell me what is tricky? It's objective information that can be easily (albeit perhaps a bit laboriously) obtained.

If the trickiness is: 

> I can't tell from the version number alone what the compatibility is and now I need to look up information elsewhere.

Then yes, but that is the nature of the beast. 

What I wanted to do here is to state the general rule people can use to determine compatibility. In addition to this rule, they can consult the table. 

It seems like you are saying "don't state a general rule at all, just give the table and tell people to always look it up in the Markdown table, rather than either in the Markdown table or on npmjs".

I think that is independent of a diagram that shows the Cloud Assembly schema, which we could do but I was also trying to keep the schema out of it as much as possible (it's an implementation detail after all) and just focus on the 2 main components.

COMPATIBILITY.md
iliapolo: > I don't want to gate publishing this information on having an answer to that question.

Thats fine. Would be good to at least reference this table in a place developers stumble upon when making schema changes, but I can' think of a good place because the version bump is automatic. If you can think of anything useful great, if not, oh well..

COMPATIBILITY.md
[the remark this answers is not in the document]
iliapolo: > Can you tell me what is tricky?

My concern is that this general rule will get people to try and lookup **when** their CLI version was released, **when** their Lib version was released, and compare. I would like to avoid that because its a lot of steps. Given they cannot infer this information from version numbers alone, I think this will happen often. 

So if we are sending some anywhere, I would prefer it to be the table below only, instead of looking up release dates on github or npm.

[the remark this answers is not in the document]
iliapolo: > (it's an implementation detail after all)

True, but I think at this point its has already become more than tha (maybe because it changes more often than we thought?) I think folks would benefit from understanding which component uses it and when. 

I agree its not necessarily related, but I think it may be more helpful.

iliapolo (approved):

iliapolo:
Dequeued to address comments at will.

iliapolo (approved):

PR #243, opened 2025-03-17
refactor(toolkit): resource metadata

mrgrain:
Pulls out resource-metadata extraction into a separate API so it can easily be reused in other places.
Also adds support for resource metadata to `EvaluateCloudFormationTemplate` (internal API) in preparation for a change in hotswap.

---

### DIFF
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/messages.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/messages.ts
@@ -21,7 +21,7 @@ import type { FileWatchEvent, WatchSettings } from '../payloads/watch';
  * - X900-X999 are reserved for results
  */
 export const IO = {
-  // Defaults
+  // Defaults (0000)
   DEFAULT_TOOLKIT_INFO: make.info({
     code: 'CDK_TOOLKIT_I0000',
     description: 'Default info messages emitted from the Toolkit',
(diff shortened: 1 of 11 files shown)


packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/messages.ts
```diff
@@ -297,7 +297,7 @@ export const IO = {
     interface: 'ErrorPayload',
   }),
 
-  // 9: Bootstrap
+  // 9: Bootstrap (9xxx)
```
kaizencc, line 300: i assume by the way you are "enforcing" this that these are not meant to be strongly enforced at all, and will rely on our attention to detail to ensure we follow.

i feel like we are getting to the point where these rules may be hard to get right when we eventually introduce new message codes.

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -181,23 +181,12 @@ export class StackActivityMonitor {
     this.scheduleNextTick();
   }
 
-  private findMetadataFor(logicalId: string | undefined): ResourceMetadata | undefined {
+  private findMetadataFor(logicalId: string | undefined) {
```
kaizencc, line 184: why delete this? seems like we are still returning the same type

kaizencc (approved):
feel free to merge once comments are responded to :)

mrgrain:
bypassing integ tests due to unrelated failure

PR #244, opened 2025-03-17
fix(cli): hotswap is reporting and running changes that don't happen

mrgrain:
This change is three fold:

- We clean up the private `HotswapOperation` interface (née `HotswappableChange`) by removing two unused fields and adding in a new filed in preparation for structured data alongside hotswap messages
- In the hotswap providers, pull gates outside of the apply function. This prevents hotswaps being reported when nothing would actually be done. These are local checks and now just run a little bit earlier.
- In `lambda-functions.ts` don't report hotswappable changes for versions and aliases. These used to be reported with an empty array for `resourceNames` and a noop `apply` function. This means that literally nothing happens with these entries, since `apply` doesn't do anything and the CLI uses `resourceNames` to print anything. Just removing these as they are reported as part of a change to the function.

---

### DIFF
--- a/packages/aws-cdk/lib/api/hotswap/common.ts
+++ b/packages/aws-cdk/lib/api/hotswap/common.ts
@@ -1,13 +1,16 @@
-import type { PropertyDifference, Resource } from '@aws-cdk/cloudformation-diff';
+import type { PropertyDifference } from '@aws-cdk/cloudformation-diff';
+import type { HotswappableChange, ResourceChange } from '../../../../@aws-cdk/tmp-toolkit-helpers/src/api/io/payloads/hotswap';
 import { ToolkitError } from '../../toolkit/error';
 import type { SDK } from '../aws-auth';
 
 export const ICON = '✨';
 
-export interface HotswappableChange {
+export interface HotswapOperation {
+  /**
+   * Marks the operation as hotswappable
+   */
   readonly hotswappable: true;
-  readonly resourceType: string;
-  readonly propsChanged: Array<string>;
+
   /**
    * The name of the service being hotswapped.
    * Used to set a custom User-Agent for SDK calls.
@@ -15,10 +18,18 @@ export interface HotswappableChange {
   readonly service: string;
 
   /**
+   * Description of the change that is applied as part of the operation
+   */
+  readonly change: HotswappableChange;
+
+  /**
    * The names of the resources being hotswapped.
    */
   readonly resourceNames: string[];
 
+  /**
+   * Applies the hotswap operation
+   */
   readonly apply: (sdk: SDK) => Promise<void>;
 }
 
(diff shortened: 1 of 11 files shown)


packages/aws-cdk/lib/api/hotswap/common.ts
```diff
@@ -1,24 +1,35 @@
...
 export const ICON = '✨';
 
-export interface HotswappableChange {
+export interface HotswapOperation {
+  /**
+   * Marks the operation as hotswappable
+   */
   readonly hotswappable: true;
-  readonly resourceType: string;
-  readonly propsChanged: Array<string>;
```
mrgrain, line 10: Surprisingly, these were unused. `propsChanged` is not made available as raw data via `change`.

packages/aws-cdk/lib/api/hotswap/common.ts
```diff
@@ -65,38 +76,6 @@ export enum HotswapMode {
   FULL_DEPLOYMENT = 'full-deployment',
 }
 
-/**
- * Represents a change that can be hotswapped.
- */
-export class HotswappableChangeCandidate {
```
mrgrain, line 71: This is a class, but was never used as one. It's now an interface as it should be.

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -1,19 +1,21 @@
...
 import {
   classifyChanges, lowerCaseFirstCharacter,
   reportNonHotswappableChange,
   transformObjectKeys,
 } from './common';
+import type { ResourceChange } from '../../../../@aws-cdk/tmp-toolkit-helpers/src/api/io/payloads/hotswap';
 import type { SDK } from '../aws-auth';
 import type { EvaluateCloudFormationTemplate } from '../evaluate-cloudformation-template';
 
+const EcsServiceResourceType = 'AWS::ECS::Service';
```
rix0rrr, line 14: 💅 Super-nit: `ECS_SERVICE_RESOURCE_TYPE`.

packages/aws-cdk/lib/api/hotswap/lambda-functions.ts
```diff
@@ -369,38 +346,44 @@ async function versionsAndAliases(logicalId: string, evaluateCfnTemplate: Evalua
...
+  functionName: string,
   evaluateCfnTemplate: EvaluateCloudFormationTemplate,
-  callbackfn: (value: any, index: number, array: any[]) => Promise<string>,
-): Promise<string[]> {
-  const aliasesNames = (await versionsAndAliases(logicalId, evaluateCfnTemplate)).aliasesNames;
+): Promise<Array<{
+    logicalId: string;
+    resourceType: string;
+    physicalName?: string;
+    description: string;
```
rix0rrr, line 361: 💅 This would be nicer with a named struct definition.

rix0rrr (approved):

packages/aws-cdk/lib/api/hotswap/lambda-functions.ts
mrgrain: Coming up next!

mrgrain:
Bypassing tests due to unrelated test failure.

PR #246, opened 2025-03-18
chore: use non-caret dependency versions

rix0rrr:
Monorepo dependencies were using carets `^`; for most of these it didn't matter because the resulting artifact was bundled anyway, but for at least one it does matter:

* `cdk -> aws-cdk`: should use the exact same version, otherwise running `npx cdk@2.1000` may accidentally execute `aws-cdk@2.1001`.

Also making the `cloud-assembly-schema` references exact; probably doesn't matter since we're bundling most of these but it does show the intention of shipping exactly one version.

---


mrgrain (approved):

mrgrain:
I'm just worried this will affect the release process (read: bump and unbump and if the repo is considered unchanged afterwards).

.projenrc.ts
```diff
@@ -1059,8 +1059,8 @@ const toolkitLib = configureProject(
     description: 'AWS CDK Programmatic Toolkit Library',
     srcdir: 'lib',
     deps: [
-      cloudAssemblySchema,
-      cloudFormationDiff,
+      cloudAssemblySchema.customizeReference({ exactVersion: true }),
+      cloudFormationDiff.customizeReference({ exactVersion: true }),
```
mrgrain, line 1063: That's not right. Toolkit Library intentionally uses dependency ranges. In fact it should probably be wider and declare the actual minimal version and not just `^current`.

PR #249, opened 2025-03-18
feat(toolkit): add typed return to bootstrap action

### LINKED ISSUE #192: Toolkit Action: typed return for bootstrap
Split-out from #63

iankhou:
Fixes #192

## Description
The `bootstrap` action in the Toolkit Library currently does not have a returned value. Although there is logged output, the `Toolkit` instance does not get any information about the result of the bootstrap action.

## Solution
Make `Toolkit.bootstrap()` Return an instance of type `Promise<BootstrapResult>`, which also uses new type `EnvironmentBootstrapResult` with metrics for each environment.

## Testing
Added unit tests.

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -177,17 +182,59 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
         await ioHelper.notify(IO.CDK_TOOLKIT_I9900.msg(chalk.green('\n' + message), { environment }));
         await bootstrapSpan.end();
       } catch (e: any) {
+        const result: EnvironmentBootstrapResult = {
+          environment: environment.name,
+          status: 'failed',
+          error: e,
+          message: formatErrorMessage(e),
+          duration: Date.now() - envStartTime,
+        };
```
mrgrain, line 208: we throw an exception three lines  further down, so this will just never happen.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
```diff
@@ -169,6 +169,25 @@ export interface BootstrapParameters {
   readonly customPermissionsBoundary?: string;
 }
 
+export interface EnvironmentBootstrapResult {
+  environment: string;
+  status: 'success' | 'failed' | 'no-op';
+  message?: string;
+  error?: Error;
```
mrgrain, line 176: See other comment, this will just never be returned.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
```diff
@@ -169,6 +169,25 @@ export interface BootstrapParameters {
   readonly customPermissionsBoundary?: string;
 }
 
+export interface EnvironmentBootstrapResult {
+  environment: string;
+  status: 'success' | 'failed' | 'no-op';
```
mrgrain, line 174: See other comment, `failed` will just never be returned.

```suggestion
  status: 'success' | 'no-op';
```

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
```diff
@@ -169,6 +169,25 @@ export interface BootstrapParameters {
   readonly customPermissionsBoundary?: string;
 }
 
+export interface EnvironmentBootstrapResult {
+  environment: string;
```
mrgrain, line 173: Would this be better types as ?
```suggestion
  environment: cxapi.Environment;
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -177,17 +182,59 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
+        case 'no-op':
+          acc.noOp++;
+          break;
+      }
+      return acc;
+    }, { total: 0, successful: 0, failed: 0, noOp: 0 });
+
+    return {
+      environments: results,
+      summary,
```
mrgrain, line 235: I think let's just loose the summary. If we return, it's successful. If someone really cares about the counts for success vs. no-op, they can do that themselves.

mrgrain (changes requested):
Looks good! Let's do a little less and re-use what's already there.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -177,17 +182,59 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
             usePreviousParameters: parameters?.keepExistingParameters,
           },
         );
+
+        const result: EnvironmentBootstrapResult = {
+          environment: environment.name,
+          status: bootstrapResult.noOp ? 'no-op' : 'success',
+          message: bootstrapResult.noOp ? 'No changes required' : 'Successfully bootstrapped',
+          duration: Date.now() - envStartTime,
```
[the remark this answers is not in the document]
iankhou: I see, I've removed this field.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
iankhou: Got it, removed this code. I also removed 'failed' as a status case because it throws an error.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
iankhou: Yup, makes sense. Removed the summary.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
iankhou: Replaced and updated tests as well.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
mrgrain: It's still there

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -177,6 +180,15 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
             usePreviousParameters: parameters?.keepExistingParameters,
           },
         );
+
+        const result: EnvironmentBootstrapResult = {
+          environment,
+          status: bootstrapResult.noOp ? 'no-op' : 'success',
+          message: bootstrapResult.noOp ? 'No changes required' : 'Successfully bootstrapped',
```
mrgrain, line 187: this seems like duplication of the status, do we need this?

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
```diff
@@ -169,6 +169,18 @@ export interface BootstrapParameters {
   readonly customPermissionsBoundary?: string;
 }
 
+export interface EnvironmentBootstrapResult {
+  environment: cxapi.Environment;
+  status: 'success' | 'no-op';
+  message?: string;
```
mrgrain, line 175: if we keep message, it should not be optional since we always set it anyway.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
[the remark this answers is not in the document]
mrgrain: We can have it, let's just don't time twice and instead do
```
const envTime = await bootstrapSpan.end();
// ...
{ duration: envTime.asMs }
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
iankhou: Agree that it presents the same information. We probably don't need message as we already have a human-readable message via the IOHost. Will remove it.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
iankhou: Removed it as per the change in toolkit.ts.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
```diff
@@ -169,6 +169,17 @@ export interface BootstrapParameters {
   readonly customPermissionsBoundary?: string;
 }
 
+export interface EnvironmentBootstrapResult {
+  environment: cxapi.Environment;
+  status: 'success' | 'no-op';
+  error?: Error;
```
mrgrain, line 175:
```suggestion
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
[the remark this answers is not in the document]
iankhou: Sorry, I had misunderstood what you meant, I'll use the time from bootstrapSpan.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
iankhou: Removed this, it will never be used.

packages/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.ts
```diff
@@ -169,6 +169,17 @@ export interface BootstrapParameters {
   readonly customPermissionsBoundary?: string;
 }
 
+export interface EnvironmentBootstrapResult {
+  environment: cxapi.Environment;
+  status: 'success' | 'no-op';
+  duration: Number;
```
mrgrain, line 175:
```suggestion
  duration: number;
```

mrgrain (changes requested):

mrgrain (approved):

PR #251, opened 2025-03-19
feat(cli): CcApi context provider can be configured to fail if listing does not find a specific count of resources

### LINKED ISSUE #257: contextprovider: CC API context provider should ensure the result has exact one value
### Describe the bug

While conversations in aws/aws-cdk#33619, implementing `PrefixList.fromLookup()`, I encountered issues how to handle unexpected result from CC API context provider with `propertyMatch`.

Other context provider returns one of the expected query result or a dummy value with provider error.

But CC API context provider returns an empty array `[]` when no resources match or an array with multiple elements when multiple resources match. It's not expected result for lookup method.

For example, when the result is `[]`, `cdk.context.json` has the fixed result `[]` and never be removed. `Vpc.fromLookup()` removes the result from `cdk.context.json` when no VPC are found.

### Regression Issue

_No response_

### Expected Behavior

CC API context provider should report ContextProviderError when the result does not have exact one resource.

### Current Behavior

CC API context provider can return an empty or more than one entry.

### Reproduction Steps

After synth the following stack,

``` ts
export class CCApiTest extends Stack {
  constructor(scope: Construct, id: string, props: StackProps) {
    super(scope, id, props);

    const value = ContextProvider.getValue(this, {
      provider: cxschema.ContextProvider.CC_API_PROVIDER,
      props: {
        typeName: 'AWS::EC2::PrefixList',
        propertyMatch: {
          PrefixListName: 'not-exist-name',
        },
        propertiesToReturn: ['PrefixListId'],
      },
      dummyValue: [{ PrefixListId: 'pl-xxxxxxxx' }],
    });

    new CfnOutput(this, 'PrefixListQueryResult', {
      value: value.PrefixListId,
    });
  }
}
```

`cdk.context.json` contains

``` json5
{
  "cc-api-provider:account=ACCOUNTID:propertiesToReturn.0=PrefixListId:propertyMatch.PrefixListName=not-exist-name:region=REGION:typeName=AWS$:$:EC2$:$:PrefixList": []
}
```

### Possible Solution

In `CcApiContextProviderPlugin.listResources()`, throws ContextProviderError when the result does not have exact one resource, instead of returning as-is.

_No response_

### CDK CLI Version

2.1004.0

### Framework Version

_No response_

### Node.js Version

v22.14.0

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

Tietew:
Fixes #257

As described in the issue, `CcApiContextProviderPlugin` should have the way to ensure the result has exact one resource.

This PR adds the `expectedMatchCount` option to restrict the length of results in `listResources()`.

Unit tests are added and modified ensure the changes.

---


mrgrain (commented):
@rix0rrr this looks okay to me, but you were more involved with the original version of this

rix0rrr:
I agree with you that persisting a 0-length array may make it harder to recover from problems in the future.

But I don't agree that a >=1 length array is automatically a failure. It is up the `lookup` method to reject >=1 element if it is expecting exactly one. It is not up to the provider to say "I will never support looking up more than one resource". For example, securitygroups associated with another resource are a perfectly legit resource to look up.

rix0rrr (changes requested):
I agree with you that persisting a 0-length array may make it harder to recover from problems in the future. 

But I don't agree that a >=1 length array is automatically a failure. It is up the `lookup` method to reject >=1 element if it is expecting exactly one. It is not up to the provider to say "I will never support looking up more than one resource". For example, securitygroups associated with another resource are a perfectly legit resource to look up. 

For your issue:

* It is the job of `PrefixList.fromLookup()` to handle the >=1 case.
* As for the 0 results case: we can fail the provider, that's an easy way of making sure that the 0-length array is not persisted. To achieve that, we should add an additional field into the query that indicates that a 0-element list should be considered an error.

packages/aws-cdk/test/context-providers/cc-api-provider.test.ts
```diff
@@ -107,10 +107,6 @@ test('looks up RDS instance using CC API listResources', async () => {
         Identifier: 'my-db-instance-2',
         Properties: '{"DBInstanceArn":"arn:aws:rds:us-east-1:123456789012:db:test-instance-2","StorageEncrypted":"false","Endpoint":{"Address":"address2.amazonaws.com","Port":"5432"}}',
       },
-      {
-        Identifier: 'my-db-instance-3',
-        Properties: '{"DBInstanceArn":"arn:aws:rds:us-east-1:123456789012:db:test-instance-3","StorageEncrypted":"true","Endpoint":{"Address":"address3.amazonaws.com","Port":"6000"}}',
-      },
```
rix0rrr, line 113: The fact that you need to break tests (i.e., behavior) to achieve what you want to do should already tell you we can't accept this 😉

Tietew:
@rix0rrr Thank you for clarification.

I've pushed following updates:
- reverted the test I changed
- added `allowEmptyResult` and `exactResult` options to `CcApiContextQuery`
- updated tests to ensure above options

Current PRs using CC API Context Provider, except for my PrefixList, use `exactIdentifier` and is not affected by this change.
For my `PrefixList.fromLookup()`, specifying `exactResult: true` could ensure to expect 1 resources.

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/context-queries.ts
```diff
@@ -385,6 +385,18 @@ export interface CcApiContextQuery extends ContextLookupRoleOptions {
...
+   * Whether to allow to return empty result when propertyMatch is specified.
+   * @default false
+   */
+  readonly allowEmptyResult?: boolean;
+
+  /**
+   * Whether to restrict to return exact one result when propertyMatch is specified.
+   * @default false
+   */
+  readonly exactResult?: boolean;
```
rix0rrr, line 399: Is this an enum in hiding?

I wonder if the following isn't more readable: 

```ts
expectedMatchCount?: 'any' | 'at-least-one' | 'at-most-one' | 'exactly-one';
```

rix0rrr (changes requested):

rix0rrr:
I'm going to wait for https://github.com/aws/aws-cdk-cli/pull/211 to be in, then refactor this PR a bit on top of that other one.

They're too uncomfortably close for me so I'm taking the reins to make them play nicely together.

Tietew:
Ok, thanks for adjustment!

rix0rrr (approved):

PR #252, opened 2025-03-19
chore: adds well-defined reason to non hotswappable resources

mrgrain:
This is in preparation of making the value available in a structured message to the IoHost.

---


otaviomacedo (approved):

mrgrain:
applied exemption label because a previous run passed

PR #255, opened 2025-03-19
refactor(cli): non hotswappable changes with structured data

mrgrain:
Refactor `NonHotswappableChange` to contain consistent, well-structured data.

---

### DIFF
--- a/packages/aws-cdk/lib/api/deployments/hotswap-deployments.ts
+++ b/packages/aws-cdk/lib/api/deployments/hotswap-deployments.ts
@@ -3,7 +3,7 @@ import * as cfn_diff from '@aws-cdk/cloudformation-diff';
 import type * as cxapi from '@aws-cdk/cx-api';
 import type { WaiterResult } from '@smithy/util-waiter';
 import * as chalk from 'chalk';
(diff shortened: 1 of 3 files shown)


packages/aws-cdk/lib/api/deployments/hotswap-deployments.ts
```diff
@@ -414,36 +427,56 @@ function makeRenameDifference(
...
       logicalId,
-      reason: NonHotswappableReason.RESOURCE_DELETION,
-      description: `resource '${logicalId}' was destroyed by this deployment`,
+      change: {
+        reason: NonHotswappableReason.RESOURCE_DELETION,
+        description: `resource '${logicalId}' was destroyed by this deployment`,
+        subject: {
+          type: 'Resource',
+          logicalId,
+          resourceType: change.oldValue?.Type,
```
iliapolo, line 459: nit (maybe): this used to be:

```ts
change.oldValue!.Type
```

So now we allow null values where we used to fail on and proceeding with this code path could be unexpected. I recognize this is probably just type-checking blindspots but wondering if there's something better to do here.

iliapolo (approved):

PR #265, opened 2025-03-20
refactor(cli): re-organize files

mrgrain:
Hopefully the final big code re-org in the CLI before we move code into the shared package.

This move focuses on moving top-level files into `commands` and `cli` respectively, to indicate they will stay within the CLI package.
Also removes the `toolkit/error.ts` re-export, which causes many of the import changes.

Includes a non-consequential bug fix, to the init command and how it's finding the new `.init-version.json` and `.recommended-feature-flags.json` files. This was incorrect before and would happened to work because `init.ts` was in the same directory as `index.ts`.  However because the code is bundled, we need to address the `init-templates` (and the new files) from the project root location. This is already done for the templates themselves, but was missed for `.init-version.json` and `.recommended-feature-flags.json`.

---


packages/aws-cdk/lib/api/deployments/asset-publishing.ts
mrgrain, line 1: This is the only larger code change in this PR

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -500,7 +500,7 @@ interface Versions {
  * This has been built into the CLI at build time.
  */
 async function loadInitVersions(): Promise<Versions> {
-  const initVersionFile = path.join(__dirname, './init-templates/.init-version.json');
+  const initVersionFile = path.join(cliRootDir(), 'lib', 'init-templates', '.init-version.json');
```
mrgrain, line 503: This was incorrect before and would have broken the toolkit-lib and cli-lib-alpha and integ-runner and cdk-cli-wrapper if they were supporting `init` command.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -524,6 +524,6 @@ async function loadInitVersions(): Promise<Versions> {
  * These have been built into the CLI at build time.
  */
 export async function currentlyRecommendedAwsCdkLibFlags() {
-  const recommendedFlagsFile = path.join(__dirname, './init-templates/.recommended-feature-flags.json');
+  const recommendedFlagsFile = path.join(cliRootDir(), 'lib', 'init-templates', '.recommended-feature-flags.json');
```
mrgrain, line 527: This was incorrect before and would have broken the toolkit-lib and cli-lib-alpha and integ-runner and cdk-cli-wrapper if they were supporting init command.

packages/aws-cdk/lib/commands/init/init.ts
mrgrain, line 1: These fixes surfaced because the file moved to a subdirectory.

packages/aws-cdk/lib/toolkit/error.ts
mrgrain, line 1: we now import directly from the shared library

rix0rrr (approved):

PR #266, opened 2025-03-20
refactor(cli): use modern messaging infrastructure

mrgrain:
Updates code in `lib/api` to use the latest version of the modern messaging infrastructure instead of legacy helpers.

---


packages/aws-cdk/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -536,11 +535,11 @@ export class GarbageCollector {
         });
 
         const deletedCount = batch.length;
-        await this.ioHelper.notify(debug(`Deleted ${deletedCount} assets`));
+        await this.ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(`Deleted ${deletedCount} assets`));
         printer.reportDeletedAsset(deletables.slice(0, deletedCount));
       }
     } catch (err) {
-      await this.ioHelper.notify(info(chalk.red(`Error deleting images: ${err}`)));
```
iliapolo, line 543: Intentional color removal?

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -119,12 +118,12 @@ export class StackActivityMonitor {
 
   public async start() {
     this.monitorId = uuid.v4();
-    await this.ioHelper.notify(debug(`Deploying ${this.stackName}`, 'CDK_TOOLKIT_I5501', {
```
iliapolo, line 122: debug to info? intentional?

packages/aws-cdk/lib/api/garbage-collection/garbage-collector.ts
mrgrain: Intentional. Turned the red info message that says "Error" into a proper error level message, which will be colored red as well.

packages/aws-cdk/lib/api/stack-events/stack-activity-monitor.ts
mrgrain: Yes, they are `info` from the toolkit perspective. In our CLI we handle the Stack Monitoring messages differently (via the activity printers `--progress=events` and `--progress=bar`), so we don't actually care about the levle here.

iliapolo (approved):

PR #270, opened 2025-03-20
refactor(cli): new internal cloudformation api module

mrgrain:
Pulls some shared "cloudformation" helpers into a new internal api module. This resolves some annoying inter-dependencies between internal modules and specifically breaks up the `deployment` mega module a bit.

Also moves some other files from `api/util` so that it's now almost gone (which is the goal).

---


packages/aws-cdk/lib/api/cloudformation/stack-helpers.ts
mrgrain, line 1: this is a collection of helpers that were previously in `api/deployments` but in reality were used across multiple modules.

otaviomacedo (approved):

PR #278, opened 2025-03-24
chore(tmp-toolkit-helpers): `formatStackDiff` and `formatSecurityDiff` moved to tmp-toolkit-helpers

kaizencc:
moves two functions that are relevant to diff into `tmp-toolkit-helpers`. this will facilitate the adoption of `diff` in the programmatic toolkit as it will now be able to reuse these functions. 

existing tests pass to ensure no breakage. additional tests have been added to `tmp-toolkit-helpers` to test `formatXxxDiff`, which were previously not directly tested.

---


packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/format/format-stack-diff.ts
```diff
@@ -0,0 +1,133 @@
...
+  // `formatStackDiff` to decide what to do with it.
+  const stream = new StringWriteStream();
+
+  let numStacksWithChanges = 0;
+  let formattedDiff = '';
+  let filteredChangesCount = 0;
+  try {
+    // must output the stack name if there are differences, even if quiet
+    if (stackName && (!quiet || !diff.isEmpty)) {
+      stream.write(format('Stack %s\n', chalk.bold(stackName)));
```
kaizencc, line 62: welp, these prints don't work yet and im worried im not doing the right thing here: 

`format` and `chalk` don't exist in this context and i'm not sure how to replicate them.

kaizencc: i guess a better question to ask was: are we certain we want it in tmp-toolkit-helpers and not imported [here](https://github.com/aws/aws-cdk-cli/blob/conroy/diffs/packages/@aws-cdk/toolkit-lib/lib/api/aws-cdk.ts) like a lot of other heavy duty imports.

but it sounds like your answer also answers that question too

packages/aws-cdk/lib/api/cloudformation/nested-stack-helpers.ts
```diff
@@ -5,15 +5,9 @@ import { formatErrorMessage } from '../../util';
...
 
-export interface NestedStackTemplates {
-  readonly physicalName: string | undefined;
-  readonly deployedTemplate: Template;
-  readonly generatedTemplate: Template;
-  readonly nestedStackTemplates: {
-    [nestedStackLogicalId: string]: NestedStackTemplates;
-  };
-}
+export { NestedStackTemplates } from '../../../../@aws-cdk/tmp-toolkit-helpers/src/api';
```
kaizencc, line 10: is it reasonable to both `export` and `import` `NestedStackTemplate` here?

packages/aws-cdk/test/cli/cdk-toolkit.test.ts
```diff
@@ -216,7 +216,7 @@ describe('deploy', () => {
       // WHEN
       await cdkToolkit.deploy({
         selector: { patterns: ['Test-Stack-A-Display-Name'] },
-        requireApproval: RequireApproval.Never,
+        requireApproval: RequireApproval.NEVER,
```
kaizencc, line 219: this is kind of silly but also a breaking change right? what should i have done instead?

.projenrc.ts
```diff
@@ -679,7 +679,9 @@ const tmpToolkitHelpers = configureProject(
     ],
     deps: [
       cloudAssemblySchema.name,
+      cloudFormationDiff,
```
kaizencc, line 682: elsewhere we are depending on `cloudformationDiff.customizeReference({ versionType: 'exact' }),` and i'm not sure why. just thought i'd flag

packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff.ts
```diff
@@ -0,0 +1,296 @@
...
+  ioHelper: IoHelper,
+  oldTemplate: any,
+  newTemplate: cxapi.CloudFormationStackArtifact,
+  strict: boolean,
+  context: number,
+  quiet: boolean,
+  stackName?: string,
+  changeSet?: DescribeChangeSetOutput,
+  isImport?: boolean,
+  nestedStackTemplates?: { [nestedStackLogicalId: string]: NestedStackTemplates }): FormatStackDiffOutput {
```
kaizencc, line 132: i agree that this sucks and should be part of a class with `formatSecurityDiff`. but i don't wanna do it in this PR, it's already too complicated.

packages/aws-cdk/lib/api/cloudformation/nested-stack-helpers.ts
mrgrain: yes, it's just going to annoy me when I do the move :D

packages/aws-cdk/test/cli/cdk-toolkit.test.ts
mrgrain: why is it breaking?

kaizencc: i suppose its not because the CLI users have no access to this..

packages/@aws-cdk/tmp-toolkit-helpers/test/api/diff/diff.test.ts
kaizencc, line 1: AI disclosure: this file was originally intended to be full AI generated and started off as such but it insisted on mocking `@aws-cdk/cloudformation-diff` functions and I really didn't want it to. So I spent twice as much time bashing the code into what I wanted. At the end of the day this is <50% AI generated.

mrgrain: There are much easier ways to achieve the same mocking results you were after. Check some other files for inspirations.

Specifically all calls end up in an IoHost, so all the helpers could just stay real.

packages/aws-cdk/lib/commands/diff.ts
```diff
@@ -1,294 +1 @@
...
-        return false;
-      }
-      if (change.oldResourceType === 'AWS::CDK::Metadata') {
-        return false;
-      }
-      return true;
-    });
-  }
-}
+export { formatStackDiff, formatSecurityDiff, RequireApproval } from '../../../@aws-cdk/tmp-toolkit-helpers/src/api';
```
mrgrain, line 1: That's fine. But you also could have updated any imports.

mrgrain (approved):

PR #279, opened 2025-03-24
feat(toolkit-lib): diff

kaizencc:
introduces diff support for programmatic toolkit.

- has feature parity with the existing CLI diff
- programmatic toolkit functionality is unit tested.
- AI disclosure: no AI

does most of https://github.com/aws/aws-cdk/issues/33182, but typed return will come later

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -249,6 +252,145 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
     return new IdentityCloudAssemblySource(assembly.assembly);
   }
 
+  /**
+   * Diff Action
+   */
+  public async diff(cx: ICloudAssemblySource, options: DiffOptions): Promise<number> {
```
mrgrain, line 258:
```suggestion
  public async diff(cx: ICloudAssemblySource, options: DiffOptions): Promise<TemplateDiff> {
```

kaizencc: ack. will do before this becomes a reviewable PR

packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
```diff
@@ -114,4 +115,32 @@ export interface DiffOptions {
...
+   * @default false
+   */
+  readonly compareAgainstProcessedTemplate?: boolean;
+
+  /**
+   * Whether or not to create, analyze, and subsequently delete a changeset
+   *
+   * @default true
+   */
+  readonly changeSet?: boolean;
```
kaizencc, line 145: need help with a decision here: the CLI default is `true`, and that is enforced via yargs default -- the code itself actually does not default to true (i.e. it is `if (options.changeSet) { ... }`). because of this, currently in this PR, programmatic toolkit updates that conditional to `if (options.changeSet ?? true)` to behave the same way.

packages/@aws-cdk/toolkit-lib/lib/api/aws-cdk.ts
```diff
@@ -12,6 +12,7 @@ export { CloudWatchLogEventMonitor, findCloudWatchLogGroups } from '../../../../
 export { type WorkGraph, WorkGraphBuilder, AssetBuildNode, AssetPublishNode, StackNode, Concurrency } from '../../../../aws-cdk/lib/api/work-graph';
 export { Bootstrapper } from '../../../../aws-cdk/lib/api/bootstrap';
 export { loadTree, some } from '../../../../aws-cdk/lib/api/tree';
+export { createDiffChangeSet } from '../../../../aws-cdk/lib/api/deployments/cfn-api';
```
kaizencc, line 15: i also need to use this function and there's not a good way of moving it to tmp-toolkit-helpers because i would basically be moving the entire `deployments` folder. wonder if this is ok for now

packages/aws-cdk/lib/api/deployments/cfn-api.ts
```diff
@@ -240,7 +240,7 @@ async function uploadBodyParameterAndCreateChangeSet(
       role: executionRoleArn,
     });
   } catch (e: any) {
-    await ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(e));
+    await ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(String(e)));
```
kaizencc, line 243: seems like a small bug i came across: the message must be a string as we eventually call `.trim()` on it

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
```diff
@@ -0,0 +1,253 @@
...
+        templatePath: path.join(__dirname, '..', '_fixtures', 'two-empty-stacks', 'cdk.out', 'Stack1.template.json'),
+      });
+
+      // THEN
+      expect(ioHost.notifySpy).toHaveBeenCalledWith(expect.objectContaining({
+        action: 'diff',
+        level: 'warn',
+        code: 'CDK_TOOLKIT_W0000',
+        message: expect.stringContaining('This deployment will make potentially sensitive changes according to your current security approval level (--require-approval broadening)'),
+      }));
```
kaizencc, line 231: sanity check: do we expect to have this as a warning notification when we run `toolkit.diff`?

packages/@aws-cdk/toolkit-lib/lib/api/aws-cdk.ts
mrgrain: Yes fine. Will move soon.

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
mrgrain: No, probably not.

packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
mrgrain: I'm guessing in the CLI this default is set via the cli option.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/messages.ts
```diff
@@ -81,6 +81,30 @@ export const IO = {
   }),
 
   // 4: Diff (4xxx)
+  CDK_TOOLKIT_I4200: make.debug({
```
mrgrain, line 84: For all of these: they should have data if useful, or not get a code

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -249,6 +253,157 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
+          'Can only select one stack when comparing to fixed template. Use --exclusively to avoid selecting multiple stacks.',
+        );
+      }
+
+      if (!(await fs.pathExists(options.templatePath))) {
+        throw new ToolkitError(`There is no file at ${options.templatePath}`);
+      }
+
+      const file = fs.readFileSync(options.templatePath).toString();
+      const template = deserializeStructure(file);
```
mrgrain, line 287: This wants to be a helper method

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -249,6 +253,157 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
+      if (options.securityOnly) {
```
mrgrain, line 288: All this smells like helper functions to me. It's so hard to read at the moment.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -249,6 +253,157 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
+          await ioHelper.notify(IO.CDK_TOOLKIT_I4401.msg(diff.formattedDiff), );
+          diffs += diff.numStacksWithChanges;
+          fullDiff = diff.fullDiff;
+        }
+      }
+    }
+    await ioHelper.notify(IO.CDK_TOOLKIT_I4402.msg(`Number of differences: ${diffs}`));
+
+    // Shouldn't happen but just in case
+    if (!fullDiff) {
```
mrgrain, line 400: Probably the code can be changed to guarantee this

packages/@aws-cdk/toolkit-lib/lib/api/aws-cdk.ts
kaizencc: ok, but doing so means that i export the `cfn-api` file out of `aws-cdk`

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
kaizencc: this comes from [here](https://github.com/aws/aws-cdk-cli/blob/af827ea95df07f6090d69aa9119043f642f72c2a/packages/%40aws-cdk/tmp-toolkit-helpers/src/api/diff/diff.ts#L79-L80)

it was previously a `warning()` before being refactored into `ioDefaultHelper.warning`, so it was always printed as a warning even when the command was `cdk diff` (and not `cdk deploy`, where presumably you would need to stop and confirm the diff). `cdk deploy` has its own confirm message that actually prompts the user.

i think we should be able to just delete those two lines -- i don't think they are useful. alternatively, i don't think its harmful to leave them in.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
kaizencc: i think this can be achieved in my subsequent PR that pulls what i can into helper methods

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
kaizencc: ack. i've decided to do a subsequent PR to refactor the code here

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
kaizencc, line 1: i removed the return statement, now all the tests are kind of ugly because the easiest thing to test for is the return `TemplateDiff`. i suppose they can stay ugly until we settle on the correct return value in a separate PR, which i agree is a map of template diffs.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/messages.ts
kaizencc: 1) i'm not sure if i should return the formatted diff (string) or the template diff. 
2) why? i see lots of other instances of coded messages that do not have any data attached.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/payloads/diff.ts
```diff
@@ -17,3 +17,18 @@ export enum PermissionChangeType {
    */
   NON_BROADENING = 'non-broadening',
 }
+
+/**
+ * Output of the diff command
+ */
+export interface DiffOutput {
```
mrgrain, line 24:
```suggestion
export interface DiffResult {
```

packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
```diff
@@ -28,6 +28,10 @@ export interface ChangeSetDiffOptions extends CloudFormationDiffOptions {
   readonly parameters?: { [name: string]: string | undefined };
 }
 
+export interface Options extends ChangeSetDiffOptions, CloudFormationDiffOptions {
```
mrgrain, line 31: Seems like a dangerous name

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -249,6 +252,143 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
+          } catch (e: any) {
+            await ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(`Checking if the stack ${stack.stackName} exists before creating the changeset has failed, will base the diff on template differences.\n`));
+            await ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(formatErrorMessage(e)));
+            stackExists = false;
+          }
+
+          if (stackExists) {
+            changeSet = await createDiffChangeSet(ioHelper, {
+              stack,
+              uuid: uuid.v4(),
```
mrgrain, line 350: such a weird thing that this needs to be inserted 😂

packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
```diff
@@ -66,7 +73,7 @@ export class DiffMethod {
 
   private constructor(
     public readonly method: 'change-set' | 'template-only' | 'local-file',
-    public readonly options: ChangeSetDiffOptions | CloudFormationDiffOptions | { path: string },
+    public readonly options: Options,
```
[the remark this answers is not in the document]
mrgrain: I mean I was going for the type cast, but I can see how this is annoying. The problem is you still need to use `!` now (e.g. on path), so I don't think this is really much better.

you create a little helper that abstracts the type cast, but same same really

```ts
function diffMethodOptions<T>(diffMethod: DiffMethod): T {
  return diffMethod.options as T;
}
```

packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/messages.ts
```diff
@@ -500,6 +516,11 @@ export const SPAN = {
     start: IO.CDK_TOOLKIT_I6100,
     end: IO.CDK_TOOLKIT_I6000,
   },
+  DIFF_STACK: {
+    name: 'Diff',
+    start: IO.CDK_TOOLKIT_I5200,
+    end: IO.CDK_TOOLKIT_I5201,
```
mrgrain, line 522: wrong codes

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -249,6 +252,145 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
+          formattedStackDiff = diff.formattedDiff;
+          diffs = diff.numStacksWithChanges;
+        }
+      }
+    }
+    await ioHelper.notify(IO.CDK_TOOLKIT_I4401.msg(`✨ Number of stacks with differences: ${diffs}`, {
+      formattedSecurityDiff,
+      formattedStackDiff,
+    }));
+    await diffSpan.end();
```
mrgrain, line 389: the first of these two message just is the the end. no need to doubling

kaizencc: youre saying i dont need diffSpan.end()?

mrgrain: You need `diffSpan.end()` but you are using that to send `CDK_TOOLKIT_I4401` with the payload

kaizencc: or i need to end the span with the payload instead

mrgrain:
```ts
    await diffSpan.end(`✨ Number of stacks with differences: ${diffs}`, {
      formattedSecurityDiff,
      formattedStackDiff,
    });
```

kaizencc: its the latter :)

mrgrain (approved):

PR #292, opened 2025-03-28
refactor(cli): aws-auth uses modern messaging infrastructure everywhere

mrgrain:
Updates `aws-auth` module to use modern messaging infrastructure.

## Testing

- Existing unit and integration tests are passing
- Successfully run the auth test suite, with the exception of `tests/test-non-commercial-region.sh` which was skipped. 
- Manually confirmed that the `AccountAccessKeyCache` is still written
- Manually executed common CLI operations and I did not spot  any additional output (as expected)

---

### DIFF
--- a/packages/aws-cdk/lib/api/aws-auth/account-cache.ts
+++ b/packages/aws-cdk/lib/api/aws-auth/account-cache.ts
@@ -1,7 +1,6 @@
 import * as path from 'path';
 import * as fs from 'fs-extra';
 import type { Account } from './sdk-provider';
-import { debug } from '../../logging';
 import { cdkCacheDir } from '../../util';
 
 /**
@@ -16,13 +15,24 @@ export class AccountAccessKeyCache {
    */
   public static readonly MAX_ENTRIES = 1000;
 
+  /**
+   * The default path used for the accounts access key cache
+   */
+  public static get DEFAULT_PATH(): string {
+    // needs to be a getter because cdkCacheDir can be set via env variable and might change
+    return path.join(cdkCacheDir(), 'accounts_partitions.json');
+  }
+
   private readonly cacheFile: string;
+
+  private readonly debug: (msg: string) => Promise<void>;
 
   /**
    * @param filePath Path to the cache file
    */
-  constructor(filePath?: string) {
-    this.cacheFile = filePath || path.join(cdkCacheDir(), 'accounts_partitions.json');
+  constructor(filePath: string = AccountAccessKeyCache.DEFAULT_PATH, debugFn: (msg: string) => Promise<void>) {
+    this.cacheFile = filePath;
+    this.debug = debugFn;
   }
 
   /**
@@ -40,7 +50,7 @@ export class AccountAccessKeyCache {
     // try to get account ID based on this access key ID from disk.
     const cached = await this.get(accessKeyId);
     if (cached) {
-      debug(`Retrieved account ID ${cached.accountId} from disk cache`);
+      await this.debug(`Retrieved account ID ${cached.accountId} from disk cache`);
       return cached;
     }
 
--- a/packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
+++ b/packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
@@ -1,15 +1,15 @@
+import { format } from 'node:util';
 import { createCredentialChain, fromEnv, fromIni, fromNodeProviderChain } from '@aws-sdk/credential-providers';
 import { MetadataService } from '@aws-sdk/ec2-metadata-service';
 import type { NodeHttpHandlerOptions } from '@smithy/node-http-handler';
 import { loadSharedConfigFiles } from '@smithy/shared-ini-file-loader';
 import type { AwsCredentialIdentityProvider, Logger } from '@smithy/types';
 import * as promptly from 'promptly';
-import { ProxyAgent } from 'proxy-agent';
 import { makeCachingProvider } from './provider-caching';
+import { ProxyAgentProvider } from './proxy-agent';
 import type { SdkHttpOptions } from './sdk-provider';
-import { readIfPossible } from './util';
 import { AuthenticationError } from '../../../../@aws-cdk/tmp-toolkit-helpers/src/api';
-import { debug } from '../../logging';
+import { IO, type IoHelper } from '../../../../@aws-cdk/tmp-toolkit-helpers/src/api/io/private';
 
 const DEFAULT_CONNECTION_TIMEOUT = 10000;
 const DEFAULT_TIMEOUT = 300000;
@@ -23,16 +23,22 @@ const DEFAULT_TIMEOUT = 300000;
  * https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
  */
 export class AwsCliCompatible {
+  private readonly ioHelper: IoHelper;
+
+  public constructor(ioHelper: IoHelper) {
+    this.ioHelper = ioHelper;
+  }
+
   /**
    * Build an AWS CLI-compatible credential chain provider
    *
    * The credential chain returned by this function is always caching.
    */
-  public static async credentialChainBuilder(
+  public async credentialChainBuilder(
     options: CredentialChainOptions = {},
   ): Promise<AwsCredentialIdentityProvider> {
     const clientConfig = {
-      requestHandler: AwsCliCompatible.requestHandlerBuilder(options.httpOptions),
+      requestHandler: await this.requestHandlerBuilder(options.httpOptions),
       customUserAgent: 'aws-cdk',
       logger: options.logger,
     };
@@ -63,7 +69,7 @@ export class AwsCliCompatible {
       return makeCachingProvider(fromIni({
         profile: options.profile,
         ignoreCache: true,
-        mfaCodeProvider: tokenCodeFn,
+        mfaCodeProvider: makeTokenCodeFn(this.ioHelper),
         clientConfig,
         parentClientConfig,
         logger: options.logger,
@@ -100,7 +106,7 @@ export class AwsCliCompatible {
       clientConfig,
       parentClientConfig,
       logger: options.logger,
-      mfaCodeProvider: tokenCodeFn,
+      mfaCodeProvider: makeTokenCodeFn(this.ioHelper),
       ignoreCache: true,
     });
 
@@ -109,8 +115,8 @@ export class AwsCliCompatible {
       : nodeProviderChain;
   }
 
-  public static requestHandlerBuilder(options: SdkHttpOptions = {}): NodeHttpHandlerOptions {
-    const agent = this.proxyAgent(options);
+  public async requestHandlerBuilder(options: SdkHttpOptions = {}): Promise<NodeHttpHandlerOptions> {
+    const agent = await new ProxyAgentProvider(this.ioHelper).create(options);
 
     return {
       connectionTimeout: DEFAULT_CONNECTION_TIMEOUT,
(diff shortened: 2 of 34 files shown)


packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
mrgrain, line 1: this is a weird change, but i think the abstraction was kind of wrong already.

packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
```diff
@@ -120,19 +126,6 @@ export class AwsCliCompatible {
     };
   }
 
-  public static proxyAgent(options: SdkHttpOptions) {
```
mrgrain, line 123: pulled out into a separate class. this had otherwise nothing to do with this file

packages/aws-cdk/lib/api/aws-auth/credential-plugins.ts
```diff
@@ -23,9 +23,11 @@ import { PluginHost } from '../plugin/plugin';
 export class CredentialPlugins {
   private readonly cache: { [key: string]: PluginCredentialsFetchResult | undefined } = {};
   private readonly host: PluginHost;
+  private readonly ioHelper: IoHelper;
 
-  constructor(host?: PluginHost) {
-    this.host = host ?? PluginHost.instance;
+  constructor(host: PluginHost, ioHelper: IoHelper) {
```
mrgrain, line 28: just forcing the call site to pass the plugin host explicitly (which is still `PluginHost.instance`)

packages/aws-cdk/lib/api/aws-auth/sdk.ts
```diff
@@ -1006,7 +1015,7 @@ export class SDK {
         if (!accountId) {
           throw new AuthenticationError("STS didn't return an account ID");
         }
-        debug('Default account ID:', accountId);
+        await this.debug(`Default account ID ${accountId}`);
```
mrgrain, line 1018:
```suggestion
        await this.debug(`Default account ID: ${accountId}`);
```

packages/aws-cdk/lib/api/aws-auth/util.ts
```diff
@@ -13,7 +12,6 @@ export function readIfPossible(filename: string): string | undefined {
     }
     return fs.readFileSync(filename, { encoding: 'utf-8' });
   } catch (e: any) {
-    debug(e);
```
mrgrain, line 16: I'm removing this debug. This only happens of the file exists but cannot be read, which seems super unliekly. The helper is called from only two places. If someone feels strongly about it, I can implement this in the call-sites.

packages/aws-cdk/lib/notices.ts
```diff
@@ -485,13 +488,18 @@ export interface NoticeDataSource {
...
-  fetch(): Promise<Notice[]> {
+  async fetch(): Promise<Notice[]> {
     const timeout = 3000;
-    return new Promise((resolve, reject) => {
+
+    const options: RequestOptions = {
+      agent: await new ProxyAgentProvider(this.ioHelper).create(this.options),
+    };
+
+    const notices = await new Promise<Notice[]>((resolve, reject) => {
```
mrgrain, line 502: bit of a rejig here to allow sending a message after the promise resolved

packages/aws-cdk/test/api/_helpers/hotswap-test-setup.ts
```diff
@@ -145,6 +145,6 @@ export class HotswapMockSdkProvider extends MockSdkProvider {
     hotswapPropertyOverrides?: HotswapPropertyOverrides,
   ): Promise<SuccessfulDeployStackResult | undefined> {
     let hotswapProps = hotswapPropertyOverrides || new HotswapPropertyOverrides();
-    return deployments.tryHotswapDeployment(this, asIoHelper(ioHost, 'deploy'), assetParams, currentCfnStack, stackArtifact, hotswapMode, hotswapProps);
+    return deployments.tryHotswapDeployment(this, asIoHelper(ioHost, 'deploy'), assetParams, currentCfnStack, stackArtifact, hotswapMode as any, hotswapProps);
```
mrgrain, line 148: this was a type error before. `tryHotswapDeployment` only takes two of the 3 enum values and in the CLI code this is enforced. Tests also don't call the method with the third value.

packages/aws-cdk/lib/api/aws-auth/sdk.ts
```diff
@@ -555,14 +555,14 @@ export interface IStepFunctionsClient {
  */
 @traceMemberMethods
 export class SDK {
-  private static readonly accountCache = new AccountAccessKeyCache();
```
mrgrain, line 558: this doesn't need to be a singleton. The class is always reading/writing from a file anywhere and does not cache values. If we ever wanted to, we can make the cache a singleton but keep the class as an instance per SDK.

packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
```diff
@@ -109,8 +115,8 @@ export class AwsCliCompatible {
       : nodeProviderChain;
   }
 
-  public static requestHandlerBuilder(options: SdkHttpOptions = {}): NodeHttpHandlerOptions {
-    const agent = this.proxyAgent(options);
+  public async requestHandlerBuilder(options: SdkHttpOptions = {}): Promise<NodeHttpHandlerOptions> {
+    const agent = await new ProxyAgentProvider(this.ioHelper).create(options);
```
otaviomacedo, line 119: Since we are changing these things, it would be better to pass then agent as a constructor argument. Let the SDK Provider, which is the factory class, to also instantiate this.

packages/aws-cdk/lib/api/aws-auth/sdk.ts
otaviomacedo: It could also be made a constructor argument. Same point as for the proxy agent.

packages/aws-cdk/lib/api/aws-auth/account-cache.ts
```diff
@@ -16,13 +15,24 @@ export class AccountAccessKeyCache {
    */
   public static readonly MAX_ENTRIES = 1000;
 
+  /**
+   * The default path used for the accounts access key cache
+   */
+  public static get DEFAULT_PATH(): string {
+    // needs to be a getter because cdkCacheDir can be set via env variable and might change
```
rix0rrr, line 22: ...during tests?

packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
rix0rrr: It wasn't an abstraction, it was a set of factory functions grouped in a namespace 😆

packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
```diff
@@ -63,7 +69,7 @@ export class AwsCliCompatible {
       return makeCachingProvider(fromIni({
         profile: options.profile,
         ignoreCache: true,
-        mfaCodeProvider: tokenCodeFn,
+        mfaCodeProvider: makeTokenCodeFn(this.ioHelper),
```
rix0rrr, line 72: `tokenCodeFn` could probably be a member function, then it would have access to `this.ioHelper` already.

packages/aws-cdk/lib/api/aws-auth/sdk-provider.ts
```diff
@@ -273,19 +282,19 @@ export class SdkProvider {
...
       try {
-        return await new SDK(this.defaultCredentialProvider, this.defaultRegion, this.requestHandler, this.logger).currentAccount();
+        return await new SDK(this.defaultCredentialProvider, this.defaultRegion, this.requestHandler, this.ioHelper, this.logger).currentAccount();
       } catch (e: any) {
         // Treat 'ExpiredToken' specially. This is a common situation that people may find themselves in, and
         // they are complaining about if we fail 'cdk synth' on them. We loudly complain in order to show that
         // the current situation is probably undesirable, but we don't fail.
         if (e.name === 'ExpiredToken') {
-          warning(
+          await this.ioHelper.notify(IO.DEFAULT_SDK_WARN.msg(
```
rix0rrr, line 291: I think I might quibble with this naming.

This is not actually a warning "from" the SDK, it's a warning from our SDK wrapper.

packages/aws-cdk/lib/notices.ts
rix0rrr: Could just have done a `.then()` at the end 🤣

packages/aws-cdk/lib/notices.ts
```diff
@@ -485,13 +488,18 @@ export interface NoticeDataSource {
 export class WebsiteNoticeDataSource implements NoticeDataSource {
   private readonly options: SdkHttpOptions;
 
-  constructor(private readonly ioMessages: IoDefaultMessages, options: SdkHttpOptions = {}) {
+  constructor(private readonly ioHelper: IoHelper, options: SdkHttpOptions = {}) {
```
rix0rrr, line 491: Looks like you didn't really need to change this? You didn't use `this.ioHelper` anywhere new and the message emitted is the same?

packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
```diff
@@ -23,16 +23,22 @@ const DEFAULT_TIMEOUT = 300000;
...
+    this.ioHelper = ioHelper;
+  }
+
   /**
    * Build an AWS CLI-compatible credential chain provider
    *
    * The credential chain returned by this function is always caching.
    */
-  public static async credentialChainBuilder(
+  public async credentialChainBuilder(
```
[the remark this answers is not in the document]
mrgrain: We chatted about this and this class seems to be only used by a very small number of mostly outdated projects.

packages/aws-cdk/lib/api/aws-auth/account-cache.ts
mrgrain: during tests, but in `toolkit-lib` it could be at any point of the code execution.

packages/aws-cdk/lib/api/aws-auth/credential-plugins.ts
mrgrain: We chatted about this and this class seems to be only used by a very small number of mostly outdated projects.

packages/aws-cdk/lib/notices.ts
mrgrain: Happy to change it, I don't mind either way. I've received feedback in the past that mixing promise styles is confusing.

packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts
mrgrain: I tried, but it has a bunch of knock-on effects that I don't want to deal with right now.

packages/aws-cdk/lib/api/aws-auth/sdk.ts
mrgrain: > It could also be made a constructor argument. Same point as for the proxy agent.

True. There is opportunity for refactoring. Right now I want to keep the changes to the minimum to get this in.

packages/aws-cdk/lib/api/aws-auth/sdk-provider.ts
mrgrain: I think my argument is that these are all prefixed `CDK_SDK` and we don't claim them to be from the SDK directly. Is there an actual benefit in separating these out?

iliapolo (approved):

rix0rrr (approved):

PR #300, opened 2025-03-31
feat(docs): add api-extractor config

iankhou:
Allows us to invoke `api-extractor` to get an outputted JSON file produced from source code and TSDocs.

Ref: https://api-extractor.com/pages/setup/invoking/#4-running-the-tool

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1178,3 +1178,65 @@ jobs:
             echo "::error::Failed to upload docs artifact"
             exit 1
           fi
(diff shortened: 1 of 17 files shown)


rix0rrr (commented):
How/where are you planning to run API Extractor? I'd expect the script and the config to live side-by-side (unless there's a very good reason not to).

Also, if it must live here: I think I'd prefer all of our JSON files that are actually config to go through projen. So create a new [projen.JsonFile](https://projen.io/docs/api/projen/#jsonfile-) in `.projenrc.ts` and put the config there I think.

PR #301, opened 2025-03-31
refactor(toolkit-lib): diff action uses helper methods

kaizencc:
refactors the diff action to use helper methods and in general cleans up the code a lot. introduces an internal concept of `TemplateInfo` that standardizes the to-be-diffed information, so there's not a bunch of repetitive code

existing tests continue to succeed. 

---


packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
mrgrain, line 1: this seems to all be private helpers, so they should go into `diff/private/helpers.ts` or something

packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
```diff
@@ -1,4 +1,4 @@
-import type { StackSelector } from '../../api/cloud-assembly';
+import type { StackSelector } from '@aws-cdk/tmp-toolkit-helpers';
```
mrgrain, line 1: or whatever the path is
```suggestion
import type { StackSelector } from '../../@aws-cdk/tmp-toolkit-helpers';
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -268,122 +266,40 @@ export class Toolkit extends CloudAssemblySourceBuilder implements AsyncDisposab
...
     const contextLines = options.contextLines || 3;
-    const diffMethod = options.method ?? DiffMethod.ChangeSet();
 
     let diffs = 0;
     let formattedSecurityDiff = '';
     let formattedStackDiff = '';
 
-    if (diffMethod.method === 'local-file') {
-      const methodOptions = diffMethod.options as LocalFileDiffOptions;
+    const templateInfos = await makeTemplateInfos(ioHelper, stacks, deployments, await this.sdkProvider('diff'), options);
```
mrgrain, line 274: So this is preparing the diff? Current names read unclear to me. Maybe something like this is clearer:

```suggestion
    const diffInput = await prepareDiff(ioHelper, stacks, deployments, await this.sdkProvider('diff'), options);
```

kaizencc: i like the prepareDiff name but also after my latest refactor i think the type of `TemplateInfo` now makes more sense -- since its only getting passed into `DiffFormatter`. lemme know if you disagree

packages/@aws-cdk/tmp-toolkit-helpers/test/api/diff/diff.test.ts
```diff
@@ -59,20 +59,18 @@ describe('formatStackDiff', () => {
...
-        template: {},
-        templateFile: 'template.json',
+      templateInfo: {
+        oldTemplate: {},
+        newTemplate: {
+          template: {},
+          templateFile: 'template.json',
+          stackName: 'test-stack',
+          findMetadataByType: () => [],
+        } as any,
```
mrgrain, line 69: there are proper helpers for this around

mrgrain: why not `mockNewTemplate`?

packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
```diff
@@ -1,4 +1,4 @@
-import type { StackSelector } from '../../api/cloud-assembly';
+import type { StackSelector } from '../../../lib/api/shared-public';
```
mrgrain, line 1: Any reason for this?

mrgrain (approved):

packages/@aws-cdk/tmp-toolkit-helpers/test/api/diff/diff.test.ts
kaizencc: oops

PR #303, opened 2025-04-01
feat(docs): api-extractor changes and gh workflow update

### LINKED ISSUE #367: Upload api-extractor output for static site generation
### Describe the feature

We want to use `api-extractor` to generate a JSON, which encapsulates information about interfaces, classes, and enums in the AWS CDK Programmatic Toolkit Library

We currently upload TypeDoc artifacts in a similar way. The difference is that the TypeDoc upload is a static-site artifact, which the `api-extractor` artifact is what will be used to generate that artifact.

### Use Case

Enables us to use an api-extractor JSON as an input to a static-site generator.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.1007.0

Not relevant

iankhou:
Fixes #367 

Uses projen to generate the api-extractor config.

Adds `api-extractor` JSON output and .md files to the S3 bucket for docs during release.

Does not affect the existing TypeDoc site workflow.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/api-extractor.json
+++ b/packages/@aws-cdk/toolkit-lib/api-extractor.json
@@ -1,0 +1,36 @@
+{
+  "projectFolder": ".",
+  "mainEntryPointFilePath": "<projectFolder>/lib/index.d.ts",
+  "bundledPackages": [],
+  "apiReport": {
+    "enabled": false
+  },
+  "docModel": {
+    "enabled": true,
+    "apiJsonFilePath": "./dist/<unscopedPackageName>.api.json",
+    "projectFolderUrl": "https://github.com/aws/aws-cdk-cli/tree/main/packages/%40aws-cdk/toolkit-lib"
+  },
+  "dtsRollup": {
+    "enabled": false
+  },
+  "tsdocMetadata": {
+    "enabled": false
+  },
+  "messages": {
+    "compilerMessageReporting": {
+      "default": {
+        "logLevel": "warning"
+      }
+    },
+    "extractorMessageReporting": {
+      "default": {
+        "logLevel": "warning"
+      }
+    },
+    "tsdocMessageReporting": {
+      "default": {
+        "logLevel": "warning"
+      }
+    }
+  }
+}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1116,8 +1116,8 @@ jobs:
         run: |-
           aws ssm put-parameter --name "/published/cdk/cli-npm/version" --type "String" --value "${{ steps.aws-cdk-version.outputs.version }}" --overwrite
           aws ssm put-parameter --name "/published/cdk/cli-npm/timestamp" --type "String" --value "$(date +%s)" --overwrite
-  aws-cdk-toolkit-lib_release_docs:
-    name: "@aws-cdk/toolkit-lib: Publish docs to S3"
+  aws-cdk-toolkit-lib_release_typedoc:
+    name: "@aws-cdk/toolkit-lib: Publish typedoc to S3"
     needs: aws-cdk-toolkit-lib_release_npm
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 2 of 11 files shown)


projenrc/api-extractor-docs-publishing.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+    this.github = gh;
+
+    this.props = props;
+
+    // Add a task to run api-extractor and zip the output
+    const apiExtractorDocsTask = project.addTask('api-extractor-docs', {
+      exec: [
+        // Ensure API Extractor is installed
+        'npm install -g @microsoft/api-extractor || true',
```
mrgrain, line 40: Don't use a global install, just add it as devDep

projenrc/api-extractor-docs-publishing.ts
```diff
@@ -0,0 +1,150 @@
...
+        // Run api-extractor to generate the API model
+        // Use || true to ensure the task continues even if api-extractor reports failures
```
mrgrain, line 42: This seems counter intuitive? Why continue on failure?

projenrc/api-extractor-docs-publishing.ts
```diff
@@ -0,0 +1,150 @@
...
+        'api-extractor run --local --diagnostics || true',
+        // Create a directory for the API model
+        'mkdir -p dist/api-extractor-docs/cdk/api/toolkit-lib',
+        // Copy the API model to the directory (with error handling)
+        'if [ -f dist/toolkit-lib.api.json ]; then cp dist/toolkit-lib.api.json dist/api-extractor-docs/cdk/api/toolkit-lib/; else echo "Warning: API JSON file not found"; fi',
+        // Add version file
+        '(cat dist/version.txt || echo "latest") > dist/api-extractor-docs/cdk/api/toolkit-lib/VERSION',
+        // Find and copy all markdown files (excluding node_modules)
```
mrgrain, line 50: I reckon we can limit this to all files in the `docs` directory and `README.md`

You will probably want _all_ files because we might have images.

packages/@aws-cdk/toolkit-lib/api-extractor.json
```diff
@@ -0,0 +1,36 @@
...
+  "projectFolder": ".",
+  "mainEntryPointFilePath": "<projectFolder>/lib/index.d.ts",
+  "bundledPackages": [],
+  "apiReport": {
+    "enabled": false
+  },
+  "docModel": {
+    "enabled": true,
+    "apiJsonFilePath": "./dist/<unscopedPackageName>.api.json",
+    "projectFolderUrl": "https://github.com/aws/aws-cdk-cli"
```
mrgrain, line 11: not this?
```suggestion
    "projectFolderUrl": "https://github.com/aws/aws-cdk-cli/tree/main/packages/%40aws-cdk/toolkit-lib"
```

mrgrain (dismissed):
left some comments

projenrc/api-extractor-docs-publishing.ts
iankhou: `api-extractor` will still output a file when there are certain failures. Will work on fixing the underlying issue, but I did this to temporarily unblock it.

projenrc/api-extractor-docs-publishing.ts
iankhou: Got it, that makes sense to me. I was thinking we might have .md files in subdirectories, but if that isn't the case, then we won't need anything outside of docs and README.md. And will include all file types.

projenrc/api-extractor-docs-publishing.ts
iankhou: Maybe a better way to handle this error is to check whether the file was outputted?

packages/@aws-cdk/toolkit-lib/api-extractor.json
iankhou: Will change it to this.

rix0rrr (approved):

PR #304, opened 2025-04-01
fix(cli): diff always prints stack name

### LINKED ISSUE #302: Stack names missing in output
### Describe the bug

CDK CLI version 2.1006.0 doesn't correctly output stack names.

In diff output, they are completely missing while during deployment it says `Stack undefined` in the headline before prompting for security-related changes.

### Regression Issue

2.1005.0

### Expected Behavior

Show stack names.

CDK <= 2.1005.0 gave the following in my project for `cdk diff --no-change-set`:
```
$ cdk diff --no-change-set
Stack Base
There were no differences
Stack CfSupportA
There were no differences
Stack CfSupportB
There were no differences
Stack IngestCore
There were no differences
Stack IngestRegional-eu-west-1
There were no differences
Stack IngestRegional-eu-central-1
There were no differences
Stack IngestCoordinatorMaster
There were no differences
Stack IngestDashboard
There were no differences
Stack Api
There were no differences
Stack Rms
There were no differences
Stack Dashboard
There were no differences

✨  Number of stacks with differences: 0
```

### Current Behavior

Missing stack names.

CDK 2.1006.0 gives this:
```
$ cdk diff --no-change-set
There were no differences

There were no differences

There were no differences

There were no differences

There were no differences

There were no differences

There were no differences

There were no differences

There were no differences

There were no differences

There were no differences


✨  Number of stacks with differences: 0
```

### Reproduction Steps

Also happens on an empty, freshly initialized project.

- `cdk init app -l typescript` (at least that's what I tried - may affect other languages, too)
- `cdk deploy CdkTestStack` (empty stack with just a single SSM parameter)
- `cdk diff --no-change-set`

Output:
```
There were no differences


✨  Number of stacks with differences: 0
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.1006.0 (build a3b9762)

### Framework Version

_No response_

### Node.js Version

22.11.0

### OS

Windows 11 14H2

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

kaizencc:
Fixes #302 

In https://github.com/aws/aws-cdk-cli/pull/264, we capture the output of the Formatter into a stream, and that is then converted to a string. This diff only gets printed if the diff isn't empty, but by some quirk `formatStackDiff` used to print the stack name _outside_ of the stream, thus creating the (correct) behavior. #264 made sure that `format` doesn't print, so we need to make sure that the consumer of `formatStackDiff` prints the relevant info.

This PR makes sure that the stack name is returned as the formatted diff even if the actual diff is empty. We will print the right information when we print the formatted diff now.

I also made similar modifications to `formatSecurityDiff`, as we should move away from `format` methods printing anything at all.

BEFORE (cdk 2.1006.0):

Screenshot 2025-04-02 at 9 04 04 AM

AFTER (my local cdk):

Screenshot 2025-04-02 at 9 03 11 AM

---


packages/aws-cdk/test/commands/diff.test.ts
```diff
@@ -906,7 +906,10 @@ Resources
 
 Stack newGrandChild
 Resources
-[+] AWS::Something SomeResource`);
+[+] AWS::Something SomeResource
+
+Stack UnChangedChild
+There were no differences`);
```
kaizencc, line 912: this was missed in https://github.com/aws/aws-cdk-cli/pull/264; it should have not been removed, and would have alerted us of the issue.

it also shows an additional thing that this PR fixes. previously, the comment `There were no differences` was printed separately from the diff (on console, it renders the same). now it is included in the formatted diff itself

mrgrain (approved):

PR #306, opened 2025-04-02
feat(toolkit-lib): default to using a non interactive IoHost

### LINKED ISSUE #157: @aws-cdk/toolkit-lib: unable to destroy stack without custom ioHost
### Describe the bug

As a programmatic user of `@aws-cdk/toolkit-lib` I'd like to be able to destroy stacks without explicit operator approval. At the moment this requires me implementing a custom `ioHost`, which can be a lot of work.

### Regression Issue

_No response_

### Expected Behavior

While I understand and appreciate the current `requestResponse` system for operations that require some kind of intervention, I'd like an easy way to tell the toolkit that I want to destroy stacks without any approval.

It's already possible to turn off approvals for stack creation, and the legacy `@aws-cdk/cli-lib-alpha` also allowed to destroy stacks without approvals.

### Current Behavior

Currently the only way to destroy stacks in a completely unattended way is to implement a passthrough `ioHost`, like:

```js
ioHost: {
  async notify(msg) {
    console.log(msg.message);
  },
  async requestResponse(msg) {
    console.log(msg.message);
    return msg.defaultResponse;
  },
},
```

This works, but requires me to now also take care of messaging and handle different levels - something that is done automatically by the default `ioHost`.

### Reproduction Steps

N/A

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.1000.2

### Framework Version

_No response_

### Node.js Version

v22.x

### OS

N/A

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

### LINKED ISSUE #289: NonInteractiveIoHost
By default, the Toolkit from Toolkit Lib currently uses the CliIoHost. However we will need a separate Non interactive copy of it to prevent a dependency on the CLI.

Also it makes more sense for the Toolkit Lib to not include any interaction by default.

mrgrain:
Replaces the current default `IoHost` which is the same as the one used by the CLI with a strictly non-interactive version. This is more in accordance with the goals of a library providing actions.

Fixes #157
Closes #289 

BREAKING CHANGE: This change updates the default `IoHost` implementation used by `Toolkit` to a version that is strictly non-interactive, i.e. there is no expectation anymore that users will respond to command-line prompts. To restore previous behavior, you will can provide a custom `IoHost` implementation to your `Toolkit` instance. You may consider extending the new `NonInteractiveIoHost` class with desired interactive prompts.

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/non-interactive-io-host.ts
```diff
@@ -0,0 +1,174 @@
...
+   * In CI mode, all non-error output goes to stdout instead of stderr.
+   * Set to false in the IoHost constructor it will be overwritten if the CLI CI argument is passed
+   *
+   * @default - determined from the environment, specifically based on `process.env.CI`
+   */
+  readonly isCI?: boolean;
+}
+
+/**
+ * A simple IO host for a non interactive CLI that writes messages to the console.
```
rix0rrr, line 39: ...and returns the default answer to all queries?

Or fails all queries?

rix0rrr (approved):

packages/@aws-cdk/toolkit-lib/lib/toolkit/non-interactive-io-host.ts
mrgrain: >...and returns the default answer to all queries?

This. There's not really a concept of failing. For booleans you might consider the inverse a failure or it could just throw.

mrgrain:
exempted integ test because a previous run passed and the latest change just fixed a comment

PR #318, opened 2025-04-03
feat(toolkit): add a return type for toolkit.destroy()

### LINKED ISSUE #33190: Toolkit Action: Typed return for destroy
Currently the destroy action isn't returning anything. It should return some useful information to the user. Make a suggestion what we should return here and discuss with the team.

Basics are stack arn, stack name etc. Could include skipped resources.

rix0rrr:
`toolkit.destroy()` now returns information about the stacks it deployed.

Closes https://github.com/aws/aws-cdk/issues/33190

---


otaviomacedo (approved):

packages/@aws-cdk/toolkit-lib/lib/toolkit/types.ts
```diff
@@ -8,6 +8,167 @@ export interface DeployResult {
...
+  /**
+   * List of stacks deployed by this operation
+   */
+  readonly stacks: DeployedStack[];
+}
+
+/**
+ * Information about a deployed stack
+ */
+export interface DeployedStack extends PhysicalStack {
```
iankhou, line 85: Looks like some stuff got repeated here, maybe because of an automerge issue?

iankhou (dismissed):

PR #323, opened 2025-04-03
refactor(cli): inject CLI version as a parameter to Notices

mrgrain:
`Notices` will eventually serve both toolkit-lib and cli. Currently the class has a hard-coded "cli version" - but really the value is reading the version from whatever the current package is. Once we start using this file in another package, it will be wrong. For now, let's just remove the hard dependency on cli version inside the class. In future it will be extended.

- Remove direct dependency on versionNumber import in Notices class
- Add cliVersion as a required parameter in NoticesProps interface
- Store cliVersion as a class property and use it in display method
- Improve testability by removing the need to mock version functions

---

### DIFF
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/util/package-info.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/util/package-info.ts
@@ -1,0 +1,20 @@
+import * as path from 'path';
+import { bundledPackageRootDir } from './directories';
+
+export function displayVersion() {
+  return `${versionNumber()} (build ${commit()})`;
+}
+
+export function isDeveloperBuild(): boolean {
+  return versionNumber() === '0.0.0';
+}
+
+export function versionNumber(): string {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  return require(path.join(bundledPackageRootDir(__dirname), 'package.json')).version.replace(/\+[0-9a-f]+$/, '');
+}
+
+function commit(): string {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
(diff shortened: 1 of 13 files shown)


packages/aws-cdk/lib/context-providers/index.ts
```diff
@@ -98,7 +97,6 @@ export async function provideContextValues(
       }
     }
 
-    ioHelper = ioHelper ?? CliIoHost.instance().asIoHelper();
```
mrgrain, line 101: this was left over. `ioHelper` is a required argument, so the second part never happens anyway

packages/@aws-cdk/tmp-toolkit-helpers/src/util/package-info.ts
kaizencc, line 1: these functions look like they're copied from `aws-cdk/cli/version.ts`. do you mean to copy them over without deleting their duplicates in aws-cdk?

mrgrain: yes, they will be removed eventually

kaizencc:
> Remove direct dependency on versionNumber import in Notices class
why? 
> Add cliVersion as a required parameter in NoticesProps interface
why?

kaizencc (approved):
approved but would like the PR description to be updated to explain why these changes are necessary

mrgrain:
> > Remove direct dependency on versionNumber import in Notices class
> 
> why?
> 
> > Add cliVersion as a required parameter in NoticesProps interface
> 
> why?

Notices will serve both toolkit-lib and cli. having a hard coded "cli version" that is actually reading the version from the current package will be wrong. In future this will be extended, for now I'm just removing the hard dependency.

PR #335, opened 2025-04-04
fix(toolkit): keep locks

rix0rrr:
Associate a read lock with the lifetime of a `CloudAssembly`. The write lock that we take during synthesis gets converted into a read lock after synthesis is done.

The presence of long-lasting read locks prevents the cloud assembly direcctory being trampled on by a new synthesis while it is still being read from.

In order to keep the return types somewhat compatible (`cxapi.CloudAssembly`) we now return a new type that represents that a `cxapi.CloudAssembly` may or may not have a read lock associated with it.

This may not be good enough because it doesn't give a good way of disposing of the lock; in order to be able to use a `using` statement we need an interface that *definitely* has `Symbol.asyncDispose`, not one that *maybe* has it.

---


packages/@aws-cdk/tmp-toolkit-helpers/src/api/rwlock.d.ts
mrgrain, line 1: d.ts???

Issue #141, opened 2025-02-25
Add a `refactor` command


otaviomacedo:
Add a new command, `refactor`. At phase 1, it should only work if passed the flag `--dry-run`, that will only show the mappings, and stop.

Issue #135, opened 2025-02-25
Implement ambiguity detection


otaviomacedo:
In case it's not possible to establish a 1:1 mapping between resources, show the ambiguous pairs, in the following format:

```
    ┌───┬──────────────────────┐
    │   │ Resource             │
    ├───┼──────────────────────┤
    │ - │ Stack.Queue1A4198146 │
    │   │ Stack.Queue2B0BA5D32 │
    ├───┼──────────────────────┤
    │ + │ Stack.Queue3C7606C37 │
    │   │ Stack.Queue4D681F510 │
    └───┴──────────────────────┘

```

Issue #134, opened 2025-02-25
Implement the display of mappings


otaviomacedo:
Display a table to the user with the [computed mapping][1]. Example:

```        
    ┌───────────────────────────────┬───────────────────────────────┬──────────────────────────┐
    │ Resource Type                 │ Old Logical ID                │ New Logical ID           │
    ├───────────────────────────────┼───────────────────────────────┼──────────────────────────┤
    │ AWS::S3::Bucket               │ MyStack.Queue1A4198146        │ Web.Bucket843D52FF       │
    ├───────────────────────────────┼───────────────────────────────┼──────────────────────────┤
    │ AWS::CloudFront::Distribution │ MyStack.DistributionE3BB089E  │ Web.Distribution7142E1F1 │
    ├───────────────────────────────┼───────────────────────────────┼──────────────────────────┤
    │ AWS::Lambda::Function         │ MyStack.FunctionA5EA2BD8      │ Service.Function8F0BB69B │
    └───────────────────────────────┴───────────────────────────────┴──────────────────────────┘

```

[1]: https://github.com/aws/aws-cdk-cli/issues/133

Issue #133, opened 2025-02-25
Implement the computation of mappings


otaviomacedo:
Using the [digest function][1], implement the computation of the mapping between resources before and after the refactoring.

More formally:

Let $G$ and $H$ be two resource graphs. Implement a function that generates a list of resource pairs $(g, h)$ with $g \in G$, $h \in H$, such that $x \equiv y$.

[1]: https://github.com/aws/aws-cdk-cli/issues/132

PR #342, opened 2025-04-08
feat: refactor (dry-run mode only)

### LINKED ISSUE #132: Implement basic resource equivalence
To detect which resources should be refactored, we need to indentify which
resources have only changed their location, but have remained "the same", in
some sense. This can be made precise by defining an [equivalence relation] on
the set of resources.

Before that, let's define a digest function, `d`:

```
d(resource) = hash(type + properties + dependencies.map(d))
```

where `hash` is a cryptographic hash function. In other words, the hash 
is computed from its type, its own properties (that is, excluding properties 
that refer to other resources), and the digests of each of its dependencies.

The digest of a resource, defined recursively this way, remains stable even if
one or more of its dependencies gets renamed. Since the resources in a
CloudFormation template form an acyclic graph, this function is well-defined.

The equivalence relation then follows directly: two resources `r1` and `r2`
are equivalent if `d(r1) = d(r2)`.

[equivalence relation]: https://en.wikipedia.org/wiki/Equivalence_relation

otaviomacedo:
Add a `refactor` command. For now, it only works in dry run mode (`-dry-run` in the command line, or `dryRun: true` to the toolkit). It computes the mappings based on the difference between the deployed stacks and the cloud assembly stacks, and shows them in a table. Example:

```
$ cdk refactor --dry-run --unstable=refactor

The following resources were moved or renamed:
┌───────────────────────┬────────────────────────────────────────┬───────────────────────────────────────┐
│ Resource Type         │ Old Construct Path                     │ New Construct Path                    │
├───────────────────────┼────────────────────────────────────────┼───────────────────────────────────────┤
│ AWS::IAM::Role        │ Consumer/Function/ServiceRole/Resource │ Consumer/NewName/ServiceRole/Resource │
├───────────────────────┼────────────────────────────────────────┼───────────────────────────────────────┤
│ AWS::Lambda::Function │ Consumer/Function/Resource             │ Consumer/NewName/Resource             │
└───────────────────────┴────────────────────────────────────────┴───────────────────────────────────────┘
```

Note that we are launching this feature under unstable mode, which requires the user to acknowledge that by passing the `--unstable=refactor` flag.

Closes https://github.com/aws/aws-cdk-cli/issues/132, https://github.com/aws/aws-cdk-cli/issues/133, https://github.com/aws/aws-cdk-cli/issues/141, https://github.com/aws/aws-cdk-cli/issues/134 and https://github.com/aws/aws-cdk-cli/issues/135.

---


packages/@aws-cdk/tmp-toolkit-helpers/src/api/io/private/messages.ts
```diff
@@ -349,6 +350,18 @@ export const IO = {
...
+    code: 'CDK_TOOLKIT_I8900',
+    description: 'Refactor result',
+    interface: 'RefactorResult',
+  }),
+
+  CDK_TOOLKIT_W8010: make.warn({
+    code: 'CDK_TOOLKIT_W8010',
+    description: 'Refactor execution not yet supported',
+  }),
+
```
otaviomacedo, line 364: Not sure if I got the pattern right, here. Someone please check me on this.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/streams.ts
```diff
@@ -0,0 +1,23 @@
+import { Writable } from 'node:stream';
```
otaviomacedo, line 1: This was moved from the `diff-formatter.ts` file.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+  Resources?: {
+    [logicalId: string]: {
+      Type: string;
+      Properties?: any;
+      Metadata?: Record<string, any>;
+    };
+  };
+}
+
+export interface BasicStack {
```
otaviomacedo, line 25: Suggestions for a better name are welcome :)

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -865,6 +910,47 @@ export class Toolkit extends CloudAssemblySourceBuilder {
     return ret;
   }
 
+  /**
+   * Refactor Action. Moves resources from one location (stack + logical ID) to another.
+   */
+  public async refactor(cx: ICloudAssemblySource, options: RefactorOptions = {}): Promise<void> {
```
rix0rrr, line 916: Refactor should probably return some useful structure as well, shouldn't it?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -865,6 +910,47 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+  }
+
+  private async _refactor(assembly: StackAssembly, ioHelper: IoHelper, options: RefactorOptions = {}): Promise<void> {
+    if (!options.dryRun) {
+      throw new ToolkitError('Refactor is not available yet. Too see the proposed changes, use the --dry-run flag.');
+    }
+
+    const strategy = options.stacks?.strategy ?? StackSelectionStrategy.ALL_STACKS;
+    if (strategy !== StackSelectionStrategy.ALL_STACKS) {
+      await ioHelper.notify(IO.CDK_TOOLKIT_W8010.msg(
```
rix0rrr, line 929: Shouldn't this just be an error?

If you want to convert user input into a warning and proceeding anyway, then do the conversion in the CLI?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1210,6 +1214,31 @@ export class CdkToolkit {
...
+      return 1;
+    }
+
+    if (options.selector.patterns.length > 0) {
+      warning('Refactor does not yet support stack selection. Proceeding with the default behavior (considering all stacks).');
+    }
+
+    const stacks = await this.selectStacksForList([]);
+    try {
+      const mappings = await detectRefactorMappings(stacks.stackArtifacts, this.props.sdkProvider);
```
rix0rrr, line 1229: Hm. I thought the CLI would use the toolkit code here, but I see we haven't gotten that far yet.

rix0rrr (commented):
Surface level stuff, still need to look at the meat & potatoes

packages/aws-cdk/lib/cli/cdk-toolkit.ts
otaviomacedo: From what I understand, the idea is that both the CLI and the toolkit library share common code, but the CLI doesn't call the toolkit library. Not yet, at least.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
otaviomacedo: Possibly. I'm confused as to what should get notified to the `IoHost`, and what should be returned.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+  formatAmbiguousMappings as fmtAmbiguousMappings,
+} from '@aws-cdk/cloudformation-diff';
+import type * as cxapi from '@aws-cdk/cx-api';
+import type { StackSummary } from '@aws-sdk/client-cloudformation';
+import { deserializeStructure } from '../../util';
+import type { SdkProvider } from '../aws-auth';
+import { Mode } from '../plugin';
+import { StringWriteStream } from '../streams';
+
+interface CloudFormationTemplate {
```
rix0rrr, line 15: Surely we have this definition somewhere else already?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+ * but is richer, since it has a reference to the stack object, rather than
+ * merely the stack name.
+ */
+export class ResourceLocation {
+  constructor(readonly stack: BasicStack, readonly logicalResourceId?: string) {
+  }
+
+  public toPath(): string {
+    const stack = this.stack;
+    const resource = stack.template.Resources?.[this.logicalResourceId ?? ''];
```
rix0rrr, line 56: This doesn't make any sense, does it? Should `logicalResourceId` really be optional?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    return {
+      // the type is the same in both source and destination,
+      // so we can use either one
+      type: this.getType(this.source),
+      sourcePath: this.source.toPath(),
+      destinationPath: this.destination.toPath(),
+    };
+  }
+
+  private getType(location: ResourceLocation): string {
```
rix0rrr, line 85: This method doesn't access `this` anywhere, so it might as well be a static helper function.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    const resource = location.stack.template.Resources?.[location.logicalResourceId ?? ''];
+    return resource?.Type ?? 'Unknown';
+  }
+}
+
+function buildRecord(entries: [string, ResourceLocation][]): Record<string, ResourceLocation[]> {
```
rix0rrr, line 91:
```suggestion
function groupByKey<A>(entries: [string, A][]): Record<string, A[]> {
```

?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    }
+  }
+
+  return result;
+}
+
+/**
+ * Computes a list of pairs [digest, location] for each resource in the stack.
+ */
+function index(stack: BasicStack): [string, ResourceLocation][] {
```
rix0rrr, line 173:
```suggestion
function resourceDigests(stack: BasicStack): [string, ResourceLocation][] {
```

?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+  if (ambiguousPairs.length > 0) {
+    throw new AmbiguityError(ambiguousPairs);
+  }
+
+  return Object.values(pairs)
+    .filter(([pre, post]) => pre.length === 1 && post.length === 1 && !equalLocations(pre[0], post[0]))
+    .map(([pre, post]) => new ResourceMapping(pre[0], post[0]));
+}
+
+function removeCommonResources(
```
rix0rrr, line 124:
```suggestion
function removeUnmovedResources(
```

?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    .filter(([pre, post]) => pre.length > 1 || post.length > 1);
+
+  if (ambiguousPairs.length > 0) {
+    throw new AmbiguityError(ambiguousPairs);
+  }
+
+  return Object.values(pairs)
+    .filter(([pre, post]) => pre.length === 1 && post.length === 1 && !equalLocations(pre[0], post[0]))
+    .map(([pre, post]) => new ResourceMapping(pre[0], post[0]));
+}
```
rix0rrr, line 122: It's unfortunate that the return type is:

* Either everything is unambiguous and here is the answer; or
* There is at least one ambiguity, so YOU GET NOTHING

Can we not return both the suggested and ambiguous mappings, and make the decision to throw somewhere else?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+  for (const [_, [before, after]] of stackGroups) {
+    result.push(...computeMappings(before, after));
+  }
+  return result;
+}
+
+async function getDeployedStacks(sdkProvider: SdkProvider, environment: cxapi.Environment): Promise<BasicStack[]> {
+  const cfn = (await sdkProvider.forEnvironment(environment, Mode.ForReading)).sdk.cloudFormation();
+  const summaries: StackSummary[] = [];
+  await paginateSdkCall(async (nextToken) => {
```
rix0rrr, line 219: Doesn't the SDK abstraction we return paginate already? Did we or did we not do that?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+    }
+  }
+
+  addToHash(obj);
+  return hash.digest('hex');
+}
+
+/**
+ * Removes sub-properties containing Ref or Fn::GetAtt to avoid hashing
+ * references themselves but keeps the property structure.
```
rix0rrr, line 364: Surely though, it matters WHICH object we reference, not just that we reference SOME object?

Wouldn't it be better to replace the logical ID with the digest of the referenced source?

Currently:

```
{ "Function": { "Ref": "MyFunction" } }
-->
{ "Function": { __cloud_ref__: "Ref" } }


{ "Function": { "Fn::GetAtt": ["MyFunction", "Arn"] } }
-->
{ "Function": { __cloud_ref__: "Fn::GetAtt" } }
```

Instead:

```
{ "Function": { "Ref": "MyFunction" } }
-->
{ "Function": { "Ref": "baf399b3cdafe24b4b6e1b0674bb59ae" } }


{ "Function": { "Fn::GetAtt": ["MyFunction", "Arn"] } }
-->
{ "Function": { "Fn::GetAtt": ["baf399b3cdafe24b4b6e1b0674bb59ae", Arn" ] } }
```

?

rix0rrr: EDIT: oh I see, you mix the dependencies in elsewhere

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
```diff
@@ -0,0 +1,415 @@
...
+      }
+    }
+  }
+
+  // 4. Compute digests in sorted order
+  const result: Record<string, string> = {};
+  for (const id of order) {
+    const resource = resources[id];
+    const depDigests = Array.from(graph[id]).map((d) => result[d]);
+    const propsWithoutRefs = JSON.stringify(stripReferences(stripConstructPath(resource.Properties)));
```
rix0rrr: What about other top-level CloudFormation attributes? Like CreationPolicy, Condition, etc? Should we not incorporate those as well?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/index.ts
rix0rrr: Seriously, I wonder if we shouldn't invest in a proper CloudFormation Template parser at some point... it kinda bothers me that we do ad-hoc JSON processing all over the place :/ but at every point it's the path of least resistance, and never worth doing anything more... resulting in loads of duplication

rix0rrr (approved):
I have some comments that I'll leave for you to decide what to do with, but I don't want to stand in the way of this, especially since you're going to iterate on it over time anyway.

rix0rrr (commented):
One more thing: in calculating the digest, should we normalize property order and dependency order? 

A reordering of the properties in the bag will affect the hash right now, but it probably shouldn't.

PR #351, opened 2025-04-09
fix(docs): api-extractor uses 'toolkit-lib-api-model' name

### LINKED ISSUE #367: Upload api-extractor output for static site generation

iankhou:
Fixes #367 

### Description
We currently use the `toolkit-lib` directory in our S3 bucket to upload TypeDoc documentation. This change uses the `toolkit-lib-api-model` directory to upload api-extractor documentation.

---


rix0rrr (approved):

rix0rrr (commented):
Well I approved it, but by changing the role SESSION name you're not changing what role you're targeting (unless we have weird IAM policies, which I hope we don't).

But why doesn't the publish role just have permissions to write to the location it needs?

rix0rrr (changes requested):
So actually I retract my approval 😅

iankhou:
> Well I approved it, but by changing the role SESSION name you're not changing what role you're targeting (unless we have weird IAM policies, which I hope we don't).
> 
> But why doesn't the publish role just have permissions to write to the location it needs?

It changes the user, which causes Access Denied:

Previously we had:
```
User: arn:aws:sts::***:assumed-role/aws-cdk-toolkit-lib-docs-publishing-role/s3-api-docs-publishing@aws-cdk-cli
```

with `api-` being added in the user name. This caused the PutObject operation to fail, because that new user doesn't have the correct permissions.

This CR changes it to:

```
User: arn:aws:sts::***:assumed-role/aws-cdk-toolkit-lib-docs-publishing-role/s3-docs-publishing@aws-cdk-cli
```

which matches the user that's used for the TypeDoc zip upload.

rix0rrr (approved):

PR #354, opened 2025-04-10
feat(toolkit-lib): fallBackToTemplate option for diffs

kaizencc:
The primary motivation for this PR is https://github.com/aws/aws-cdk-cli/issues/914 and the ensuing PR https://github.com/aws/aws-cdk/pull/32830. 

When implementing diff in the programmatic toolkit library we included the diff method options, and this PR adds the `fallBackToTemplate` property to the `ChangeSet` mode. The idea is that people can specify `fallBackToTemplate=false` to explicitly fail if we cannot successfully create the changeset. The current behavior is to fallback to the template-only method. 

---


packages/@aws-cdk/tmp-toolkit-helpers/src/api/deployments/cfn-api.ts
```diff
@@ -142,6 +142,14 @@ export type PrepareChangeSetOptions = {
...
   parameters: { [name: string]: string | undefined };
   resourcesToImport?: ResourcesToImport;
+  /**
+   * Enable falling back to template-based diff in case creating the changeset is not possible or results in an error.
+   *
+   * Should be used for stacks containing nested stacks or when change set permissions aren't available.
+   *
+   * @default true
+   */
+  fallBackToTemplate?: boolean;
```
rix0rrr, line 152: This doesn't seem to be added on a type that is only used for the `diff` operation. 

For example, `resourcesToImport` definitely should not be part of the `diff` operation. Can't we place `fallBackToTemplate` elsewhere?

And if it is the case that `uploadBodyParameterAndCreateChangeSet` is also used in `cdk import`, then the error message "we will fall back to a template diff" doesn't make any sense.

Can you categorize the uses of this function a bit better?

rix0rrr (changes requested):

packages/@aws-cdk/tmp-toolkit-helpers/src/api/deployments/cfn-api.ts
kaizencc: right now, `createDiffChangeSet` is the only one that uses `PrepareChangeSetOptions`. and in turn, `createDiffChangeSet` is only used in the diff operation. The error message / IOHost messages have existed previously, and are diff specific. are you asking me to refactor this to be more generic? i'd rather not in this PR.

would it be better if I renamed `fallBackToTemplate` as a more generic `failOnError`?

rix0rrr (approved):

PR #360, opened 2025-04-11
feat(refactor): resources with the same physical ID are considered equivalent

otaviomacedo:
Incorporate the physical ID into the computation of resource digests. This gives the user more flexibility: if a resource with physical ID set is moved and updated at the same time, the algorithm can still detect that it is the same resource.

---

### DIFF
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/digest.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/digest.ts
@@ -7,12 +7,18 @@ import type { CloudFormationTemplate } from './cloudformation';
  *
  * Conceptually, the digest is computed as:
  *
- *     digest(resource) = hash(type + properties + dependencies.map(d))
+ *     d(resource) = hash(type + physicalId)                       , if physicalId is defined
+ *                 = hash(type + properties + dependencies.map(d)) , otherwise
  *
- * where `hash` is a cryptographic hash function. In other words, the digest of a
- * resource is computed from its type, its own properties (that is, excluding
- * properties that refer to other resources), and the digests of each of its
- * dependencies.
+ * where `hash` is a cryptographic hash function. In other words, if a resource has
+ * a physical ID, we use the physical ID plus its type to uniquely identify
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/digest.ts
```diff
@@ -82,10 +89,29 @@ export function computeResourceDigests(template: CloudFormationTemplate): Record
   const result: Record<string, string> = {};
   for (const id of order) {
     const resource = resources[id];
-    const depDigests = Array.from(graph[id]).map((d) => result[d]);
-    const propsWithoutRefs = hashObject(stripReferences(stripConstructPath(resource)));
-    const toHash = resource.Type + propsWithoutRefs + depDigests.join('');
-    result[id] = crypto.createHash('sha256').update(toHash).digest('hex');
+
+    const model = loadResourceModel(resource.Type);
+    const primaryIdentifier = intersection(Object.keys(resource.Properties ?? {}), model?.primaryIdentifier ?? []);
```
rix0rrr, line 94: If `primaryIdentifier` is compound, it's not "one of", it's "all of".

An `intersection` would be nonempty if at least one of the keys matched, but don't we need all keys to match?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/digest.ts
```diff
@@ -82,10 +89,29 @@ export function computeResourceDigests(template: CloudFormationTemplate): Record
...
+      // The resource has a physical ID defined, so we can
+      // use the ID and the type as the identity of the resource.
+
+      const toHash =
+        resource.Type +
+        primaryIdentifier
+          .sort()
+          .map((attr) => JSON.stringify(resource.Properties[attr]))
+          .join('');
+      result[id] = crypto.createHash('sha256').update(toHash).digest('hex');
```
rix0rrr, line 105: 💅 The exact hash calculation is the same in both cases so could be factored out of the `if`.

rix0rrr (approved):
Conditionally approved!

PR #368, opened 2025-04-14
feat(toolkit-lib): add a return type for toolkit.diff()

### LINKED ISSUE #33182: Toolkit Action: diff & typed return
Figure out how we can best handle notifying people of [stack diff changes ](https://github.com/aws/aws-cdk/blob/946b7488e680051ccdecb6a7f932d98ab6b6ce6f/packages/aws-cdk/lib/diff.ts#L29). This may require some thought on what the best UX would be

Follow the pattern of https://github.com/aws/aws-cdk/pull/33298 but for the diff action.

We are looking for:
- a copy-paste-and-adjust implementation of the existing [diff action](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L163-L277) for the new Toolkit
- corresponding new types in `lib/actions/diff`
- 100% test coverage for the "new" code
- A suggestion for what this method should return (new feature)

kaizencc:
`toolkit.diff()` now returns a `TemplateDiff` for each stack it diffs.

Closes https://github.com/aws/aws-cdk/issues/33182

---

### DIFF
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
@@ -121,11 +121,6 @@ export interface TemplateInfo {
   readonly changeSet?: any;
 
   /**
-   * The name of the stack
-   */
-  readonly stackName: string;
-
-  /**
    * Whether or not there are any imported resources
    *
    * @default false
@@ -164,7 +159,7 @@ export class DiffFormatter {
     this.ioHelper = props.ioHelper;
     this.oldTemplate = props.templateInfo.oldTemplate;
     this.newTemplate = props.templateInfo.newTemplate;
-    this.stackName = props.templateInfo.stackName;
+    this.stackName = props.templateInfo.newTemplate.stackName;
     this.changeSet = props.templateInfo.changeSet;
     this.nestedStacks = props.templateInfo.nestedStacks;
     this.isImport = props.templateInfo.isImport ?? false;
--- a/packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
+++ b/packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
@@ -122,10 +122,8 @@ export interface TemplateInfo {
 
   /**
    * The name of the stack
-   *
-   * @default undefined
    */
-  readonly stackName?: string;
+  readonly stackName: string;
 
   /**
    * Whether or not there are any imported resources
@@ -151,10 +149,16 @@ export class DiffFormatter {
   private readonly ioHelper: IoHelper;
   private readonly oldTemplate: any;
   private readonly newTemplate: cxapi.CloudFormationStackArtifact;
-  private readonly stackName?: string;
+  private readonly stackName: string;
   private readonly changeSet?: any;
   private readonly nestedStacks: { [nestedStackLogicalId: string]: NestedStackTemplates } | undefined;
   private readonly isImport: boolean;
+
+  /**
+   * Stores the TemplateDiffs that get calculated in this DiffFormatter,
+   * indexed by the stack name.
+   */
+  private _diffs: { [name: string]: TemplateDiff } = {};
 
   constructor(props: DiffFormatterProps) {
     this.ioHelper = props.ioHelper;
@@ -164,6 +168,10 @@ export class DiffFormatter {
     this.changeSet = props.templateInfo.changeSet;
     this.nestedStacks = props.templateInfo.nestedStacks;
     this.isImport = props.templateInfo.isImport ?? false;
+  }
+
+  public get diffs() {
+    return this._diffs;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
```diff
@@ -184,11 +192,12 @@ export class DiffFormatter {
 
   private formatStackDiffHelper(
     oldTemplate: any,
-    stackName: string | undefined,
+    stackName: string,
```
iankhou, line 195: What if it's an empty string?

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
```diff
@@ -54,38 +53,49 @@ describe('diff', () => {
...
-  //           },
-  //         }),
-  //       }),
-  //     },
-  //   }));
-  // });
+    // THEN
+    expect(result['Stack1']).toBeDefined();
+    expect(result['Stack2']).toBeDefined();
+  });
```
iankhou, line 93: Would like to see more thorough testing for multiple template diffs, i.e. more than just that the result is defined.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
```diff
@@ -122,10 +122,8 @@ export interface TemplateInfo {
 
   /**
    * The name of the stack
-   *
-   * @default undefined
    */
-  readonly stackName?: string;
+  readonly stackName: string;
```
iankhou, line 126: Breaking change? Not sure if it matters, though.

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
```diff
@@ -235,71 +245,70 @@ describe('diff', () => {
...
-    //         removals: [],
-    //       }),
-    //     }),
-    //   }));
-    // });
+      // THEN
+      expect(ioHost.notifySpy).toHaveBeenCalledWith(expect.objectContaining({
+        action: 'diff',
+        level: 'warn',
+        code: 'CDK_TOOLKIT_W0000',
```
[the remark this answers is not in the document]
kaizencc: Momo specifically wanted this lol. The new directive is default warning code unless its returning something

packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
kaizencc: diff-formatter isn't public so this isn't breaking. i made this change because we do always have the stackName, since we have the stack

packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
kaizencc: it shouldn't be. but even if it is, i don't think my code would fail. obviously right now if you somehow call `formatStackDiff()` or `formatSecurityDiff` on the same stack in the same `DiffFormatter`, we'd write over the previous diff object. 

1) we are not calling `formatXxx` multiple times per CLI / toolkit call, so I don't expect this to happen in practice at all.
2) even if we did overwrite, we'd be writing the same object as the `fullDiff` is the same.

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
```diff
@@ -54,38 +53,83 @@ describe('diff', () => {
...
+  });
+
+  test('returns multiple template diffs', async () => {
+    // WHEN
+    const cx = await builderFixture(toolkit, 'two-different-stacks');
+    const result = await toolkit.diff(cx, {
+      stacks: { strategy: StackSelectionStrategy.ALL_STACKS },
+    });
+
+    console.log(JSON.stringify(result));
```
iankhou, line 90: Build is failing due to this

iankhou (changes requested):

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
kaizencc, line 90:
```suggestion
```

packages/@aws-cdk/tmp-toolkit-helpers/src/api/diff/diff-formatter.ts
```diff
@@ -151,21 +144,31 @@ export class DiffFormatter {
...
+   * indexed by the stack name.
+   */
+  private _diffs: { [name: string]: TemplateDiff } = {};
+
   constructor(props: DiffFormatterProps) {
     this.ioHelper = props.ioHelper;
     this.oldTemplate = props.templateInfo.oldTemplate;
     this.newTemplate = props.templateInfo.newTemplate;
-    this.stackName = props.templateInfo.stackName;
+    this.stackName = props.templateInfo.newTemplate.stackName;
```
kaizencc, line 162: we are now always taking the stack name from the new template. this always exists. this is what we were doing in practice previously, we are now codifying that and removing `stackName` as an option

iankhou (approved):

PR #374, opened 2025-04-16
chore: syntax check test files

rix0rrr:
Currently, the test files are not syntax checked during TypeScript compilation, only during test running. This makes it hard to pick up on necessary changes during large-scale refactors.

You will only notice them during test runs, and then every single test file will fail with a parse error (instead of potentially a single typechecking error during typechecking phase).

Another complication with the trivial solution of just running `tsc -p tsconfig.dev.json` is that VSCode will not detect that `test/` files are included in a tsconfig file if it's not literally called `tsconfig.json`, so we have to do some trickery to type-check at compilation time AND convince VSCode to type check as well.

---


projenrc/TypecheckTests.ts
```diff
@@ -0,0 +1,58 @@
...
+ * extend the parent `tsconfig.dev.json`, but we have to change the rootDir
+ * and also we have to copy over references because those are not inherited.
+ */
+export class TypecheckTests extends Component {
+  constructor(repo: TypeScriptProject) {
+    super(repo);
+
+    new JsonFile(repo, 'test/tsconfig.json', {
+      obj: {
+        extends: '../tsconfig.dev.json',
```
iliapolo, line 32: Will this cause a recompilation of non test files though? 

https://github.com/aws/aws-cdk-cli/blob/25db152d769ab547fe71259fb2d9e020cbb0a46b/packages/aws-cdk/tsconfig.dev.json#L34-L35

If so - lets override `include` to only point to test files?

rix0rrr: It will cause a recompilation of test files. If we change `include` to non-test files, compilation will fail because we typically `import` files from `lib` and they need to part of the "compiled source set".

The 100% correct way of doing is this to put an independent `tsconfig.json` in every directory, and use project references to link them up. So:

```
packages/
    lib1/
        src/
	    tsconfig.json
	test/
	    tsconfig.json  { references: ../src }
    lib2/
        src/
	    tsconfig.json  { references: ../../lib1/src }
	test/
	    tsconfig.json  { references: ../src, ../../lib1/src }
```

I don't want to spend a week getting that right.

iliapolo (approved):

PR #399, opened 2025-04-22
fix(cli): `cdk notices` ignores failures and calling cli with `--no-notices` makes an unexpected network call

### LINKED ISSUE #370: CLI: --no-notices won't supress call to https://cli.cdk.dev-tools.aws.dev/notices.json
### Describe the bug

the `--no-notices` parameter does not supress the call to https://cli.cdk.dev-tools.aws.dev/notices.json

### Regression Issue

https://github.com/aws/aws-cdk/issues/19357

### Expected Behavior

When --no-notices is passed there will be no call to /notices.json

### Current Behavior

synth with --no-notices calls `https://cli.cdk.dev-tools.aws.dev/notices.json`

### Reproduction Steps

CDK_HOME=$(mktemp -d) NODE_DEBUG=http npx cdk synth --no-notices

### Possible Solution

when notices is supressed also don't call the notices endpoint

_No response_

### CDK CLI Version

2.1007.0 (build d3f6c3c)

### Framework Version

_No response_

### Node.js Version

v22.13.0

### OS

macos - 15.4

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

### LINKED ISSUE #440: (cli): `cdk notices` ignores any failure fetching notices
### Describe the bug

`cdk notices` is a command to explicitly list all relevant (unacknowledged) notices. In an environment without internet connection, it currently displays nothing, and doesn't fail. Only with increased verbosity there is a message indicating the failure. This is unexpected for the explicit command.

Note that the implicit update that's running for every other command should still silently pass. E.g. `cdk synth` we don't explicitly request notices so it's okay to continue. 

### Regression Issue

_No response_

### Expected Behavior

`cdk notices` should fail if it fails to update the notices.

### Current Behavior

It silently continues and pretends no relevant notices are found.

### Reproduction Steps

`cdk notices`

### Possible Solution

_No response_

_No response_

### CDK CLI Version

any

### Framework Version

_No response_

### Node.js Version

any

### OS

any

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

mrgrain:
Fixes #370 
Fixes #440

Previously when the `--no-notices` option was passed, the CLI would still attempt to refresh notices in the background. This is unnecessary and an unexpected behavior to users, especially when running the CLI in environments with no internet access.

On the other hand, running the explicit `cdk notices` command did previously not fail if the command failed to refresh notices. Instead the command would pretend no notices are found and complete successfully.

This PR updates the code to not refresh notices when notices are disabled. Unless the explicit `cdk notices` command is called, which will always refresh and now error if fetching notices fails. It also removes a double refresh of the notices (which didn't have an effect because they were cached, but still).

`cdk notices` with an error now looks like this:

image

Before the command would incorrectly and silently succeed and not show any notices:

image

Only when run with the `-v` flag, it would hint at the failure:

```console
$ cdk notices -v
...
[12:38:29] Could not refresh notices: Error: getaddrinfo ENOTFOUND cli.cdk.dev-tools
```
---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/notices/cached-data-source.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/notices/cached-data-source.ts
@@ -1,6 +1,7 @@
 import * as fs from 'fs-extra';
 import type { Notice, NoticeDataSource } from './types';
 import type { IoDefaultMessages } from '../io/private';
+import { ToolkitError } from '../toolkit-error';
 
 interface CachedNotices {
   expiration: number;
@@ -15,7 +16,8 @@ export class CachedDataSource implements NoticeDataSource {
     private readonly ioMessages: IoDefaultMessages,
     private readonly fileName: string,
     private readonly dataSource: NoticeDataSource,
-    private readonly skipCache?: boolean) {
+    private readonly skipCache?: boolean,
+  ) {
   }
 
   async fetch(): Promise<Notice[]> {
@@ -24,9 +26,21 @@ export class CachedDataSource implements NoticeDataSource {
     const expiration = cachedData.expiration ?? 0;
 
     if (Date.now() > expiration || this.skipCache) {
-      const freshData = await this.fetchInner();
-      await this.save(freshData);
-      return freshData.notices;
+      let updatedData: CachedNotices = cachedData;
+
+      try {
+        updatedData = await this.fetchInner();
+      } catch (e) {
+        this.ioMessages.debug(`Could not refresh notices: ${e}`);
+        updatedData = {
+          expiration: Date.now() + TIME_TO_LIVE_ERROR,
+          notices: [],
+        };
+        throw ToolkitError.withCause('Failed to load CDK notices. Please try again later.', e);
+      } finally {
+        await this.save(updatedData);
+      }
+      return updatedData.notices;
     } else {
       this.ioMessages.debug(`Reading cached notices from ${this.fileName}`);
       return data;
@@ -34,18 +48,10 @@ export class CachedDataSource implements NoticeDataSource {
   }
 
   private async fetchInner(): Promise<CachedNotices> {
-    try {
-      return {
-        expiration: Date.now() + TIME_TO_LIVE_SUCCESS,
-        notices: await this.dataSource.fetch(),
-      };
-    } catch (e) {
-      this.ioMessages.debug(`Could not refresh notices: ${e}`);
-      return {
-        expiration: Date.now() + TIME_TO_LIVE_ERROR,
-        notices: [],
-      };
-    }
+    return {
+      expiration: Date.now() + TIME_TO_LIVE_SUCCESS,
+      notices: await this.dataSource.fetch(),
+    };
   }
 
   private async load(): Promise<CachedNotices> {
--- a/packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
@@ -1,13 +1,13 @@
 import type { ClientRequest } from 'http';
 import type { RequestOptions } from 'https';
 import * as https from 'node:https';
-import { formatErrorMessage } from '../../util';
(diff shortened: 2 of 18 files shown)


rix0rrr:
> This is unexpected to users and prevents the CLI from running in environments with no internet access.

If this prevents the CLI from running in those environments we have even more problems, because the behavior *should* be that failures get swallowed. This is not critical but best-effort behavior, so it's fine.

It's just for `cdk notices` that a fetch failure should be fatal.

mrgrain:
> > This is unexpected to users and prevents the CLI from running in environments with no internet access.
> 
> If this prevents the CLI from running in those environments we have even more problems, because the behavior _should_ be that failures get swallowed. This is not critical but best-effort behavior, so it's fine.
> 
> It's just for `cdk notices` that a fetch failure should be fatal.

Ah yes, you are correct. `CachedDataSource` makes the call non-fatal. I had only looked at the `WebsiteNoticeDataSource` code. But that means it's also not fatal for `cdk notices`.

rix0rrr:
> But that means it's also not fatal for cdk notices.

We could make a distinction in the internal state between "successful fetch with no notices" and "no fetch", and do something like:

```ts
notices.assertLastFetchSuccessful();
```

In the handling of `cdk notices`.

packages/@aws-cdk/toolkit-lib/lib/api/notices/filter.ts
mrgrain, line 1: extracted unchanged

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
mrgrain, line 1: In this file I mostly make the error messages consistent. Variations of `Failed to fetch notices` are moved to one level up. Any errors from this file will show up as the cause printed in yellow.

packages/@aws-cdk/toolkit-lib/lib/util/net.ts
mrgrain, line 1: AI generated. Needed an easy way to make errors more human.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -104,21 +104,28 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
-  await notices.refresh();
+  const refreshNotices = (async () => {
+    if (shouldDisplayNotices) {
+      try {
+        return await notices.refresh();
+      } catch (e: any) {
+        await ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(`Could not refresh notices: ${e}`));
+      }
+    }
+  })();
```
mrgrain, line 127: `refresh` now throws

packages/aws-cdk/lib/cli/pretty-print-error.ts
```diff
@@ -9,7 +9,7 @@ export function prettyPrintError(error: unknown, debug = false) {
   if (err.cause) {
     const cause = ensureError(err.cause);
     console.error(chalk.yellow(cause.message));
-    printTrace(err, debug);
+    printTrace(cause, debug);
```
mrgrain, line 12: Unrelated bug, but previously in dev mode we would print the stack trace of the parent error twice.

packages/@aws-cdk/toolkit-lib/lib/api/notices/cached-data-source.ts
```diff
@@ -24,28 +26,32 @@ export class CachedDataSource implements NoticeDataSource {
...
-      await this.save(freshData);
-      return freshData.notices;
+      let updatedData: CachedNotices = cachedData;
+
+      try {
+        updatedData = await this.fetchInner();
+      } catch (e) {
+        this.ioMessages.debug(`Could not refresh notices: ${e}`);
+        updatedData = {
+          expiration: Date.now() + TIME_TO_LIVE_ERROR,
```
rix0rrr, line 36: Caching a negative result eh?  Bold choice.

rix0rrr (approved):

PR #417, opened 2025-04-27
feat(cli): configurable timeout for ECS hotswap operation

iliapolo:
During hotswap deployment, we wait for the ECS service to become stable:

https://github.com/aws/aws-cdk-cli/blob/f6bd7b9056186c00160ece09ac2d4ddc2ad72f35/packages/%40aws-cdk/tmp-toolkit-helpers/src/api/hotswap/ecs-services.ts#L153-L156

The timeout is hardcoded to 600 seconds, which is a very long time to **have** to wait. This PR add an additional property to the already existing ECS overrides bag, to configure this timeout. 

### Notes

Our integ tests currently suffer from this because they needlessly wait the hard-coded 10 minute timeout in a test that asserts ECS updates error handling. With this option, we set the operation timeout to 10 seconds and reduce the total integ suite duration **from 17min to 9min**.

---


packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-hotswap-deployment-for-ecs-service-detects-failed-deployment-and-errors.integtest.ts
```diff
@@ -10,7 +10,7 @@ integTest(
 
     // WHEN
     const deployOutput = await fixture.cdkDeploy('ecs-hotswap', {
-      options: ['--hotswap'],
+      options: ['--hotswap', '--hotswap-operation-timeout-seconds', '10'],
```
rix0rrr, line 13: 🤢

`--hotswap-timeout=10s` ?

I know you probably want `seconds` in there to be quite clear... but this is an extremely long flag name to type.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -72,7 +72,7 @@ export async function isHotswappableAppSyncChange(
       },
       hotswappable: true,
       service: 'appsync',
-      apply: async (sdk: SDK) => {
+      apply: async (sdk: SDK, _timeoutSeconds?: number /* unused because we don't currently wait for anything */) => {
```
rix0rrr, line 75: 💅 You can leave out the argument if you don't use it.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/common.ts
```diff
@@ -26,7 +26,7 @@ export interface HotswapOperation {
   /**
    * Applies the hotswap operation
    */
-  readonly apply: (sdk: SDK) => Promise<void>;
+  readonly apply: (sdk: SDK, timeoutSeconds?: number) => Promise<void>;
```
rix0rrr, line 29: Should we predict that there will be more options in the future, and move to an `Options` struct straight away?

packages/aws-cdk/lib/cli/user-input.ts
```diff
@@ -727,6 +727,13 @@ export interface DeployOptions {
    */
   readonly hotswap?: boolean;
 
+  /**
+   * Number of seconds to wait for a hotswap operation to complete. Normally this applies to resource stabilization.
```
rix0rrr, line 731: "Normally"? What are the non-normal cases?

packages/aws-cdk/lib/cli/user-input.ts
```diff
@@ -727,6 +727,13 @@ export interface DeployOptions {
...
+   *
+   * @default - undefined
```
rix0rrr, line 733: "undefined" is not good value for `@default`. *I already knew that*. Tell me what it does instead.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/ecs-services.ts
```diff
@@ -153,7 +153,7 @@ export async function isHotswappableEcsServiceChange(
             await sdk.ecs().waitUntilServicesStable({
               cluster: update.service?.clusterArn,
               services: [service.serviceArn],
-            });
+            }, timeoutSeconds);
```
rix0rrr, line 156: Can we also bring down the default to something reasonable? Like 2 minutes?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/lambda-functions.ts
```diff
@@ -318,7 +319,7 @@ async function waitForLambdasPropertiesUpdateToFinish(
 
   await lambda.waitUntilFunctionUpdated(delaySeconds, {
     FunctionName: functionName,
-  });
+  }, timeoutSeconds);
```
rix0rrr, line 322: Can we also bring down the default to something reasonable? Like 1 minute?

rix0rrr (dismissed):

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-hotswap-deployment-for-ecs-service-detects-failed-deployment-and-errors.integtest.ts
mrgrain: +1

packages/@aws-cdk/tmp-toolkit-helpers/src/api/deployments/deploy-stack.ts
```diff
@@ -167,6 +167,14 @@ export interface DeployStackOptions {
    */
   readonly hotswap?: HotswapMode;
 
+  /**
+   * Number of seconds to wait for a hotswap operation to complete.
+   * This normally controls the amount of time to wait for resource stabilization.
+   *
+   * @default - operation dependant
+   */
+  readonly hotswapOperationTimeoutSeconds?: number;
```
mrgrain, line 176:
```suggestion
  readonly hotswapTimeout?: number;
```

or at least

```suggestion
  readonly hotswapTimeoutSeconds?: number;
```

mrgrain, line 176: Also wondering if we shoud start a generic `hotswapOptions` instead 🤔 WDYT?

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/common.ts
mrgrain: +1 (also this not public currently so maybe not super important)

packages/aws-cdk/lib/cli/user-input.ts
mrgrain: What happens if the timeout is exceeded?

mrgrain, line 731: Maybe something like this?
```suggestion
   * Number of seconds to wait for a single hotswap resource update to complete. Normally this applies to resource stabilization.
```

mrgrain (changes requested):
see inline

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-hotswap-deployment-for-ecs-service-detects-failed-deployment-and-errors.integtest.ts
iliapolo: Changed to `--hotswap-timeout=10s` in the CLI layer only, APIs still accept a concrete number of seconds.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/common.ts
iliapolo: Added a` HotswapOperationOptions` struct.

packages/aws-cdk/lib/cli/user-input.ts
iliapolo: Rephrased.

packages/aws-cdk/lib/cli/user-input.ts
iliapolo: This is code generated from the yargs definitions, so every string option gets `undefined` as its `@default` value. I don't believe there's a way to inject here a "description of the default".

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/ecs-services.ts
iliapolo: I would like to, but lets be careful about this because its a breaking change. History tells us that a small section of customers will indeed be broken. 

- **On one hand**, the folks benefiting from a reduced timeout are those who's operation fails right now anyway, so presumably they have fixed whatever the problem was and it doesn't hurt them that much.
- **On the other hand**, users who's update takes >2 minutes will now start seeing failures.

All in all: broken operations continue to be broken, and working operations may fail. Doesn't seem worth it...

@mrgrain wanna be the tie breaker?

packages/@aws-cdk/toolkit-lib/lib/index.ts
```diff
@@ -17,3 +17,4 @@ export * from './api/cloud-assembly';
 export * from './api/io';
 export * from './api/tags';
 export * from './api/plugin';
+export { HotswapOperationOptions } from './api/hotswap/common';
```
iliapolo, line 20: Explicitly exporting only this type because there are duplicate exports if we export the entire `common` module.

```console
Module './actions' has already exported a member named 'EcsHotswapProperties'. Consider explicitly re-exporting to resolve the ambiguity.ts(2308)
Module './actions' has already exported a member named 'HotswapMode'. Consider explicitly re-exporting to resolve the ambiguity.ts(2308)
module "/Users/epolon/dev/src/github.com/aws/aws-cdk-cli/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/common"

We should cleanup this up because there are currently duplicate definitions of the `HotswapMode` and `EcsHotswapProperties` class/interface. Didn't want to venture into that in this PR thought. 
```

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
```diff
@@ -93,6 +94,7 @@ export async function tryHotswapDeployment(
   stackArtifact: cxapi.CloudFormationStackArtifact,
   hotswapMode: HotswapMode,
   hotswapPropertyOverrides: HotswapPropertyOverrides,
+  hotswapOperationOptions?: HotswapOperationOptions,
```
mrgrain, line 97: It might be easier to start combining these two (`hotswapPropertyOverrides` and `hotswapOperationOptions`) into a single object.

packages/@aws-cdk/toolkit-lib/lib/index.ts
mrgrain: I tend to put the export into the action file where it's needed. This is a bit awkward in this case because of the private baseoptions, but maybe in `packages/@aws-cdk/toolkit-lib/lib/actions/deploy/index.ts` would be okay.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-hotswap-deployment-for-ecs-service-detects-failed-deployment-and-errors.integtest.ts
mrgrain: Nice! That's exactly what we have the separate layers for.

packages/aws-cdk/lib/cli/user-input.ts
mrgrain: Not yet, we could add that though. We could also just explicitly set a default global timeout as part of the config.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/hotswap/ecs-services.ts
mrgrain: I only realize this now: The way we are using this value isn't really a generic hotswap timeout, isn't it? It's a very specific option on how long to wait for ECS service stabilization - which can be very different depending on your actual service.  

Tend to agree with @iliapolo that this isn't worth the risk of breaking.

mrgrain (changes requested):
1/ I think we should align how we handle `HotswapPropertyOverrides` and `HotswapOperationOptions`. Both are basically serving the same end goal (customize a specific hotswapper) but are implemented differently. Neither way is better or worse but I think we need to pick one and have it be used by both.

2/ I'm now convinced this the wrong the name for the option. ~At least it should be called something like `--hotswap-stabilization`~. But I'm also thinking it should at least be on a per service level. If not per resource.


Edit: Thinking more about this - the stabilization timeout seems to be exactly the same as what `hotswapPropertyOverrides` are already trying to solve today. A way to override deployment config on a per use basis. We currently have `minimumHealthyPercent`/`maximumHealthyPercent`. Adding `waitTimeForStabilization` (or whatever) feels exactly the same to me! (`hotswapPropertyOverrides` should have probably operated on a resource level from the beginning, but we can add that later)

packages/@aws-cdk/toolkit-lib/test/api/hotswap/ecs-services-hotswap-deployments.test.ts
```diff
@@ -724,5 +724,96 @@ describe.each([
       },
       forceNewDeployment: true,
     });
+    expect(mockECSClient).toHaveReceivedCommandWith(DescribeServicesCommand, {
+      cluster: 'arn:aws:ecs:region:account:service/my-cluster',
+      services: ['arn:aws:ecs:region:account:service/my-cluster/my-service'],
+    });
```
iliapolo, line 730: I just noticed this assertion was missing, to make sure we actually call the waiter.

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -305,8 +305,9 @@ export function commandLineArgumentsToSettings(argv: Arguments): Settings {
     ignoreNoStacks: argv['ignore-no-stacks'],
     hotswap: {
       ecs: {
-        minimumEcsHealthyPercent: argv.minimumEcsHealthyPercent,
-        maximumEcsHealthyPercent: argv.maximumEcsHealthyPercent,
```
iliapolo, line 309: This allows for passing the hotswap override properties via the command line. However, it never actually worked because the keys it sets are not actually the keys we read:

https://github.com/aws/aws-cdk-cli/blob/382393730b6e8870cfb5d22aaf40c51b0b73b887/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L387-L393

- `minimumEcsHealthyPercent` !== `minimumHealthyPercent`
- `maximumEcsHealthyPercent` !== `maximumHealthyPercent`

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -305,8 +305,9 @@ export function commandLineArgumentsToSettings(argv: Arguments): Settings {
...
+        minimumHealthyPercent: argv.hotswapEcsMinimumHealthyPercent,
+        maximumHealthyPercent: argv.hotswapEcsMaximumHealthyPercent,
```
iliapolo, line 309: So actually a user could never override this behavior from the CLI. Which means, we can change the argument names to be more descriptive:

- `hotswap-ecs-minimum-healthy-percent`
- `hotswap-ecs-maximum-healthy-percent`

These names are derived and match the argument names that I think we initially wanted, based on:

https://github.com/aws/aws-cdk-cli/blob/382393730b6e8870cfb5d22aaf40c51b0b73b887/packages/%40aws-cdk/toolkit-lib/lib/api/hotswap/common.ts#L91-L96

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -158,6 +158,18 @@ export async function makeConfig(): Promise<CliConfig> {
...
               'Do not use this in production environments',
           },
+          'hotswap-ecs-minimum-healthy-percent': {
+            type: 'string',
+            desc: 'Lower limit on the number of your service\'s tasks that must remain in the RUNNING state during a deployment, as a percentage of the desiredCount',
+          },
+          'hotswap-ecs-maximum-healthy-percent': {
+            type: 'string',
+            desc: 'Upper limit on the number of your service\'s tasks that are allowed in the RUNNING or PENDING state during a deployment, as a percentage of the desiredCount',
+          },
```
iliapolo, line 168: These were just flat out missing from the yargs definition, even though we did (try to) handle them in:

https://github.com/aws/aws-cdk-cli/blob/382393730b6e8870cfb5d22aaf40c51b0b73b887/packages/aws-cdk/lib/cli/user-configuration.ts#L306-L310

iliapolo:
@mrgrain Apparently passing ECS overrides was supposed to be supported (see comments) - but didn't actually work. Right now, this PR both adds a new override and fixes the previous ones. I created this combined PR so that it will be easy to see the entire context, but I'm happy to split the PR into:

- First fix the current bugs in the CLI handling.
- Then add the new override.

Also, since CLI support doesn't actually work now, we can drop that entire code path and give it some more thought before re-introducing it.

WDYT?

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-hotswap-deployment-for-ecs-service-detects-failed-deployment-and-errors.integtest.ts
```diff
@@ -10,7 +10,7 @@ integTest(
 
     // WHEN
     const deployOutput = await fixture.cdkDeploy('ecs-hotswap', {
-      options: ['--hotswap'],
+      options: ['--hotswap', '--hotswap-ecs-stabilization-timeout-seconds', '10'],
```
iliapolo, line 13: Decided not to do the `hotswap-ecs-stabilization-timeout=10s` option now because:

- The argument is also quite long even without the `-seconds` suffix.
- This can also be defined in `cdk.json` - so no excessive typing needed.

mrgrain (approved):

mrgrain (approved):

Issue #379, opened 2025-04-16
Add support for skip files


otaviomacedo:
Support user provided skip files, that work as deny lists, to remove individual computed mappings.

PR #421, opened 2025-04-28
feat(refactor): ability to exclude selected resources from refactoring

### LINKED ISSUE #412: Define and implement a skipping contract in Cloud Assembly
CDK applications should be able to communicate to the CLI which constructs should not be subject to refactoring. Define a contract in Cloud Assembly (probably using metadata in the manifest), and implement the honoring of that contract in the CLI.

This is useful for constructs such as API Gateway `Deployment`, that intentionally change the logical ID to force a resource replacement.

otaviomacedo:
Allow the user to select resources to be skipped for refactoring. There are two ways to do this:

- By adding the `aws:cdk:skip-refactor = true` metadata to each resource to be skipped, in the cloud assembly manifest.
- By passing a skip fie via command line or the toolkit library API. The file should be in JSON format and contain a list of destination locations. Entries in this list can be either in the form `"<stack name>.<logical ID>"` or a construct path.

The main abstraction is the `SkipList` interface, used by the mapping detection algorithm to query for mappings that should be excluded.

Closes https://github.com/aws/aws-cdk-cli/issues/412, https://github.com/aws/aws-cdk-cli/issues/379

---


packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -14,4 +14,13 @@ export interface RefactorOptions {
...
   stacks?: StackSelector;
+
+  /**
+   * The absolute path to a file that contains a list of
+   * resources to skip during the refactor. The file should
+   * be in JSON format and contain an array of _destination_
+   * logical IDs, that is, the logical IDs of the resources
+   * as they would be after the refactor.
+   */
+  skipFile?: string;
```
otaviomacedo, line 25: I'm looking for a better name. If anyone has any suggestions, let me know.

packages/@aws-cdk/tmp-toolkit-helpers/src/api/refactoring/cloudformation.ts
```diff
@@ -15,3 +16,54 @@ export interface CloudFormationStack {
   readonly stackName: string;
   readonly template: CloudFormationTemplate;
 }
+
```
otaviomacedo, line 19: This was just moved from `index.ts`

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -15,3 +16,54 @@ export interface CloudFormationStack {
   readonly stackName: string;
   readonly template: CloudFormationTemplate;
 }
+
```
otaviomacedo, line 19: This code was just moved from `index.ts`.

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
mrgrain: For toolkit-lib, this should be an array. We can make the read from file a CLI feature.

mrgrain: > I'm looking for a better name. If anyone has any suggestions, let me know.

`exclude`?

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -14,4 +14,18 @@ export interface RefactorOptions {
    * @default - all stacks
    */
   stacks?: StackSelector;
+
+  /**
+   * The absolute path to a file that contains a list of resources to
+   * skip during the refactor. The file should be in JSON format and
+   * contain an array of _destination_ locations that should be skipped,
+   * i.e., the location to which a resource would be moved if the
+   * refactor were to happen.
```
mrgrain, line 23: Why is _destination_ the better choice over _source_? My gut reaction was: "huh, but how do I know what the destination is?" Is this a concern?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/skip.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  constructor(private readonly filePath?: string) {
+    this.skippedLocations = [];
+    this.skippedPaths = [];
+
+    if (!this.filePath) {
+      return;
+    }
+
+    const parsedData = JSON.parse(fs.readFileSync(this.filePath, 'utf-8'));
```
mrgrain, line 65: Is a JSON with a single array better than a newline separate file?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/skip.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (!isValidSkipFileContent(parsedData)) {
+      throw new ToolkitError('The content of a skip file must be a JSON array of strings');
+    }
+
+    const locationRegex = /^[A-Za-z0-9]+\.[A-Za-z0-9]+$/;
```
mrgrain, line 70: Double checking: Can a valid path really never look like this?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1218,30 +1211,23 @@ export class CdkToolkit {
   }
 
   public async refactor(options: RefactorOptions): Promise<number> {
-    if (!options.dryRun) {
-      info('Refactor is not available yet. Too see the proposed changes, use the --dry-run flag.');
+    const tk = new Toolkit();
```
mrgrain, line 1214: We are already setting this up in L186 and make it available as `this.toolkit` - any reason we cannot use that instance here?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1218,30 +1211,23 @@ export class CdkToolkit {
...
+
+    const casmSource = await tk.fromAssemblyBuilder(() => this.assembly());
+    try {
+      await tk.refactor(casmSource, {
+        dryRun: options.dryRun,
+        skipFile: options.skipFile,
+        stacks: {
+          patterns: options.selector.patterns,
+          strategy: options.selector.patterns.length > 0 ? StackSelectionStrategy.PATTERN_MATCH : StackSelectionStrategy.ALL_STACKS,
+        },
```
mrgrain, line 1224: Nice! This could maybe be helper so other (future) actions can re-use it. Not important for this PR though.

mrgrain (changes requested):
High-level items:
- I also don't like `skip` that much. Maybe `exclude` or `ignore` is a more common terminology
- Let's keep the toolkit-lib interface truly programmatic and accept a list, the CLI can handle the reading from a file.
- `aws:cdk:skip-refactor = true` I know this is needed to communicate between the two code bases. But how is the actual user experience going to look like? Like what code do I (eventually) need to write as a user to skip a resource?

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
mrgrain: To be clear: I genuinely don't know the answer here. But if there's two choices we always run the risk of picking the less natural one.

otaviomacedo:
>`aws:cdk:skip-refactor = true` I know this is needed to communicate between the two code bases. But how is the actual user experience going to look like? Like what code do I (eventually) need to write as a user to skip a resource?

The idea is that the end user will never have to do anything. Library authors will exclude their constructs by calling

```ts
resource.addMetadata('aws:cdk:skip-refactor', true);
```

inside the construct class.

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
mrgrain: Okay, so the thinking here is: The code base (and therefore the output assembly) have already been changed by the user. Refactoring is matching the new target state with the previous state from CFN. We are addressing resources by what they are currently called in the users local code base?

I can dig that. We might want to tweak the wording and explanation here a bit (maybe AI can help?), but otherwise all good.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/skip.ts
mrgrain: Might be worth adding that explanation as comment.

mrgrain:
> The idea is that the end user will never have to do anything. Library authors will exclude their constructs by calling
> 
> ```ts
> resource.addMetadata('aws:cdk:skip-refactor', true);
> ```
> 
> inside the construct class.

Should we add a helper method for this? Is there precedence for helpers that add "magic" metadata values?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/metadata-schema.ts
```diff
@@ -312,6 +312,11 @@ export enum ArtifactMetadataEntryType {
    * Represents tags of a stack.
    */
   STACK_TAGS = 'aws:cdk:stack-tags',
+
+  /**
+   * Whether the resource should be skipped during refactoring.
+   */
+  SKIP_REFACTOR = 'aws:cdk:skip-refactor',
```
mrgrain, line 319: Should this be called exclude now? Or

```suggestion
  SKIP_REFACTOR = 'aws:cdk:no-refactor',
```

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -14,4 +14,17 @@ export interface RefactorOptions {
...
+   * A list of resources to skip during the refactor. Elements of
+   * this list should be _destination_ locations that should be skipped,
+   * i.e., the location to which a resource would be moved if the
+   * refactor were to happen.
+   *
+   * The format of the locations in the file can be either:
+   *
+   * - Stack name and logical ID (e.g. `Stack1.MyQueue`)
+   * - A construct path (e.g. `Stack1/Foo/Bar/Resource`).
+   */
```
mrgrain, line 28:
```suggestion
  /**
   * A list of resources that will not be part of the refactor. Elements of
   * this list must be the _destination_ locations that should be skipped,
   * i.e., the location to which a resource would be moved if the
   * refactor were to happen.
   *
   * The format of the locations in the file can be either:
   *
   * - Stack name and logical ID (e.g. `Stack1.MyQueue`)
   * - A construct path (e.g. `Stack1/Foo/Bar/Resource`).
   */
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -964,11 +967,13 @@ export class Toolkit extends CloudAssemblySourceBuilder {
 
     const stacks = await assembly.selectStacksV2(ALL_STACKS);
     const sdkProvider = await this.sdkProvider('refactor');
-    const movements = await findResourceMovements(stacks.stackArtifacts, sdkProvider);
+    const skipList = fromManifestAndExclusionList(assembly.cloudAssembly.manifest, options.exclude);
```
mrgrain, line 970:
```suggestion
    const excludeList = fromManifestAndExclusionList(assembly.cloudAssembly.manifest, options.exclude);
```

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1218,30 +1211,28 @@ export class CdkToolkit {
...
   public async refactor(options: RefactorOptions): Promise<number> {
-    if (!options.dryRun) {
-      info('Refactor is not available yet. Too see the proposed changes, use the --dry-run flag.');
-      return 1;
+    let exclude: string[] = [];
+    if (options.excludeFile != null) {
+      if (!(await fs.pathExists(options.excludeFile))) {
+        throw new ToolkitError(`The exclude file ${options.excludeFile} does not exist`);
+      }
+      exclude = fs.readFileSync(options.excludeFile).toString('utf-8').split('\n');
```
mrgrain, line 1219: Not super important, but this could probably done with a single "read" operation and catching the `ENOENT` error.

mrgrain (approved):

PR #423, opened 2025-04-29
feat(cli-integ): calculate `maxWorkers` based on machine spec

iliapolo:
In order to be able to run on various machine types, calculate the max number of workers based on the total memory and CPU cores available in the machine. With this calculation we get:

- CLI Repo (64GB & 16 Cores) = `Math.min(16 * 15, (64000 - 2000) / (0.7 * 500)) = Math.min(240, 177) = 177`
- CodeBuild Canary (16GB & 8 Cores) = `Math.min(8 * 15, (16000 - 2000) / (0.7 * 500)) = Math.min(120, 40) = 40`
- CodeBuild Release & Main (72GB & 32 Cores) = `Math.min(32 * 15, (72000 - 2000) / (0.7 * 500)) = Math.min(480, 200) = 200`

### Memory Data

Here is some memory usage sampling i've [added](https://github.com/aws/aws-cdk-cli/pull/422) to confirm our memory consumption stats:

```console

2025-04-29T09:50:51.525Z | [pre-callback] [test] Memory RSS: 152 MB
2025-04-29T09:50:51.200Z | [pre-callback] [test] Memory RSS: 150 MB
2025-04-29T09:50:50.836Z | [pre-callback] [test] Memory RSS: 150 MB
2025-04-29T09:54:08.055Z | [post-callback] [test] Memory RSS: 122 MB
2025-04-29T09:54:08.797Z | [post-callback] [test] Memory RSS: 114 MB
2025-04-29T09:54:12.054Z | [post-callback] [test] Memory RSS: 113 MB

2025-04-29T09:53:06.601Z | [pre-tree] [app.js] Memory RSS: 49 MB
2025-04-29T09:53:10.468Z | [pre-tree] [app.js] Memory RSS: 49 MB
2025-04-29T09:54:04.682Z | [pre-tree] [app.js] Memory RSS: 49 MB
2025-04-29T09:53:07.836Z | [post-tree] [app.js] Memory RSS: 105 MB
2025-04-29T09:54:05.071Z | [post-tree] [app.js] Memory RSS: 106 MB
2025-04-29T09:53:11.895Z | [post-tree] [app.js] Memory RSS: 105 MB
2025-04-29T09:53:08.864Z | [post-synth] [app.js] Memory RSS: 139 MB
2025-04-29T09:53:13.215Z | [post-synth] [app.js] Memory RSS: 139 MB
2025-04-29T09:54:05.367Z | [post-synth] [app.js] Memory RSS: 138 MB

2025-04-29T09:53:09.219Z | [post-synth] [cli] Memory RSS: 200 MB
2025-04-29T09:53:13.359Z | [post-synth] [cli] Memory RSS: 203 MB
2025-04-29T09:54:05.437Z | [post-synth] [cli] Memory RSS: 202 MB
2025-04-29T09:53:10.358Z | [pre-synth] [cli] Memory RSS: 201 MB
2025-04-29T09:54:04.627Z | [pre-synth] [cli] Memory RSS: 183 MB

```  

From here we see that:

- Memory consumed by jest process: ~150MB
- Memory consumed by CDK App subprocess: ~140MB
- Memory consumed by CLI subprocess: ~200MB

## Notes

- These numbers suggest a total of about 500MB per worker...multiple that by 50 and you get 25GB - way above the 16GB we were running with. Stands to reason that a simple 500 X 50 multiplication doesn't accurately describe the total max memory consumption of the suite.
- Its surprising that the memory consumed by jest after the callback finishes is less than before the callback starts.  

---


packages/@aws-cdk-testing/cli-integ/resources/integ.jest.config.js
```diff
@@ -19,9 +20,42 @@ module.exports = {
...
+
+  // empirically observed. this includes:
+  // - 150 jest test process
+  // - 140 app synthesis subprocess  
+  // - 200 cli subprocess
+  const observedWorkerMemoryMB = 500;
+
+  // GC occasionally gets a chance to clear this amount of memory
+  // so the workers memory doesnt need to be so tight.
+  const estimatedWorkerFluctuationMB = 150;
```
rix0rrr, line 46: This is not entirely accurate I don't think, so I wonder if this helps. GC plays a part, but I think a small part. My guess is that the bigger part of it is that 2 of the memory consumers (CLI and App) only live for very short bursts, and afterwards their memory gets cleaned up. So the actual (average) consumption is a function of how much the "synths" overlap.

image

The average memory consumption in the right-hand case is lower, even though both are running 2 tests and the per-test memory is 500MB. That is not really caused by "GC".

Also I think this should probably be expressed as a fraction, not an absolute decrease, since it's a function of process overlap.

```ts
const theoreticalMaxTestMemoryMB = 500;

const estimatedTestMemoryMB = theoreticalMaxTestMemoryMB * 0.7; // Observed by very scientific eyeballing
```

packages/@aws-cdk-testing/cli-integ/resources/integ.jest.config.js
```diff
@@ -19,9 +20,42 @@ module.exports = {
...
+
+  // leave 3GB for the OS and other external processes
+  const reservedMemoryMB = 3000;
```
rix0rrr, line 49: This is a lot, no?

packages/@aws-cdk-testing/cli-integ/resources/integ.jest.config.js
```diff
@@ -19,9 +20,42 @@ module.exports = {
...
+
+  // our processes don't take up much CPU so we allow for a large factor.
+  const cpuScaleFactor = 10;
```
rix0rrr, line 52: I'm tempted to leave this out altogether?

rix0rrr (changes requested):

packages/@aws-cdk-testing/cli-integ/resources/integ.jest.config.js
iliapolo: Is it? Lowered to 2GB

packages/@aws-cdk-testing/cli-integ/resources/integ.jest.config.js
iliapolo: We probably can...but it doesn't jive well with me saying we are completely not bound by CPU (thinking about has calculation and regex matching). I've increased to 15.

rix0rrr (approved):

PR #425, opened 2025-04-29
chore(integ-tests): decouple CLI and Library sources, improve documentation

rix0rrr:
We used to make assumptions in the integ tests that the CLI and library versions were the same unless specified otherwise, and that they both would come from the same repo.

These assumptions are no longer true: the CLI and the library have their own version numbers, and only the CLI can come from a source repo (or at least, that's how it works in practice).

Tear apart `ReleasePackageSource` and `RepoPackageSource` into (currently):

- `CliNpmSource` / `CliRepoSource`
- `LibraryNpmSource`

Make a more clear naming distinction between the part that runs in the runner and the part that runs in the test.

Explain in documentation what the options are, and how the use of the given component is achieved at runtime.

Make the argument names and default values a little more sensible given the new source locations.

CHANGES:

- Add `--cli-version` to mirror `--framework-version`.
- Default to `--auto-source` if no CLI version given.
- Add `-F` alias to specify a single test file.
- Resolve the framework version immediately.
- Deprecate `--use-source` and replace with `--cli-source`.
- Deprecate `--auto-source` and replace with `--cli-source=auto`.
- Change to a different `yargs` version and twiddle all the options slightly differently to enable strict parsing mode.

---


.projenrc.ts
```diff
@@ -1589,17 +1589,20 @@ const cliInteg = configureProject(
...
       include: ['**/*.ts'],
       exclude: ['resources/**/*'],
     },
     jestOptions: jestOptionsForProject({
       jestConfig: {
         coverageThreshold: {
-          statements: 40,
-          lines: 40,
+          statements: 25,
+          lines: 25,
```
mrgrain, line 1603: Not loving this. Can you at least explain why these are lowered?

packages/@aws-cdk-testing/cli-integ/lib/cli/run-suite.ts
```diff
@@ -2,115 +2,126 @@
...
+    if (args[flagAlias]) {
+      cliSource.set(new RunnerCliNpmSource(args[flagAlias]), `--${flagAlias}`);
     }
-    packageSource = s;
   }
-
-  if (args['use-source'] || args['auto-source']) {
-    if (args['framework-version']) {
-      throw new Error('Cannot use --framework-version with --use-source');
+  for (const flagAlias of ['cli-source', 'use-source'] as const) {
```
mrgrain, line 107: I think this gives `--use-source` precedence over `--cli-source` if both are defined.

packages/@aws-cdk-testing/cli-integ/lib/package-sources/find-root.ts
```diff
@@ -0,0 +1,14 @@
+import * as path from 'path';
+import { findUp } from '../files';
+
+/**
+ * Find the root directory of the repo from the current directory
+ */
+export async function autoFindRoot() {
```
mrgrain, line 7: Maybe this? Because it could also be package root.

```suggestion
export async function autoFindRepoRoot() {
```

packages/@aws-cdk-testing/cli-integ/lib/package-sources/library-npm-source.ts
```diff
@@ -0,0 +1,55 @@
...
+export class TestLibraryNpmSource implements ITestLibrarySource {
+  constructor(public readonly packageName: string, public readonly version: string) {
+  }
+
+  public requestedVersion(): string {
+    return this.version;
+  }
+
+  public assertJsiiPackagesAvailable(): void {
+    // FIXME: Always a no-op.
```
mrgrain, line 35: Is this a `FIXME` then if it's always a no-op?

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-deploy-with-role.integtest.ts
mrgrain, line 1: Why is this test removed?

.projenrc.ts
rix0rrr: Yeah. There's a bunch of new code that involves calling NPM and making symlinks on disk, that doesn't have unit tests because it's either awkward to test or introduces sources of network calls and latency that we don't really want in unit tests (imo). 

I figured since this is for the integ tests, it is basically tested as part of every PR already.

packages/@aws-cdk-testing/cli-integ/lib/cli/run-suite.ts
rix0rrr: `cliSource` is of type `UniqueOption<T>` which will throw an error if it's set more than once.

mrgrain (approved):

PR #429, opened 2025-04-29
feat(cli): new typescript projects use modern tsconfig

### LINKED ISSUE #427: Proposal: Update default `tsconfig.json` for `cdk init --language=typescript` to align with modern module resolution
### Describe the feature

Currently, the default `tsconfig.json` generated by `cdk init --language=typescript` uses older TypeScript settings, particularly for `module` and `moduleResolution`.

While this works for many scenarios, it can cause compatibility issues with newer versions of popular libraries within the AWS ecosystem, such as Powertools for AWS Lambda (TypeScript). These libraries are increasingly relying on the `exports` map in `package.json` for defining entry points, which requires modern `moduleResolution` strategies like `node16`, `nodenext`, or `bundler` to be correctly interpreted by TypeScript.

A recent issue ([aws-powertools/powertools-lambda-typescript#3865](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3865)) highlighted this problem. While Powertools temporarily restored `typeVersions` for backward compatibility in a minor release, they intend to rely solely on the `exports` map in their next major version (v3). This means CDK users who haven't updated their `tsconfig.json` manually will likely encounter TS2307 import errors when upgrading Powertools or other similar libraries in the future.

**Describe the solution you'd like**

Update the default `tsconfig.json` generated by `cdk init --language=typescript` to use more modern and forward-compatible settings. Specifically, consider aligning with the recommended settings for projects using modern Node.js features and libraries like Powertools:

Reference: [Powertools for AWS Lambda (TypeScript) - Recommended Settings](https://docs.powertools.aws.dev/lambda/typescript/latest/getting-started/typescript-settings/)

Key recommended changes often include:

*   `"module": "Node16"` or `"module": "NodeNext"`
*   `"moduleResolution": "Node16"` or `"moduleResolution": "NodeNext"` (or `"Bundler"` depending on the target environment and bundler usage)
*   `"target": "ES2022"` (or a similarly modern target)

Using settings like `Node16` or `NodeNext` for `module` and `moduleResolution` ensures that TypeScript correctly interprets the `exports` map in dependencies.

### Use Case

See above.

### Proposed Solution

See above.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.192.0

24.04

iankhou:
Changed target, module, moduleResolution, and lib to more modern TS settings.

This sets a newer default, which prevents import errors when importing other packages that use the `exports` map for defining entrypoints.

While the settings in this template will not fit all use-cases, it sets a better example for future modifications (i.e. should specify moduleResolution).

Fixes #427 

---


packages/aws-cdk/lib/init-templates/app/typescript/tsconfig.json
```diff
@@ -20,11 +21,11 @@
     "experimentalDecorators": true,
     "strictPropertyInitialization": false,
     "typeRoots": [
-      "./node_modules/@types"
+      "./node_modules/@types",
```
rix0rrr, line 24: Trailing commas make this non-standard JSON. It might work due to the specific TSC parser but I'm not entirely comfortable with it.

rix0rrr:
Would love a motivation for the change in the PR body

rix0rrr (dismissed):
See previous comments

packages/aws-cdk/lib/init-templates/lib/typescript/tsconfig.json
```diff
@@ -25,6 +26,6 @@
   },
   "exclude": [
     "node_modules",
-    "cdk.out",
```
iankhou, line 28: There was originally a trailing comma here, which I have also removed.

mrgrain (approved):

PR #437, opened 2025-04-30
feat(toolkit-lib): deploy prompt prints security diff and returns template diff

### LINKED ISSUE #33435: Diff and security diff during deploy
We currently do not send a message for the diff on deploy, only if we need to confirm approval.

We should send
- [ ] a info message with the formatted security diff
- [ ] a request response with the full diff to confirm
- [x] require approval should move into the CliIoHost

kaizencc:
closes https://github.com/aws/aws-cdk/issues/33435

Does the following:

- sends the formatted security diff in the request response
- returns the full diff to in the request response

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -512,13 +512,17 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
 
       const securityDiff = formatter.formatSecurityDiff();
-      const permissionChangeType = securityDiff.permissionChangeType;
+
+      // Send a request response with the formatted security diff as part of the message,
+      // and the template diff as data
+      // (IoHost decides whether to print depending on permissionChangeType)
       const deployMotivation = '"--require-approval" is enabled and stack includes security-sensitive updates.';
-      const deployQuestion = `${deployMotivation}\nDo you wish to deploy these changes`;
+      const deployQuestion = `${securityDiff.formattedDiff}\n\n${deployMotivation}\nDo you wish to deploy these changes`;
```
kaizencc, line 520: i don't have a good way of testing this because formatted diff has a bunch of bold/underlined/colored words so it's hard to unit test

mrgrain: run `stripAnsi` over it

mrgrain (approved):

mrgrain:
Approved but leaving @kaizencc opportunity to improve test.

PR #441, opened 2025-05-01
chore(cli): lint test files

mrgrain:
These were previously ignored, presumingly because they did not follow most of our linting rules. However now a lot less tests are in the CLI package (most moved to toolkit-lib) so we can safely re-enable linting of test files.

Added `'@typescript-eslint/unbound-method': 'off'` for test files, since there were many violations of this and it doesn't seem necessary to enforce this rule in test code.

Most test code changes are automated. I've annotated the few manual fixes I made.

---


packages/aws-cdk/test/cli/io-host/cli-io-host.test.ts
```diff
@@ -429,7 +430,7 @@ describe('CliIoHost', () => {
       });
 
       test('require approval by default - respond no', async () => {
-        expect(() => requestResponse('n', plainMessage({
+        await expect(() => requestResponse('n', plainMessage({
```
mrgrain, line 433: this was missing before, all promises must be awaited

packages/aws-cdk/test/cli/cdk-toolkit.test.ts
```diff
@@ -276,6 +278,7 @@ describe('deploy', () => {
         deployments: new FakeCloudFormation({}),
       });
       stderrMock.mockImplementation((...x) => {
+        // eslint-disable-next-line no-console
```
mrgrain, line 281: allow in this case

rix0rrr (approved):

PR #442, opened 2025-05-01
feat(cli,toolkit-lib): drift detection via `cdk drift` and `toolkit.drift()`

Leo10Gama:
Added an additional action and command `toolkit.drift()` and `cdk drift`, which invokes drift detection. Prints an output of the drift results listing out any resources in the stack that may have drifted.

Tested via `yarn build && yarn test` and adding in a few integ tests.

Running the command locally shows the following. With one resource drifted:
image

With no resources drifted: (image outdated)
image

---


rix0rrr:
Thanks! This is a great feature to have!

I understand that we're detecting some kind of "diff" from a user point of view, but I'm very put-off by drift detection being part of `cdk diff`. That's just not what CDK diff historically did, and it's confusing to me. I wonder if it will be confusing to more people?

```
┌────────────┐            ╔═════════════╗            ┌────────────┐
│            │    diff    ║             ║    drift   │            │
│  Template  │◀──────────▶║    Stack    ║◀──────────▶│  Reality   │
│            │            ║             ║            │            │
└────────────┘            ╚═════════════╝            └────────────┘
                                                                   
                   ^                                               
               cdk diff                                            
            does this today                                        
```

Can we make this its own subcommand to begin with? `cdk drift` ?

> Tested via yarn build && yarn test, which both compile.

Surely there's more? 🥹 An integ test would be nice.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -466,6 +468,86 @@ export async function stabilizeStack(
...
+      return response;
+    }
+
+    if (response.DetectionStatus === 'DETECTION_FAILED') {
+      throw new ToolkitError(`Drift detection failed: ${response.DetectionStatusReason}`);
+    }
+
+    if (attempts++ > 30) {
+      throw new ToolkitError('Drift detection timed out after 30 attempts');
+    }
```
rix0rrr, line 540: This is a bit hard to reason about, what with the backoff and all. What I'd recommend is doing a deadline-driven loop.

Something like this:

```ts
const deadline = Date.now() + SOME_TIME_THAT_SEEMS_REASONABLE;
while (true) {
  if (Date.now() > deadline) {
    throw new Error(`Took more than ${SOME_TIME_THAT_SEEMS_REASONABLE}!`);
  }

  // Normal work...
}
```

Advantages:

* Need to keep track of `attempts`
* Very clear what the maximum runtime is
* Error message is more useful to users

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -466,6 +468,86 @@ export async function stabilizeStack(
...
+
+    // Calculate backoff with jitter
+    const jitter = Math.random() * 1000;
```
rix0rrr, line 543: Jitter is usually calculated over the entire potentially delay period.

Here are the two different methods of calculating sleep (yours and the standard one):

image

For the first one (`baseDelay + jitter(1s)`), you will see that as `baseDelay` grows the jitter will become a negligible part of the total sum, so you might as well not have it.

(Since you have a `maxDelay`, also make sure to jittering over that as well)

-----

There is a standard way of calculating exponential backoff. Don't try to outsmart that standard way. 

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
```diff
@@ -329,6 +363,36 @@ export class DiffFormatter {
     const formattedDiff = stream.toString();
     return { formattedDiff, permissionChangeType: this.permissionType() };
   }
+
+  public formatStackDrift(options: FormatStackDriftOptions): FormatStackDriftOutput {
```
rix0rrr, line 367: Do we get information about resources that don't support drift detection anywhere? 

I wonder if this gives a false sense of confidence over resources that don't support the operation at all, and whether we should tell users that we didn't check 'em.

packages/@aws-cdk/cloudformation-diff/lib/format.ts
```diff
@@ -19,6 +20,16 @@ export interface FormatStream extends NodeJS.WritableStream {
...
 
+/**
+ * Resource drift status types from CloudFormation
+ */
+export enum ResourceDriftStatus {
+  MODIFIED = 'MODIFIED',
+  DELETED = 'DELETED',
+  NOT_CHECKED = 'NOT_CHECKED',
+  IN_SYNC = 'IN_SYNC',
+}
```
mrgrain, line 31: We prefer discriminated string unions for internal types these days. Also the enum seems unused outside this file, so no need to export

```suggestion
type ResourceDriftStatus = 'MODIFIED' | 'DELETED' | 'NOT_CHECKED' | 'IN_SYNC';
```

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk.ts
```diff
@@ -88,6 +88,14 @@ import type {
...
   UpdateTerminationProtectionCommandOutput,
   StackSummary,
+  DescribeStackDriftDetectionStatusCommandInput,
+  DescribeStackDriftDetectionStatusCommandOutput,
+  DescribeStackResourceDriftsCommandOutput,
+  DetectStackDriftCommandInput,
+  DetectStackDriftCommandOutput,
+  DetectStackResourceDriftCommandInput,
+  DetectStackResourceDriftCommandOutput,
+  DescribeStackResourceDriftsCommandInput,
```
mrgrain, line 98: These surely will require new permissions? If not can you make a note why? Or are you preparing a corresponding change to the bootstrap stack?

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -466,6 +468,86 @@ export async function stabilizeStack(
...
+ * Wait for a drift detection operation to complete
+ */
+async function waitForDriftDetection(
+  cfn: ICloudFormationClient,
+  ioHelper: IoHelper,
+  driftDetectionId: string,
+): Promise<DescribeStackDriftDetectionStatusCommandOutput | undefined> {
+  await ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(format('Waiting for drift detection %s to complete...', driftDetectionId)));
+
+  const maxDelay = 30_000; // 30 seconds max delay
```
mrgrain, line 521: This is an awwful long time without any visual status update.

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
mrgrain: It's quite limited these days what is not supported, but yes.

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
```diff
@@ -329,6 +363,36 @@ export class DiffFormatter {
...
+    const stream = new StringWriteStream();
+    let driftCount = 0;
+
+    if (!this.driftResults?.StackResourceDrifts) {
+      return { formattedDrift: '', numResourcesWithDrift: 0 };
+    }
```
mrgrain, line 373: These seems incorrect. It's not `0` resources with drift - it's unkown.

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
```diff
@@ -60,10 +77,15 @@ interface DiffFormatterProps {
    * Includes the old/current state of the stack as well as the new state.
    */
   readonly templateInfo: TemplateInfo;
+
+  /**
+   * The results of stack drift
+   */
+  readonly driftResults?: DescribeStackResourceDriftsCommandOutput;
```
mrgrain, line 84: why not in template info?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -264,6 +258,19 @@ export class CdkToolkit {
...
+
+        if (options.detectDrift) {
+          const env = await this.props.deployments.resolveEnvironment(stack);
+          const cfn = (await this.props.sdkProvider.forEnvironment(env, Mode.ForReading)).sdk.cloudFormation();
+
+          driftResults = await detectStackDrift(
+            cfn,
+            asIoHelper(this.ioHost, 'diff'),
+            stack.stackName,
+          );
```
mrgrain, line 271: This typically goes into a helper that receives `this.props.deployments` and `this.props.sdkProvider`

mrgrain (changes requested):
Also needs integ tests and added to `toolkit-lib` (see Rico's comments).

If we are adding this as a new action/command, the action code should live in toolkit-lib and CLI just be calling it.

GavinZZ:
> Thanks! This is a great feature to have!
> 
> I understand that we're detecting some kind of "diff" from a user point of view, but I'm very put-off by drift detection being part of `cdk diff`. That's just not what CDK diff historically did, and it's confusing to me. I wonder if it will be confusing to more people?
> 
> ```
> ┌────────────┐            ╔═════════════╗            ┌────────────┐
> │            │    diff    ║             ║    drift   │            │
> │  Template  │◀──────────▶║    Stack    ║◀──────────▶│  Reality   │
> │            │            ║             ║            │            │
> └────────────┘            ╚═════════════╝            └────────────┘
>                                                                    
>                    ^                                               
>                cdk diff                                            
>             does this today                                        
> ```
> 
> Can we make this its own subcommand to begin with? `cdk drift` ?

Let's discuss this further during Tuesday's meeting. While I'm open to making this a separate command, I'd like to explain my original reasoning for including it in `cdk diff`:

I envisioned `cdk diff` as a broader capability for detecting any differences affecting a deployed stack, whether those differences:
- Come from changes in local CDK code
- Result from manual modifications to deployed resources through console or CLI (Drifts)

However, I recognize your point about the historical context of `cdk diff` and the potential for user confusion. This would be a good topic for us to explore during the meeting, weighing the benefits of command consolidation against clarity and user expectations.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
Leo10Gama: Good point, and thanks for the quick crash course! I think in conjunction with your above comment (re: keeping track of attempts) that simpler deadline-driven loop approach makes a lot more sense. I can revamp this chunk of the code

packages/@aws-cdk/cloudformation-diff/lib/format.ts
Leo10Gama: Actually, looking further into it, we can probably just import the `StackResourceDriftStatus` directly from the SDK instead of redefining it here, which might be cleaner still

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
Leo10Gama: I had debated on including it in there vs outside, and my reasoning was that it wasn't as closely related to the template diff things being computed to warrant grouping them. Though going through and refactoring it now, I can see how combining them might make more sense in the grand scheme of things, and might make some of the logic in the toolkit-lib simpler as well, since it seems like that's meant to return the template info only

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk.ts
Leo10Gama: They will need new permissions, yes. Specifically, after looking into it, it seems I'll need to add these:
```
cloudformation:DescribeStackDriftDetectionStatus
cloudformation:DescribeStackResourceDrifts
cloudformation:DetectStackDrift
cloudformation:DetectStackResourceDrift
```
I am preparing to make those changes to the PR as well to update the bootstrap stack.

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
Leo10Gama: Closest thing I'm able to find about resources without drift detection support is [this doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import-supported-resources.html). I tried testing locally with one of these values (specifically `AWS::EC2::SubnetRouteTableAssociation` since it's only listed to not support drift detection) and in both the console and the command return, the resource itself just doesn't come up:
image
image


If we want to show users which resources we haven't checked, we can probably cross-reference the output list with the list of resources we can see in the stack, but since drift detection already doesn't check mention them, it might make things more confusing for users. I'd opt to leave the messages be without noting those resources

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
Leo10Gama: Good point. The chunk is supposed to catch if the drift results are empty when the method is called, which right now doesn't seem like it will happen with my implementation, but I agree it would return 0 when really drift detection hasn't run. Would it be better to return a more descriptive message, maybe with `numResourcesWithDrift: -1`?

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk.ts
Leo10Gama: Testing this locally, it looks like I don't need to add these permissions to give access for drift detection. Do you know what role `cdk diff` is using to execute these changes? It may be that it either has these permissions already or is using the admin role.

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
rix0rrr: It's not that hard, right?

If we get the list of all resources, and **remove** those that are `IN_SYNC` (or drifted), then we are left with the list of resources that we don't know? And we can print that list? 

Also, I wouldn't show the `IN_SYNC` resources at all. If the question is `cdk diff`, then the answer "here are 99 resources that are a-ok" isn't that interesting.

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk.ts
rix0rrr: I think `cdk diff` might be using the CLI's credentials, or perhaps the lookup role.

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
mrgrain: Yeah, that kind of highlights a weakness in the approach of introducing an optional prop into the formatter class constructor, to later maybe call the format function.

As far as I can tell, there is no reason this needs to be part of `DiffFormatter`. So let's just make it it's own file and api: `drift/formatter.ts` or whatever. You can probably also simplify how the class handles the formatting since the detour via `stream` is just an artifact of legacy code. Instead pushing onto an array and joining with newlines will do the same.

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
Leo10Gama: Agreed, I'm not arguing that it's difficult to implement, I just don't see the reasoning why we'd want to mention those resources. If drift detection already ignores these resources, it'd be weird for us to highlight them, since it could lead to the confusion of "why does the console say I have no drift but CDK says some resources weren't checked for drift?" 

Happy to add that implementation, I just want to understand better why this is something customers will want to see, as opposed to something that may clutter the output and make customers think there's something they need to fix, when in some cases (like the `AWS::EC2::SubnetRouteTableAssociation`) they can't even touch those resources.

Re: the `IN_SYNC` resources, the reason I had the output show all resources in the above image was for debugging purposes to show that the SDK call doesn't return the `AWS::EC2::SubnetRouteTableAssociation`. Final implementation will of course not contain these, I should've mentioned that with the image 😅

GavinZZ:
I think having a separate command like `cdk drift` and support `cdk diff --drift` meanwhile make sense to me. @Leo10Gama Just a note that if we're going to make drift a standalone CLI command, I think we also need to have a separate AppSec review and Talos engagement for the new CLI command.

mrgrain:
@Leo10Gama FYI new actions should primarily be implemented in `toolkit-lib` and than called from the CLI. See the refactor command for an example.

Leo10Gama:
Still implementing an integ test but pushing changes thus far

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
Leo10Gama: I see your point, yup. I'll whip up some changes to include that logic

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk.ts
Leo10Gama: Confirmed with AppSec, this uses the lookup role, which has the following permissions:
```yml
...
"cloudformation:Describe*",
"cloudformation:Detect*",
...
```
Hence, no bootstrap changes are necessary.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Wait for a drift detection operation to complete
+ */
+async function waitForDriftDetection(
+  cfn: IDriftCloudFormationClient,
+  ioHelper: IDriftIoHelper,
+  driftDetectionId: string,
+): Promise<DescribeStackDriftDetectionStatusCommandOutput | undefined> {
+  await ioHelper.notify(IO.DEFAULT_TOOLKIT_DEBUG.msg(format('Waiting for drift detection %s to complete...', driftDetectionId)));
+
+  const timeout = 10_000; // if takes longer than 10s, fail
```
rix0rrr, line 56: Is it realistic for a drift detection operation to finish in 10s? This seems short to me.

I think we probably want a max timeout of a minute or more, and periodically print something like `Still checking...`. We could even print

```
Checking resource drift (5/20)
```

Maybe the timeout even needs to be configurable; it's going to be the first change request coming in on this I'd imagine.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -675,6 +676,48 @@ export class CdkToolkit {
     });
   }
 
+  /**
+   * Detect infrastructure drift for the given stack(s)
+   */
+  public async drift(options: DriftCommandOptions): Promise<number> {
```
rix0rrr, line 682: I think we'd want this code to live in `@aws-cdk/toolkit-lib`, and then call it from the CLI.

Right, @mrgrain ?

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
Leo10Gama: Good point, I'll see if I can include some way to have it track resource count as they're being checked, or just something to let the user know the command is still churning away

packages/@aws-cdk/cloudformation-diff/lib/format.ts
```diff
@@ -1,9 +1,10 @@
 import { format } from 'util';
+import { StackResourceDriftStatus, type DescribeStackResourceDriftsCommandOutput } from '@aws-sdk/client-cloudformation';
 import * as chalk from 'chalk';
 import type { DifferenceCollection, TemplateDiff } from './diff/types';
 import { deepEqual } from './diff/util';
-import type { Difference, ResourceDifference } from './diff-template';
```
mrgrain, line 5: This seems unintentional.

.projenrc.ts
```diff
@@ -431,6 +431,7 @@ const cloudFormationDiff = configureProject(
     deps: [
       '@aws-cdk/aws-service-spec',
       '@aws-cdk/service-spec-types',
+      '@aws-sdk/client-cloudformation',
```
mrgrain, line 434: If this is now a runtime dep, than it cannot be a peer dep anymore.

packages/@aws-cdk/toolkit-lib/lib/index.ts
```diff
@@ -14,6 +14,7 @@ export * from './payloads';
 // Supporting acts
 export * from './api/aws-auth';
 export * from './api/cloud-assembly';
+export * from './api/drift';
```
mrgrain, line 17: These are not public APIs, let's not export them

```suggestion
```

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -0,0 +1,146 @@
...
+   */
+  readonly quiet?: boolean;
+
+  /**
+   * Whether to be verbose
+   *
+   * @default false
+   */
+  readonly verbose?: boolean;
+}
```
mrgrain, line 73: I'm aware this is just copying what `diff` currently does for legacy reasons. But I don't want to carry this forward to a brand new API. If you are really keen on having this, let's get on a call together to figure out a solution. Otherwise just drop this please.

packages/@aws-cdk/toolkit-lib/lib/api/drift/interfaces.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Interface for IO operations used in drift detection
+ */
+export interface IDriftIoHelper {
+  /**
+   * Send a notification
+   */
+  notify(msg: any): Promise<void>;
+}
```
mrgrain, line 9: I don't see the need for this being different than the regular IoHelper. AI gone rouge?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
mrgrain: Correct. See `refactor()` for how to do this correctly.

packages/@aws-cdk/cloudformation-diff/lib/format.ts
```diff
@@ -67,6 +68,89 @@ export function formatSecurityChanges(
...
+/**
+ * Renders stack drift information to the given stream
+ *
+ * @param stream The stream to write the formatted drift to
+ * @param driftResults The stack resource drifts from CloudFormation
+ * @param allStackResources A map of all stack resources
+ * @param verbose Whether to output more verbose text (include undrifted resources)
+ * @param logicalToPathMap A map from logical ID to construct path
+ */
+export function formatStackDriftChanges(
```
mrgrain, line 80: I don't think this function should live in this package since it's not used in `@aws-cdk/cloudformation-diff` it at all. We can move this to `toolkit-lib`.  WDYT @rix0rrr ?

rix0rrr: Fair point. I guess this now lives with the other formatting code, but I do in general like scoping code to the smallest scope possible. Which means a private function in `toolkit-lib` is better than a shared function in `cloudformation-diff` (that's only intended to be used from `toolkit-lib`).

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
Leo10Gama: Yup, was mainly just copying over the flags and functionality from`diff`. I'm not particularly attached to them if that's not the direction we want for the new API, so I'm happy to drop them. Though in the case of `verbose`, I do want to keep that functionality of showing all resources, including those unchecked (addressing [this concern](https://github.com/aws/aws-cdk-cli/pull/442#discussion_r2081366824)) but maybe to prevent overloading the flag, we can rename `--verbose` to `--show-unchecked` in this command, or something along those lines?

packages/@aws-cdk/toolkit-lib/lib/api/drift/interfaces.ts
mrgrain: This change will become clearer once the function is moved to the `toolkit-lib` module. We no longer conditionally emit output. Instead, all output is emitted as messages with an appropriate log level. If you need to support additional output when running `cdk drift -v`, you can assign those messages a `debug` log level.

Your code appears to have 4 sections, which I would classify like this:
- unchanged -> debug
- unchecked -> debug
- modified -> result
- deleted -> result

Leo10Gama:
AppSec review signed off

.projenrc.ts
```diff
@@ -810,7 +808,7 @@ const toolkitLib = configureProject(
     },
     tsJestOptions: {
       transformOptions: {
-        isolatedModules: false, // we use the respective tsc setting
+        isolatedModules: true,
```
mrgrain, line 811: pls revert

packages/@aws-cdk/cloudformation-diff/lib/format.ts
```diff
@@ -2,7 +2,7 @@ import { format } from 'util';
 import * as chalk from 'chalk';
 import type { DifferenceCollection, TemplateDiff } from './diff/types';
 import { deepEqual } from './diff/util';
-import type { Difference, ResourceDifference } from './diff-template';
+import type { ResourceDifference, Difference } from './diff-template';
```
mrgrain, line 5: pls revert

.projenrc.ts
```diff
@@ -426,12 +426,10 @@ const cloudFormationDiff = configureProject(
     devDeps: [
       'fast-check',
     ],
-    peerDeps: [
-      '@aws-sdk/client-cloudformation@^3',
-    ],
     deps: [
       '@aws-cdk/aws-service-spec',
       '@aws-cdk/service-spec-types',
+      '@aws-sdk/client-cloudformation',
```
mrgrain, line 432: this change and removal from peerDep is not needed anymore. pls revert

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-cdk-drift---fail-throws-when-drift-is-detected.integtest.ts
mrgrain, line 1: please move these tests to a directory called `drift` (see other tests)

packages/@aws-cdk/toolkit-lib/lib/actions/drift/index.ts
```diff
@@ -0,0 +1,81 @@
...
+   * Criteria for selecting stacks to check for drift
+   */
+  readonly stacks: StackSelector;
+
+  /**
+   * Whether to fail with exit code 1 if drift is detected
+   *
+   * @default false
+   */
+  readonly fail?: boolean;
```
mrgrain, line 14: This is not a toolkit-lib option, but a CLI feature. please remove here.

packages/@aws-cdk/toolkit-lib/lib/actions/drift/index.ts
```diff
@@ -0,0 +1,81 @@
...
+   */
+  readonly numResourcesUnchecked?: number;
+
+  /**
+   * Complete formatted drift
+   */
+  readonly formattedDrift: FormattedDrift;
+}
+
+export interface DriftCommandResult {
```
mrgrain, line 66: Looks like these are the same, so let's just use `DriftResult`

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -0,0 +1,87 @@
...
+  stackName: string,
+): Promise<DescribeStackResourceDriftsCommandOutput> {
+  await ioHelper.notify(IO.CDK_TOOLKIT_I4506.msg(format('Starting drift detection for stack %s...', stackName)));
+
+  // Start drift detection
+  const driftDetection = await cfn.detectStackDrift({
+    StackName: stackName,
+  });
+
+  await ioHelper.notify(IO.CDK_TOOLKIT_I4506.msg(format('Detecting drift for stack %s...', stackName)));
```
mrgrain, line 28: Is this good UX? Seems like we are just printing a non actionable message twice. And as soon as we are hitting `waitForDriftDetection` we are sending a third message that's also just "Waiting for"

I'd suggest to keep the first message but use the wording if the second.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Wait for a drift detection operation to complete
+ */
+async function waitForDriftDetection(
+  cfn: ICloudFormationClient,
+  ioHelper: IoHelper,
+  driftDetectionId: string,
+): Promise<DescribeStackDriftDetectionStatusCommandOutput | undefined> {
+  await ioHelper.notify(IO.CDK_TOOLKIT_I4506.msg(format('Waiting for drift detection %s to complete...', driftDetectionId)));
```
mrgrain, line 57: Can we skip the initial waiting message? We just printed a message that we are starting the drift.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @param ioHelper - helper for IO operations
+ * @param stackName - the name of the stack to check for drift
+ * @returns the CloudFormation description of the drift detection results
+ */
+export async function detectStackDrift(
+  cfn: ICloudFormationClient,
+  ioHelper: IoHelper,
+  stackName: string,
+): Promise<DescribeStackResourceDriftsCommandOutput> {
+  await ioHelper.notify(IO.CDK_TOOLKIT_I4506.msg(format('Starting drift detection for stack %s...', stackName)));
```
mrgrain, line 21: Please either include a useful payload here (maybe `StackSelectionDetails`?) or make it a default message.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
mrgrain: Is the driftDetectionId useful here? I feel like this would be better as a single debug or trace message right after we get this.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
mrgrain: Actually the second message be turned into a debug containing the StackDriftDetectionId

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
mrgrain, line 57: Make this a default message pls

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (response.DetectionStatus === 'DETECTION_FAILED') {
+      throw new ToolkitError(`Drift detection failed: ${response.DetectionStatusReason}`);
+    }
+
+    if (Date.now() > deadline) {
+      throw new ToolkitError(`Drift detection failed: Timed out after ${timeout / 1000} seconds.`);
+    }
+
+    if (Date.now() > checkIn) {
+      await ioHelper.notify(IO.CDK_TOOLKIT_I4506.msg('Waiting for drift detection to complete...'));
```
mrgrain, line 82: make this a default message

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -0,0 +1,87 @@
...
+      checkIn = Date.now() + timeBetweenOutputs;
+    }
+    await new Promise(resolve => setTimeout(resolve, 500));
```
mrgrain, line 85: this seems like it's doing nothing?

packages/@aws-cdk/toolkit-lib/test/api/drift/drift.test.ts
```diff
@@ -0,0 +1,640 @@
...
+  return {
+    ...originalModule,
+    IO: {
+      ...originalModule.IO,
+      DEFAULT_TOOLKIT_DEBUG: {
+        msg: jest.fn().mockReturnValue('mocked-message'),
+      },
+    },
+  };
+});
```
mrgrain, line 25: There is absolutely no need to mock this module. Look for `TestIoHost` if you need to assert messages.

packages/@aws-cdk/toolkit-lib/test/api/drift/drift.test.ts
```diff
@@ -0,0 +1,640 @@
...
+
+    // WHEN & THEN
+    await expect(detectStackDrift(mockCfn, mockIoHelper, stackName))
+      .rejects.toThrow(`Drift detection failed: ${failureReason}`);
+  });
+});
+
+describe('formatStackDrift', () => {
+  let mockIoHelper: IoHelper;
+  let mockNewTemplate: cxapi.CloudFormationStackArtifact;
```
mrgrain, line 292: Checkout `TestIoHost` for a much easier way of mocking and testing.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -339,6 +339,16 @@ export async function makeConfig(): Promise<CliConfig> {
...
         },
       },
+      drift: {
+        description: 'Detect drifts in the given CloudFormation stack(s)',
+        arg: {
+          name: 'STACKS',
+          variadic: true,
+        },
+        options: {
+          fail: { type: 'boolean', desc: 'Fail with exit code 1 if drift is detected' },
```
mrgrain, line 349: please add a default

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,85 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+    const selectStacks = options.stacks ?? ALL_STACKS;
+    await using assembly = await assemblyFromSource(ioHelper, cx);
+    const stacks = await assembly.selectStacksV2(selectStacks);
+
+    let drifts = 0;
+    let uncheckedResources = 0;
+    const output = [];
+
+    for (const stack of stacks.stackArtifacts) {
+      const cfn = (await (await this.sdkProvider('drift')).forEnvironment(stack.environment, Mode.ForReading)).sdk.cloudFormation();
```
mrgrain, line 395: Can this be outside the loop?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,85 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+
+      // Get all stack resources if we want to show all
+      const allStackResources = new Map<string, string>();
+      Object.keys(stack.template.Resources || {}).forEach(id => {
+        const resource = stack.template.Resources[id];
+        allStackResources.set(id, resource.Type);
+      });
+
+      if (!driftResults) {
+        await ioHelper.notify(IO.CDK_TOOLKIT_I4505.msg(`${chalk.bold(stack.displayName)}: unable to detect drift`));
```
mrgrain, line 411: make this a default message if we really need it

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,85 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+        continue;
+      }
+
+      const formatter = new DriftFormatter({
+        ioHelper: asIoHelper(this.ioHost, 'drift'),
```
mrgrain, line 416: You already have one
```suggestion
        ioHelper,
```

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -0,0 +1,283 @@
...
+  constructor(props: DriftFormatterProps) {
+    this.stack = props.stack;
+    this.driftResults = props.driftResults;
+    this.allStackResources = props.allStackResources;
+  }
+
+  /**
+   * Format the stack drift detection results
+   */
+  public formatStackDrift(): DriftResult {
```
mrgrain, line 76: Please keep internal types and public types separate

packages/@aws-cdk/toolkit-lib/lib/actions/drift/index.ts
mrgrain: Okay I see now `DriftResult` is an internal type and `DriftCommandResult` is public. 

Make sure `DriftResult` lives with the formatter (and probably rename it slightly). 

For the public type `DriftResult` is the better name.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,85 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+        output.push(driftOutput.formattedDrift.deleted);
+      }
+      if (driftOutput.formattedDrift.finalResult) {
+        await ioHelper.notify(IO.CDK_TOOLKIT_I4500.msg(driftOutput.formattedDrift.finalResult, driftOutput));
+        output.push(driftOutput.formattedDrift.finalResult);
+      }
+      drifts += driftOutput.numResourcesWithDrift === -1 ? 0 : driftOutput.numResourcesWithDrift || 0;
+      uncheckedResources += driftOutput.numResourcesUnchecked === -1 ? 0 : driftOutput.numResourcesUnchecked || 0;
+      output.push(driftOutput.formattedDrift);
+    }
```
mrgrain, line 451: We don't need to send the same payload multiple times and also it's the same stuff as in the message already.

Header and footer (finalResult?) can just be default messages without code and payload, they are not super interesting. Let's just turn them all into default messages.

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
Leo10Gama: When you say default messages, are you referring to making `CDK_TOOLKIT_I4500` take no payload? The default messages seem to have been removed from `IO`

Leo10Gama: Nevermind, discovered it as `ioHelper.defaults.<loggingLevel>`

packages/@aws-cdk/integ-runner/THIRD_PARTY_LICENSES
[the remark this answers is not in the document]
Leo10Gama: Honestly not sure how this came about 😅 was a result of the `yarn build` script, and I'm not sure how to get it fixed

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,79 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      });
+
+      const driftOutput = formatter.formatStackDrift();
+
+      if (driftOutput.sections?.stackHeader) {
+        await ioHelper.defaults.result(driftOutput.sections?.stackHeader);
+        output.push(driftOutput.sections?.stackHeader);
+      }
+      if (driftOutput.sections?.unchanged) {
+        await ioHelper.notify(IO.CDK_TOOLKIT_I4500.msg(driftOutput.sections?.unchanged));
```
mrgrain, line 424: Either default message or a payload

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -0,0 +1,305 @@
...
+        output.push(formatterOutput[section]);
+      }
+    });
+
+    if (drifts.length === 0) {
+      const finalResult = chalk.green('No drift detected\n');
+      output.push(finalResult);
+      return {
+        numResourcesWithDrift: 0,
+        numResourcesUnchecked: this.allStackResources ? this.allStackResources.size - this.driftResults.StackResourceDrifts.length : -1,
```
mrgrain, line 115:
```suggestion
        numResourcesUnchecked: this.allStackResources ? this.allStackResources.size - this.driftResults.StackResourceDrifts.length : undefined,
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,79 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+        ioHelper,
+        stack.stackName,
+      );
+
+      // Get all stack resources if we want to show all
+      const allStackResources = new Map<string, string>();
+      Object.keys(stack.template.Resources || {}).forEach(id => {
+        const resource = stack.template.Resources[id];
+        allStackResources.set(id, resource.Type);
+      });
```
mrgrain, line 408: This should be movable to inside the DriftFormatter

packages/@aws-cdk/toolkit-lib/lib/actions/drift/index.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Formatted sections of the drift results
+   */
+  readonly sections?: FormattedDrift;
+
+  /**
+   * Complete formatted drift
+   */
+  readonly formattedDrift: string;
```
mrgrain, line 61:
```suggestion
  /**
   * Complete formatted drift
   */
  readonly formattedDrift: FormattedDrift;
```

packages/@aws-cdk/toolkit-lib/lib/actions/drift/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * The header, containing the stack name
+   */
+  readonly stackHeader?: string;
+
+  /**
+   * The final results (summary) of the drift results
+   */
+  readonly finalResult?: string;
+}
```
mrgrain, line 40:
```suggestion
export interface FormattedDrift {
  /**
   * Resources that have not changed
   */
  readonly unchanged?: string;

  /**
   * Resources that were not checked for drift
   */
  readonly unchecked?: string;

  /**
   * Resources with drift
   */
  readonly modified?: string;

  /**
   * Resources that have been deleted (drift)
   */
  readonly deleted?: string;
}
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,79 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+        output.push(driftOutput.sections?.finalResult);
+      }
+      drifts += driftOutput.numResourcesWithDrift === -1 ? 0 : driftOutput.numResourcesWithDrift || 0;
+      uncheckedResources += driftOutput.numResourcesUnchecked === -1 ? 0 : driftOutput.numResourcesUnchecked || 0;
+    }
+
+    return {
+      numResourcesWithDrift: drifts,
+      numResourcesUnchecked: uncheckedResources,
+      formattedDrift: output.join('\n\n'),
```
mrgrain, line 450:
```suggestion
      formattedDrift: {
        unchanged: driftOutput. sections?.unchanged,
        unchecked: driftOutput. sections?.unchecked,
        modified: driftOutput. sections?.modified,
        deleted: driftOutput. sections?.deleted,
      },
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -375,6 +378,79 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      if (driftOutput.sections?.deleted) {
+        await ioHelper.notify(IO.CDK_TOOLKIT_I4503.msg(driftOutput.sections?.deleted, { stacks: selectStacks }));
+        output.push(driftOutput.sections?.deleted);
+      }
+      if (driftOutput.sections?.finalResult) {
+        await ioHelper.defaults.result(driftOutput.sections?.finalResult);
+        output.push(driftOutput.sections?.finalResult);
+      }
+      drifts += driftOutput.numResourcesWithDrift === -1 ? 0 : driftOutput.numResourcesWithDrift || 0;
+      uncheckedResources += driftOutput.numResourcesUnchecked === -1 ? 0 : driftOutput.numResourcesUnchecked || 0;
```
mrgrain, line 444: not -1

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: maybe just add a comment pointing out that `stack.environment` might be different

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    if (Date.now() > deadline) {
+      throw new ToolkitError(`Drift detection failed: Timed out after ${timeout / 1000} seconds.`);
+    }
+
+    if (Date.now() > checkIn) {
+      await ioHelper.defaults.trace('Waiting for drift detection to complete...');
+      checkIn = Date.now() + timeBetweenOutputs;
+    }
+    await new Promise(resolve => setTimeout(resolve, 500));
```
mrgrain, line 82:
```suggestion
    const timeBetweenApiCall = 1000;
    await new Promise(resolve => setTimeout(resolve, timeBetweenApiCall));
```

mrgrain (approved):

PR #444, opened 2025-05-02
fix(toolkit-lib): fromCdkApp fails if working directory is given and outdir is relative

rix0rrr:
If a working directory is given and outdir is relative, then the CDK app synths relative to the supplied working directory, but the toolkit tries to load the assembly relative to the process' `cwd` and fails.

Also, the documentation of `fromCdkApp` falsely claimed that by default it will create a temporary directory and clean it up. It does not, by default it emits into a relative `cdk.out` directory which won't get cleaned up. The `cdk.out` directory also was not entirely consistently evaluated, so we specify that it must be evaluated against `workingDirectory`.

While we're at it, immediately add a flag that allows explicit directories to be cleaned up as well so that our tests (which explicitly specify temporary directories and so they don't get cleaned up) don't spam all the way into the system's tempdir.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
@@ -80,6 +80,13 @@ export interface AssemblySourceProps {
    * Options to configure loading of the assembly after it has been synthesized
    */
   readonly loadAssemblyOptions?: LoadAssemblyOptions;
+
+  /**
+   * Delete the `outdir` when the assembly is disposed
+   *
+   * @default - `true` if `outdir` is not given, `false` otherwise
+   */
+  readonly disposeOutdir?: boolean;
 }
 
 /**
--- a/packages/@aws-cdk/toolkit-lib/test/_helpers/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/_helpers/index.ts
@@ -1,7 +1,7 @@
 import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
-import type { AssemblyDirectoryProps, FromCdkAppOptions, ICloudAssemblySource } from '../../lib';
+import type { AssemblyBuilder, AssemblyDirectoryProps, FromCdkAppOptions, ICloudAssemblySource } from '../../lib';
 import { ToolkitError } from '../../lib';
 import type { CloudAssemblySourceBuilder } from '../../lib/api/cloud-assembly/private';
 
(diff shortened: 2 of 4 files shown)


packages/@aws-cdk/toolkit-lib/test/_helpers/index.ts
```diff
@@ -8,19 +8,31 @@ import type { CloudAssemblySourceBuilder } from '../../lib/api/cloud-assembly/pr
...
-  const appPath = fixturePath(name, 'app.js');
-  if (!fs.existsSync(appPath)) {
-    throw new ToolkitError(`App Fixture ${name} does not exist in ${appPath}`);
-  }
-  const app = `cat ${appPath} | node --input-type=module`;
-  return toolkit.fromCdkApp(app, {
+/**
+ * Return config we can send into `fromCdkApp` to execute a given app fixture
+ */
+export function appFixtureCommand(name: string) {
```
mrgrain, line 22:
```suggestion
export function appFixtureConfig(name: string) {
```

packages/@aws-cdk/toolkit-lib/test/api/cloud-assembly/source-builder.test.ts
```diff
@@ -114,6 +116,24 @@ describe('fromCdkApp', () => {
...
+      outdir: 'relative.dir',
+    });
+
+    try {
+      await using assembly = await cx.produce();
+
+      // THEN - synth succeeds
+      expect(assembly.cloudAssembly.stacksRecursively.map(s => s.hierarchicalId)).toEqual(['Stack1', 'Stack2']);
+    } finally {
+      await fs.rm(path.join(app.workingDirectory, 'relative.dir'), { force: true, recursive: true });
```
mrgrain, line 133: This should be automatic with `deleteOutdir`, right?

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
```diff
@@ -80,6 +80,13 @@ export interface AssemblySourceProps {
    * Options to configure loading of the assembly after it has been synthesized
    */
   readonly loadAssemblyOptions?: LoadAssemblyOptions;
+
+  /**
+   * Delete the `outdir` when the assembly is disposed
+   *
+   * @default - `true` if `outdir` is not given, `false` otherwise
+   */
+  readonly deleteOutdir?: boolean;
```
mrgrain, line 89: Can we have a test for this?

packages/@aws-cdk/toolkit-lib/test/api/cloud-assembly/source-builder.test.ts
rix0rrr: Good point!

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
rix0rrr: Yes!

mrgrain (approved):

PR #447, opened 2025-05-02
feat(cli): allow change set imports resources that already exist

GavinZZ:
Allow `cdk diff` command to create a change set that imports existing resources.

The current `cdk diff` command implicitly calls
CloudFormation change set creation, providing high-level details such as "add", "delete", "modify", "import",
and etc. like the following:

```s
$ cdk diff
[-] AWS::DynamoDB::Table MyTable orphan
[+] AWS::DynamoDB::GlobalTable MyGlobalTable add
```

However, when the resource is meant to be imported, the `cdk diff` command still shows this as add. Adding `cdk diff --import-existing-resources` flag to show the new resource being imported instead of `add`.

```s
$ cdk diff --import-existing-resources
[-] AWS::DynamoDB::Table MyTable orphan
[←] AWS::DynamoDB::GlobalTable MyGlobalTable import
```

Here is the underlying CFN change set JSON output
```json
[
  {
    "type": "Resource",
    "resourceChange": {
      "action": "Import", # NOTE THAT THIS SHOWS "Import"
      "logicalResourceId": "MyTable794EDED1",
      "physicalResourceId": "DemoStack-MyTable794EDED1-11W4MR8VZ0UPE",
      "resourceType": "AWS::DynamoDB::GlobalTable",
      "replacement": "True",
      "scope": [],
      "details": [],
      "afterContext": "..."
    }
  },
  {
    "type": "Resource",
    "resourceChange": {
      "policyAction": "Retain", # Note that this is "Retain"
      "action": "Remove",
      "logicalResourceId": "MyTable794EDED1",
      "physicalResourceId": "DemoStack-MyTable794EDED1-11W4MR8VZ0UPE",
      "resourceType": "AWS::DynamoDB::Table",
      "scope": [],
      "details": [],
      "beforeContext": "..."
    }
  }
]
```

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-cdk-diff--import-existing-resources-shows-import.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-cdk-diff--import-existing-resources-shows-import.integtest.ts
@@ -19,6 +19,7 @@ integTest(
       },
     });
 
+    console.log('first diff' + diff);
     // Assert there are no changes and diff shows import
     expect(diff).not.toContain('There were no differences');
     expect(diff).toContain('[←]');
@@ -31,6 +32,7 @@ integTest(
       },
     });
 
+    console.log('second diff' + diff);
     // Assert there are no changes and diff shows add
     expect(diff).not.toContain('There were no differences');
     expect(diff).toContain('[+]');
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-cdk-diff--import-existing-resources-shows-import.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-cdk-diff--import-existing-resources-shows-import.integtest.ts
@@ -1,0 +1,45 @@
+import { integTest, withSpecificFixture } from '../../lib';
+
+jest.setTimeout(2 * 60 * 60_000); // Includes the time to acquire locks, worst-case single-threaded runtime
+
+integTest(
+  'cdk diff --import-existing-resources show resource being imported',
+  withSpecificFixture('import-app', async (fixture) => {
+    // GIVEN
+    await fixture.cdkDeploy('base-1', {
+      modEnv: {
+        VERSION: 'v1',
+      },
+    });
+
+    // THEN
+    let diff = await fixture.cdk(['diff --import-existing-resources', fixture.fullStackName('base-1')], {
+      modEnv: {
+        VERSION: 'v2',
+      },
+    });
+
+    // Assert there are no changes and diff shows import
+    expect(diff).not.toContain('There were no differences');
+    expect(diff).toContain('[←]');
+    expect(diff).toContain('import');
+
+    // THEN
+    diff = await fixture.cdk(['diff', fixture.fullStackName('base-1')], {
+      modEnv: {
+        VERSION: 'v2',
+      },
+    });
+
+    // Assert there are no changes and diff shows add
+    expect(diff).not.toContain('There were no differences');
+    expect(diff).toContain('[+]');
+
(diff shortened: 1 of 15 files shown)


GavinZZ:
Pending adding unit and integ tests

mrgrain (changes requested):
Please also add this feature to `toolkit-lib/toolkit/toolkit.ts`

rix0rrr:
Is there any downside to just switching this on by default?

rix0rrr:
Do we also already have this feature in `cdk deploy` ?

GavinZZ:
> Is there any downside to just switching this on by default?

I think it's not enabled by default because this is also the CFN behaviour and CDK deploy behaviour. CDK deploy already supports `--import-existing-resources` option which will create the change set with `--import-existing-resources` flag. This is not the default behaviour to use by `cdk deploy` and I think it makes sense to do the same in `cdk diff`.

mrgrain:
> Do we also already have this feature in `cdk deploy` ?

yes

mrgrain:
> > toolkit-lib/toolkit/toolkit.ts
> 
> For my understanding, is this for programmatic SDK call?

Correct, that's for the Programmatic Toolkit. Unfortunately we have some unavoidable duplication at the moment.

mrgrain:
> Is there any downside to just switching this on by default?

The main risk is if the real resource and the CDK definition don't match. Would have to test, but I think CFN can get in a weird state then or at least it's unexpected if an existing (production) resource is changed.

packages/@aws-cdk/toolkit-lib/lib/actions/diff/index.ts
```diff
@@ -129,4 +129,11 @@ export interface DiffOptions {
    * @deprecated implement in IoHost
    */
   readonly securityOnly?: boolean;
+
+  /**
+   * Whether or not the change set imports resources that already exist
+   *
+   * @default false
+   */
+  readonly importExistingResources?: boolean;
```
mrgrain, line 138: This needs to go into `method` and `DiffMethod` (and ultimately `ChangeSetDiffOptions`) because it's only available for changesets. See deploy action and `ChangeSetDeploymentMethod`.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-cdk-diff--import-existing-resources-shows-import.integtest.ts
```diff
@@ -0,0 +1,45 @@
...
+    expect(diff).toContain('import');
+
+    // THEN
+    diff = await fixture.cdk(['diff', fixture.fullStackName('base-1')], {
+      modEnv: {
+        VERSION: 'v2',
+      },
+    });
+
+    // Assert there are changes
```
mrgrain, line 34: Like this?
```suggestion
    // Assert there are no changes
```

For my understanding: Is this test just trying to show the difference in diff output between `diff` and `diff --import-existing-resources`?

release.json
mrgrain, line 1: Pls remove

mrgrain (changes requested):
Almost there.

mrgrain (changes requested):

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-cdk-diff--import-existing-resources-shows-import.integtest.ts
GavinZZ: Yes, the first diff expects to show resource being imported with `[←]` and `import` substring. The second diff expects `[+]` to make sure `--import-existing-resoruces` is working.

mrgrain (approved):

tmokmss:
I just noticed this feature got merged (finally after https://github.com/aws/aws-cdk/pull/32831)! Great job team! I'll write a blog about this 😄

Issue #418, opened 2025-04-27
(toolkit-lib): build fails due to missing tmp-toolkit-helpers types


corymhall:
### Describe the bug

`tsc --build` fails on projects trying to use `@aws-cdk/tookit-lib`. Error message is:

```
node_modules/@aws-cdk/toolkit-lib/lib/actions/bootstrap/index.d.ts:3:15 - error TS2305: Module '"../../api/shared-private"' has no exported member 'Tag'.

3 import type { Tag } from '../../api/shared-private';
                ~~~

node_modules/@aws-cdk/toolkit-lib/lib/actions/deploy/index.d.ts:2:15 - error TS2305: Module '"../../api/shared-private"' has no exported member 'Tag'.

2 import type { Tag } from '../../api/shared-private';
                ~~~

node_modules/@aws-cdk/toolkit-lib/lib/actions/deploy/private/deploy-options.d.ts:3:15 - error TS2305: Module '"../../../api/shared-private"' has no exported member 'CloudWatchLogEventMonitor'.

3 import type { CloudWatchLogEventMonitor } from '../../../api/shared-private';
                ~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@aws-cdk/toolkit-lib/lib/actions/deploy/private/helpers.d.ts:2:10 - error TS2305: Module '"../../../api/shared-private"' has no exported member 'HotswapPropertyOverrides'.

2 import { HotswapPropertyOverrides } from '../../../api/shared-private';
           ~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@aws-cdk/toolkit-lib/lib/actions/deploy/private/helpers.d.ts:3:15 - error TS2305: Module '"../../../api/shared-private"' has no exported member 'Deployments'.

3 import type { Deployments, WorkGraph } from '../../../api/shared-private';
                ~~~~~~~~~~~

node_modules/@aws-cdk/toolkit-lib/lib/actions/deploy/private/helpers.d.ts:3:28 - error TS2305: Module '"../../../api/shared-private"' has no exported member 'WorkGraph'.

3 import type { Deployments, WorkGraph } from '../../../api/shared-private';
                             ~~~~~~~~~

node_modules/@aws-cdk/toolkit-lib/lib/api/aws-auth/types.d.ts:2:15 - error TS2305: Module '"../shared-private"' has no exported member 'SdkProviderServices'.

2 import type { SdkProviderServices } from '../shared-private';
[...]

PR #452, opened 2025-05-05
refactor(toolkit-lib): remove shared-public and other re-export files

mrgrain:
Related to #418

---


packages/@aws-cdk/toolkit-lib/lib/index.ts
```diff
@@ -7,11 +7,13 @@ import './private/dispose-polyfill';
 
 // The main show
 export * from './toolkit';
+export * from './toolkit/toolkit-error';
```
iliapolo, line 10: Why like this and not via the previous `./toolkit` export?

packages/aws-cdk/lib/cli/root-dir.ts
```diff
@@ -1,6 +1,6 @@
 import * as fs from 'fs';
 import * as path from 'path';
-import { ToolkitError } from '../../../@aws-cdk/toolkit-lib/lib/api';
+import { ToolkitError } from '../../../@aws-cdk/toolkit-lib';
```
iliapolo, line 3: Hmm - not caused by this PR, but why isn't this just:

```ts
import { ToolkitError } from '@aws-cdk/toolkit-lib`
```

?

Where does the relative path come from?

mrgrain: Relative path as in import via the file system. We need this for imports that are not publicly exported from `@aws-cdk/toolkit-lib`. Mostly lower level code in `cdk-toolkit.ts` that is still needed while the CLI is not using `toolkit-lib`. 

I'm leaning towards keeping it consistent, and that means the relative import for now. Eventually (far future) the CLI would also just use what is publicly available, but we are not there yet.

packages/@aws-cdk/toolkit-lib/lib/index.ts
mrgrain: This one's a bit annoying. Previously `ToolkitError` et.al. where exported via `shared-public.ts`. Adding it to the barrel file `lib/toolkit/index.ts` can too easily cause circular imports, because `ToolkitError` is used all over the place. So its best to import it directly. My idea/hope was that only exporting `ToolkitError` from the top-level public export will discourage a potentially problematic import from the lower level.

iliapolo (approved):

PR #460, opened 2025-05-07
feat(cli): use explicit mapping when provided by the user

### LINKED ISSUE #378: Add support for mapping files
Support user provided mapping files that will take precedence over the on-the-fly computation of mappings.

otaviomacedo:
This PR introduces two new options to the `refactor` command: `--mapping-file` and `--revert`. When a mapping file is provided, the CLI won't do try to compute the refactor mappings, and just use what the user instructed it to. The `--revert` can only be used in conjunction with `--mapping-file`, and modifies its behavior by swapping source and destination. 

Closes #378.

---


packages/aws-cdk/README.md
```diff
@@ -1145,6 +1145,36 @@ The pattern language is the same as the one used in the `cdk deploy` command.
...
+{
+  "mappings": [
+    {
+      "source": "Foo.OldName",
+      "destination": "Bar.NewName",
+      "environment": {
+        "name": "test",
+        "account": "123456789012",
+        "region": "us-east-1"
+      }
```
rix0rrr, line 1163: Oof! If we need to pass the environment for every resource, that is going to be an annoying file to work with if the list of resources grows, right?

How about something like this:

```json
{ 
  "environments": [
    {
      "account": "123456789012",
      "region": "us-east-1",
      "resources": {
        "Bar.NewName": "Foo.OldName"
      } 
    }
  ]
}
```

(Or reverse the mapping, old-to-new might be more logical to people's minds)

rix0rrr (approved):

PR #466, opened 2025-05-07
chore: reorg eslint rules into separate files

mrgrain:
The rules file got a bit large and hard to maintain/understand what's going on. In this change we are splitting the file up into sensible categories. 

Also includes a projen upgrade to uses `@stylistic` for formatting rules and fixes a couple of rules to not use an array when they prefer to not use one (based on VSCode eslint plugin warnings). Places where some of the affected rules are disabled needed updating as well.

There are a bunch of small formatting changes to type statements. These didn't get picked up previously because the built-in eslint rules don't support TypeScript, but the `@stylistic` ones do.

The lack of any other changes to the codebase proves that these fixes are alright.

---


packages/aws-cdk/test/_helpers/jest-bufferedconsole.ts
```diff
@@ -99,7 +99,7 @@ export default class TestEnvironment extends NodeEnvironment implements JestEnvi
 type TestDescription = PartialBy<Pick<Circus.TestEntry, 'name' | 'parent'>, 'parent'>;
 
 // Utility type to make specific fields optional
-type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
```
mrgrain, line 102: This wasn't picked up before because the default eslint `semi` rules doesn't do TypeScript.

packages/aws-cdk/lib/legacy-types.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * @deprecated
  */
-export type Obj<T> = {[key: string]: T};
+export type Obj<T> = { [key: string]: T };
```
mrgrain, line 4: This wasn't picked up before because the default eslint `key-spacing` rules doesn't do TypeScript.

packages/aws-cdk/lib/legacy-aws-auth.ts
```diff
@@ -25,7 +25,7 @@ export function cached<A extends object, B>(obj: A, sym: symbol, fn: () => B): B
  * @deprecated
  */
 export interface ContextProviderPlugin {
-  getValue(args: {[key: string]: any}): Promise<any>;
+  getValue(args: { [key: string]: any }): Promise<any>;
```
mrgrain, line 28: This wasn't picked up before because the default eslint `key-spacing` rules doesn't do TypeScript.

iliapolo (approved):

PR #479, opened 2025-05-08
feat: exclude unsupported resources from refactoring

otaviomacedo:
CloudFormation has a [list] of resource types that are not available for the stack refactoring. Exclude resources of these types when detecting resource movements.

[list]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stack-refactoring.html

---


rix0rrr (commented):
Why are these not available? What if the list updates?

PR #499, opened 2025-05-15
chore: use AWS_MANAGED_KEY for integ tests

rix0rrr:
Use an AWS-managed KMS key for bucket encryption in the bootstrapping of integ tests.

This should shave ~1 minute off of the bootstrapping of every environment.

---


mrgrain (approved):

rix0rrr:
Update: a lot of the time seems to be in `npm install` of dependencies. 

```
💻 node /home/runner/work/aws-cdk-cli/aws-cdk-cli/node_modules/npm/index.js install

added 5 packages in 3m

4 packages are looking for funding
```

Might be lack of cache, or the 70 tests in parallel are overloading Verdaccio.

We might consider priming the cache as well.

mrgrain:
Sounds like there is more to do, but this is still worth doing. Will merge.

Issue #396, opened 2025-04-22
Remove `promptly` from toolkit-lib


mrgrain:
There are a few remaining of uses of the `promptly` library inside toolkit-lib. Replace them with a coded message and an IoHost feature.

PR #508, opened 2025-05-19
fix(toolkit-lib): MFA token cannot be provided through IoHost

mrgrain:
Relates to #396 

Considering this a bug fix, since it is currently not possible to integrate an IoHost with this. While this does affect the CLI, the DX is virtually unchanged. I run the auth test suite to ensure everything is still working as expected.

Before:

before

After:

after

---

### DIFF
--- a/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
+++ b/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
@@ -357,10 +357,12 @@ export class CliIoHost implements IIoHost {
       const data: {
         motivation?: string;
         concurrency?: number;
+        responseDescription?: string;
       } = msg.data ?? {};
 
       const motivation = data.motivation ?? 'User input is needed';
       const concurrency = data.concurrency ?? 0;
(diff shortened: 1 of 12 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -392,8 +394,10 @@ export class CliIoHost implements IIoHost {
 
       // Asking for a specific value
       const prompt = extractPromptInfo(msg);
-      const answer = await promptly.prompt(`${chalk.cyan(msg.message)} (${prompt.default})`, {
+      const desc = responseDescription ?? prompt.default;
+      const answer = await promptly.prompt(`${chalk.cyan(msg.message)}${desc ? ` (${desc})` : ''}`, {
         default: prompt.default,
+        trim: true,
```
mrgrain, line 400: I cannot imagine a situation we don't want to trim input. At least none of the current prompts want untrimmed.

rix0rrr (approved):

PR #513, opened 2025-05-20
refactor(toolkit-lib,cli): access default messages through IoHelper

mrgrain:
This change is in preparation of work to have all default message go through `IoHelper` instead of using `IO.DEFAULT_XYZ`. It contains a number of clean-up and alignment changes:

- `IoDefaultMessages` is now a local-only class, that is only accessed through the new `IoHelper.defaults` getter.
- Removed unused usage of `IoHelper` and `IoDefaultMessages` in `DiffFormatter`.
- Changed every use of `IoDefaultMessages` to `IoHelper.defaults`
- Made `IoDefaultMessages` methods async to be consistent with the rest, we can `void` at the place of use if we really need to
- Also fixed most of the remaining `void`'ed call-sites for the above. Only the legacy logging is now left.
- Removed unused `messages.ts` file
- Removed unused `trace` from `logging.ts` (and respective test) to not encourage further use

---


packages/aws-cdk/lib/logging.ts
```diff
@@ -129,21 +128,6 @@ export const debug = (input: LogInput<'I'>, ...args: unknown[]) => {
...
- *
- * Can be used in multiple ways:
- * ```ts
- * trace(`entered ${name} with ${args}`) // infers default info code `CDK_TOOLKIT_I000`
- * trace('method: %s, args: %j', name, args) // infers default info code `CDK_TOOLKIT_I000`
- * trace({ message: 'entered', code: 'CDK_TOOLKIT_I001' }) // specifies info code `CDK_TOOLKIT_I001`
- * trace({ message: 'method: %s', code: 'CDK_TOOLKIT_I001' }, name) // specifies info code `CDK_TOOLKIT_I001`
- * ```
- */
-export const trace = (input: LogInput<'I'>, ...args: unknown[]) => {
```
mrgrain, line 143: This was not used anymore. Removing to discourage any further use.

packages/aws-cdk/test/logging.test.ts
```diff
@@ -93,18 +93,6 @@ describe('logging', () => {
         expect.stringMatching(/^\[\d{2}:\d{2}:\d{2}\] debug message\n$/),
       );
     });
-
-    test('trace messages only show at trace level with both styles', () => {
```
mrgrain, line 97: removing test for the removed `trace` function

packages/aws-cdk/lib/cli/messages.ts
mrgrain, line 1: Temporary implementation during the code moving phase and not used anymore.

iliapolo (approved):

PR #517, opened 2025-05-21
feat(toolkit-lib)!: hotswap is now a deployment method

mrgrain:
Historically we modeled hotswap deployments in the CLI as a boolean flag on the deploy command. At some point we then added control over the procedure for not hotswappable resources: fallback to a regular CFN deployment or do nothing. Most of this design was carried over to `toolkit-lib`.

This change re-models hotswap deployments as a deployment method on the same level as change-set and direct deployments. This make sense, since hotswap also takes additional input that is otherwise ignored. For its fallback mode, the hotswap deployment method can take any of the other deployment methods (but not itself).

BREAKING CHANGE: hotswap deployments are now modeled as a deployment method. Instead of `DeployOptions.hotswap`, use `DeployOptions.deploymentMethod = { method: 'hotswap' }`. Similarly, `DeployOptions.hotswapProperties` is now `DeployOptions.deploymentMethod = { method: 'hotswap', properties: myHotswapProperties }`.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
@@ -26,14 +26,14 @@ import {
 } from './cfn-api';
 import { determineAllowCrossAccountAssetPublishing } from './checks';
 import type { DeployStackResult, SuccessfulDeployStackResult } from './deployment-result';
-import type { ChangeSetDeploymentMethod, DeploymentMethod } from '../../actions/deploy';
+import type { ChangeSetDeployment, DeploymentMethod, DirectDeployment } from '../../actions/deploy';
 import { ToolkitError } from '../../toolkit/toolkit-error';
 import { formatErrorMessage } from '../../util';
 import type { SDK, SdkProvider, ICloudFormationClient } from '../aws-auth/private';
 import type { TemplateBodyParameter } from '../cloudformation';
 import { makeBodyParameter, CfnEvaluationException, CloudFormationStack } from '../cloudformation';
 import type { EnvironmentResources, StringWithoutPlaceholders } from '../environment';
-import { HotswapMode, HotswapPropertyOverrides, ICON } from '../hotswap/common';
+import { HotswapPropertyOverrides, HotswapMode, ICON, createHotswapPropertyOverrides } from '../hotswap/common';
 import { tryHotswapDeployment } from '../hotswap/hotswap-deployments';
 import { type IoHelper } from '../io/private';
 import type { ResourcesToImport } from '../resource-import';
@@ -158,17 +158,21 @@ export interface DeployStackOptions {
    */
   readonly rollback?: boolean;
 
-  /*
+  /**
    * Whether to perform a 'hotswap' deployment.
    * A 'hotswap' deployment will attempt to short-circuit CloudFormation
    * and update the affected resources like Lambda functions directly.
    *
    * @default - `HotswapMode.FULL_DEPLOYMENT` for regular deployments, `HotswapMode.HOTSWAP_ONLY` for 'watch' deployments
+   *
+   * @deprecated  Use 'deploymentMethod' instead
    */
   readonly hotswap?: HotswapMode;
 
   /**
    * Extra properties that configure hotswap behavior
+   *
+   * @deprecated Use 'deploymentMethod' instead
    */
   readonly hotswapPropertyOverrides?: HotswapPropertyOverrides;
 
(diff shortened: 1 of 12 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -202,9 +206,31 @@ export interface DeployStackOptions {
 
 export async function deployStack(options: DeployStackOptions, ioHelper: IoHelper): Promise<DeployStackResult> {
   const stackArtifact = options.stack;
-
   const stackEnv = options.resolvedEnvironment;
 
+  let deploymentMethod = options.deploymentMethod ?? { method: 'change-set' };
+  // Honour old hotswap option
+  // @TODO remove when we don't care about this export anymore
```
iliapolo, line 213: Why not remove now?

packages/@aws-cdk/toolkit-lib/test/actions/deploy-hotswap.test.ts
```diff
@@ -33,42 +32,82 @@ beforeEach(() => {
...
-  test('does print hotswap warnings for FALL_BACK mode', async () => {
+  test('does print hotswap warnings for hotswap with fallback', async () => {
     // WHEN
     const cx = await builderFixture(toolkit, 'two-empty-stacks');
     await toolkit.deploy(cx, {
-      hotswap: HotswapMode.FALL_BACK,
+      deploymentMethod: {
+        method: 'hotswap',
+        fallback: { method: 'change-set' },
+      },
```
iliapolo, line 42: So previously the fallback method would be derived from the default of `deploymentMethod` - right? And now it seems like if a user wants to hotswap with fallback, they **must** also provide the fallback method as well. This is a slightly more verbose API and I wonder if we can instead have:

```suggestion
      deploymentMethod: {
        method: 'hotswap',
        fallback: true, // default to change-set
        // if a user wants
        fallbackMethod: 'direct'
      },
```

Also not super elegant. Your call.

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
```diff
@@ -9,6 +9,8 @@ import { Toolkit } from '../../lib/toolkit';
 import { builderFixture, disposableCloudAssemblySource, TestIoHost } from '../_helpers';
 import { MockSdk, restoreSdkMocksToDefault, setDefaultSTSMocks } from '../_helpers/mock-sdk';
 
+jest.setTimeout(10_000);
```
iliapolo, line 12: Intentional?

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
mrgrain: This is API is currently exported as part of the legacy API exports from the CLI. I think we said we don't want to break them. Will update the comment to be more explicit.

packages/@aws-cdk/toolkit-lib/test/actions/deploy-hotswap.test.ts
mrgrain: I thought about this. In the CLI if you run `cdk deploy --hotswap` it does hotswap-only aka no fallback. To get the fallback you have to pass `cdk deploy --hotswap-fallback`. You cannot pass both flags at the same time. `cdk watch` defaults to hotswap-only for all deployments unless `cdk watch --no-hotswap` is passed.

So really the CLI doesn't have a default but a slight bias towards hotswap-only, it being the more prominent name and the watch default.

This was replicated in the toolkit-lib with the `HotswapMode` enum. You had to make an explicit choice, but you got the style of fallback method for free from the default for `deploymentMethod`.

---

The problem with your suggestion is that `deploymentMethod` has additional parameters that can be set, so it's not enough to just have a single `fallbackMethod` field.

I considered defaulting fallback to `{ method: 'change-set' }` and introduce something like `{ method: 'none' }` but that is against the existing bias towards no fallback and also pollutes the `DeploymentMethod` type with a none variant.

Another option is the introduction of enum-like helpers: `DeploymentMethod.changeSet()` and `DeploymentMethod.direct()` etc. which makes this slightly nicer to write. We do this elsewhere but ultimately I thought it's not needed here. This can be added afterwards as well.

packages/@aws-cdk/toolkit-lib/test/actions/diff.test.ts
mrgrain: Yes, the test frequently fails locally for me. This is a QoL improvement, happy to take it out if you care strongly.

iliapolo (approved):

PR #518, opened 2025-05-21
refactor(toolkit-lib): all default messages go through the IoHelper

mrgrain:
Follow-up from #513, this PR removes the `DEFAULT_TOOLKIT_*` message codes and changes all usage to go through the `IoHelper` instead. Needed to make some additional changes to message spans and the asset logging mechanism.

Next step will be the same treatment for SDK and ASSEMBLY messages.

---


packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
```diff
@@ -500,10 +478,6 @@ export const IO = {
   }),
 
   // SDK codes
-  DEFAULT_SDK_TRACE: make.trace({
-    code: 'CDK_SDK_I0000',
-    description: 'An SDK trace message.',
-  }),
```
mrgrain, line 506: Eagerly removed because not currently used

iliapolo (approved):

PR #519, opened 2025-05-21
refactor(cli): hotswap is now a deployment method

mrgrain:
Follow-up from #517 to make the same changes to the CLI. This has no user impact.

---


packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -518,8 +507,6 @@ export class CdkToolkit {
             parameters: Object.assign({}, parameterMap['*'], parameterMap[stack.stackName]),
             usePreviousParameters: options.usePreviousParameters,
             rollback,
-            hotswap: options.hotswap,
-            hotswapPropertyOverrides: hotswapPropertyOverrides,
```
mrgrain, line 522: this is now handled in `cli.ts`

rix0rrr (approved):

PR #524, opened 2025-05-22
fix: property references are skipped when `DependsOn` is present (refactor)

otaviomacedo:
When computing the graph for the stack, if there is a `DependsOn`, the algorithm will ignore other dependencies. Also, `DependsOn` may be a string or an array, and this was not being properly handled.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/digest.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/digest.ts
@@ -49,16 +49,16 @@ export function computeResourceDigests(template: CloudFormationTemplate): Record
       const refTarget = Array.isArray(value['Fn::GetAtt']) ? value['Fn::GetAtt'][0] : value['Fn::GetAtt'].split('.')[0];
       return [refTarget];
     }
+    const result = [];
     if ('DependsOn' in value) {
-      const result = Object.values(value).flatMap(findDependencies);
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/refactoring/digest.ts
```diff
@@ -50,7 +50,13 @@ export function computeResourceDigests(template: CloudFormationTemplate): Record
       return [refTarget];
     }
     if ('DependsOn' in value) {
-      return [value.DependsOn];
+      const result = Object.values(value).flatMap(findDependencies);
```
iliapolo, line 53: Bit strange we have to call this again, since we are already doing it in line 61. Can you not:

```ts
const result = [];
if ('DependsOn' in value) {
  if (Array.isArray(value.DependsOn)) {
    result.push(...value.DependsOn);
  } else {
    result.push(value.DependsOn);
  }
}
result.push(...Object.values(value).flatMap(findDependencies))
return result;
```

?

iliapolo (approved):

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/digest.ts
iliapolo: Approved anyway - take it or leave it.

PR #528, opened 2025-05-23
fix(toolkit-lib): `fromAssemblyBuilder` mutates globally shared `process.env`

rix0rrr:
The Toolkit communicates with the CDK app passing environment variables (containing context, and others).

If the CDK app is executed as a subprocess it is trivial to set those environment variables when executing the subprocess; but for an in-memory synthesis as allowed by `fromAssemblyBuilder` mutating `process.env` is not the correct approach: it is clobbering global state, and multiple async assembly builders might fight with each other over the single shared global object.

At the same time, it is *extreeeeeemely* convenient for an app author to not have to think about passing props.

So we leave the current behavior as a convenient default, but add the option `clobberEnv: false` to explicitly disable this.

In `fromAssemblyBuilder()`, we now pass an additional `env` prop containing the environment variables we applied (or didn't). There were already parametes for `outdir` and `context`; with `clobberEnv: false` those *must* now be passed to an `App` instantiated inside the builder (whereas otherwise the `App` will pick up on the globally mutated `process.env` changes).

The rest of this PR is reorganizing the code to make the code that touches the environment return values instead of directly applying them. 

* Remove `withEnv()` and `withContext()`.
* Move logging of the final context object out to the call site instead of the builder function.

---


rix0rrr:
TODO: this pessimizes the default API for an edge case.

Retain all the code changes, but make it an option; leave the legacy behavior by default. This will require capturing the OUTDIR and CONTEXT envvars in the `env` bundle to `fromAssemblyBuilder()` as well.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/prepare-source.ts
```diff
@@ -167,59 +167,45 @@ export class ExecutionEnvironment implements AsyncDisposable {
...
-    } finally {
-      if (contextOverflowLocation) {
-        fs.removeSync(path.dirname(contextOverflowLocation));
-      }
+  return async () => {
+    if (contextOverflowLocation) {
+      await fs.promises.rm(path.dirname(contextOverflowLocation), { recursive: true, force: true });
     }
-  }
+  };
```
mrgrain, line 208: Why not implement AsyncDisposable?

packages/@aws-cdk/cli-lib-alpha/lib/cli.ts
mrgrain, line 1: The package is deprecated. Can we just not change anything here please? You know let's not create work for ourselves.

packages/@aws-cdk/cli-lib-alpha/lib/cli.ts
```diff
@@ -122,10 +123,23 @@ export class AwsCdkCli implements IAwsCdkCli {
...
+        };
+        await debugFn(format('context:', fullCtx));
+
+        const env = {
+          ...process.env,
+          ...await prepareDefaultEnvironment(sdk, debugFn),
+          ...params.env,
+        };
+
+        await writeContextToEnv(env, fullCtx);
```
mrgrain, line 140: no cleanup here?

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/source-builder.ts
```diff
@@ -219,3 +259,53 @@ export abstract class CloudAssemblySourceBuilder {
...
+
+  process.env = immutableEnv({
+    ...process.env,
+    ...env,
+  });
+
+  return () => {
+    process.env = oldEnv;
+  };
+}
```
mrgrain, line 290: Nit: I think this pattern is actually quite a bite worse than the previous pattern of passing a block. Now you always have to remember to call the cleanup function. Seems rife for errors.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/prepare-source.ts
rix0rrr: Led to pains in using it in `aws-cdk` that I didn't want to deal with.

mrgrain (approved):

Issue #398, opened 2025-04-22
Remove ProxyAgent from being included in Toolkit Lib by default


mrgrain:
Toolkit Lib should not force users to take a dependency on Proxy Agent. Instead it should accept an Agent (Node http Agent type) where it makes sense. The CLI as an implementor of the toolkit lib can and should still ship with proxy agent.

This will be some sort of API that allows the user to provide or return a [https.Agent](https://nodejs.org/api/https.html#class-httpsagent) instance. That instance should be used where ever it is needed.

PR #532, opened 2025-05-26
refactor(cli): create `ProxyAgent` in the CLI in preparation of `proxy-agent` removal from toolkit-lib

mrgrain:
Updates the `CdkToolkit` and `Notices` APIs to take a Node https.Agent instead of options. This gives more flexibility to integrators while retaining proxy support. We can make the changes to `Notices` since it is not a public API.

The CLI changed need to be split out from the toolkit-lib changes, so the latter can be safely marked as breaking without touching CLI files. The respective `toolkit-lib` changes are in #533 

Relates to #398 

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/notices/notices.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/notices/notices.ts
@@ -1,6 +1,6 @@
+import type { Agent } from 'https';
 import * as path from 'path';
 import { cdkCacheDir } from '../../util';
-import type { SdkHttpOptions } from '../aws-auth';
 import type { Context } from '../context';
 import type { IIoHost } from '../io';
 import { CachedDataSource } from './cached-data-source';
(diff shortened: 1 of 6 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/notices/notices.ts
```diff
@@ -27,9 +41,9 @@ export interface NoticesProps {
   readonly output?: string;
 
   /**
-   * Options for the HTTP request
+   * Options for the HTTPS requests made by Notices
    */
-  readonly httpOptions?: SdkHttpOptions;
+  readonly httpsOptions?: NoticesHttpsOptions;
```
rix0rrr, line 46: If you call it `httpOptions` I'll assume it applies to all flavors of HTTP requests.

If you call it `httpsOptions` I'll wonder where the `httpOptions` are.

I see how this is strictly correct but is it helpful?

rix0rrr (approved):

packages/@aws-cdk/toolkit-lib/lib/api/notices/notices.ts
mrgrain: Will revert back.

PR #534, opened 2025-05-26
feat(toolkit-lib)!: parameterize context storage

rix0rrr:
Previously, context was always being written to `cdk.context.json` in some directory, even for `fromAssemblyBuilder()` or `fromAssemblyDirectory()` which should probably both not write context anywhere.

In this PR, we replace the context parameter to the toolkit with the concept of a `contextStore`, which has two operations: `read()` and `update()`.

There are simple file and in-memory implementations for simple cases, as well as the more complex CDK app case which is implemented as `CdkAppMultiContext`. That last one is the default for `fromCdkApp()`.

The complex one is implemented in terms of the prior `Context/ContextBag/Settings` classes, which are probably overly complex for the job this class is doing but keeping the old classes is an easy way to make sure that the behavior stays compatible.

BREAKING CHANGES: `context` is now called `contextStore`. Message `CDK_ASSEMBLY_I0042` no longer knows about the filename where changes get written.

---


packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/context-store.ts
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * Updates will be written to `<appDirectory>/cdk.context.json`.
+ */
+export class CdkAppMultiContext implements IContextStore {
+  private _context?: Context;
+  private configContextFile: string;
+  private projectContextFile: string;
+  private userConfigFile: string;
+
+  constructor(appDirectory: string, private readonly initialContext?: Record<string, unknown>) {
```
mrgrain, line 51: Only `FileContext` does not get `initialContext`?

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/source-builder.ts
```diff
@@ -165,22 +174,29 @@ export abstract class CloudAssemblySourceBuilder {
...
+   * to the app's `workingDirectory`. This means that context will be loaded from
+   * all the CDK's default context sources, and updates will be written to
+   * `cdk.context.json`.
+   *
    * @param props - additional configuration properties
    * @returns the CloudAssembly source
    */
   public async fromCdkApp(app: string, props: FromCdkAppOptions = {}): Promise<ICloudAssemblySource> {
     const services: ToolkitServices = await this.sourceBuilderServices();
     // @todo this definitely needs to read files from the CWD
```
mrgrain, line 189:
```suggestion
```

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/context-store.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+/**
+ * On-disk context stored in a single file
+ */
+export class FileContext implements IContextStore {
+  private _cache?: Record<string, unknown>;
+
+  constructor(private readonly fileName: string) {
```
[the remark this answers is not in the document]
mrgrain: Yes true. That's confusing then.

mrgrain (approved):

PR #550, opened 2025-05-28
feat(toolkit-lib): emit drift in span

mrgrain:
When we merged #442 we forgot to add message spans to the drift command. This PR rectifies this.

Also contains a refactor of `MessageSpan` to be a separate class and to represent itself as `IoHelper`. With that we can now pass a `Span` down into subroutines and have those messages also emitted as part of the span. Start using that in some places.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/io/private/io-helper.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/io/private/io-helper.ts
@@ -59,7 +59,7 @@ export class IoHelper implements IIoHost, IActionAwareIoHost {
    * Create a new marker from a given registry entry
    */
   public span<S extends object, E extends SpanEnd>(definition: SpanDefinition<S, E>) {
-    return new SpanMaker(this, definition);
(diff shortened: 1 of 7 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/io/private/io-helper.ts
```diff
@@ -59,7 +59,7 @@ export class IoHelper implements IIoHost, IActionAwareIoHost {
    * Create a new marker from a given registry entry
    */
   public span<S extends object, E extends SpanEnd>(definition: SpanDefinition<S, E>) {
-    return new SpanMaker(this, definition);
+    return new SpanMaker(this, definition, (ioHost) => IoHelper.fromActionAwareIoHost(ioHost));
```
mrgrain, line 62: This is needed because otherwise there's a circular dependency between `io-helper.ts` and `span.ts`. The alternative is one big file, but that's also nasty.

kaizencc (approved):

PR #555, opened 2025-05-29
fix: stack definitions for refactor are incorrect when matching resources are different

otaviomacedo:
In most cases, when a resource exists both in the deployed stacks and the local stacks, they will be identical (up to references). But if a resource has a physical ID defined, we take use it to identify the resource, instead of computing a hash from the resource's contents. This allows the toolkit to detect correspondence between these resources even if users changed some properties while refactoring.

When it comes to generating the stacks to send to the refactor API, we need to take this into account. This means taking the deployed resource as the source of truth, and only updating its references, taking them from the local stacks. To do this, we traverse the deployed resource, and every time we find a reference (either within the same stack or cross-stack), transform it to the corresponding local one (also either within- or cross-stack), using the set of resource mappings we already have.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
@@ -6,6 +6,7 @@ export interface CloudFormationResource {
   Type: string;
   Properties?: any;
   Metadata?: Record<string, any>;
+  DependsOn?: string | string[];
 }
 
 export interface CloudFormationTemplate {
@@ -82,7 +83,7 @@ export abstract class ResourceReference {
   abstract toCfn(): any;
 }
 
-class Ref extends ResourceReference {
+export class Ref extends ResourceReference {
   public static fromCfn(stackName: string, value: any): ResourceReference {
     if (!('Ref' in value)) {
       throw new ToolkitError(`Expected a Ref object, got ${JSON.stringify(value)}`);
@@ -109,7 +110,7 @@ class Ref extends ResourceReference {
   }
 }
 
-class GetAtt extends ResourceReference {
+export class GetAtt extends ResourceReference {
   public static fromCfn(stackName: string, value: any): ResourceReference {
     if (!('Fn::GetAtt' in value)) {
       throw new ToolkitError(`Expected a Fn::GetAtt object, got ${JSON.stringify(value)}`);
@@ -148,6 +149,30 @@ class GetAtt extends ResourceReference {
 
   toCfn(): any {
     return { 'Fn::GetAtt': [this.logicalResourceId, this.attributeName] };
+  }
+}
+
+export class DependsOn extends ResourceReference {
+  public static fromString(stackName: string, logicalId: string): ResourceReference {
+    return new DependsOn(stackName, logicalId);
+  }
+
+  constructor(public readonly stackName: string, public readonly logicalResourceId: string) {
+    super(stackName, logicalResourceId);
+  }
+
+  public equals(other: ResourceReference): boolean {
+    return (
+      other instanceof DependsOn && this.stackName === other.stackName && this.logicalResourceId === other.logicalResourceId
+    );
+  }
+
+  public map(stackName: string, logicalId: string): ResourceReference {
+    return new DependsOn(stackName, logicalId);
+  }
+
+  toCfn(): any {
+    return this.logicalResourceId;
   }
 }
 
--- a/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
@@ -1,5 +1,6 @@
 import type { TypedMapping } from '@aws-cdk/cloudformation-diff';
 import type * as cxapi from '@aws-cdk/cx-api';
+import { ToolkitError } from '../../toolkit/toolkit-error';
 
 export interface CloudFormationResource {
   Type: string;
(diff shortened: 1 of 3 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -1,10 +1,13 @@
 import type { TypedMapping } from '@aws-cdk/cloudformation-diff';
+import { deepEqual } from '@aws-cdk/cloudformation-diff';
 import type * as cxapi from '@aws-cdk/cx-api';
+import { ToolkitError } from '../../toolkit/toolkit-error';
 
 export interface CloudFormationResource {
```
rix0rrr, line 6: This is not intended to be complete, right?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -70,3 +73,201 @@ export class ResourceMapping {
...
+
+export class Ref implements ResourceReference {
+  public static fromCfn(stackName: string, value: any): ResourceReference {
+    if (!('Ref' in value)) {
+      throw new ToolkitError(`Expected a Ref object, got ${JSON.stringify(value)}`);
+    }
+    return new Ref(stackName, value.Ref);
+  }
+
+  constructor(public readonly stackName: string, public readonly logicalResourceId: string) {
```
rix0rrr, line 92: You might also consider modeling it as a reference to an object that has a logicalID as an attribute.

Then the rename operation is as simple as changing the `logicalID` field of the resource, and re-rendering. You then do not need the `map` and `replace` operations anymore.

image

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -70,3 +73,201 @@ export class ResourceMapping {
...
+}
+
+export class GetAtt implements ResourceReference {
+  public static fromCfn(stackName: string, value: any): ResourceReference {
+    if (!('Fn::GetAtt' in value)) {
+      throw new ToolkitError(`Expected a Fn::GetAtt object, got ${JSON.stringify(value)}`);
+    }
+    const att = value['Fn::GetAtt'];
+    if (typeof att === 'string') {
+      const [id, attributeName] = att.split('.');
```
rix0rrr, line 131: I think `attributeName` can contain `.` as well, so you need to make sure to max split once.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -70,3 +73,201 @@ export class ResourceMapping {
...
+  }
+
+  public static fromArray(stackName: string, logicalIds: string[]): ResourceReference {
+    return new DependsOn(stackName, logicalIds);
+  }
+
+  constructor(private readonly stackName: string, private readonly logicalIds: string[]) {
+  }
+
+  public equals(other: ResourceReference): boolean {
```
rix0rrr, line 187: Is this what we need to say? That the entire array is the same?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -70,3 +73,201 @@ export class ResourceMapping {
...
+    return (
+      other instanceof DependsOn &&
+      this.stackName === other.stackName &&
+      other.logicalIds.length === this.logicalIds.length &&
+      this.logicalIds.every((id) => other.logicalIds.includes(id))
```
rix0rrr, line 192: But we do allow reordering?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -70,3 +73,201 @@ export class ResourceMapping {
...
+    return (
+      other instanceof FnSub &&
+      this.stackName === other.stackName &&
+      deepEqual(this.inputString, other.inputString) &&
+      deepEqual(this.variableMap, other.variableMap)
+    );
+  }
+
+  replace(mappings: ResourceMapping[]): ResourceReference {
+    if (typeof this.inputString !== 'string') {
```
rix0rrr, line 232: When can this be true?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -70,3 +73,201 @@ export class ResourceMapping {
...
+      return this;
+    }
+
+    const newInputString = this.inputString.replace(/\${([a-zA-Z0-9_.]+)}/g, (_: any, varName: string) => {
+      if (varName.includes('.')) {
+        const [logicalId, attr] = varName.split('.');
```
rix0rrr, line 238: Same thing

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
```diff
@@ -70,3 +73,201 @@ export class ResourceMapping {
...
+  if ('Ref' in value) {
+    return Ref.fromCfn(stackName, value);
+  } else if ('Fn::GetAtt' in value) {
+    return GetAtt.fromCfn(stackName, value);
+  } else {
+    throw new ToolkitError(`Unsupported resource reference type: ${JSON.stringify(value)}`);
+  }
+}
+
+function replaceFirst(mappings: ResourceMapping[], stackName: string, logicalId: string): string {
```
rix0rrr, line 266: This name makes me think there can be more than one replacement; but that seems like it would be illegal anyway?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/execution.ts
```diff
@@ -56,7 +70,147 @@ export function generateStackDefinitions(
...
+        [inputString, variables] = value['Fn::Sub'];
+      }
+
+      const updatedVariables = updateReferences(variables, stackName, mapReference);
+      const updatedInputString = updateReferences(inputString, stackName, mapReference);
+      const fnSub = new FnSub(stackName, updatedInputString, updatedVariables);
+      return resolveLocalReference(fnSub);
+    }
+
+    if ('DependsOn' in value) {
```
rix0rrr, line 160: `DependsOn` only has a special meaning if it is at the top-level of a resource definition, not anywhere in the template. It might not be happening (yet), but I still wouldn't want to rely on it never happening.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/execution.ts
```diff
@@ -16,32 +20,42 @@ export function generateStackDefinitions(
...
+    const deployedResource = deployedStacks.find((s) => s.stackName === location.stackName)?.template.Resources?.[
+      location.logicalResourceId
+    ];
 
     if (deployedResource == null) {
       delete localTemplates[stackName].Resources?.[logicalResourceId];
+    } else {
+      // We've got a deployed resource matching a local resource.
+      // The final template should contain the deployed resource, but with local references.
+      localTemplates[stackName].Resources![logicalResourceId] = updateReferences(
```
rix0rrr, line 42: Don't forget references in the `Output` block.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/execution.ts
```diff
@@ -16,32 +20,42 @@ export function generateStackDefinitions(
...
+    const location = correspondingLocation(stackName, logicalResourceId, 'backward');
 
-    const deployedResource = deployedStacks.find((s) => s.stackName === location.stackName)?.template
-      .Resources?.[location.logicalResourceId];
+    const deployedResource = deployedStacks.find((s) => s.stackName === location.stackName)?.template.Resources?.[
+      location.logicalResourceId
+    ];
 
     if (deployedResource == null) {
       delete localTemplates[stackName].Resources?.[logicalResourceId];
```
rix0rrr, line 38: I suppose we just assume/assert there will be no references to this resource when we are done with the loop, eh? Worth a comment somewhere?

BTW, thinking about this some more:

- We now take the "new" template, removing all "new" resources. So effectively, we will be left with the "old" template + modifications.
- But, any modifications aren't allowed and will lead to a deployment error.
- So if the deployment were to succeed, we would have produced the "old" template, given the new template.

If that's true, why don't we take the "old" template and apply the renames there?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/execution.ts
rix0rrr: Also, apparently `UpdatePolicy` can contain references as well. But since you're processing the entire resource's JSON object that should all automatically work already.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
otaviomacedo: No. Just enough for what we need here.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/execution.ts
```diff
@@ -16,32 +20,42 @@ export function generateStackDefinitions(
...
+      // The final template should contain the deployed resource, but with local references.
+      localTemplates[stackName].Resources![logicalResourceId] = updateReferences(
+        deployedResource,
+        location.stackName,
+        mapper(mappings),
+      );
+
+      // After copying (an updated version of) the deployed resource,
+      // we may end up with dangling dependencies. Sanitize them.
+      sanitizeDependencies(localTemplates[stackName]);
```
rix0rrr, line 50: How is it possible that we have problems with dangling `DependsOn`s, but we don't have any problems with dangling `Ref`s?

otaviomacedo: If a resource that was pointed at via a `Ref` gets moved to another stack, that `Ref` will be converted into a `Fn::ImportValue`, which doesn't happen for `DependsOn`.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/cloudformation.ts
otaviomacedo: CloudFormation doesn't accept it, but the CDK can generate things like:

```json
"Fn::Sub": {
  "Fn::GetAtt": [
   "Queue83E4F04D",
   "QueueName"
  ]
}
```

PR #559, opened 2025-05-30
fix(toolkit-lib): new `mappingSource` option to replace multiple mutual exclusive options in the refactor API

otaviomacedo:
Introduce a new type, `MappingSource`, that allows the user to tell how they want the toolkit to get the mappings: computing it, using an explicitly provided one, or using the reverse of an explicitly provided one.

This replaces three previous separate properties: `exclude`, `mappings` and `revert` with a single `mappingSource`.

---


packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -1,4 +1,43 @@
...
+    }));
+
+    return new MappingSource('explicit', reverseGroups, new NeverExclude());
+  }
+
+  public readonly source: MappingType;
+  public readonly groups: MappingGroup[];
+  public readonly exclude: ExcludeList;
+
+  constructor(source: MappingType, groups: MappingGroup[], exclude: ExcludeList) {
```
mrgrain, line 31:
```suggestion
  private constructor(source: MappingType, groups: MappingGroup[], exclude: ExcludeList) {
```

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -1,4 +1,43 @@
+import type { ExcludeList } from '../../api';
+import { InMemoryExcludeList, NeverExclude } from '../../api';
 import type { StackSelector } from '../../api/cloud-assembly';
+import type { ResourceLocation } from '../../api/refactoring/cloudformation';
+
+export type MappingType = 'auto' | 'explicit';
```
mrgrain, line 6:
```suggestion
type MappingType = 'auto' | 'explicit';
```

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -1,4 +1,43 @@
...
+
+export class MappingSource {
```
mrgrain, line 8: docblocks please!

mrgrain (changes requested):
just some small stuff

mrgrain (approved):

PR #564, opened 2025-06-02
fix(toolkit-lib): cannot use `watch` action without explicitly providing `include` or `exclude` patterns

mrgrain:
This PR improves the watch action implementation in the toolkit-lib package by:

## Changes
- When neither `include` or `exclude` are been provided, the action now doesn't fail anymore. Instead sensible defaults are used.
- `exclude` now uses a list of default excludes as fallback. These defaults aim to cover common development patterns for all CDK languages.
- `node_modules` is no longer forcibly excluded, but part of the exclude defaults instead. This is still the behavior in the CLI but seems overly unexpected to users, especially to toolkit-lib users. We now only force exclude hidden files.

## Why these changes address the issue
Previously, the watch action would throw an error if no include/exclude patterns were specified, requiring users to always configure these settings. This change makes the watch action more user-friendly by:
1. Using sensible defaults that work out-of-the-box
2. Providing better exclusion patterns for common non-source files
3. Following the principle of least surprise by watching all files by default

## Design decisions
- Created a reusable constant  that can be used in other parts of the codebase
- Improved code comments to better explain the behavior
- Maintained backward compatibility while enhancing usability
- Organized exclude patterns by language/ecosystem for better maintainability

## Alternatives considered
- Keeping the error but providing more guidance - rejected because a sensible default is more user-friendly
- Using a more complex configuration system - rejected in favor of simplicity
- Language-specific exclude patterns - kept generic to work across all supported CDK languages

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
@@ -31,8 +31,8 @@ import { type ListOptions } from '../actions/list';
 import type { MappingGroup, RefactorOptions } from '../actions/refactor';
 import { type RollbackOptions } from '../actions/rollback';
 import { type SynthOptions } from '../actions/synth';
-import type { WatchOptions } from '../actions/watch';
-import { patternsArrayForWatch } from '../actions/watch/private';
+import type { IWatcher, WatchOptions } from '../actions/watch';
+import { WATCH_EXCLUDE_DEFAULTS } from '../actions/watch/private';
 import {
   BaseCredentials,
   type IBaseCredentialsProvider,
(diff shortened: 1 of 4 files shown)


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -854,42 +854,29 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
-    const watchExcludes = patternsArrayForWatch(options.exclude, {
-      rootDir,
-      returnRootDirIfEmpty: false,
-    });
-
-    // only exclude the outdir if it is under the rootDir
-    const relativeOutDir = path.relative(rootDir, outdir);
+    // For the "exclude" setting, the behavior is to add some default excludes in addition to
+    // patterns specified by the user orr the default patterns:
+    const watchExcludes = options.exclude ?? WATCH_EXCLUDE_DEFAULTS;
```
iliapolo: So we don't mutate the constant.

```suggestion
    const watchExcludes = options.exclude ?? [...WATCH_EXCLUDE_DEFAULTS];
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -854,42 +854,29 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
       watchExcludes.push(`${relativeOutDir}/**`);
     }
-
-    watchExcludes.push('**/.*', '**/.*/**', '**/node_modules/**');
+    // 2. Any file whose name starts with a dot.
+    watchExcludes.push('.*', '**/.*');
+    // 3. Any directory's content whose name starts with a dot.
+    watchExcludes.push('**/.*/**');
+    // 4. Any node_modules and its content (even if it's not a JS/TS project, you might be using a local aws-cli package)
+    watchExcludes.push('**/node_modules/**');
```
iliapolo, line 879: This is an interesting one...i get it but it also means that an a dependency upgrade won't cause a redeployment, even though it definitely might affect your infra. 

Thoughts? Maybe provide a toggle for this?

iliapolo (changes requested):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: The toggle would be to not force exclude it and move it to the constant. 

FWIW, this is exactly the same what the CLI currently does.

iliapolo: I actually didn't notice `node_modules` is already excluded by the toolkit-lib. So removing it would be a breaking change...Its probably fine the way it is now. Not worth our effort. If we see a need we can add a specific flag to remove `node_modules` from the exclude list.

mrgrain: Keeping it like this then. 👍🏻

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: fixed

iliapolo (approved):

PR #568, opened 2025-06-03
feat(cli): metadata can live in a separate file

rix0rrr:
When CDK apps grow extremely large (think 10-100 stacks, 1000-10000 constructs), all metadata together begins to exceed 512MB, the maximum string size in NodeJS. People usually deal with this by disabling metadata, but they shouldn't have to.

In addition, even for manifests that don't exceed 512MB the extremely large size of the single JSON object slows down its writing and reading every time, even if the metadata doesn't need to be accessed.

An effective solution is to write the metadata of an artifact to a separate file. This PR introduces the ability for that into the Cloud Assembly schema, and updates the CLI to read from both sources if available.

Removes an unused class from the refactoring project that we found through this project because it would have to be refactored. Rather than refactoring, bye-bye!

Relates to https://github.com/aws/aws-cdk/pull/34480.

---


packages/@aws-cdk/cloud-assembly-schema/schema/cloud-assembly.schema.json
```diff
@@ -58,6 +58,10 @@
                         }
                     }
                 },
+                "additionalMetadataFile": {
```
otaviomacedo, line 61: Shouldn't this be a file path?

rix0rrr: This is a JSON schema file describing the data structure, not an instance of the data structure

otaviomacedo: Ah, but of course!

otaviomacedo (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "eslint.useFlatConfig": false,
+}
```
mrgrain, line 3: suggestion: add to the workspace settings instead

Having this file in vcs is extremely annoying because now I cannot have my local config here.

packages/@aws-cdk/cloud-assembly-api/lib/artifacts/cloudformation-artifact.ts
mrgrain, line 1: strong suggestion: separate PR for separate change log

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/exclude.ts
```diff
@@ -20,20 +21,21 @@ abstract class AbstractExcludeList implements ExcludeList {
...
+  private getExcludedLocations(directory: string, asmManifest: AssemblyManifest): CfnResourceLocation[] {
     // First, we need to filter the artifacts to only include CloudFormation stacks
     const stackManifests = Object.entries(asmManifest.artifacts ?? {}).filter(
       ([_, manifest]) => manifest.type === ArtifactType.AWS_CLOUDFORMATION_STACK,
     );
 
     const result: CfnResourceLocation[] = [];
     for (let [stackName, manifest] of stackManifests) {
-      const locations = Object.values(manifest.metadata ?? {})
+      // FIXME: OH NO
```
mrgrain: FIXME

packages/@aws-cdk/cloud-assembly-api/lib/cloud-artifact.ts
```diff
@@ -36,6 +38,34 @@ export interface AwsCloudFormationStackProperties {
...
+   *
+   * HISTORICAL OR PRIVATE USE ONLY
+   *
+   * This is publicly exposed as a static function for downstream libraries that
+   * don't use the `CloudAssembly`/`CloudArtifact` API, yet still need to read
+   * an artifact's metadata.
+   *
+   * 99% of consumers should just access `artifact.metadata`.
+   */
+  public static readMetadata(assemblyDirectory: string, x: cxschema.ArtifactManifest): Record<string, cxschema.MetadataEntry[]> {
```
mrgrain, line 52: Question: I guess I am wondering why our code is not included in "99% of consumers should just access `artifact.metadata`"

mrgrain (dismissed):

.vscode/settings.json
rix0rrr: Aren't these the workspace settings?

packages/@aws-cdk/cloud-assembly-api/lib/cloud-artifact.ts
rix0rrr: The places where we call this function we don't actually have a `CloudArtifact` object; we instead have the underlying "plain JSON data" type.

Loading a `CloudAssembly` access object, and using that to access the `CloudArtifact` classes was a bigger change I wasn't prepared to make.

packages/@aws-cdk/cloud-assembly-api/lib/artifacts/cloudformation-artifact.ts
rix0rrr: *sigh*. You are right of course. This came up in the context of "I don't want to maintain this crazy behavior" for the new bit of metadata. Let's see if I can split it out without too much effort.

packages/@aws-cdk/cloud-assembly-api/lib/cloud-artifact.ts
rix0rrr: Image

.vscode/settings.json
rix0rrr: I've moved it to a separate file, with the expectation that every developer keeps their own per-workspace settings in sync with the recommended per-workspace settings.

mrgrain: True. I meant the `aws-cdk-cli.code-workspace.json` file.

PR #570, opened 2025-06-03
chore(toolkit-lib): improve CCAPI context provider error handling

mrgrain:
Just noticed some missed opportunities in the error handling for the CCAPI provider.

This only has limited end-user impact:
- instead of `error: cause` it's now `error\ncause`
- Some not enough results found error messages now include the expectation

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit-error.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit-error.ts
@@ -4,6 +4,7 @@ const TOOLKIT_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.ToolkitError');
 const AUTHENTICATION_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.AuthenticationError');
 const ASSEMBLY_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.AssemblyError');
 const CONTEXT_PROVIDER_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.ContextProviderError');
+const NO_RESULTS_FOUND_ERROR_SYMBOL = Symbol.for('@aws-cdk/toolkit-lib.NoResultsFoundError');
 
 /**
  * Represents a general toolkit error in the AWS CDK Toolkit.
@@ -132,6 +133,13 @@ export class AssemblyError extends ToolkitError {
  */
 export class ContextProviderError extends ToolkitError {
   /**
+   * Determines if a given error is an instance of AssemblyError.
+   */
+  public static isNoResultsFoundError(x: any): x is NoResultsFoundError {
+    return this.isContextProviderError(x) && NO_RESULTS_FOUND_ERROR_SYMBOL in x;
+  }
+
+  /**
    * Denotes the source of the error as user.
    */
   public readonly source = 'user';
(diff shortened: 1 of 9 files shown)


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit-error.ts
```diff
@@ -131,6 +132,13 @@ export class AssemblyError extends ToolkitError {
  * Represents an error originating from a Context Provider
  */
 export class ContextProviderError extends ToolkitError {
+  /**
+   * Determines if a given error is an instance of AssemblyError.
+   */
+  public static isNoResultsFoundError(x: any): x is NoResultsFoundError {
+    return this.isContextProviderError(x) && NO_RESULTS_FOUND_ERROR_SYMBOL in x;
```
iliapolo: I remember us having issues with accessing static functions using `this`...

```suggestion
    return ToolkitError.isContextProviderError(x) && NO_RESULTS_FOUND_ERROR_SYMBOL in x;
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit-error.ts
```diff
@@ -131,6 +132,13 @@ export class AssemblyError extends ToolkitError {
  * Represents an error originating from a Context Provider
  */
 export class ContextProviderError extends ToolkitError {
+  /**
+   * Determines if a given error is an instance of AssemblyError.
```
iliapolo:
```suggestion
   * Determines if a given error is an instance of NoResultsFoundError.
```

packages/@aws-cdk/toolkit-lib/test/context-providers/cc-api-provider.test.ts
```diff
@@ -386,7 +386,7 @@ describe('dummy value', () => {
           },
       ],
     }),
-  ).rejects.toThrow('Encountered CC API error while getting AWS::RDS::DBInstance resource bad-identifier: Other error');
+  ).rejects.toThrow('Encountered CC API error while getting AWS::RDS::DBInstance resource bad-identifier');
```
iliapolo: Why remove the assertion on cause?

```suggestion
  ).rejects.toThrow('Encountered CC API error while getting AWS::RDS::DBInstance resource bad-identifier\nOther error');
```

iliapolo (changes requested):

packages/@aws-cdk/toolkit-lib/test/context-providers/cc-api-provider.test.ts
mrgrain: Because the error object doesn't have that as a message anymore. Instead the `cause` is a specific property on the error object.

Checking the nested cause property is possible, but quite annoying and verbose. I wanted to avoid work. Instead I have added a new test to ensure that an Error with a cause will be returned correctly in the context.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit-error.ts
mrgrain: Fixed it here and in the rest of the file.

packages/@aws-cdk/toolkit-lib/test/context-providers/cc-api-provider.test.ts
mrgrain: Fixed

iliapolo (approved):

PR #571, opened 2025-06-03
chore(cli): type `currentAction` as `CliAction`

kaizencc:
`CliAction` type feels slightly off as it should be `CliAction` and not `ToolkitAction`. This makes the default make sense. Either way, the internal property of `CliIoHost` is typed to `CliAction`.
 
In practice this probably doesn't matter because I don't think we are initializing `CliIoHost` with anything but `none`.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/io/toolkit-action.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/io/toolkit-action.ts
@@ -1,5 +1,5 @@
 /**
- * The current action being performed by the Toolkit or CLI.
+ * The current action being performed by the Toolkit.
  */
 export type ToolkitAction =
 | 'assembly'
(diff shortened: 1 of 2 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -10,7 +10,10 @@ import { StackActivityProgress } from '../../commands/deploy';
 
 export type { IIoHost, IoMessage, IoMessageCode, IoMessageLevel, IoRequest };
 
-type CliAction =
+/**
+ * The current action being performed by the CLI. 'none' represents the absence of an action.
+ */
+export type CliAction =
```
mrgrain:
```suggestion
type CliAction =
```

packages/@aws-cdk/toolkit-lib/lib/api/io/toolkit-action.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * The current action being performed by the CLI. 'none' represents the absence of an action.
+ * The current action being performed by the Toolkit or CLI.
```
mrgrain: Technically the Toolkit is both.
```suggestion
 * The current action being performed by the Toolkit.
```

mrgrain (approved):

PR #576, opened 2025-06-04
Add --context-file param to override cdk.context.json

mavwolverine:
Fixes #https://github.com/aws/aws-cdk-cli/issues/632

This is replacement for PR https://github.com/aws/aws-cdk/pull/23912

---


packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -2,16 +2,18 @@ import * as os from 'os';
...
-import { debug, warning } from '../logging';
+import { info, debug, warning } from '../logging';
 
 export const PROJECT_CONFIG = 'cdk.json';
-export { PROJECT_CONTEXT } from '../api/context';
 export const USER_DEFAULTS = '~/.cdk.json';
 const CONTEXT_KEY = 'context';
 
+let PROJECT_CONTEXT = DEFAULT_PROJECT_CONTEXT;
+export { PROJECT_CONTEXT };
```
mrgrain, line 15: Let's not do this. It's going to be confusing if this value changes bases on whether `load()` has been called or not.

You can just update the name of the variable to `DEFAULT_PROJECT_CONTEXT` everywhere, or maybe for now just update the docblock.

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -122,6 +124,20 @@ export class Configuration {
...
     this._projectConfig = await loadAndLog(PROJECT_CONFIG);
+    var userProjectContextFile = this.commandLineArguments.subSettings(['contextFile']).get([]);
+    // Check if file exists
+    if (
+      userProjectContextFile &&
+      typeof userProjectContextFile === 'string' &&
+      fs.existsSync(expandHomeDir(userProjectContextFile))
+    ) {
+      info(`Using specified project context ${userProjectContextFile}`);
+      PROJECT_CONTEXT = userProjectContextFile;
```
mrgrain, line 135: see above

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -122,6 +124,20 @@ export class Configuration {
   public async load(): Promise<this> {
     const userConfig = await loadAndLog(USER_DEFAULTS);
     this._projectConfig = await loadAndLog(PROJECT_CONFIG);
+    var userProjectContextFile = this.commandLineArguments.subSettings(['contextFile']).get([]);
+    // Check if file exists
+    if (
+      userProjectContextFile &&
+      typeof userProjectContextFile === 'string' &&
+      fs.existsSync(expandHomeDir(userProjectContextFile))
```
mrgrain, line 132: I don't think t the filepath should be considered relative to the user's home directory. It's also not what is done later on.

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -122,6 +124,20 @@ export class Configuration {
   public async load(): Promise<this> {
     const userConfig = await loadAndLog(USER_DEFAULTS);
     this._projectConfig = await loadAndLog(PROJECT_CONFIG);
+    var userProjectContextFile = this.commandLineArguments.subSettings(['contextFile']).get([]);
```
mrgrain, line 127: I think you'll need to store the value as a class property

mrgrain (changes requested):
Thanks for getting this raised again. Made some comments, we also need unit and integration tests for this.

mavwolverine:
@mrgrain 
Interest - yes
time - currently started work on a new project, so little short on time.

PR #579, opened 2025-06-05
feat(cli): telemetry opt-out command

### LINKED ISSUE #575: cdk telemetry - enable or disable telemetry
Introduce a new command `telemetry` that will set a context flag in `cdk.json`.

This boolean flag will determine whether we collect anonymous CLI telemetry from the user.

Resolving this issue will not begin telemetry collection, but will allow users to opt out beforehand. Telemetry will be enabled by default, when implemented.

iankhou:
Fixes #575 

Enables users of the CDK CLI to opt out of CDK CLI telemetry collection. More details in the [RFC](https://github.com/aws/aws-cdk-rfcs/pull/734).

Telemetry will be enabled by default.

The following command will disable it, local to your CDK App:
`cdk cli-telemetry --disable`

The following command will enable it again:
`cdk cli-telemetry --enable`

To globally enable/disable telemetry add `{ "cli-telemetry": false }` to the context values of your `~/.cdk.json` file.

> At the time of writing this PR - the CDK CLI still does not yet send any telemetry data. We are providing the opt-out capability now to give enough time to opt-out in advance if they so choose. 

---


packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -44,7 +44,7 @@ export async function makeConfig(): Promise<CliConfig> {
       'unstable': { type: 'array', desc: 'Opt in to unstable features. The flag indicates that the scope and API of a feature might still change. Otherwise the feature is generally production ready and fully supported. Can be specified multiple times.', default: [] },
     },
     commands: {
-      list: {
+      'list': {
```
iliapolo, line 47: Mmm how come?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -454,6 +454,15 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
         ioHost.currentAction = 'notices';
         return cli.acknowledge(args.ID);
 
+      case 'cli-telemetry':
+        ioHost.currentAction = 'cli-telemetry';
```
iliapolo, line 458: Seems unnecessary because of:

https://github.com/aws/aws-cdk-cli/blob/7cac5d2d828800b500f80fe00aa0c0ab885b56c5/packages/aws-cdk/lib/cli/cli.ts#L183

But since all other commands are doing it, ok.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -454,6 +454,15 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
+        if (args.enable === undefined && args.disable === undefined) {
+          throw new ToolkitError('Must specify either \'--enable\' or \'--disable\'');
```
iliapolo, line 460: Do we have auto negation for these too? like, is it possible to run `cdk cli-telemetry --no-disable`?

iliapolo (changes requested):
Nice - just needs integ test.

packages/aws-cdk/lib/cli/cli.ts
kaizencc: yes that is provided via yargs for free. i dont like those options so they are unadvertised, but will add them in integ tests jic

packages/aws-cdk/lib/cli/cli-config.ts
kaizencc: because `'cli-telemetry'` needs it and apparently some pretty print decided that every other command needed to follow the same convention

iliapolo (approved):

PR #580, opened 2025-06-06
fix: change all @default tags to @defaultValue for api-extractor compatibility

iankhou:
Fixes internal issue with api-extractor (used to generate the Toolkit Library API Reference) not recognizing the `@default` tag, instead requiring the `@defaultValue` tag.

#### Sample of new api-extractor output JSON that has `defaultValue`:
Screenshot 2025-06-06 at 09 19 20
---


mrgrain:
Not needed.

PR #585, opened 2025-06-09
chore(cli): telemetry sink

kaizencc:
Introduces a generic telemetry sink to the cli. it is tested but not used anywhere.

There are 3 base types of telemetry sinks:

- `FileSink` -> gathers events and writes them to a file 
- `IoHostSink` -> gathers events and sends them to the IoHost (for writing to stdout/err)
- `EndpointSink` -> gathers events and sends them to an external endpoint, batching at intervals of 30 seconds. 

---


packages/aws-cdk/test/cli/telemetry/file-client.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  afterEach(() => {
+    // Clean up temp directory after each test
+    if (fs.existsSync(tempDir)) {
+      fs.rmdirSync(tempDir, { recursive: true });
+    }
+  });
+
+  test('saves data to a file', async () => {
```
kaizencc: add 2nd event

packages/aws-cdk/lib/cli/telemetry/file-client.ts
```diff
@@ -0,0 +1,67 @@
...
+  public async emit(event: TelemetrySchema): Promise<void> {
+    try {
+      // Format the events as a JSON string with pretty printing
+      const output = JSON.stringify(event, null, 2);
+
+      // Write to file
+      fs.appendFileSync(this.logFilePath, output);
+    } catch (e: any) {
+      // Never throw errors, just log them via ioHost
+      await this.ioHelper.defaults.warn(`Failed to add telemetry event: ${e.message}`);
```
kaizencc: test that this sends a trace msg but does not throw

packages/aws-cdk/lib/cli/telemetry/file-client.ts
```diff
@@ -0,0 +1,67 @@
...
+  public async emit(event: TelemetrySchema): Promise<void> {
+    try {
+      // Format the events as a JSON string with pretty printing
+      const output = JSON.stringify(event, null, 2);
+
+      // Write to file
+      fs.appendFileSync(this.logFilePath, output);
+    } catch (e: any) {
+      // Never throw errors, just log them via ioHost
+      await this.ioHelper.defaults.trace(`Failed to add telemetry event: ${e.message}`);
```
kaizencc: test that this writes to trace and does not throw

packages/aws-cdk/lib/cli/telemetry/file-client.ts
kaizencc, line 1: file-sink.ts

packages/aws-cdk/lib/cli/telemetry/url.ts
kaizencc, line 1: move this to other pr

kaizencc: also why are we caching

packages/aws-cdk/lib/cli/telemetry/schema.ts
```diff
@@ -0,0 +1,63 @@
...
+  readonly eventId: string;
+  readonly installationId: string;
+  readonly timestamp: string;
+  readonly accountId?: string;
+  readonly region?: string;
+}
+
+interface Event {
+  readonly state: 'ABORTED' | 'FAILED' | 'SUCCEEDED';
+  readonly eventType: string;
```
iliapolo, line 15: Shouldn't this be an string union as well? Our event types are well known.

packages/aws-cdk/lib/cli/telemetry/endpoint-sink.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  public async flush(): Promise<void> {
+    try {
+      if (this.events.length === 0) {
+        return;
+      }
+
+      const res = await this.https(this.endpoint, this.events);
```
iliapolo, line 73: Should be:

```suggestion
      const res = await this.https(this.endpoint, { events: this.events });
```

Because this is that server (will) expect.

packages/aws-cdk/test/cli/telemetry/file-sink.test.ts
```diff
@@ -0,0 +1,186 @@
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { IoHelper } from '../../../lib/api-private';
+import { CliIoHost } from '../../../lib/cli/io-host';
+import { FileTelemetrySink } from '../../../lib/cli/telemetry/file-sink';
+import type { TelemetrySchema } from '../../../lib/cli/telemetry/schema';
+
+describe('FileTelemetrySink', () => {
```
iliapolo, line 9: I think you're missing a test to validate that we throw if the file already exists.

packages/aws-cdk/lib/cli/telemetry/endpoint-sink.ts
```diff
@@ -0,0 +1,141 @@
...
+   */
+  readonly agent?: Agent;
+}
+
+/**
+ * The telemetry client that hits an external endpoint.
+ */
+export class EndpointTelemetrySink implements ITelemetrySink {
+  private events: TelemetrySchema[] = [];
+  private endpoint: UrlWithStringQuery;
```
iliapolo, line 42: Should accept a string and `parse` internally. It would avoid exposing nodejs types in the props interface. Also - external callers are not the ones who want to parse, its this class that needs it, so it should do it.

iliapolo (approved):
Approving in spite of comments to start creating smaller PRs. Looks good overall and all comments can be addressed in a quick follow.

packages/aws-cdk/lib/cli/telemetry/schema.ts
kaizencc: i believe i made that change in the other PR and forgot to backport

PR #590, opened 2025-06-10
feat(cloud-assembly-schema): add feature flag report as a possible artifact

rix0rrr:
This gives libraries the opportunity to report the feature flags they support, and what values they saw during synthesis.

I purposely did not bump the schema version, since I want this feature to be fully incremental without disturbing existing users: there is no consumption of this new schema yet, that will come later, but we don't want to force people to upgrade their CLI's unnecessarily. That's bad DX.

---


packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/artifact-schema.ts
```diff
@@ -223,11 +223,62 @@ export interface NestedCloudAssemblyProperties {
...
+   * @default - Not configured by the user
+   */
+  readonly userValue?: any;
+
+  /**
+   * Explanation about the purpose of this flag that can be shown to the user.
+   *
+   * @default - No description
+   */
+  readonly explanation?: string;
```
mrgrain, line 273: is this not going to be quite a lot of text?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/artifact-schema.ts
```diff
@@ -223,11 +223,62 @@ export interface NestedCloudAssemblyProperties {
...
+ * Artifact properties for a feature flag report
+ *
+ * A feature flag report is small enough that all the properties can be inlined
+ * here, and doesn't need an additional file.
+ */
+export interface FeatureFlagReportProperties {
+  /**
+   * The library that this feature flag report applies to.
+   */
+  readonly libraryName: string;
```
mrgrain: not sure I like this name. we could use jsii terminology `module` or something more generic like `source`

mrgrain (changes requested):

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/artifact-schema.ts
rix0rrr: I was wondering the same thing, but then I came down on, it's probably going to be ~1kB per flag. It's not going to break the bank in terms of file size.

mrgrain (approved):

PR #595, opened 2025-06-11
fix(cli): retain type of context values and not convert them to string

### LINKED ISSUE #30583: (integ-tests-alpha,custom-resources): snapshots are always outdated
### Describe the bug

image

After upgrading to `2.146.0` from `2.145.0` where https://github.com/aws/aws-cdk/issues/30418 was introduced I am observing that running integration tests via `integ-runner` always fails. Snapshot is being reported as outdated, like on screenshot above. Although I can successfully rebuild/rerun tests and I can observe a change in snapshots next time I run tests I once again observe situation from screenshot

Once I had downgraded to `2.145.0` on both `aws-cdk-lib`, `@aws-cdk/integ-runner` and `@aws-cdk/integ-tests-alpha` to `2.145.0` tests are passing without a problem.

Behavior suggests a regression of sort around: https://github.com/aws/aws-cdk/issues/30418
Also the problem occurs, I believe, around custom resources I am using.

### Expected Behavior

Screenshot are deterministically stored and next checks aren't failing.

### Current Behavior

image

### Reproduction Steps

N/A

### Possible Solution

N/A

N/A

### CDK CLI Version

2.146.0

### Framework Version

_No response_

### Node.js Version

18.16.0

### OS

MacOS sierra

### Language

TypeScript

### Language Version

5.4.5

### Other information

_No response_

kumsmrit:
Fixes [#30583](https://github.com/aws/aws-cdk/issues/30583)

CLI’s context-passing mechanism relied on serialising every entry to a "KEY=VALUE" string, leading CLI to only allows passing string context, so even JSON booleans (true/false) and numbers became plain strings. 
This resulted in losing the typed context, like boolean feature-flags are always being passed as string "true"/"false". This incorrectly passed feature flags caused certain integration tests snapshot in the [aws-cdk](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ) library to be updated with incorrect values causing the integration tests to fail. 

This fix retains the type of context values when passed via CLI, by correctly JSON parsing the context while building user configuration. Values that aren’t valid JSON will fall back to plain strings, so existing workflows should continue working as is. This will restore correct typing for feature flags without disrupting existing tests.

---


mrgrain (approved):

mrgrain (changes requested):
Requires changes to integ tests. Need to check with team if we are okay with this fix.

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -341,7 +341,13 @@ function parseStringContextListToObject(argv: Arguments): any {
           `User-provided context cannot use keys prefixed with 'aws:', but ${parts[0]} was provided.`,
         );
       }
-      context[parts[0]] = parts[1];
+      let parsedValue: any = parts[1];
+      try {
+        parsedValue = JSON.parse(parts[1]);
```
mrgrain, line 346: note to myself, we might want to limit this to booleans and numbers

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -341,7 +341,13 @@ function parseStringContextListToObject(argv: Arguments): any {
...
+      } catch {
+        debug('Non-JSON context value for %s, leaving as string: %s', parts[0], parts[1]);
+      }
```
mrgrain, line 349: I think we need to be more clever here. `JSON.parse(parts[1])` will fail if the value is a regular string like `test`. This will print debug logs for perfectly fine values:

```
cdk deploy --context foobar=bar
```

mrgrain (approved):

mrgrain:
Discussed with the team and this is a great fix!

mrgrain (changes requested):
Too soon. Team is still discussing some open questions. But the general direction is okay.

mrgrain (approved):

Issue #597, opened 2025-06-11
Error [ERR_REQUIRE_ESM]: require() of ES Module /usr/local/lib/node_modules/cdk-assets/node_modules/brace-expansion/index.js


theprivileges:
Assets step on CDK Codepipeline started to fail today due to a breaking change in `brace-expansion`.

Logs are displaying the following error:

```
/usr/local/lib/node_modules/cdk-assets/node_modules/minimatch/dist/commonjs/index.js:7
const brace_expansion_1 = __importDefault(require("brace-expansion"));
                                          ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /usr/local/lib/node_modules/cdk-assets/node_modules/brace-expansion/index.js from /usr/local/lib/node_modules/cdk-assets/node_modules/minimatch/dist/commonjs/index.js not supported.
Instead change the require of /usr/local/lib/node_modules/cdk-assets/node_modules/brace-expansion/index.js in /usr/local/lib/node_modules/cdk-assets/node_modules/minimatch/dist/commonjs/index.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/usr/local/lib/node_modules/cdk-assets/node_modules/minimatch/dist/commonjs/index.js:7:43) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.18.2
```

I managed to find this issue in `minimatch` - https://github.com/isaacs/minimatch/issues/257

rd-joni:
Same here

ubaid-github:
Same here

dcaruso-pathpoint:
Between this issue and the various times aws-sdk client packages were not released by the time a new version of cdk-assets is released. 

I am about ready to move us away from using the cdk "pipelines" construct. 

We can't have all of our pipelines broken multiple times a week because of the installation of "latest" packages that we cannot version lock.

dcaruso-pathpoint:
Should this be closed if we are getting the same error?

miekassu:
How long it takes for #599 to be updated?

gakinson:
Looks like their release for the change ran and completed about 10 minutes ago:

https://github.com/aws/aws-cdk-cli/actions/runs/15597715873

dcaruso-pathpoint:
@rix0rrr just an FYI we are still having these issues even after your fix https://github.com/aws/aws-cdk-cli/pull/603

[...]

PR #603, opened 2025-06-12
fix(cdk-assets): fails installation when new dependencies are released

rix0rrr:
Because `cdk-assets` CLI depends on floating versions of dependencies, it may break when new 3rd party dependencies are released, or when it is installed just around the time a new version of AWS SDKv3 gets released (because that has hundreds of packages that all refer to each other by specific version, but they're subject to publishing/replication latency so they don't all become instantaneously available).

We used to generate a shrinkwrap file for the `cdk-assets` CLI but apparently that didn't work because we're still getting reports of breakage.

Instead, we'll resort to bundling the dependencies of `cdk-assets` into the CLI itself, so it cannot go wrong.

Because the `cdk-assets` package does double duty as a CLI package as well as a library package, we can't just naively stick the `BundleCli` job onto the package: doing that would remove all the bundled dependencies from the `package.json` list, which breaks its function as a library.

Instead, we're splitting the `cdk-assets` package in two:

* `cdk-assets`: will continue to contains the CLI, which will be bundled
* `@aws-cdk/cdk-assets-lib`: the core portion of the assets publishing, available as a library.

Because `cdk-assets` can no longer be used as a library, we have bumped the major version to `4` to indicate this change in API surface. In CDK pipelines we install `cdk-assets@latest`, so it will automatically pick up major version 4 once released.

(While this PR is strictly speaking a breaking change, I'm not adding any "breaking change" markers into the PR description, because this is changing many files across multiple packages, and I don't trust the automatic major version bumping script to not accidentally MV bump something I didn't intend to)

---


rix0rrr:
Relates to #597

.projenrc.ts
```diff
@@ -73,8 +83,6 @@ const ADDITIONAL_CLI_IGNORE_PATTERNS = [
   'synth.lock',
 ];
 
-const CLI_SDK_V3_RANGE = '^3';
```
mrgrain, line 76: What's the reason for getting rid of this? Seemed like an easy way to control this centrally. Also `toolkitLib` probably doesn't want to force the latest version on customers but have a very open contraint.

.projenrc.ts
```diff
@@ -647,36 +653,91 @@ const cdkAssets = configureProject(
...
+        outDir: undefined,
+      },
+      include: ['bin/**/*.ts'],
     },
-  ],
-});
+    releaseWorkflowSetupSteps: [
+      {
+        name: 'Shrinkwrap',
+        run: 'npx projen shrinkwrap',
```
mrgrain: still needed if we bundle?

packages/@aws-cdk/cdk-assets-lib/README.md
```diff
@@ -0,0 +1 @@
+# replace this
```
mrgrain: replace this

packages/@aws-cdk/cdk-assets-lib/package.json
mrgrain, line 1: probably should add `exports` in here (well in .projenrc.ts) to avoid deep path imports

.projenrc.ts
```diff
@@ -647,36 +653,91 @@ const cdkAssets = configureProject(
...
-    },
-    {
-      spawn: 'unbump',
+cdkAssetsLib.gitignore.addPatterns(
+  '*.js',
+  '*.d.ts',
+);
+
+// This package happens do something only slightly naughty
+cdkAssetsLib.eslint?.addRules({ 'jest/no-export': ['off'] });
```
mrgrain, line 677: lol

.projenrc.ts
```diff
@@ -581,33 +589,32 @@ const cliPluginContract = configureProject(
...
     parent: repo,
-    name: 'cdk-assets',
-    description: 'CDK Asset Publishing Tool',
+    name: '@aws-cdk/cdk-assets-lib',
+    majorVersion: 1,
+    description: 'CDK Asset Publishing Library',
     srcdir: 'lib',
     deps: [
       cloudAssemblySchema.customizeReference({ versionType: 'any-future' }),
       cxApi,
```
mrgrain: As a library probably this? Same as toolkitLib?
```suggestion
      `${cxApi}@^2`, // stay within the same MV, otherwise any should work
```

mrgrain (changes requested):

.projenrc.ts
rix0rrr: No

.projenrc.ts
rix0rrr: Shit, I forgot if there was a good reason for doing this, or whether it was just cleanup.

This was convenient when we needed to control the version specifically because there were some incompatible changes in the SDKv3 across versions (we had a point version in there for a while).

In CLIs, there doesn't seem any value in having this constraint be literally in `package.json`; we upgrade with `projen update` anyway.

I see your point about the lib though.

mrgrain (approved):

PR #611, opened 2025-06-13
feat(toolkit-lib): `fromAssemblyDirectory` can ignore missing context

### LINKED ISSUE #549: From assembly dir no need to run lookup

corymhall:
Previously `fromAssemblyDirectory` created a `ContextAwareCloudAssemblySource` with lookups disabled. This would always fail if there was any missing context in the assembly, because you can't perform lookups with just the CloudAssembly directory.

This PR updates the logic to use a `CachedCloudAssembly` instead and check for missing context. The missing context check can be disabled using the new `failOnMissingContext` option (defaults to fail).

Resolves #549

---


mrgrain (changes requested):
Oh cool,  thank you for that! I think we still want to check for missing context values though, just don't pretend they can be looked-up. From your issue I can tell that we should make the check for missing context values optional though (default to check).

rix0rrr:
Haha we were talking about this, just hadn’t gotten around to it. Thanks for the help ☺️

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
```diff
@@ -379,30 +388,32 @@ export abstract class CloudAssemblySourceBuilder {
...
-            await readLock.release();
-            throw e;
+    return {
+      async produce() {
+        await services.ioHelper.notify(IO.CDK_ASSEMBLY_I0150.msg('--app points to a cloud assembly, so we bypass synth'));
+        const readLock = await new RWLock(directory).acquireRead();
+        try {
+          const asm = await assemblyFromDirectory(directory, services.ioHelper, props.loadAssemblyOptions);
+          const assembly = new ReadableCloudAssembly(asm, readLock, { deleteOnDispose: false });
+          if (assembly.cloudAssembly.manifest.missing && assembly.cloudAssembly.manifest.missing.length > 0) {
```
corymhall, line 399: Let me know if you want me to extract this into a new `ICloudAssemblySource`. I don't know that this needs to be reusable so I just inlined it here.

mrgrain: inline is fine. we can extract this later

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
```diff
@@ -379,30 +388,32 @@ export abstract class CloudAssemblySourceBuilder {
...
+            if (props.failOnMissingContext ?? true) {
+              const missingKeysSet = missingContextKeys(assembly.cloudAssembly.manifest.missing);
+              const missingKeys = Array.from(missingKeysSet);
+              throw new ToolkitError(
```
mrgrain: Let's make this an `AssemblyError`. Otherwise PR is ready.

mrgrain (changes requested):
One small change left. Otherwise ready.

Thank you so much and nice working with you again :D

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
mrgrain:
```suggestion
              throw new AssemblyError(
```

mrgrain (approved):
:shipit:

PR #614, opened 2025-06-15
chore: slugify artifact ID

rix0rrr:
In #613 we made the artifact ID unique for different Node versions, but the characters in `lts/*` aren't all safe for inclusion in the artifact ID.

Slugify the artifact ID first before using it.

---

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -139,12 +139,19 @@ jobs:
           if compgen -G "logs/md/*.md" > /dev/null; then
             cat logs/md/*.md >> $GITHUB_STEP_SUMMARY;
           fi
+      - name: Slugify artifact id
+        id: artifactid
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/integ.yml
```diff
@@ -139,12 +139,17 @@ jobs:
           if compgen -G "logs/md/*.md" > /dev/null; then
             cat logs/md/*.md >> $GITHUB_STEP_SUMMARY;
           fi
+      - name: Prepare artifact id
+        id: artifactid
+        uses: Clovis-team/slugify-string@v1.0.0
+        with:
+          value: logs-${{ matrix.suite }}-${{ matrix.node }}
```
mrgrain: can we not do this ourselves ?

rix0rrr: I suppose we could. This just seemed like a hassle-free way to do it.

mrgrain (approved):

PR #623, opened 2025-06-17
fix(cli): logs entire environment

rix0rrr:
Right now the CLI logs the entire environment to debug logging.  This is a lot of data, and potentially also includes tokens we'd rather don't end up in logging.

Instead, just log our additions to the existing environment.

---


mrgrain (changes requested):
Any chance we can make this change in a way that the next person working on this code doesn't see the double env assignment and "optimizes" it away? Thus creating a regression?

A test case would help. Or variable naming maybe or moving `process.env` into the `writeContextToEnv` helper.

rix0rrr:
I did both.

mrgrain (approved):

PR #631, opened 2025-06-18
feat(cli): send telemetry events to local file

kaizencc:
Introduces two telemetry events: invoke and synth:

`Invoke`: all cli commands send an invoke event. this event gets sent at the conclusion of the command and includes the entire duration of the command.
`Synth`: some commands synthesize a cloud assembly (`cdk synth`, `cdk deploy`, `cdk watch`, `cdk list`, `cdk diff`). these events also send a synth event that includes the duration of the atomic synthesis function that all commands share.

> This PR sends telemetry events to a local file. It does not send to an external endpoint yet.

Example: `cdk deploy` with no credentials (synth event succeeds, invoke event fails)

-- 

```bash
> cdk deploy MyStack --unstable=telemetry --telemetry-file=my/local/file
```

```json
[
  {
    "event": {
      "command": {
        "path": [
          "deploy",
          "$STACK1"
        ],
        "parameters": {
          "telemetry-file": "<redacted>",
          "unstable": "<redacted>",
          "verbose": 3,
          "lookups": true,
          "ignore-errors": false,
          "json": false,
          "debug": false,
          "staging": true,
          "notices": true,
          "no-color": false,
          "ci": false,
          "all": false,
          "build-exclude": "<redacted>",
          "import-existing-resources": false,
          "force": false,
          "parameters": "<redacted>",
          "previous-parameters": true,
          "logs": true,
          "concurrency": 1,
          "asset-prebuild": true,
          "ignore-no-stacks": false
        },
        "config": {
          "bags": true,
          "fileNames": true
        }
      },
      "state": "SUCCEEDED",
      "eventType": "SYNTH"
    },
    "identifiers": {
      "installationId": "7cb67dad-250e-46fa-a28f-ea365b604b8d",
      "sessionId": "5f1cf087-44dc-4ba5-8f2b-520a7fbe32b4",
      "telemetryVersion": "1.0",
      "cdkCliVersion": "0.0.0",
      "region": "us-east-1",
      "eventId": "5f1cf087-44dc-4ba5-8f2b-520a7fbe32b4:1",
      "timestamp": "2025-07-11T13:16:49.201Z"
    },
    "environment": {
      "ci": false,
      "os": {
        "platform": "darwin",
        "release": "node"
      },
      "nodeVersion": "v20.19.1"
    },
    "project": {},
    "duration": {
      "total": 4350
    }
  },
  {
    "event": {
      "command": {
        "path": [
          "deploy",
          "$STACK1"
        ],
        "parameters": {
          "telemetry-file": "<redacted>",
          "unstable": "<redacted>",
          "verbose": 3,
          "lookups": true,
          "ignore-errors": false,
          "json": false,
          "debug": false,
          "staging": true,
          "notices": true,
          "no-color": false,
          "ci": false,
          "all": false,
          "build-exclude": "<redacted>",
          "import-existing-resources": false,
          "force": false,
          "parameters": "<redacted>",
          "previous-parameters": true,
          "logs": true,
          "concurrency": 1,
          "asset-prebuild": true,
          "ignore-no-stacks": false
        },
        "config": {
          "bags": true,
          "fileNames": true
        }
      },
      "state": "FAILED",
      "eventType": "INVOKE"
    },
    "identifiers": {
      "installationId": "7cb67dad-250e-46fa-a28f-ea365b604b8d",
      "sessionId": "5f1cf087-44dc-4ba5-8f2b-520a7fbe32b4",
      "telemetryVersion": "1.0",
      "cdkCliVersion": "0.0.0",
      "region": "us-east-1",
      "eventId": "5f1cf087-44dc-4ba5-8f2b-520a7fbe32b4:2",
      "timestamp": "2025-07-11T13:16:49.385Z"
    },
    "environment": {
      "ci": false,
      "os": {
        "platform": "darwin",
        "release": "node"
      },
      "nodeVersion": "v20.19.1"
    },
    "project": {},
    "duration": {
      "total": 4609
    },
    "error": {
      "name": "ExpiredToken"
    }
  }
]
```

----


packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1341,6 +1357,61 @@ export class CdkToolkit {
...
+            duration: elapsedSynthTime,
+          },
+        },
+      });
+
+      return stacks.withDependencies();
+    } catch (e: any) {
+      const currentTime = new Date();
+      await this.ioHost.defaults.notify({
+        level: 'info',
```
kaizencc:
```suggestion
        level: 'trace',
```

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1341,6 +1357,61 @@ export class CdkToolkit {
...
+
+      const stacks = await assembly.selectStacks({
+        patterns: options.selectors,
+      }, {
+        extend: ExtendedStackSelection.Upstream,
+        defaultBehavior: DefaultSelection.AllStacks,
+      });
+
+      const elapsedSynthTime = new Date().getTime() - startSynthTime;
+      await this.ioHost.defaults.notify({
```
kaizencc: needs to be in doSynthesize

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1341,6 +1357,61 @@ export class CdkToolkit {
...
+        level: 'info',
+        message: `\n✨  Synthesis time: ${formatTime(elapsedSynthTime)}s\n`,
+        time: new Date(),
+        data: {
```
kaizencc: data needs to be typed by creating a specific code

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -154,6 +168,46 @@ export class CliIoHost implements IIoHost {
...
+    });
+
+    // sanitize the raw cli input
+    const command = redactCommmandLineArguments(props.argv, await makeConfig());
+
+    this.sessionInfo.event = {
+      command: {
+        path: command.path,
+        parameters: command.parameters,
+        // config: props.settings, // TODO: sanitize
```
kaizencc: can only grab feature flags, either booleans or truthy

kaizencc:
FAILED deploy telemetric events:

```
> cdk deploy Whatever
```

```
--- TELEMETRY EVENT ---
{
  "event": {
    "command": {
      "path": [
        "--lookups=true",
        "--ignore-errors=false",
        "--json=false",
        "--verbose=false",
        "--debug=false",
        "--staging=true",
        "--notices=true",
        "--no-color=false",
        "--ci=false",
        "--unstable=<redacted>",
        "--all=false",
        "--build-exclude=<redacted>",
        "--import-existing-resources=false",
        "--force=false",
        "--parameters=<redacted>",
        "--previous-parameters=true",
        "--logs=true",
        "--concurrency=1",
        "--asset-prebuild=true",
        "--ignore-no-stacks=false"
      ],
      "parameters": [
        "deploy",
        "<redacted-STACKS>"
      ],
      "config": {
        "@aws-cdk/aws-lambda:recognizeLayerVersion": true,
        "@aws-cdk/core:checkSecretUsage": true,
        "@aws-cdk/core:target-partitions": true,
        "@aws-cdk-containers/ecs-service-extensions:enableDefaultLogDriver": true,
        "@aws-cdk/aws-ec2:uniqueImdsv2TemplateName": true,
        "@aws-cdk/aws-ecs:arnFormatIncludesClusterName": true,
        "@aws-cdk/aws-iam:minimizePolicies": true,
        "@aws-cdk/core:validateSnapshotRemovalPolicy": true,
        "@aws-cdk/aws-codepipeline:crossAccountKeyAliasStackSafeResourceName": true,
        "@aws-cdk/aws-s3:createDefaultLoggingPolicy": true,
        "@aws-cdk/aws-sns-subscriptions:restrictSqsDescryption": true,
        "@aws-cdk/aws-apigateway:disableCloudWatchRole": true,
        "@aws-cdk/core:enablePartitionLiterals": true,
        "@aws-cdk/aws-events:eventsTargetQueueSameAccount": true,
        "@aws-cdk/aws-iam:standardizedServicePrincipals": true,
        "@aws-cdk/aws-ecs:disableExplicitDeploymentControllerForCircuitBreaker": true,
        "@aws-cdk/aws-iam:importedRoleStackSafeDefaultPolicyName": true,
        "@aws-cdk/aws-s3:serverAccessLogsUseBucketPolicy": true,
        "@aws-cdk/aws-route53-patters:useCertificate": true,
        "@aws-cdk/customresources:installLatestAwsSdkDefault": false,
        "@aws-cdk/aws-rds:databaseProxyUniqueResourceName": true,
        "@aws-cdk/aws-codedeploy:removeAlarmsFromDeploymentGroup": true,
        "@aws-cdk/aws-apigateway:authorizerChangeDeploymentLogicalId": true,
        "@aws-cdk/aws-ec2:launchTemplateDefaultUserData": true,
        "@aws-cdk/aws-secretsmanager:useAttachedSecretResourcePolicyForSecretTargetAttachments": true,
        "@aws-cdk/aws-redshift:columnId": true,
        "@aws-cdk/aws-stepfunctions-tasks:enableEmrServicePolicyV2": true,
        "@aws-cdk/aws-ec2:restrictDefaultSecurityGroup": true,
        "@aws-cdk/aws-apigateway:requestValidatorUniqueId": true
      }
    },
    "state": "SUCCEEDED",
    "eventType": "Synth"
  },
  "identifiers": {
    "installationId": "7cb67dad-250e-46fa-a28f-ea365b604b8d",
    "sessionId": "6923cc1f-ff83-4f12-912d-24d07aefd69b",
    "telemetryVersion": "1.0",
    "cdkCliVersion": "0.0.0",
    "region": "us-east-1",
    "eventId": "6923cc1f-ff83-4f12-912d-24d07aefd69b:0",
    "timestamp": "2025-06-20T20:01:43.080Z"
  },
  "environment": {
    "ci": false,
    "os": {
      "platform": "darwin",
      "release": "node"
    },
    "nodeVersion": "v20.19.1"
  },
  "project": {},
  "duration": {
    "total": 4884
  }
}
-----------------------

--- TELEMETRY EVENT ---
{
  "event": {
    "command": {
      "path": [
        "--lookups=true",
        "--ignore-errors=false",
        "--json=false",
        "--verbose=false",
        "--debug=false",
        "--staging=true",
        "--notices=true",
        "--no-color=false",
        "--ci=false",
        "--unstable=<redacted>",
        "--all=false",
        "--build-exclude=<redacted>",
        "--import-existing-resources=false",
        "--force=false",
        "--parameters=<redacted>",
        "--previous-parameters=true",
        "--logs=true",
        "--concurrency=1",
        "--asset-prebuild=true",
        "--ignore-no-stacks=false"
      ],
      "parameters": [
        "deploy",
        "<redacted-STACKS>"
      ],
      "config": {
        "@aws-cdk/aws-lambda:recognizeLayerVersion": true,
        "@aws-cdk/core:checkSecretUsage": true,
        "@aws-cdk/core:target-partitions": true,
        "@aws-cdk-containers/ecs-service-extensions:enableDefaultLogDriver": true,
        "@aws-cdk/aws-ec2:uniqueImdsv2TemplateName": true,
        "@aws-cdk/aws-ecs:arnFormatIncludesClusterName": true,
        "@aws-cdk/aws-iam:minimizePolicies": true,
        "@aws-cdk/core:validateSnapshotRemovalPolicy": true,
        "@aws-cdk/aws-codepipeline:crossAccountKeyAliasStackSafeResourceName": true,
        "@aws-cdk/aws-s3:createDefaultLoggingPolicy": true,
        "@aws-cdk/aws-sns-subscriptions:restrictSqsDescryption": true,
        "@aws-cdk/aws-apigateway:disableCloudWatchRole": true,
        "@aws-cdk/core:enablePartitionLiterals": true,
        "@aws-cdk/aws-events:eventsTargetQueueSameAccount": true,
        "@aws-cdk/aws-iam:standardizedServicePrincipals": true,
        "@aws-cdk/aws-ecs:disableExplicitDeploymentControllerForCircuitBreaker": true,
        "@aws-cdk/aws-iam:importedRoleStackSafeDefaultPolicyName": true,
        "@aws-cdk/aws-s3:serverAccessLogsUseBucketPolicy": true,
        "@aws-cdk/aws-route53-patters:useCertificate": true,
        "@aws-cdk/customresources:installLatestAwsSdkDefault": false,
        "@aws-cdk/aws-rds:databaseProxyUniqueResourceName": true,
        "@aws-cdk/aws-codedeploy:removeAlarmsFromDeploymentGroup": true,
        "@aws-cdk/aws-apigateway:authorizerChangeDeploymentLogicalId": true,
        "@aws-cdk/aws-ec2:launchTemplateDefaultUserData": true,
        "@aws-cdk/aws-secretsmanager:useAttachedSecretResourcePolicyForSecretTargetAttachments": true,
        "@aws-cdk/aws-redshift:columnId": true,
        "@aws-cdk/aws-stepfunctions-tasks:enableEmrServicePolicyV2": true,
        "@aws-cdk/aws-ec2:restrictDefaultSecurityGroup": true,
        "@aws-cdk/aws-apigateway:requestValidatorUniqueId": true
      }
    },
    "state": "FAILED",
    "eventType": "Exit"
  },
  "identifiers": {
    "installationId": "7cb67dad-250e-46fa-a28f-ea365b604b8d",
    "sessionId": "6923cc1f-ff83-4f12-912d-24d07aefd69b",
    "telemetryVersion": "1.0",
    "cdkCliVersion": "0.0.0",
    "region": "us-east-1",
    "eventId": "6923cc1f-ff83-4f12-912d-24d07aefd69b:1",
    "timestamp": "2025-06-20T20:01:43.102Z"
  },
  "environment": {
    "ci": false,
    "os": {
      "platform": "darwin",
      "release": "node"
    },
    "nodeVersion": "v20.19.1"
  },
  "project": {},
  "duration": {
    "total": 4924
  },
  "error": {
    "name": "ToolkitError"
  }
}
-----------------------
```

packages/aws-cdk/lib/cli/telemetry/sanitation-utils.ts
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+    if (globalOptions.includes(argName)) {
+      const type = config.globalOptions[argName].type;
+      if (['number', 'boolean', 'count'].includes(type)) {
+        path.push(`--${argName}=${argv[argName]}`);
+      } else {
+        path.push(`--${argName}=<redacted>`);
+      }
+    }
```
iankhou: Shouldn't parameters be things like `--json`, `--ci`, while path would be `list`?

i.e., they are swapped.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -154,6 +168,45 @@ export class CliIoHost implements IIoHost {
...
+      },
+      event: {
+        command: {
+          path: command.path,
+          parameters: command.parameters,
+          config: sanitizeContext(context),
+        },
+      },
+      environment: {
+        ci: Boolean(process.env.CI),
```
kaizencc: `isCI();` and double check that the codebuild / github wf / codebuild within pipeline have this flag

kaizencc: look at other tools that do ci detection

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -154,6 +168,45 @@ export class CliIoHost implements IIoHost {
...
+    // sanitize the raw cli input
+    const command = sanitizeCommandLineArguments(argv, await makeConfig());
+    this.telemetryInfo = {
+      identifiers: {
+        installationId: getInstallationId(this.asIoHelper()),
+        sessionId: randomUUID(),
+        telemetryVersion: '1.0',
+        cdkCliVersion: version.versionNumber(),
+        accountId: await new AccountIdFetcher().fetch(),
+        region: await new RegionFetcher().fetch(),
```
kaizencc: cdklibraryversion

packages/aws-cdk/lib/cli/telemetry/sanitation-utils.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (commandOptions.includes(argName)) {
+      const type = config.commands[command].options![argName].type;
```
kaizencc: use `typeOf`

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -82,50 +83,77 @@ export class CloudExecutable implements ICloudAssemblySource {
...
+          `\n✨  Synthesis time: ${formatTime(elapsedSynthTime)}s\n`,
+          {
+            telemetry: {
+              eventType: 'Synth',
+              state: 'SUCCEEDED',
+              duration: elapsedSynthTime,
+            },
+          },
+        ));
+        return new CloudAssembly(assembly, this.props.ioHelper);
```
kaizencc, line 133: move this up

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -82,50 +83,77 @@ export class CloudExecutable implements ICloudAssemblySource {
     // but it missing. We'll then look up the context and run the executable again, and
     // again, until it doesn't complain anymore or we've stopped making progress).
     let previouslyMissingKeys: Set<string> | undefined;
-    while (true) {
-      const assembly = await this.props.synthesizer(this.props.sdkProvider, this.props.configuration);
+    const startSynthTime = new Date().getTime();
```
kaizencc, line 86: no try, wrap inside the call in synthesis

packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
```diff
@@ -33,6 +33,12 @@ export const IO = {
     description: 'Credential plugin warnings',
   }),
 
+  CDK_TOOLKIT_I0050: make.debug<Telemetry>({
```
kaizencc: trace

kaizencc: plus run by momo about the toolkit messages

packages/@aws-cdk/toolkit-lib/lib/payloads/types.ts
```diff
@@ -70,6 +70,19 @@ export interface SingleStack {
...
 }
 
+export interface Telemetry {
+  readonly telemetry: {
+    readonly eventType: string,
+    readonly state: 'SUCCEEDED' | 'FAILED' | 'ABORTED';
+    readonly duration: number,
+    readonly error?: {
+      readonly name?: string;
+      readonly trace?: string;
```
kaizencc: stackTrace everywhere

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -613,17 +620,58 @@ function determineHotswapMode(hotswap?: boolean, hotswapFallback?: boolean, watc
...
 }
 /* c8 ignore stop */
+
+async function successfulTelemetryExitEvent(args: string[], startTime: number) {
+  await CliIoHost.instance().asIoHelper().notify(IO.CDK_TOOLKIT_I0050.msg(
+    `Exiting ${args[1]}`,
+    {
+      telemetry: {
+        duration: new Date().getTime() - startTime,
+        eventType: 'Exit',
```
kaizencc: string union

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -613,17 +620,58 @@ function determineHotswapMode(hotswap?: boolean, hotswapFallback?: boolean, watc
...
+      if (process.env.TELEMETRY_TEST_ENV) {
+        await failedTelemetryExitEvent(args, startTime, err);
+      }
       process.exitCode = 1;
     });
 }
 /* c8 ignore stop */
+
+async function successfulTelemetryExitEvent(args: string[], startTime: number) {
+  await CliIoHost.instance().asIoHelper().notify(IO.CDK_TOOLKIT_I0050.msg(
```
kaizencc: method retrieves instance but not create a new one

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: ciSystem

packages/aws-cdk/lib/cli/telemetry/sanitation-utils.ts
```diff
@@ -0,0 +1,64 @@
+import { CliConfig } from "@aws-cdk/user-input-gen";
+import { Command } from "./schema";
+
+/**
+ * argv is the output of yargs
+ */
+export function sanitizeCommandLineArguments(argv: any, config: CliConfig): Command {
+  const command = argv._[0];
+  const path: string[] = [];
+  const parameters: string[] = [command];
```
iankhou: This is a dictionary according to the RFC? https://github.com/aws/aws-cdk-rfcs/pull/734/files#r2162360512

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: put in todo -- requires cloud assembly stack metadata OR package.json (and then we drop for non ts packages)

packages/aws-cdk/lib/cli/cli.ts
kaizencc: I think I will do the `.get()` and change `.instance()` into `.getOrCreate()` in a separate PR. @iliapolo you good with that?

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: i think what i have now is best i can do without doing any actual work :)

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -82,50 +83,78 @@ export class CloudExecutable implements ICloudAssemblySource {
...
-
-          // Execute again
-          continue;
         }
-      }
 
-      return new CloudAssembly(assembly, this.props.ioHelper);
+        const cloudAssembly = new CloudAssembly(assembly, this.props.ioHelper);
+        const elapsedSynthTime = new Date().getTime() - startSynthTime;
+        await this.props.ioHelper.notify(IO.CDK_TOOLKIT_I0050.msg(
```
kaizencc:
```ts
await this.props.ioHelper.notify(CLI_PRIVATE_IO.CLI_I1234.msg(
'message',
{
  state: 'SUCCEEDED',
  duration: elapsedSynthTime,
}
));
```

packages/aws-cdk/lib/cli/io-host/messages.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface Start {}
+
+/**
+ * Private message types specific to the CLI
+ */
+export const CLI_PRIVATE_IO = {
+  CDK_CLI_I1000: make.trace<Start>({
+    code: 'CDK_CLI_I1000',
+    description: 'Cloud Execution is starting',
+    interface: 'Start',
```
kaizencc: this is a requirement but i dont need it...

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -613,17 +620,49 @@ function determineHotswapMode(hotswap?: boolean, hotswapFallback?: boolean, watc
 
 /* c8 ignore start */ // we never call this in unit tests
 export function cli(args: string[] = process.argv.slice(2)) {
+  const startTime = new Date().getTime();
```
kaizencc: cant span here because iohost does not exist

packages/aws-cdk/lib/cli/telemetry/installation-id.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as fs from 'fs';
+import { cdkCacheDir } from "../../util";
+import { IoHelper } from "../../api-private";
+
+const INSTALLATION_ID_PATH = path.join(cdkCacheDir(), 'installation-id.json');
+
+/**
+ * Get or create installation id
+ */
+export function getInstallationId(ioHost: IoHelper) {
```
kaizencc: amplify just does this, maybe its better:

```
export const getInstallationUuid = (
  namespace: string = AMPLIFY_CLI_UUID_NAMESPACE,
) => {
  return uuidV5(__dirname + hostname(), namespace);
};
```

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: Amplify achieves this by having an `NpmLockFileReader` class and a `YarnLockFileReader` class

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -88,6 +88,13 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
   });
   await configuration.load();
 
+  if (process.env.TELEMETRY_TEST_ENV) {
```
iliapolo: Convert this to a `Teletemry.isEnabled` function somewhere and reuse.

packages/aws-cdk/lib/cli/io-host/messages.ts
```diff
@@ -0,0 +1,41 @@
+import { Duration, ErrorPayload } from '@aws-cdk/toolkit-lib';
+import * as make from '../../api-private';
+import type { SpanDefinition } from '../../api-private';
+
+export interface EventResult extends Duration, Partial<ErrorPayload> {
+  readonly success: boolean;
```
kaizencc:
```suggestion
```

packages/aws-cdk/lib/cli/io-host/messages.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export interface Start {}
+
+/**
+ * Private message types specific to the CLI
+ */
+export const CLI_PRIVATE_IO = {
```
kaizencc:
```suggestion
export const CLI_TELEMETRY_IO = {
```

packages/aws-cdk/lib/cli/io-host/messages.ts
```diff
@@ -0,0 +1,41 @@
...
+    code: 'CDK_CLI_I2001',
+    description: 'Command has finished executing',
+    interface: 'EventResult',
+  }),
+}
+
+/**
+ * Payload type of the end message must extend Duration
+ */
+export const CLI_PRIVATE_SPAN = {
```
kaizencc:
```suggestion
export const CLI_TELEMETRY_SPAN = {
```

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: `node -e 'require.resolve'`

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -55,13 +55,24 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
   const ioHost = CliIoHost.instance({
     logLevel: ioMessageLevel,
     isTTY: process.stdout.isTTY,
     isCI: Boolean(argv.ci),
     currentAction: cmd,
     stackProgress: argv.progress,
+    arguments: argv,
+    context: configuration.context,
   }, true);
+  await ioHost.begin();
```
kaizencc: needed for asynchronous construction events

packages/aws-cdk/lib/cli/io-host/messages.ts
```diff
@@ -0,0 +1,49 @@
+import { Duration, ErrorPayload } from '@aws-cdk/toolkit-lib';
+import * as make from '../../api-private';
+import type { SpanDefinition } from '../../api-private';
+
+export interface EventResult extends Duration, Partial<ErrorPayload> {}
+
+export interface EventStart {}
```
kaizencc: still couldn't think of what to put here

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -153,6 +193,86 @@ export class CliIoHost implements IIoHost {
...
+    this.arguments = props.arguments;
+    this.context = props.context;
+  }
+
+  /**
+   * Asynchronous constuction that needs to be called right after construction.
+   */
+  public async begin() {
+    if (true) {
+      this.commandSpan = await this.asIoHelper().span(CLI_PRIVATE_SPAN.COMMAND).begin({});
```
kaizencc: telemetry specific span. also if(True)

packages/aws-cdk/lib/cli/telemetry/collect-telemetry.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Whether or not we collect telemetry
+ */
+export function canCollectTelemetry(context: Context): boolean {
+  // TODO: remove this at launch. for now, this is an opt-in
+  if (!process.env.CLI_TELEMETRY) {
+    return false;
+  }
+
+  if (process.env.DISABLE_CLI_TELEMETRY || context.get('cli-telemetry') !== true) {
```
kaizencc: CDK_ - change in rfc and isseu

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -153,6 +193,86 @@ export class CliIoHost implements IIoHost {
...
+    }
+  }
+
+  /**
+   * When the command is complete, so is the CliIoHost. Ends the span of the entire CliIoHost
+   * and notifies with an optional error message in the data.
+   */
+  public async end(error?: ErrorDetails) {
+    await this.commandSpan?.end({ error });
+    this.commandSpan = undefined;
```
kaizencc: explain

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: move function into the telem session

packages/aws-cdk/lib/cli/telemetry/collect-telemetry.ts
kaizencc, line 1: this gets called in iohost constructor then we call `iohost.telemetry?.begin()`

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -153,6 +193,86 @@ export class CliIoHost implements IIoHost {
...
+        // start the telemetry session
+        await this.bindTelemetrySession(this.arguments, this.context.all);
+
+        // connect ctrl-c to ABORT event
+        process.on('SIGINT', async () => {
+          // Send a special Error 
+          await this.end({
+            name: 'AbortedError',
+          });
+        });
```
kaizencc: find the specific error and then get synth event to be abort

kaizencc: check if this is necessary -- does node process exit always handle `finally` codes

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -271,6 +393,20 @@ export class CliIoHost implements IIoHost {
...
+    try {
+      if (this.telemetrySession && isTelemetryMessage(msg)) {
+        await this.telemetrySession.emit({
+          eventType: getEventType(msg),
+          duration: msg.data.duration,
+          error: msg.data.error,
+        });
+      }
+    } catch (e: any) {
+      this.defaults.debug(`Emit Telemetry Failed ${e.message}`);
```
kaizencc: trace

packages/aws-cdk/lib/cli/telemetry/sanitation-utils.ts
```diff
@@ -0,0 +1,70 @@
+import { Command } from './schema';
+import * as fs from 'fs-extra';
+import { join } from 'path';
+
+export const CLI_TYPE_REGISTRY_FILE = 'cli-type-registry.json';
+
+/**
+ * argv is the output of yargs
+ */
+export async function sanitizeCommandLineArguments(argv: any): Promise<Command> {
```
kaizencc: lose async

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -153,6 +193,86 @@ export class CliIoHost implements IIoHost {
...
+   */
+  private async bindTelemetrySession(argv: any, context: {[key: string]: any}) {
+    // TODO: change this to EndpointTelemetryClient
+    const telemetryClient = new IoHostTelemetryClient({
+      ioHost: this,
+    });
+
+    // sanitize the raw cli input
+    const command = await sanitizeCommandLineArguments(argv);
+    const telemetryInfo: SessionSchema = {
```
kaizencc: ioHost.telemetry?.begin creates session info

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -0,0 +1,61 @@
...
+  private readonly client: ITelemetrySink;
+  private readonly info: SessionSchema;
+  private count = 0;
+
+  constructor(props: TelemetrySessionProps) {
+    this.client = props.client;
+    this.info = props.info;
+  }
+
+  public async emit(event: TelemetryEvent): Promise<boolean> {
```
kaizencc: return void

packages/aws-cdk/test/commands/diff.test.ts
```diff
@@ -560,6 +560,7 @@ describe('non-nested stacks', () => {
       configuration: cloudExecutable.configuration,
       sdkProvider: cloudExecutable.sdkProvider,
     });
+    notifySpy.mockClear();
```
kaizencc:
```suggestion
```

packages/aws-cdk/lib/init-templates/.init-version.json
```diff
@@ -1 +1 @@
-{"aws-cdk-lib": "2.202.0", "constructs": "^10.0.0"}
+{"aws-cdk-lib": "2.201.0", "constructs": "^10.0.0"}
```
kaizencc:
```suggestion
{"aws-cdk-lib": "2.202.0", "constructs": "^10.0.0"}
```

packages/aws-cdk/test/cli/telemetry/session.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    });
+    await ciSession.begin();
+
+    const privateCiInfo = (ciSession as any)._sessionInfo;
+    const privateInfo = (session as any)._sessionInfo;
+
+    // THEN
+    expect(privateCiInfo).toEqual(expect.objectContaining({
+      environment: expect.objectContaining({
+        ci: NOT_CI,
```
kaizencc:
```suggestion
        ci: Boolean(NOT_CI),
```

packages/aws-cdk/test/cli/telemetry/session.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      }),
+    }));
+    expect(privateInfo).toEqual(expect.objectContaining({
+      environment: expect.objectContaining({
+        ci: CI,
```
kaizencc:
```suggestion
        ci: Boolean(CI),
```

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -622,17 +632,28 @@ function determineHotswapMode(hotswap?: boolean, hotswapFallback?: boolean, watc
...
       process.exitCode = 1;
+    })
+    .finally(async () => {
+      if (!error && process.exitCode === 1) {
+        // The existence of an error determines if telemetry is successful or not so we create a
+        // dummy error in the event that exit code is 1 but no error is thrown
+        error = { name: 'ExitCode1Error' };
+      }
+
+      await CliIoHost.get()?.telemetry?.end(error);
```
mrgrain: Checking for `exitCode === 1` is not the correct metric for success/failure of the command. We have commands that deliberately return `1` as exit code, but the command succeeds.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -245,6 +281,8 @@ export class CliIoHost implements IIoHost {
    * The caller waits until the notification completes.
    */
   public async notify(msg: IoMessage<unknown>): Promise<void> {
+    await this.maybeEmitTelemetry(msg);
```
mrgrain, line 285: **issue:** In the RFC an design doc, we said Telemetry is not affecting user experience, i.e. performance. This is the main entry point to telemetry and it should NOT be awaited.

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -0,0 +1,162 @@
...
+  return false;
+}
+
+/**
+ * Returns true if the current process is running in a CI environment
+ * @returns true if the current process is running in a CI environment
+ */
+export function isCI(): boolean {
+  return process.env.CI !== undefined && process.env.CI !== 'false' && process.env.CI !== '0';
+}
```
mrgrain: todo: split this out into a separate file since this is now a shared function.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -55,14 +57,30 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
     isTTY: process.stdout.isTTY,
     isCI: Boolean(argv.ci),
     currentAction: cmd,
     stackProgress: argv.progress,
+    arguments: argv,
+    context: configuration.context,
   }, true);
 
+  try {
+    await ioHost.telemetry?.begin();
```
mrgrain: issue: as per design, we don't want telemetry to impact performance, this does.

packages/aws-cdk/lib/cli/version.ts
mrgrain, line 1: question: why these changes?

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -82,51 +83,64 @@ export class CloudExecutable implements ICloudAssemblySource {
...
+    } catch (e: any) {
+      error = e;
+    } finally {
+      await synthSpan.end({ error });
 
-      return new CloudAssembly(assembly, this.props.ioHelper);
+      if (error) {
+        throw error;
+      }
     }
```
mrgrain: quibble: this doesn't ready nice at all

if you want to different kind of endings, why not call `synthSpan.end()` twice: Once in the try block and once in the catch?

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -82,51 +83,64 @@ export class CloudExecutable implements ICloudAssemblySource {
...
+
+    throw new ToolkitError('Unreachable Code');
```
mrgrain: note: this is always a sign of a code smell.

mrgrain (dismissed):

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: determined that this isn't doing anything truly async, but could in the future.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/synth/cdk-synth-telemetry-with-errors.integtest.ts
```diff
@@ -0,0 +1,127 @@
...
+              staging: true,
+              notices: true,
+              ['no-color']: false,
+              ci: expect.anything(), // changes based on where this is called
+              validation: true,
+              quiet: false,
+            },
+            config: {
+              bags: true,
+              fileNames: true,
```
iliapolo: This doesn't seem right - are we accidentally recording properties of the class, instead of the actual context? 

https://github.com/aws/aws-cdk-cli/blob/666f1b9362930cf8602c1cc3ab1e3af54aa3fca3/packages/%40aws-cdk/toolkit-lib/lib/api/context.ts#L29-L30

packages/aws-cdk/README.md
```diff
@@ -1254,6 +1254,15 @@ cdk gc --unstable=gc
 The command will fail if `--unstable=gc` is not passed in, which acknowledges that the user
 is aware of the caveats in place for the feature.
 
+### `telemetry-file`
```
iliapolo, line 1257: Elaborate on:

- Path must be to a non existing file (and what happens if it does exist)
- Relation to telemetry status (does it consider it?)

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -648,17 +661,28 @@ function determineHotswapMode(hotswap?: boolean, hotswapFallback?: boolean, watc
...
       if (typeof value === 'number') {
         process.exitCode = value;
       }
     })
-    .catch((err) => {
+    .catch(async (err) => {
       // Log the stack trace if we're on a developer workstation. Otherwise this will be into a minified
       // file and the printed code line and stack trace are huge and useless.
       prettyPrintError(err, isDeveloperBuildVersion());
+      error = err;
```
iliapolo: Why do we think `error` and `err` have the same shape? Instead, extract the required information from `err`, and explicitly build an `ErrorDetails` struct.

packages/aws-cdk/lib/cli/cli.ts
iliapolo: The problem is we also have commands that swallow the actual error and return `1` as an indication of failure, for example [refactor](https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L1251-L1265):

https://github.com/aws/aws-cdk-cli/blob/666f1b9362930cf8602c1cc3ab1e3af54aa3fca3/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L1262-L1265

I do agree though that in general an exit code of `1` should not map to an event failure, and in this case the `refactor` command is actually incorrectly implemented. 

Lets do this then:

- An event failure is recorded only if `error !== undefined`, i.e only from the `catch` block.
- Fix the refactor command in a follow up (or before) and quickly check if there are other commands behaving like this.

Note that we can also optionally record a new `exitCode` metric that will be sent only in the `INVOKE` event, though I don't currently see what would we derive from it, given it doesn't indicate success/failure.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -155,6 +171,36 @@ export class CliIoHost implements IIoHost {
...
+    // }
+
+    if (sink) {
+      this.telemetry = new TelemetrySession({
+        ioHost: this,
+        client: sink,
+        arguments: args,
+        context: context,
+      });
+    }
```
iliapolo, line 199: nit (can ignore): All of this can be in the constructor because its not `async`.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -155,6 +171,36 @@ export class CliIoHost implements IIoHost {
...
+
+    await this.telemetry?.begin();
```
iliapolo, line 201: nit (can ignore): Then all this function does is this, and in that case maybe not even necessary, since the caller can just call this.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
iliapolo: @kaizencc 

> but could in the future.

Question is - do we want to allow it in the future? We could make `ITelemetrySync.emit` not a return a promise, and then:

```suggestion
    this.maybeEmitTelemetry(msg);
```

But this would make the `IoHostTelemetrySink` implementation kind of awkward, because it would have to implement `flush`, instead of just printing to screen on `emit`.

@mrgrain 

> we said Telemetry is not affecting user experience, i.e. performance. This is the main entry point to telemetry and it should NOT be awaited.

Regardless of this, the main CLI entry-point still has to `await` on the `ITelemetrySync.flush` when the invocation ends (even the periodic flush has a non zero impact since node is single threaded). 

Best we can do is ensure performance is not impacted **in a noticeable way**; which we do by setting a very small timeout on the network call. 

In any case, I don't think this is a blocker. @mrgrain ?

packages/aws-cdk/lib/cli/telemetry/installation-id.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as path from 'path';
+import type { IoHelper } from '../../api-private';
+import { cdkCacheDir } from '../../util';
+
+const INSTALLATION_ID_PATH = path.join(cdkCacheDir(), 'installation-id.json');
+
+/**
+ * Get or create installation id
+ */
+export async function getInstallationId(ioHelper: IoHelper) {
```
iliapolo:
```suggestion
export async function getOrCreateInstallationId(ioHelper: IoHelper) {
```

packages/aws-cdk/lib/cli/telemetry/library-version.ts
```diff
@@ -0,0 +1,30 @@
+import { exec } from 'child_process';
+import * as path from 'path';
+import { promisify } from 'util';
+import * as fs from 'fs-extra';
+import type { IoHelper } from '../../api-private';
+
+export async function getLibraryVersion(ioHelper: IoHelper): Promise<string | undefined> {
+  try {
+    const command = "node -e 'console.log(require.resolve(\"aws-cdk-lib\"))'";
```
iliapolo:
```suggestion
    const command = "node -e 'process.stdout.write(require.resolve(\"aws-cdk-lib\"))'";
```

Otherwise, you need to `.trim()` the `stdout`.

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -0,0 +1,147 @@
...
+        },
+        nodeVersion: process.version,
+      },
+      project: {},
+    };
+
+    // If SIGINT has a listener installed, its default behavior will be removed (Node.js will no longer exit).
+    // This ensures that on SIGINT we process safely close the telemetry session before exiting.
+    process.on('SIGINT', async () => {
+      await this.end({
```
iliapolo: Shouldn't we `try/catch` here?

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -0,0 +1,147 @@
...
+        name: 'ToolkitError',
```
iliapolo:
```suggestion
        name: ToolkitError.name,
```

Why not have a special name here? because we intentionally want an abort to be counted as a `ToolkitError`?

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -82,51 +83,64 @@ export class CloudExecutable implements ICloudAssemblySource {
...
-          // Cache the new context to disk
-          await this.props.configuration.saveContext();
-
-          // Execute again
-          continue;
         }
+        return new CloudAssembly(assembly, this.props.ioHelper);
       }
+    } catch (e: any) {
+      error = e;
```
iliapolo:
```suggestion
      error = e;
      throw e;
```

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -82,51 +83,64 @@ export class CloudExecutable implements ICloudAssemblySource {
...
+    } finally {
+      await synthSpan.end({ error });
 
-      return new CloudAssembly(assembly, this.props.ioHelper);
+      if (error) {
+        throw error;
+      }
```
iliapolo: No need if you re-throw from the `catch`.

packages/aws-cdk/lib/cxapp/cloud-executable.ts
iliapolo: No need if you re-throw from the `catch`.

packages/aws-cdk/test/cli/io-host/cli-io-host.test.ts
```diff
@@ -262,6 +279,110 @@ describe('CliIoHost', () => {
...
   });
 
+  describe('telemetry', () => {
+    let telemetryIoHost: CliIoHost;
+    let telemetryEmitSpy: jest.SpyInstance;
+    let telemetryFilePath: string;
+
+    beforeEach(async () => {
+      // Create a telemetry file to satisfy requirements; we are not asserting on the file contents
+      telemetryFilePath = path.join(os.tmpdir(), 'telemetry-file.json');
```
iliapolo: This won't create a unique file per execution. You need to use `mkdtempSync`.

packages/aws-cdk/test/cli/io-host/cli-io-host.test.ts
```diff
@@ -262,6 +279,110 @@ describe('CliIoHost', () => {
     });
   });
 
+  describe('telemetry', () => {
```
iliapolo, line 282: What about a test that ensures telemetry **isn't** emitted for non telemetry message codes?

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/synth/cdk-synth-telemetry-with-errors.integtest.ts
```diff
@@ -0,0 +1,127 @@
...
+            },
+          }),
+          state: 'SUCCEEDED',
+          eventType: 'SYNTH',
+        }),
+        identifiers: expect.objectContaining({
+          installationId: expect.anything(),
+          sessionId: expect.anything(),
+          telemetryVersion: '1.0',
+          cdkCliVersion: expect.anything(),
```
iliapolo, line 52: how come `cdkLibraryVersion` isn't in here? we should be able to know its exact value to assert on.

packages/aws-cdk/test/cli/telemetry/session.test.ts
```diff
@@ -0,0 +1,165 @@
...
+  test('can emit data to the client', async () => {
+    // WHEN
+    await session.emit({
+      eventType: 'SYNTH',
+      duration: 1234,
+    });
+
+    // THEN
+    expect(clientEmitSpy).toHaveBeenCalledWith(expect.objectContaining({
+      event: expect.objectContaining({
```
iliapolo, line 42: I would expect asserting on `eventType` and `duration` as well, to ensure you actually pass them through.

packages/aws-cdk/test/cli/telemetry/session.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    const ciSession = new TelemetrySession({
+      ioHost,
+      client: new IoHostTelemetrySink({ ioHost }),
+      arguments: { _: ['deploy'], STACKS: ['MyStack'] },
+      context: new Context(),
+    });
+    await ciSession.begin();
+
+    const privateCiInfo = (ciSession as any)._sessionInfo;
+    const privateInfo = (session as any)._sessionInfo;
```
iliapolo: Why rely on internal stuff and not use `emit`?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -648,17 +661,28 @@ function determineHotswapMode(hotswap?: boolean, hotswapFallback?: boolean, watc
...
       process.exitCode = 1;
+    })
+    .finally(async () => {
+      if (!error && process.exitCode === 1) {
+        // The existence of an error determines if telemetry is successful or not so we create a
+        // dummy error in the event that exit code is 1 but no error is thrown
+        error = { name: 'ExitCode1Error' };
+      }
+
+      await CliIoHost.get()?.telemetry?.end(error);
```
iliapolo: What ensures this doesn't throw? I think it makes sense to have any call that involves `.telemetry` to catch errors - not relying on the internal implementation.

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  return {
+    path,
+    parameters,
+  };
+}
+
+export function sanitizeContext(context: { [key: string]: any }) {
```
iliapolo:
```suggestion
export function sanitizeContext(context: Context) {
```

This will then surface the bug.

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,70 @@
...
+  const sanitizedContext: { [key: string]: boolean } = {};
+  for (const [flag, value] of Object.entries(context)) {
```
iliapolo:
```suggestion
  for (const [flag, value] of Object.entries(context.all)) {
```

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -0,0 +1,147 @@
...
+        sessionId: randomUUID(),
+        telemetryVersion: '1.0',
+        cdkCliVersion: versionNumber(),
+        cdkLibraryVersion: await getLibraryVersion(this.ioHost.asIoHelper()),
+      },
+      event: {
+        command: {
+          path,
+          parameters,
+          config: sanitizeContext(this.props.context),
```
iliapolo: This says `config`, but sanitizes `context`. Are we not sending general configuration thats not context? If not - lets call this `context`.

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+  return sanitizedContext;
+}
+
+function isBoolean(value: any): value is boolean {
+  return typeof value === 'boolean';
+}
+
+function isNumberOrBoolean(value: any): boolean {
+  return typeof value === 'number' || typeof value === 'boolean';
```
iliapolo:
```suggestion
  return typeof value === 'number' || isBoolean(value);
```

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,70 @@
...
+          path.push(`$${arg}${i+1}`);
+        }
+      } else {
+        path.push(`$${arg}1`);
+      }
+    }
+
+    // Continue if the arg name is not a global option or command option
+    if (argv[argName] === undefined || (!globalOptions.includes(argName) && !commandOptions.includes(argName))) {
+      continue;
```
iliapolo, line 35: Can this happen? give an example?

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,70 @@
...
+  const globalOptions: any[] = Object.keys(config.globalOptions);
+  const commandOptions: any[] = Object.keys(config.commands[command]?.options ?? {});
+  const commandArg: { name: string; variadic: string } | undefined = config.commands[command]?.arg;
+
+  for (const argName of Object.keys(argv)) {
+    if (argName === commandArg?.name) {
+      const arg = dropDuplicate(argName);
+      if (commandArg.variadic) {
+        for (let i = 0; i < argv[argName].length; i++) {
+          path.push(`$${arg}${i+1}`);
```
iliapolo:
```suggestion
          path.push(`$${argName}_${i+1}`);
```

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,70 @@
...
+  const path: string[] = [command];
+  const parameters: { [key: string]: string } = {};
+
+  const globalOptions: any[] = Object.keys(config.globalOptions);
+  const commandOptions: any[] = Object.keys(config.commands[command]?.options ?? {});
+  const commandArg: { name: string; variadic: string } | undefined = config.commands[command]?.arg;
+
+  for (const argName of Object.keys(argv)) {
+    if (argName === commandArg?.name) {
+      const arg = dropDuplicate(argName);
```
iliapolo: Pretty, but also very somewhat fragile because it assumes naming.

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (commandArg.variadic) {
+        for (let i = 0; i < argv[argName].length; i++) {
+          path.push(`$${arg}${i+1}`);
+        }
+      } else {
+        path.push(`$${arg}1`);
```
iliapolo:
```suggestion
        path.push(`$${argName}`);
```

iliapolo (changes requested):

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
kaizencc: yargs returns an object with aliases possible as well. so it could look like:

```
{
  'all': true',
  'a': true',
}
```

No reason to record the aliases; yargs will make sure that the "real" property name holds the same value.

kaizencc: i will update the comment to explain better

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: the reason why its here and not in the constructor is because `context` cannot get created before `IoHost` (since the `configuration` class logs errors to `ioHost` and there is a real dependency there)

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
kaizencc: judging by the other comments you want to remove `dropDuplicate` and i'm not married to it so i will

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/synth/cdk-synth-telemetry-with-errors.integtest.ts
kaizencc: forgot, so i added it in, but its `expect.anything()` because the best we can do is assert that its recorded.

packages/aws-cdk/test/cli/telemetry/session.test.ts
kaizencc: changed

packages/aws-cdk/lib/cli/telemetry/session.ts
kaizencc: we would have to rename this in the schema. however i think we may want to hold off on this entirely now

packages/aws-cdk/lib/cli/cli.ts
kaizencc: agree w the action items. will follow up with separate PRs for refactor and potentially others

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
mrgrain: Not a blocker. @kaizencc and I had discussed this in person. 

> We could make ITelemetrySync.emit not a return a promise, and then:

I'm very much in favour of this. It makes this non-async by design and carries less risk of introducing a blocking call later on. Of course sync code can also be blocking and impacting performance. An (easier) alternative would be to use `void` instead of `await` and add a comment explaining the why.

> But this would make the IoHostTelemetrySink implementation kind of awkward, because it would have to implement flush, instead of just printing to screen on emit.

See other comment, you could use `void` on the printing calls. We don't need to guarantee the order of these and since it's CLI only, it has no impact on integrators.

> Regardless of this, the main CLI entry-point still has to await on the ITelemetrySync.flush when the invocation ends (even the periodic flush has a non zero impact since node is single threaded).
> 
> Best we can do is ensure performance is not impacted in a noticeable way; which we do by setting a very small timeout on the network call.

Yes the flush is totally fine. The network call is not problematic (unless we `await` it). I/O including network calls are implemented asynchronously in native code and are actually non-blocking.

iliapolo: @mrgrain 

> I'm very much in favour of this. It makes this non-async by design and carries less risk of introducing a blocking call later on. Of course sync code can also be blocking and impacting performance. An (easier) alternative would be to use void instead of await and add a comment explaining the why.

I'm not (strongly) opposed to it. 

**For later consideration**: It would be nice to not have to introduce `void` calls, and instead maybe have async/sync variations in IoHost?

> (unless we await it)

Mm what do you mean here? we do actually `await` on the network promise:

https://github.com/aws/aws-cdk-cli/blob/8b5b952abd9d34bac21872de822e750d91657e85/packages/aws-cdk/lib/cli/telemetry/endpoint-sink.ts#L93

Do you mean we could just drop the `await`?

mrgrain: > I'm not (strongly) opposed to it.
> 
> **For later consideration**: It would be nice to not have to introduce `void` calls, and instead maybe have async/sync variations in IoHost?

"Not opposed" is not outright in favour though, is it? Sync cod has the problem that it can equally be blocking. Someone might see that this is sync, and then decide to do something like `fs.readFileSync('very-large-file-that-takes-ages')`. Of course nothing can really prevent that, but I feel an explicit `void` to explicitly indicate that "this code should be running outside the normal loop and not block anything", might be better. But I don't have strong feelings either way.

> Mm what do you mean here? we do actually await on the network promise:
> Do you mean we could just drop the await?

No of course you have to wait something (unless it's truly fire and forget). This is still fine because you are calling `flush` only from `setInterval` and don't (cannot) await the flush there.

iliapolo: As discussed offline:

> "Not opposed" is not outright in favour though, is it?

Not in favor because I think using `void` introduces some bad hygiene into the code, allowing async functions to disguise as sync. The specific situation we are in though seems to merit it. @kaizencc its your  call. 

> No of course you have to wait something

👍

packages/aws-cdk/README.md
```diff
@@ -1254,6 +1254,23 @@ cdk gc --unstable=gc
...
+cdk list --telemetry-file=my/file/path --unstable=telemetry
+```
+
+The supplied path must be a non existing file. If the file exists, it will fail to log telemetry
+data but the command itself will continue uninterrupted.
+
+> Note: [opting out](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0732-cdk-cli-telemetry.md#opt-out)
+> of sending telemetry data to AWS has no effect on the `--telemetry-file` option, and vice versa.
+> If you specify `--telemetry-file`, we will log your telemetry data to the given file path.
+> Independently, we will send telemetry data to AWS if `cdk cli-telemetry --status` says it is enabled.
```
kaizencc:
```suggestion
> Note: The file will be written to regardless of your opt-out status.
```

packages/aws-cdk/lib/cli/telemetry/error.ts
```diff
@@ -0,0 +1,12 @@
+import { ErrorName } from "./schema";
+
+export function cdkCliErrorName(name: string): ErrorName {
+  if (!isErrorName(name)) {
```
kaizencc:
```suggestion
  // We only record error names that we control. Errors coming from dependencies
  // contain text that we have no control over so it is safer to not send it.
  if (!isKnownErrorName(name)) {
```

packages/aws-cdk/lib/cli/telemetry/error.ts
```diff
@@ -0,0 +1,12 @@
...
+    return ErrorName.UNKNOWN_ERROR;
+  }
+  return name;
+}
+
+function isErrorName(name: string): name is ErrorName {
```
kaizencc:
```suggestion
function isKnownErrorName(name: string): name is ErrorName {
```

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -0,0 +1,78 @@
...
+import type { Context } from '../../api/context';
+
+export const CLI_TYPE_REGISTRY_FILE = 'cli-type-registry.json';
+
+/**
+ * argv is the output of yargs
+ */
+export function sanitizeCommandLineArguments(argv: any): { path: string[]; parameters: { [key: string]: string } } {
+  // Get the configuration of the arguments
+  const config = fs.readJSONSync(join(__dirname, '..', CLI_TYPE_REGISTRY_FILE));
```
iliapolo: I believe this won't bundle properly. @kaizencc it might be why your integ tests are failing. Consider using `require('cli-type-registry.json')`.

iliapolo (approved):

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
kaizencc: good catch :)

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
kaizencc: keeping the `await` for now -- i think its cleaner short-term hygiene and we can always make the change later.

PR #645, opened 2025-06-19
fix: CDK app fails to launch if paths contain spaces

### LINKED ISSUE #636: aws-cdk: cdk synth/deploy command fails with "C:\Program" is not recognized as an internal or external command
### Describe the bug

Short log:
```
PS1> cdk synth --all --app MyApp\dist\bin\app.js

'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

Subprocess exited with error 1
```

This happens every time I try to run `cdk` from a Powershell terminal, same happens when run from a cmd terminal.

Debug log:
```
PS1> cdk --version
2.43.1 (build c1ebb85)
PS1> node --version
v16.15.0
PS1> npm --version
8.13.1
PS1> cdk --verbose --debug synth --all --app C:\Users\sabatin\MyApp\dist\bin\app.js
CDK toolkit version: 2.43.1 (build c1ebb85)
Command line arguments: {
  _: [ 'synth' ],
  verbose: 1,
  v: 1,
  debug: true,
  all: true,
  app: 'C:\\Users\\sabatin\\MyApp\\dist\\bin\\app.js',
  a: 'C:\\Users\\sabatin\\MyApp\\dist\\bin\\app.js',
  lookups: true,
  'ignore-errors': false,
  ignoreErrors: false,
  json: false,
  j: false,
  ec2creds: undefined,
  i: undefined,
  'version-reporting': undefined,
  versionReporting: undefined,
  'path-metadata': true,
  pathMetadata: true,
  'asset-metadata': true,
  assetMetadata: true,
  'role-arn': undefined,
  r: undefined,
  roleArn: undefined,
  staging: true,
  'no-color': false,
  noColor: false,
  ci: false,
  validation: true,
  quiet: false,
  q: false,
  '$0': '..\\..\\AppData\\Roaming\\npm\\node_modules\\aws-cdk\\bin\\cdk'
}
cdk.json: {
  "app": "node dist/bin/app",
  "context": {
    "@aws-cdk/core:newStyleStackSynthesis": true,
    "@aws-cdk/aws-ec2:uniqueImdsv2TemplateName": true
  },
  "output": "cdk.out"
}
merged settings: {
  versionReporting: true,
  pathMetadata: true,
  output: 'cdk.out',
  app: 'C:\\Users\\sabatin\\MyApp\\dist\\bin\\app.js',
  context: {
    '@aws-cdk/core:newStyleStackSynthesis': true,
    '@aws-cdk/aws-ec2:uniqueImdsv2TemplateName': true
  },
  debug: true,
  assetMetadata: true,
  toolkitBucket: {},
  staging: true,
  bundlingStacks: [ '*' ],
  lookups: true
}
Determining if we're on an EC2 instance.
Reading cached notices from C:\Users\sabatin\.cdk\cache\notices.json
Does not look like an EC2 instance.
Toolkit stack: CDKToolkit
Setting "CDK_DEFAULT_REGION" environment variable to eu-west-1
Resolving default credentials
Retrieved account ID XXXXX from disk cache
Setting "CDK_DEFAULT_ACCOUNT" environment variable to XXXXX
context: {
  '@aws-cdk/core:newStyleStackSynthesis': true,
  '@aws-cdk/aws-ec2:uniqueImdsv2TemplateName': true,
  'aws:cdk:enable-path-metadata': true,
  'aws:cdk:enable-asset-metadata': true,
  'aws:cdk:version-reporting': true,
  'aws:cdk:bundling-stacks': [ '*' ]
}
outdir: cdk.out
env: {
  CDK_DEFAULT_REGION: 'eu-west-1',
  CDK_DEFAULT_ACCOUNT: 'XXXXX',
  CDK_DEBUG: 'true',
  CDK_CONTEXT_JSON: '{"@aws-cdk/core:newStyleStackSynthesis":true,"@aws-cdk/aws-ec2:uniqueImdsv2TemplateName":true,"aws:cdk:enable-path-metadata":true,"aws:cdk:enable-asset-metadata":true,"aws:cdk:version-reporting":true,"aws:cdk:bundling-stacks":["*"]}',
  CDK_OUTDIR: 'cdk.out',
  CDK_CLI_ASM_VERSION: '21.0.0',
  CDK_CLI_VERSION: '2.43.1'
}
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
Reading cached notices from C:\Users\sabatin\.cdk\cache\notices.json

Subprocess exited with error 1
Error: Subprocess exited with error 1
    at ChildProcess.<anonymous> (C:\Users\sabatin\AppData\Roaming\npm\node_modules\aws-cdk\lib\api\cxapp\exec.ts:131:23)
    at ChildProcess.emit (node:events:527:28)
    at ChildProcess.emit (node:domain:475:12)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:291:12)
```

Note also that the `exec.ts` file is missing:
```
C:\Users\sabatin\AppData\Roaming\npm\node_modules\aws-cdk\lib\api\cxapp\exec.ts
```

although the corresponding `exec.js` file is there (tried to edit but it's not affecting the output of the command). What is this `exec.ts` file about and why is it missing?

Possibly related to:
https://github.com/serverless-stack/sst/issues/73

BTW adding more logs would be useful to diagnose the issue.

### Expected Behavior

The command should not fail.

### Current Behavior

The command fail.

### Reproduction Steps

See above.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.43.1 (build c1ebb85)

### Framework Version

_No response_

### Node.js Version

v16.15.0

### OS

Windows

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

rix0rrr:
When executing CDK apps, users specify the command as a string. The only feasible interpretation of that is by executing the command line through a shell, either `bash` or `cmd.exe`. We are using shell execution on purpose; it's used for tests, it's necessary on Windows to properly execute `.bat` and `.cmd` files, and since we have historically offered it you can bet dollars to doughnuts that customers have built workflows depending on that.

Note that the source of the string and the machine it's executing on are part of the same security domain: it's all "the customer". The customer writes the command string, then executes it on their own machine. Customers don't build strings from untrusted sources.

However, in code we did do trivial parsing of that `string` to `string[]`, do some manipulation, and then combine it back to a `string`, by doing splitting and joining on `" "`. This led to an error on Windows systems where we would build an invalid command line if the path points to a `.js` file and the path  to the Node interpreter has spaces in it (`C:\Program Files`) (or the path of the `.js` file itself).

Instead of splitting and joining on on `" "`, in this PR we parse and join according to the rules of the shell (be it either POSIX-compatible or cmd.exe-compatible). This is quite tricky and it might be incorrect. Still, we are not doing this with the goal of nullifying the active shell characters; we are just trying to account for quotes and spaces. 

This PR fixes several instances where we were building a command line by joining an array of command line arguments with spaces. This would then cause problems if any of the parts of the command line contain spaces themselves, because we have lost information about the argument separators.

Instead, keep the structure information (keeping the command line in `string[]` form as long as possible) so that on Windows, where spaces in command lines are common, executions don't fail.

Closes https://github.com/aws/aws-cdk-cli/issues/636

---


packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/command-line.ts
```diff
@@ -0,0 +1,266 @@
...
+  return ret;
+}
+
+function isWhitespace(char: string): boolean {
+  return char === ' ' || char === '\t';
+}
+
+function isWindows(x?: ShellSyntax) {
+  return x ? x === 'windows' : process.platform === 'win32';
+}
```
mrgrain: I don't think we can make this assumption. What about Windows users using `bash`?

rix0rrr: I know, but `subprocess.spawn(..., { shell: true })` will always use `cmd.exe` on Windows.

If you want bash you'll need to explicitly do `subprocess.spawn("bash -c \"...\"", { ... })`.

rix0rrr: Well actually it does respect `%COMSPEC%`, so let us also respect that.

rix0rrr:
Tested on a Windows machine:

```sh
C:\...> node_modules\.bin\cdk -a "hello.js" synth
# ✅ works

C:\...> node_modules\.bin\cdk -a "dir with spaces\hello.js" synth
# ✅ doesn't work, but that's expected because the argument to `-a` must be a command line, not a filename, so we need to put "" inside the command line string

C:\...> node_modules\.bin\cdk -a "\"dir with spaces\hello.js\"" synth -v
C:\...> node_modules\.bin\cdk -a "^"dir with spaces\hello.js^"" synth -v
# ❌ doesn't work, because I can't figure out how to pass quotes inside quotes in a way that passes both through the initial parsing of cmd.exe and then the GetCommandLineArgvW used inside Node.js. I can tell from the verbose logging that the quotes get lost before they get to the CLI code.

# ...but his is not really incorrect and I'm sure someone smarter than me can do it, because...
{
  "app": "\"dir with spaces\hello.js\""
}
C:\...> node_modules\.bin\cdk synth
# ✅ works! The JSON serves as a reliable way to pass those quotes to the CLI and if they are there, it all works correctly.
```

PR #656, opened 2025-06-24
fix: longer timeout for notices in the testing environment

rix0rrr:
We're observing a lot of these errors in our integration tests:

```
[09:29:38] Could not refresh notices: _ToolkitError: Network error: Request timed out
```

It could be that our timeout is too short on a heavily congested machine. Increase the timeout specifically for those cases to improve the reliability of our tests.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
@@ -44,7 +44,11 @@ export class WebsiteNoticeDataSource implements NoticeDataSource {
   }
 
   async fetch(): Promise<Notice[]> {
-    const timeout = 3000;
+    // We are observing lots of timeouts when running in a massively parallel
+    // integration test environment, so wait for a longer timeout there.
+    //
+    // In production, have a short timeout to not hold up the user experience.
+    const timeout = process.env.TESTING_CDK ? 30_000 : 3_000;
 
     const options: RequestOptions = {
       agent: this.agent,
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
```diff
@@ -44,7 +44,11 @@ export class WebsiteNoticeDataSource implements NoticeDataSource {
   }
 
   async fetch(): Promise<Notice[]> {
-    const timeout = 3000;
+    // We are observing lots of timeouts when running in a massively parallel
+    // integration test environment, so wait for a longer timeout there.
+    //
+    // In production, have a short timeout to not hold up the user experience.
+    const timeout = process.env.TESTING ? 30_000 : 3_000;
```
mrgrain: Should we have a more unique env name here? This name feels generic enough for customers to have it set in their enviornment.

mrgrain (changes requested):
Is this a public facing `fix` or a `chore`?

rix0rrr:
> Is this a public facing `fix` or a `chore`?

I want to make sure it gets released, that's why it's a fix.

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
rix0rrr: I was debating that. I was actually looking for something like `CI`, an officiously standardized environment variable, but couldn't find one. 

Obviously I can turn it into `CDK_TESTING` if you'd prefer.

mrgrain (approved):

PR #663, opened 2025-06-27
fix(integ-testing): add test that forces cdk-assets to use the Docker credentials helper

rix0rrr:
This runs `docker-credential-cdk-asset` (the Docker helper tool) and checks its output to see that it actually usefully does something.

- I would have liked to do a full integration between `cdk-assets` and the correct Docker config, but unfortunately this only makes sense for a 2nd registry (not the primary registry we're pushing to). So we would need either:
  - A fake docker registry; I tried looking for one or faking one via a static HTTP server, but the requirement of signing the image manifest got me stuck. Also, the Docker registry/HTTP server needs to be accessible from the Docker Host, so it can't easily run inside the CodeBuild/GHA container; we would need to set up AWS infra to host this static server which implies S3 + CloudFront or APIGW + Lambda, which starts to get annoying.
  - A real Docker registry that requires user/pass authentication; but now we have to manage a username and password and security is going to make us rotate it.
  - An ECR repository; it must be in a different account or region, because the registry name (`<account>.dkr.ecr.<region>.amazonaws.com`) must be different. To make it in a different account or region, we would need to run Atmosphere twice, but that doesn't help when we're running locally or in a non-Atmosphere enabled environment.

- I would also have liked to do a test that reads from SecretsManager, but at least the Atmosphere roles don't have permissions to use SecretsManager and I'm not sure what other environments these tests run in, so this test is commented out for now.

## Also in this PR

- Features to create and clean up a role in the integ tests.
- Properly configure the `cdk-assets` version to test (not just `latest`, that will work in our online builds with a fake NPM repo but won't work locally, for example). This needed some refactoring of the "package sources" code to generalize the classes that were dealing with the `aws-cdk` package to take a package name.
- Refactored the assets helper code a bit to pull out some helper functions used to prepare assets.
- Fix a bug in `yarn-cling` that allows hoisting conflicting dependencies into certain places if `yarn` has already hoisted them during the proper install.

---


packages/@aws-cdk/cli-lib-alpha/THIRD_PARTY_LICENSES
mrgrain, line 1: That doesn't look right

packages/@aws-cdk/integ-runner/THIRD_PARTY_LICENSES
mrgrain, line 1: That doesn't look right

packages/aws-cdk/THIRD_PARTY_LICENSES
mrgrain, line 1: That doesn't look right

packages/cdk-assets/THIRD_PARTY_LICENSES
mrgrain, line 1: That doesn't look right

packages/@aws-cdk-testing/cli-integ/lib/aws.ts
```diff
@@ -50,8 +54,10 @@ export class AwsClients {
   public readonly iam: IAMClient;
   public readonly lambda: LambdaClient;
   public readonly sts: STSClient;
+  public readonly secretsManager: SecretsManagerClient;
 
-  constructor(
+  private constructor(
+    private readonly randomString: string,
```
mrgrain, line 61: Does this have to be externalized? Or could this just be internal to the class?

rix0rrr: This is the unique identifier assigned for the test. Yes, we could make a fresh unique identifier for "just" temporary AWS resources, it wouldn't really matter. In fact we could generate a unique identifier for every resource.

But it just seems nice for traceability if we use the same unique string everywhere, so that we can relate test executions and logs and resources to each other.

packages/@aws-cdk-testing/cli-integ/package.json
```diff
@@ -76,6 +76,7 @@
     "@aws-sdk/client-iam": "3.839.0",
     "@aws-sdk/client-lambda": "3.839.0",
     "@aws-sdk/client-s3": "3.839.0",
+    "@aws-sdk/client-secrets-manager": "^3.840.0",
```
mrgrain, line 78: this should probably be the same version and constraint as the others

rix0rrr: I made them all `^3.840.0`

mrgrain (approved):

PR #665, opened 2025-06-27
chore: cache nx across builds on the same PR

mrgrain:
Fixes #

---

### DIFF
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -1770,6 +1770,7 @@ const disallowed = new Set([
 ]);
 repoProject.github?.tryFindWorkflow('pull-request-lint')?.file?.patch(
   pj.JsonPatch.replace('/jobs/validate/steps/0/with/scopes', [
+    'ci',
     'cli',
     'deps',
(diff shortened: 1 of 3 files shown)


.projenrc.ts
```diff
@@ -276,12 +272,22 @@ const repoProject = new yarn.Monorepo({
 
   buildWorkflowOptions: {
     preBuildSteps: [
+      {
+        name: 'Nx cache',
+        uses: 'actions/cache@v4',
+        with: {
+          'key': '${{ runner.os }}-nxcache',
+          'path': ['.nx/cache'].join('\n'),
```
rix0rrr: I'm not sure this is safe. I regularly need to blow the nx cache away in order to get things to builder properly.

mrgrain: I would be safe since `integ` and `merge-queue` builds won't be using the cache. It might however be annoying. Currently doing this to suss out what the potential gains would be.

rix0rrr (approved):
Mmmmkay

PR #666, opened 2025-06-27
fix(integ-runner): integ-runner produces snapshot that doesn't validate

rix0rrr:
The integ-runner uses `synthFast` to produce the assembly that will be used to compare the snapshot to, to determine if there is "a change".

However, when a snapshot has changed and a test is run, there is a fork:

- In dry-run mode: `synthFast` will be used to produce the updated snapshot.
- In real-mode, the result of `this.deploy()` will be the updated snapshot (but also: the next time it will be compared to the result of `synthFast`).

We are running into a situation where the results of `synthFast` and `this.deploy` are different, and snapshot validation fails.

In this change, we always generate the snapshot using `synthFast`, and ignore the template produced by `this.deploy()`. This makes sure that the templates will at least compile equal.

I'm not sure why this ever worked this way, because the template that is deployed can be specialized to an account and region, but the snapshot can never be, so we need to investigate a little more.

---

### DIFF
--- a/packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
+++ b/packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
@@ -7,7 +7,7 @@ import * as chokidar from 'chokidar';
 import * as fs from 'fs-extra';
 import * as workerpool from 'workerpool';
 import type { IntegRunnerOptions } from './runner-base';
-import { IntegRunner, DEFAULT_SYNTH_OPTIONS } from './runner-base';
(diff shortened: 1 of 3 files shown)


rix0rrr:
Okay the actual issue we had related to the flag `@aws-cdk/aws-lambda:createNewPoliciesWithAddToRolePolicy` being different between the 2 modes.

Don't know what that was about, but I still feel the snapshot-that-gets-commited and the snapshot-that-gets-compared should use the same code paths.

mrgrain (approved):

PR #667, opened 2025-06-27
feat: modification is forbidden during a refactor

### LINKED ISSUE #657: Forbid modifications
## What's wrong

Suppose you have two resources, with logical IDs `Foo` and `Bar`, living in the same stack. `Foo` has a reference to `Bar`:

```
Resources:
  Foo:
    ...
    Properties:
      SomeProp:
        Ref: Bar # <-- in-stack reference

  Bar: ...
```

Using CloudFormation's stack refactoring API, you can move `Bar` to another stack, as long as you keep `Foo` pointing to it. The way to represent this cross-stack reference in CloudFormation is to create an output in the target stack referencing the resource you want. This output must have an export name which can then be imported using the `Fn::ImportValue` intrinsic function in the consuming stack:

```
# Original stack
Resources:
  Foo:
    ...
    Properties:
      SomeProp:
        Fn::ImportValue: RefBarExport # <-- replaced with a cross-stack one
```

```
# New stack
Resources:
  Bar: ...
Outputs:
  Value:
    Ref: Bar
  Export:
    Name: RefBarExport
```

Converting an in-stack reference into a cross-stack reference like this poses no problems when using CloudFormation templates in isolation. But when the templates are the output of a CDK application, there is a case where we may run into problems. Here is an example:

Suppose you have a Lambda function and a DynamoDB table in the same stack, with the function referencing the table (via a `Ref`, for example). This is the deployed state of the infrastructure. Then, in their CDK application, the user moves the table to another stack and, at the same time, replaces the function with another one, `Lambda'`. The matching algorithm will detect that the table was moved from Stack X to Stack Y. But since `Lambda'` is different from `Lambda`, the function is not going to be part of the mappings sent to the stack refactoring API. The result of this refactor is that the table will be moved to the other stack, but the function will stay where it was. Since the stack refactoring API doesn’t allow modifications, `Lambda` still needs to reference `Table`, even if they are in different stacks. The natural thing to do would be to convert the `Ref` into an `Fn::ImportValue`, as described above.

But here is the problem: `Fn::ImportValue` needs a corresponding export. And, in this case, there is none. Note how Stack Y, in the CDK output has no need to export anything, even after the move. We could create an ad-hoc export just for the refactor, but that would cause another problem: the next deploy would fail, as it would try to remove the export while an import for it (in `Lambda`) still exists. We could then try to solve that, by doing a corrective deployment immediately after. First deploying Stack X, thus removing the importer, then deploying Stack Y, which would safely remove the export and add the new function. But in the period between the two deploys, there would be no Lambda function at all, potentially causing downtime for the customer.

In summary: **if a refactor would create cross-stack references that don’t exist in the local CDK output, that refactor is invalid, and cannot be executed.**

## Solution

Add a validation to only allow a refactor if the deployed stack and the CDK output are exactly the same, up to resource locations. In other words, you can move stuff around, but if you change anything, like add, remove or change a resource, you'll get an error message. So, instead of building a more flexible experience on top of CFN, we impose the same constraints they do. In exchange, we gain predictability and also, when we refuse to do a refactor, it's easy to explain to the user why.

otaviomacedo:
CloudFormation’s stack refactoring API does not allow modifications (other than resource moves). Initially we wanted to give CDK users a more flexible experience, by allowing them to move resources while also adding and removing other resources (and, in some cases, updating as well). These additions and removals were ignored by the toolkit, and only what has moved was sent to the API for refactoring. As a result of this flexibility, the refactor operation may leave the infrastructure in a state that is not the same as the CDK application.

This becomes a problem when this hypothetical intermediate state has cross-stack references that don’t exist in the CDK application. Due to the way CloudFormation deals with references and exports, we would either not be able to create this kind of reference, or create it in a way that blocks subsequent deployments.

This PR changes the comparison algorithm, to enforce the same constraint that the CloudFormation API does, but at the CDK level. Now the set of deployed resources and the set of local resources should be the same. Only resource moves are allowed.

The filtering strategy has also changed. Now, the stacks included in the comparison are as follows:

**Local**: all local stacks, filtered according to the provided patterns (if any)

**Deployed**: all deployed stacks whose name satisfy at least one of these properties:
- there exists a local stack (after the filter) with the same name.
- was explicitly included by the user

Filters are provided to the CLI in the usual way, as a positional parameter to the `refactor` command:

```shell
cdk refactor *Gamma
```

Additional stack names are provided via a new CLI option:

```shell
cdk refactor --additional-stack-name=Foo --additional-stack-name=Bar
```

Closes https://github.com/aws/aws-cdk-cli/issues/657.

---


packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -71,16 +70,25 @@ export interface RefactorOptions {
...
    */
-  stacks?: StackSelector;
+  mappingSource?: MappingSource;
 
   /**
-   * How the toolkit should obtain the mappings
+   * List of patterns for filtering local stacks. If no patterns are passed,
+   * then all stacks, except the bootstrap stacks are considered. If you want
+   * to consider all stacks (including bootstrap stacks), pass the wildcard
+   * '*'.
```
mrgrain: question(blocking): Why are we considering the bootstrap stack for this? Seems to be the first time we are doing this?

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -71,16 +70,25 @@ export interface RefactorOptions {
   readonly dryRun?: boolean;
 
   /**
-   * Criteria for selecting stacks to deploy
-   *
-   * @default - All stacks
+   * How the toolkit should obtain the mappings
    */
-  stacks?: StackSelector;
```
mrgrain, line 78: > Because of this more stringent validation, users are more likely to run into scenarios that block a refactor. In particular, the deployed infrastructure may have more stacks, that are not part of the CDK application at hand. 

question(blocking): I'm not sure I follow this reasoning. It sounds to me like we want to offer refactoring for stacks that are not part of the cdk. That's an interesting change to how we do things traditionally.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/context.ts
```diff
@@ -135,35 +131,55 @@ function resourceDigests(stacks: CloudFormationStack[]): [string, ResourceLocati
...
     .filter(([pre, post]) => pre.length === 1 && post.length === 1 && !pre[0].equalTo(post[0]))
-    .map(([pre, post]) => new ResourceMapping(pre[0], post[0]))
-    .filter(stacksPredicate);
+    .map(([pre, post]) => new ResourceMapping(pre[0], post[0]));
+}
+
+/**
+ * Are two sets equal to each other
+ */
+function equalSets<A>(a: Set<A>, b: Set<A>) {
```
mrgrain: suggestion: I'm sure we already have this somewhere. if not, let's move it to utils

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/index.ts
```diff
@@ -135,30 +143,86 @@ export async function getDeployedStacks(
...
+    return (s) => !Object.entries(s.template?.Resources ?? {})
+      .some(([logicalId, resource]: [string, any]) => logicalId === 'CdkBootstrapVersion' && resource.Type === 'AWS::SSM::Parameter');
+  }
+
+  if (patterns.includes('*')) {
+    return () => true;
+  }
+
+  return (s: S) => patterns.some(pattern => minimatch(id(s), pattern));
+}
```
mrgrain: note: I really don't like that refactoring is having it's own stack selection

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/index.ts
```diff
@@ -135,30 +143,86 @@ export async function getDeployedStacks(
...
+    const key = await fn(x);
+    const group = ret.get(key);
+    if (group) {
+      group.push(x);
+    } else {
+      ret.set(key, [x]);
+    }
+  }
+  return ret;
+}
```
mrgrain: suggestion: seems like a generally useful util that can go into utils

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1072,73 +1064,51 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
-      return result;
+    }
+
+    async function notifyInfo(message: string, data: any = {}) {
+      await ioHelper.notify(IO.CDK_TOOLKIT_I8900.msg(message, data));
+    }
+
+    async function notifyError(message: string, error: Error) {
+      await ioHelper.notify(IO.CDK_TOOLKIT_E8900.msg(message, { error }));
     }
```
mrgrain: quibble: Eh, really? Seems like we are not saving that much keystrokes here.

mrgrain (changes requested):
Code looks good, but I'm not convinced of this new `localStack` vs `deployedStack` thing. I'm sure I'm missing something, but feels like this warrants a discussion.

Also honestly could be a separate PR.

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
otaviomacedo: The problem here is this: we want to compare all the deployed stacks with the local stacks to come up with a list of resources that moved locations. In the usual case, the user has at least two stacks in their target account: the bootstrap stack and at least another stack for the application. So the deployed resource graph will be different from the local graph most of the time (deployed graph has resources that are not present locally). Since we are requiring that the two graphs be the same, up to resource moves, this will lead to a failure. Hence the default filter.

Unlike other commands, that operate on each stack separately, `refactor` operates on many stacks at once.

The ideal solution, really, would be to have a reliable way to group CFN stacks by CDK application, and then only compare stacks that belong to the same application.

I'm open to considering alternative solutions, though :-)

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
otaviomacedo: In-lining vs. calling a function?

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/index.ts
otaviomacedo: Where should we put it? (Having in mind that this logic applies uniformly both to stacks in the cloud assembly and deployed stacks)

mrgrain: > (Having in mind that this logic applies uniformly both to stacks in the cloud assembly and deployed stacks)

No, that's my issue really. Cloud assembly stack selection now is different than for other actions.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: yeah

```ts 
await notifyInfo("foo", { bar: "baz" });
```

vs

```ts 
await ioHelper.notify(IO.CDK_TOOLKIT_I8900.msg("foo", { bar: "baz" }));
```

Just seems like it's abstracting away the message code/typing at the cost of forcing the user to think about the message code and payload.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1072,73 +1064,51 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+        await notifyInfo(formatTypedMappings(typedMappings), { typedMappings });
+
+        if (context.ambiguousPaths.length > 0) {
+          const paths = context.ambiguousPaths;
+          await notifyInfo(formatAmbiguousMappings(paths), { ambiguousPaths: paths });
+        }
+      } catch (e: any) {
+        await notifyError(e.message, e);
       }
-      return result;
```
mrgrain, line 1141: removed on purpose?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1072,73 +1064,51 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+        const typedMappings = mappings
+          .map(m => m.toTypedMapping())
+          .filter(m => m.type !== 'AWS::CDK::Metadata');
+
+        await notifyInfo(formatTypedMappings(typedMappings), { typedMappings });
+
+        if (context.ambiguousPaths.length > 0) {
+          const paths = context.ambiguousPaths;
+          await notifyInfo(formatAmbiguousMappings(paths), { ambiguousPaths: paths });
+        }
```
mrgrain: See other thread. Try to have a single result (or error) per action. This is a shift in mindset compared to imperative logging. What's better for a consumer of these messages? Multiple messages of the same code that have different payloads and one of them might not even come through? Or a single, coded message that always has the full payload and is guaranteed to be receieved?

```suggestion
        let refactorMessage = formatTypedMappings(typedMappings);
        const refactorResult = { typedMappings };
        await notifyInfo(formatTypedMappings(typedMappings), { typedMappings });

        if (context.ambiguousPaths.length > 0) {
          const paths = context.ambiguousPaths;
          refactorMessage += "\n" + formatAmbiguousMappings(paths);
          refactorResult.ambiguousPaths = paths;
        }
        
        await ioHelper.notify(IO.CDK_TOOLKIT_I8900.msg(refactorMessage, refactorResult));
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: In fact, it's also not used correctly anyway. `CDK_TOOLKIT_I8900` is the specific message used for the `RefactorResult`. It should only be used for that. That message also is a `result` level and not `info`.

As a rule of thumb: Have one result or error per action. Or at least per '|thing" if that makes more sense (stack group in this case).

For general messages you can use `await ioHelper.defaults.info(...)`

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1072,73 +1064,51 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
-    for (const context of nonAmbiguousContexts) {
-      const mappings = context.mappings.filter((m) => !exclude.isExcluded(m.destination));
-      const typedMappings = mappings.map(m => m.toTypedMapping());
-      const environment = context.environment;
-      await ioHelper.notify(IO.CDK_TOOLKIT_I8900.msg(formatTypedMappings(environment, typedMappings), {
-        typedMappings,
-      }));
-    }
+    for (let { environment, localStacks, deployedStacks } of groups) {
+      await notifyInfo(formatEnvironmentSectionHeader(environment));
```
mrgrain:
```suggestion
      await ioHelper.defaults.info(formatEnvironmentSectionHeader(environment));
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1072,73 +1064,51 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
-        const key = hashObject(environment);
-        environments.set(key, environment);
-        if (stackGroups.has(key)) {
-          stackGroups.get(key)![1].push(stack);
-        } else {
-          // The first time we see an environment, we need to fetch all stacks deployed to it.
-          const before = await getDeployedStacks(sdkProvider, environment);
-          stackGroups.set(key, [before, [stack]]);
+        if (mappings.length === 0 && context.ambiguousPaths.length === 0) {
+          await notifyInfo('Nothing to refactor.');
```
mrgrain:
```suggestion
          await ioHelper.defaults.info('Nothing to refactor.');
```

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
otaviomacedo: This was part of a function, that was moved and renamed.

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/index.ts
otaviomacedo: I see. But what's the alternative?

otaviomacedo: I guess it would be ok to have the filters work differently for local and deployed stacks... for local, use the regular cloud assembly filter. For deployed, use this one.

mrgrain (approved):

PR #673, opened 2025-06-30
feat(cli): user-provided cdk init templates

rohang9000:
Updated functionality allows CDK CLI users to pass in custom templates written in any CDK supported language from local directories, Git repositories, or NPM packages as parameters to the "cdk init" command.

---


custom-template-test3/README.md
iankhou, line 1: Why 3? Let's have descriptive names if we're going to commit these template examples.

packages/aws-cdk/lib/commands/init/github-template.ts
```diff
@@ -0,0 +1,30 @@
...
+  const tempDir = path.join(os.tmpdir(), `cdk-template-${Date.now()}`);
+  await fs.mkdirp(tempDir);
+
+  try {
+    // Clone the repository
+    await execAsync(`git clone --depth 1 ${githubUrl} ${tempDir}`);
+    return tempDir;
+  } catch (e) {
+    // Clean up on error
+    await fs.remove(tempDir).catch(() => {
```
iankhou, line 26: Why empty catch block here?

init-fixed.ts
```diff
@@ -0,0 +1,8 @@
+// Fixed version of the NPM template section with language auto-detection
+
+        template = await InitTemplate.fromPath(templatePath, templateName);
+
+        // Auto-detect language if not specified
+        if (!language && template.languages.length === 1) {
+          language = template.languages[0];
```
iankhou: I remember you mentioning we need to specify language, so autodetect can be removed?

template-name-summary.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+### 4. Implementation Logic
+When `--template-name` is provided:
+1. Clone the Git repository to temp directory
+2. Look for subdirectory matching the template name
+3. Validate it contains required `typescript/` directory and `info.json`
+4. Use that specific template instead of root directory
+
+### 5. Error Handling
```
iankhou: No need to specify here. How the error is handled should be evident to the user when they attempt to use the CLI.

debug-init.sh
iliapolo, line 1: Remove this file

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -15,83 +16,340 @@ const camelCase = require('camelcase');
...
-    await printAvailableTemplates(options.language);
-    throw new ToolkitError(`Unknown init template: ${type}`);
-  }
-  if (!options.language && template.languages.length === 1) {
-    const language = template.languages[0];
-    warning(
-      `No --language was provided, but '${type}' supports only '${language}', so defaulting to --language=${language}`,
+  try {
+    // BRANCH 1: Git Repository Template Source
+    if (options.gitUrl) {
```
iliapolo, line 79: The new options should be gated with the `unstable` feature protection. For example:

https://github.com/aws/aws-cdk-cli/blob/666f1b9362930cf8602c1cc3ab1e3af54aa3fca3/packages/aws-cdk/lib/cli/cli.ts#L280-L282

iliapolo (changes requested):

PR #678, opened 2025-07-02
fix: CCAPI provider does not do pagination

### LINKED ISSUE #587: (context-providers): Support pagination for listresources in CC API
### Describe the bug

When the lookups happen for resources like AWS::EC2::PrefixList, the lack of pagination leads to false errors like `RuntimeError: ValidationError: Could not find any managed prefix lists matching {"prefixListName":"xxxxxxxxxxxxx"}`

Internal TT - V1780034086, D249642744

### Regression Issue

_No response_

### Expected Behavior

The List API handles pagination to search through the pages and find the resource, and also make one call for common types. 

### Current Behavior

If there are 10 prefixlist lookups in the stack, now it's making 10 list calls which are throttled down the line and due to lack of pagination, it fails with validation error.

### Reproduction Steps

NA

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.195.0

### Framework Version

_No response_

### Node.js Version

-

### OS

macos

### Language

TypeScript, Python

### Language Version

_No response_

### Other information

_No response_

rix0rrr:
The CCAPI provider stops after the first page of results. This may cause for example EC2 Prefix Lists that exist to not be found, if they don't occur in the first page of results.

Make the provider retrieve all pages of results.

Also in this PR:

- New SDK mocks had been added without them being added to all the places where they needed to be added to be reset properly. Instead, put them all into an object so we can do a reliable `for` loop that will never go out of date again.
- Up the `MaxResults` to 100, so we need to do fewer paginations and the chances of the error message finding too many elements has a reasonable indicator of the number of resources actually found.

Fixes https://github.com/aws/aws-cdk-cli/issues/587

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/context-providers/cc-api-provider.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/context-providers/cc-api-provider.ts
@@ -101,23 +101,38 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
     expectedMatchCount?: CcApiContextQuery['expectedMatchCount'],
   ): Promise<FoundResource[]> {
     try {
-      const result = await cc.listResources({
-        TypeName: typeName,
-      });
-      const found = (result.ResourceDescriptions ?? [])
-        .map(foundResourceFromCcApi)
-        .filter((r) => {
-          return Object.entries(propertyMatch).every(([propPath, expected]) => {
-            const actual = findJsonValue(r.properties, propPath);
-            return propertyMatchesFilter(actual, expected);
-          });
+      const found: FoundResource[] = [];
+      let nextToken: string | undefined = undefined;
+
+      do {
+        const result = await cc.listResources({
+          TypeName: typeName,
+          MaxResults: 100,
+          ...nextToken ? { NextToken: nextToken } : {},
         });
+
+        found.push(
+          ...(result.ResourceDescriptions ?? [])
+            .map(foundResourceFromCcApi)
+            .filter((r) => {
+              return Object.entries(propertyMatch).every(([propPath, expected]) => {
+                const actual = findJsonValue(r.properties, propPath);
+                return propertyMatchesFilter(actual, expected);
+              });
+            }),
+        );
+
+        nextToken = result.NextToken;
+
+        // This allows us to error out early, before we have consumed all pages.
+        if ((expectedMatchCount === 'at-most-one' || expectedMatchCount === 'exactly-one') && found.length > 1) {
+          const atLeast = nextToken ? 'at least ' : '';
+          throw new ContextProviderError(`Found ${atLeast}${found.length} resources matching ${JSON.stringify(propertyMatch)}; expected ${expectedMatchCountText(expectedMatchCount)}. Please narrow the search criteria`);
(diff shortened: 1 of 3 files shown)


packages/@aws-cdk/toolkit-lib/lib/context-providers/cc-api-provider.ts
```diff
@@ -101,24 +101,39 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
+      const found: FoundResource[] = [];
+      let nextToken: string | undefined = undefined;
+
+      do {
+        const result = await cc.listResources({
+          TypeName: typeName,
+          ...nextToken ? { NextToken: nextToken } : {},
         });
+        // eslint-disable-next-line no-console
+        console.log(result);
```
mrgrain:
```suggestion
```

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/context-providers/cc-api-provider.ts
```diff
@@ -101,24 +101,37 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
+        );
+
+        nextToken = result.NextToken;
+
+        // This allows us to error out early, before we have consumed all pages.
+        if ((expectedMatchCount === 'at-most-one' || expectedMatchCount === 'exactly-one') && found.length > 1) {
+          throw new ContextProviderError(`Found ${found.length} resources matching ${JSON.stringify(propertyMatch)}; expected ${expectedMatchCountText(expectedMatchCount)}. Please narrow the search criteria`);
+        }
+      } while (nextToken);
+
```
mrgrain, line 133: Should we be looping through all results if we know we only want 0 or 1 results?

mrgrain: Never mind. I cannot read comments apparently.

packages/@aws-cdk/toolkit-lib/lib/context-providers/cc-api-provider.ts
```diff
@@ -101,24 +101,37 @@ export class CcApiContextProviderPlugin implements ContextProviderPlugin {
...
+                return propertyMatchesFilter(actual, expected);
+              });
+            }),
+        );
+
+        nextToken = result.NextToken;
+
+        // This allows us to error out early, before we have consumed all pages.
+        if ((expectedMatchCount === 'at-most-one' || expectedMatchCount === 'exactly-one') && found.length > 1) {
+          throw new ContextProviderError(`Found ${found.length} resources matching ${JSON.stringify(propertyMatch)}; expected ${expectedMatchCountText(expectedMatchCount)}. Please narrow the search criteria`);
```
mrgrain: The error messages is technically wrong now. Maybe something like?

```suggestion
          throw new ContextProviderError(`Found more than one resources matching ${JSON.stringify(propertyMatch)}; expected ${expectedMatchCountText(expectedMatchCount)}. Please narrow the search criteria`);
```

mrgrain (approved):

PR #679, opened 2025-07-02
feat: overrides to resolve mapping ambiguities

otaviomacedo:
Previously the CLI gave the user the option to pass a mapping file, that would be considered the source of truth for the refactor operation. When a list of mappings was passed, the CLI would use it instead of computing the mappings itself. 

But now we've put in place a more strict constraint: that the deployed set of resources and the output of `synth` should be the same, up to location changes. That means that the user doesn't have a choice anymore: the only valid mappings are the ones that satisfy this property. The only case in which it makes sense for the user to influence the mapping is in case of ambiguity.

In graph terms, we are requiring the resource graphs to be isomorphic. (In fact, the requirement is even stronger than that: every pair of resources in the isomorphism must also have the same content). Given this requirement, the only choice the user has is _which_ isomorphism they want, in case there is more than one. Multiple isomorphisms manifest themselves as an ambiguity, from the user's perspective.

For example, for a given pair of resource graphs, we could have two isomorphisms:

image

image

The labels in each node represent the location `StackName.LogicalId`. In this example, resources with logical ID `r4` and `r5`, `r6` and `r7` all have the same digest, which is the source of the ambiguity. 

This change replaces the idea of a source-of-truth-mapping-file with an override file, that can only be used to resolve ambiguities in the computed mappings.

By specifying an override like:

```json
{
    "environments": [
        {
            "account": "669420849322",
            "region": "us-east-2",
            "resources": {
                "S4.r4": "S4.r7"
            }
        }
    ]
}
```

the user is effectively choosing the second isomorphism.

For the same reason, this change also removes the ability to exclude resources from a refactor operation.

---


mrgrain (changes requested):
PR looks good, but

> For the same reason, this change also removes the ability to exclude resources from a refactor operation.

question: Is there really never a situation where a user legitimately only wants to refactor a subset of resources? 

At this stage less concerned about implementation (`--exclude` or `--override` or `-only), but the specific user story.

mrgrain (approved):

PR #680, opened 2025-07-02
chore(cli): diff tests should not depend on message order

kaizencc:
This turns into an error every time we add a new message to the ioHost.

---


packages/aws-cdk/test/commands/diff.test.ts
```diff
@@ -1425,7 +1426,7 @@ describe('stack display names', () => {
     });
 
     // THEN
-    const output = notifySpy.mock.calls[0][0].message.replace(/\x1B\[[0-?]*[ -/]*[@-~]/g, '');
+    const output = notifySpy.mock.calls.map(x => x[0].message).join('\n').replace(/\x1B\[[0-?]*[ -/]*[@-~]/g, '');
```
otaviomacedo: Can't you call `output()` here too?

rix0rrr (approved):

iliapolo (approved):

PR #689, opened 2025-07-07
fix(toolkit-lib): make default environment resolution optional

mrgrain:
This PR makes the default environment resolution optional in the toolkit-lib. It allows avoiding unnecessary STS calls when the environment is explicitly specified or when local actions are performed without internet access.

Then uses the new feature in `integ-runner` to disable the default env lookup in the toolkit-lib engine. The existing engine doesn't provide the env variables either, so it's clearly not needed. We don't want to add to this contract at this time and also resolving the account adds an unnecessary delay.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/prepare-source.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/prepare-source.ts
@@ -20,6 +20,22 @@ import type { Context, Env } from '../environment';
 import { prepareDefaultEnvironment, spaceAvailableForContext, guessExecutable, synthParametersFromSettings } from '../environment';
 import type { AppSynthOptions, LoadAssemblyOptions } from '../source-builder';
 
+export interface ExecutionEnvironmentOptions {
+  /**
+   * The directory the cloud assembly will be written to.
+   *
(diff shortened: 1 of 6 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/prepare-source.ts
```diff
@@ -33,18 +49,39 @@ export class ExecutionEnvironment implements AsyncDisposable {
...
-  public static async create(services: ToolkitServices, props: { outdir?: string } = {}) {
-    let tempDir = false;
-    let dir = props.outdir;
+  public static async create(services: ToolkitServices, options: ExecutionEnvironmentOptions) {
+    let outDirIsTemporary = false;
+    let dir = options.outdir;
     if (!dir) {
-      tempDir = true;
+      outDirIsTemporary = true;
       dir = fs.mkdtempSync(path.join(fs.realpathSync(os.tmpdir()), 'cdk.out'));
```
iankhou, line 57: What happens here if there is already a `cdk.out` directory in the tmp dir location?

otaviomacedo: `fs.mkdtempSync` will create a new directory under `/tmp/cdk.out`.

otaviomacedo (approved):

PR #699, opened 2025-07-08
feat: "cdk flags" command reports active and missing feature flags (unstable)

vivian12300:
- Read feature flag artifacts from the Cloud Assembly
- Introduces `cdk flags`, which displays user's current feature flag configuration in the form of a table

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1281,6 +1283,43 @@ export class Toolkit extends CloudAssemblySourceBuilder {
     }
   }
 
+  /**
+   * Retrieve feature flag information from the cloud assembly
+   */
+  public async flags(cx: ICloudAssemblySource): Promise<FeatureFlagReportProperties[]> {
+    const ioHelper = asIoHelper(this.ioHost, 'flags');
+    const assembly = await assemblyFromSource(ioHelper, cx);
```
rix0rrr: Needs an `await using`.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1281,6 +1283,43 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+    const artifacts = assembly.cloudAssembly.manifest.artifacts;
+    const allFeatureFlags: ArtifactManifest[] = [];
+
+    for (const [_, artifactInfo] of Object.entries(artifacts!)) {
```
rix0rrr: This works, but here's  a way to do this in fewer lines:

```ts
const allFeatureFlags = Object.values(artifacts).filter(a => a.type === ArtifactType.FEATURE_FLAG_REPORT);
```

The pattern "take a collection, iterate over it, and accumulate some entries into another collection" is very common and can be expressed more directly (imo, at least).

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1281,6 +1283,43 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      if (artifactInfo.type === ArtifactType.FEATURE_FLAG_REPORT) {
+        allFeatureFlags.push(artifactInfo);
+      }
+    }
+
+    const featureFlagReport: FeatureFlagReportProperties[] = [];
+
+    for (const report of allFeatureFlags) {
+      const flagData = report.properties as any;
```
rix0rrr: Don't do `as any`. Type errors hide every time you use `as any`. Cast it to the `cxschema.FeatureFlagReportProperties` object you know that it is.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1281,6 +1283,43 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      let individualReport: FeatureFlagReportProperties = {
+        module: flagData.module,
+        flags: {},
+      };
+      for (const [flagName, flagInfo] of Object.entries(flagData.flags)) {
+        const flag = flagInfo as FeatureFlag;
+        individualReport.flags[flagName] = {
+          userValue: flag.userValue ?? '-',
```
rix0rrr: I think deciding to render this as a `-` in a table is a frontend decision, not a backend decision.

We've now lost the ability to distinguish between users literally configuring `"-"` and the flag not being set.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1281,6 +1283,43 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+          recommendedValue: flag.recommendedValue,
+          explanation: flag.explanation ?? '',
+        };
+      }
+      featureFlagReport.push(individualReport);
```
rix0rrr: Again with a `push`.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1281,6 +1283,43 @@ export class Toolkit extends CloudAssemblySourceBuilder {
     }
   }
 
+  /**
+   * Retrieve feature flag information from the cloud assembly
+   */
+  public async flags(cx: ICloudAssemblySource): Promise<FeatureFlagReportProperties[]> {
```
rix0rrr: The return type here should not be the same type as the type in `cxschema`. I've told you before, but I've told you a lot of things so I don't blame you if you don't remember that :).

Define your own type, which looks a lot like (but may be different from) the upstream type. I was personally thinking something like this:

```ts
export interface FeatureFlag {
  readonly module: string;
  readonly name: string;
  readonly recommendedValue: unknown; // Can also be any, but 'unknown' is nice and accurate
  readonly userValue?: unknown;
  readonly explanation?: string;
}
```

One thing you'll notice is that I've changed the data model slightly from:

```
┌──────────┐         ┌──────────┐         ┌──────────┐
│          │    0..* │          │    0..* │          │
│ Assembly │────────▶│ Artifact │────────▶│   Flag   │
│          │         │          │         │          │
└──────────┘         └──────────┘         └──────────┘
```

To

```
┌──────────┐         ┌──────────┐
│          │    0..* │          │
│ Assembly │────────▶│   Flag   │
│          │         │          │
└──────────┘         └──────────┘
```

I'm figuring that at this level, it will be most convenient to just get a list of flags we can treat as a giant table. We don't care about the grouping by artifact anymore; instead, we'll lift the module name into the flag data.

It won't be much harder for filtering and grouping downstream if people want to do that (grouping and filtering tables is easy and well-understood), and for the common use cases it will be trivial to use.

.projenrc.ts
```diff
@@ -1543,7 +1543,7 @@ const integRunner = configureProject(
       '@aws-sdk/client-cloudformation',
     ],
     devDeps: [
-      'aws-cdk-lib',
+      'aws-cdk-lib@2.203.0',
```
rix0rrr: I think we only need this dependency to synthesize some assemblies with flags during tests, and only for the duration that the PR isn't released again.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -2039,10 +2039,10 @@ export interface DriftOptions {
 
 function buildParameterMap(
   parameters:
-  | {
-    [name: string]: string | undefined;
-  }
-  | undefined,
+    | {
```
rix0rrr, line 2025: Where are all these whitespace formatting changes coming from?

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -112,6 +112,9 @@ export async function makeConfig(): Promise<CliConfig> {
           'bootstrap-stack-name': { type: 'string', desc: 'The name of the CDK toolkit stack, if different from the default "CDKToolkit"', requiresArg: true },
         },
       },
+      'flags': {
+        description: 'Modify feature flag configurations.',
```
rix0rrr:
```suggestion
        description: 'View and toggle feature flags.',
```

?

packages/aws-cdk/lib/commands/flags.ts
```diff
@@ -0,0 +1,37 @@
...
+  info('Feature Flags Report:');
+  const headers = ['Feature Flag Name', 'Recommended Value', 'User Value'];
+
+  for (const report of flagsData) {
+    const flags = Object.entries(report.flags);
+    const columnWidths = [
+      Math.max(headers[0].length, ...flags.map(([name]) => name.length)),
+      Math.max(headers[1].length, ...flags.map(([, flag]) => String(flag.recommendedValue).length)),
+      Math.max(headers[2].length, ...flags.map(([, flag]) => String(flag.userValue).length)),
+    ];
```
rix0rrr: Good stuff, but this should probably be in a helper function.

And while you're writing that, search the code base because it would surprise me if we didn't have something like this already.

packages/@aws-cdk/toolkit-lib/test/toolkit-flags.test.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    flags.forEach((report) => {
+      expect(Object.keys(report.flags).length).toBeGreaterThan(0);
+      expect(report).toHaveProperty('module');
+      expect(report).toHaveProperty('flags');
+      expect(typeof report.module).toBe('string');
+      expect(typeof report.flags).toBe('object');
+
+      Object.entries(report.flags).forEach(([flagName, flag]) => {
+        expect(typeof flagName).toBe('string');
```
rix0rrr: This can never really fail, so it's really interesting to test for (unless you're going be telling me you are preventing the value from being a `symbol`...)

packages/@aws-cdk/toolkit-lib/test/toolkit-flags.test.ts
```diff
@@ -0,0 +1,251 @@
...
+      '@aws-cdk/aws-apigateway:usagePlanKeyOrderInsensitiveId',
+    ];
+
+    let foundFlags = 0;
+
+    flags.forEach((report) => {
+      expect(Object.keys(report.flags).length).toBeGreaterThan(0);
+      expect(report).toHaveProperty('module');
+      expect(report).toHaveProperty('flags');
+      expect(typeof report.module).toBe('string');
```
rix0rrr: This implies the `toHaveProperty`, so if you do this you don't need the other one. But why are we doing this?

packages/@aws-cdk/toolkit-lib/test/toolkit-flags.test.ts
```diff
@@ -0,0 +1,251 @@
...
+        expect(typeof flagName).toBe('string');
+        expect(flag).toHaveProperty('userValue');
+        expect(flag).toHaveProperty('recommendedValue');
+        expect(flag).toHaveProperty('explanation');
+
+        expect(typeof flag.explanation).toBe('string');
+
+        if (flag.userValue === null || flag.userValue === undefined) {
+          expect(flag.userValue).toBe('-');
+        }
```
rix0rrr:
```ts
x === null || x === undefined  /* implies that */ x === '-';
```

If this test passes, it's because `x` is never nullish.

packages/@aws-cdk/toolkit-lib/test/toolkit-flags.test.ts
```diff
@@ -0,0 +1,251 @@
...
+          expect(flag.explanation).toBe('');
+        }
+
+        if (expectedFlags.includes(flagName)) {
+          foundFlags++;
+        }
+      });
+    });
+
+    expect(foundFlags).toBe(expectedFlags.length);
```
rix0rrr: If you want to say "I want the full set of flags to include at least these 3" there is a more direct way of expressing that.

For example:

```ts
const foundFlagNames = /* ... fill this in ... */;

// Method 1
for (const expected of EXPECTED_FLAG_NAMES) {
  expect(foundFlagNames).toContain(expected);
}

// Method 2
expect(isSubset(EXPECTED_FLAG_NAMES, foundFlagNames)).toBeTrue();
```

But why do you want to assert that?

packages/@aws-cdk/toolkit-lib/test/toolkit-flags.test.ts
```diff
@@ -0,0 +1,251 @@
...
+    const mockFlags = await toolkit.flags(mockCloudAssemblySource as any);
+
+    expect(mockFlags.length).toBe(2);
+    expect(mockFlags[0].module).toBe('module1');
+    expect(mockFlags[1].module).toBe('module2');
+    expect(mockFlags[0].flags).toHaveProperty('flag1');
+    expect(mockFlags[1].flags).toHaveProperty('flag2');
+  });
+
+  test('filters out non-feature-flag artifacts', async () => {
```
rix0rrr: This test is not so interesting. The fact that you were able to read flags from a real manifest already tells me that.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -190,7 +190,7 @@ export class CdkToolkit {
       emojis: true,
       ioHost: this.ioHost,
       toolkitStackName: this.toolkitStackName,
-      unstableFeatures: ['refactor'],
+      unstableFeatures: ['refactor', 'flags'],
```
rix0rrr: question: Is this necessary? Where is `this.toolkit` used? Does it ever use the `flags` command?

packages/aws-cdk/lib/commands/flags.ts
```diff
@@ -0,0 +1,55 @@
...
+  rows.forEach(row => {
+    table += formatRow(row) + '\n';
+  });
+
+  table += separator;
+  return table;
+}
+
+export async function displayFlags(flagsData: FeatureFlag[]): Promise<void> {
+  info('Feature Flags Report:');
```
rix0rrr: question: I think we can potentially do without this header? Not sure.

packages/aws-cdk/lib/commands/flags.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  flagsData.forEach((flag, index) => {
+    if (index === 0 || flagsData[index].module !== flagsData[index - 1].module) {
+      rows.push([chalk.bold(`Module: ${flag.module}`), '', '']);
+    }
+
+    rows.push([
+      flag.name,
+      String(flag.recommendedValue),
+      flag.userValue === undefined ? '- <unset>' : String(flag.userValue),
```
rix0rrr: suggestion:

```suggestion
      flag.userValue === undefined ? '<unset>' : String(flag.userValue),
```

packages/aws-cdk/test/commands/flags.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+      }];
+
+    await displayFlags(flagsData);
+
+    const plainTextOutput = notifySpy.mock.calls.map(x => x[0].message).join('\n').replace(/\x1B\[[0-?]*[ -/]*[@-~]/g, '');
+    expect(plainTextOutput).toContain('Feature Flags Report:');
+    expect(plainTextOutput).toContain('Feature Flag Name');
+    expect(plainTextOutput).toContain('Recommended Value');
+    expect(plainTextOutput).toContain('User Value');
```
rix0rrr: You are asserting a lot of things here. Why are you asserting this, and what are you protecting against?

packages/aws-cdk/test/commands/flags.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        module: 'aws-cdk-lib',
+        name: '@aws-cdk/aws-s3:createDefaultLoggingPolicy',
+        recommendedValue: 'true',
+        userValue: null,
+        explanation: 'Create default logging policy for S3 buckets',
+      }];
+
+    await displayFlags(flagsData);
+
+    const plainTextOutput = notifySpy.mock.calls.map(x => x[0].message).join('\n').replace(/\x1B\[[0-?]*[ -/]*[@-~]/g, '');
```
rix0rrr: This regex seems like a `stripAnsi` sort of thing. First: do we not have one like those already? Second, all this "plainTextOutput" wrangling is not trivial and it's duplicated all over the place. You could stick this all into helper functions to make it more readable what's going on.

```ts
function plainTextOutputLines(): string[] { 
  /* logic here */
```

packages/aws-cdk/test/commands/flags.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    expect(plainTextOutput).toContain('v2');
+    expect(plainTextOutput).toContain('v1');
+    expect(plainTextOutput).toContain('42');
+    expect(plainTextOutput).toContain('0');
+  });
+
+  test('handles mixed data types in flag values', async () => {
```
rix0rrr, line 84: Does this test not subsume the previous one?

packages/aws-cdk/test/commands/flags.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    expect(plainTextOutput).toContain('Feature Flags Report:');
+    expect(plainTextOutput).toContain('Feature Flag Name');
+    expect(plainTextOutput).toContain('Recommended Value');
+    expect(plainTextOutput).toContain('User Value');
+    expect(plainTextOutput).toContain('@aws-cdk/core:enableStackNameDuplicates');
+    expect(plainTextOutput).toContain('true');
+    expect(plainTextOutput).toContain('false');
+  });
+
+  test('displays multiple feature flags', async () => {
```
rix0rrr, line 40: Does this test not subsume the previous one?

packages/aws-cdk/test/commands/flags.test.ts
vivian12300: I want to make sure the output being displayed to the user matches what I'm expecting (the headers and the information about the flag).

rix0rrr: Yes, but why? Or put another way: what are you protecting that someone might accidentally break? Because a lot of this is now double bookkeeping, but not in a way that seems immediately valueable. Also, testing that the text you unconditionally print did actually get unconditionally printed... is that worth it? 

If you just want to eyeball the output, `expect(...).toMatchSnapshot()` or `expect(...).toMatchInlineSnapshot()` might be appropriate.

If someone wants to change a column header, they now need to change that in two places (change it in the code, and change it here). That may be worth it if that column name encodes a public contract, but it's just annoying busywork if it doesn't.

My point is that tests aren't free: they are additional code to write, compile, run, and maintain. So I'd hope they assert something worthwhile, to pull their weight

(And this is a very trivial case, but that generalizes to more complicated tests)

rix0rrr (approved):

PR #708, opened 2025-07-11
feat: `diff` shows moved resources

### LINKED ISSUE #362: Add refactoring information to the `diff` command
The output of `cdk diff` should include a section for refactoring candidates.

otaviomacedo:
When a resource is moved from one location to another, the `diff` command will show two entries: an addition and a removal. This PR augments those entries to show where they moved to and from, respectively:

Drawing 2025-07-15 14 35 02 excalidraw

This information is not shown by default. To enable it, the user has to pass `--include-moves` to the CLI, and `includeMoves: true` to the toolkit.

Closes https://github.com/aws/aws-cdk-cli/issues/362

Depends on https://github.com/aws/aws-cdk-cli/pull/679.

---


packages/@aws-cdk/cloudformation-diff/lib/format.ts
```diff
@@ -239,6 +246,10 @@ export class Formatter {
     }
   }
 
+  private formatMove(move?: Move): string {
+    return !move ? '' : chalk.gray('(OR', chalk.italic(chalk.bold('move')), `${move.direction} ${move.stackName}.${move.resourceLogicalId} via refactoring)`);
```
rix0rrr: Can we make this slightly more visible than grey? Maybe yellow or orange or something?

packages/@aws-cdk/toolkit-lib/lib/actions/refactor/index.ts
```diff
@@ -142,3 +144,28 @@ export function parseMappingGroups(s: string) {
...
+): Promise<EnvironmentSpecificMappings[]> {
+  const groups = await groupStacks(sdkProvider, stackArtifacts, []);
+  return groups.map((group) => {
+    const context = new RefactoringContext({
+      ...group,
+      ignoreModifications,
+    });
+    return {
+      environment: context.environment,
+      mappings: Object.fromEntries(
```
rix0rrr, line 166: Rather than splatting down the object to a string here, aren't we better off maintaining the object and rendering it later when it comes time to present the diff?

rix0rrr (approved):
Conditionally approved, give my comments a thought

PR #709, opened 2025-07-11
refactor(cli): split version functions into those that display and those that do not

kaizencc:
Pulled out of #631. Needed so that we can access version functions within the IoHost without offending the circular dependency checker (which checks file imports)

---

### DIFF
--- a/packages/aws-cdk/lib/cli/version.ts
+++ b/packages/aws-cdk/lib/cli/version.ts
@@ -2,7 +2,7 @@
 import * as path from 'path';
 import { cliRootDir } from './root-dir';
 
-export function version() {
+export function versionWithBuild() {
   return `${versionNumber()} (build ${commit()})`;
 }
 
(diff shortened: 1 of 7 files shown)


packages/aws-cdk/lib/cli/version.ts
```diff
@@ -1,26 +1,12 @@
...
-import { execNpmView } from './util/npm';
 
-const ONE_DAY_IN_SECONDS = 1 * 24 * 60 * 60;
-
-const UPGRADE_DOCUMENTATION_LINKS: Record<number, string> = {
-  1: 'https://docs.aws.amazon.com/cdk/v2/guide/migrating-v2.html',
-};
-
-export function displayVersion() {
+export function version() {
```
mrgrain: Maybe this?
```suggestion
export function versionWithBuild() {
```

mrgrain (approved):

PR #710, opened 2025-07-11
chore(cli): file sink telemetry client writes valid json

kaizencc:
Pulled out of #631. Now writes valid json to the telemetry file

---

### DIFF
--- a/packages/aws-cdk/lib/cli/telemetry/file-sink.ts
+++ b/packages/aws-cdk/lib/cli/telemetry/file-sink.ts
@@ -1,6 +1,5 @@
-import * as fs from 'fs';
-import * as path from 'path';
 import { ToolkitError, type IIoHost } from '@aws-cdk/toolkit-lib';
+import * as fs from 'fs-extra';
 import type { TelemetrySchema } from './schema';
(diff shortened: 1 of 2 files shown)


packages/aws-cdk/lib/cli/telemetry/file-sink.ts
```diff
@@ -38,23 +37,19 @@ export class FileTelemetrySink implements ITelemetrySink {
...
   public async emit(event: TelemetrySchema): Promise<void> {
     try {
-      // Format the events as a JSON string with pretty printing
-      const output = JSON.stringify(event, null, 2) + '\n';
-
-      // Write to file
-      fs.appendFileSync(this.logFilePath, output);
+      const json = fs.readJsonSync(this.logFilePath);
+      json.push(event);
+      fs.writeJSONSync(this.logFilePath, json, { spaces: 2 });
```
kaizencc, line 52: I originally used `appendFileSync()` because it's ~50% faster than read + write. However we are not emitting that many telemetry events at the moment and I think this is fine unless we add many many more events. The payoff is that we can write valid json and the telemetry file can be `.json`.

otaviomacedo (approved):

PR #711, opened 2025-07-11
chore(cli): allow valid enum values in telemetry data

kaizencc:
Enum values are defined as `choices` in the `cli-type-registry.json` file. If the choices array exists, and if the value we get is included as a choice, then we allow that raw value to be recorded in telemetry because it is not free text. Invalid values passed in an enum property are not recorded. In practice, the CLI fails before telemetry starts with 

```
Invalid values:
  Argument: require-approval, Given: "neveasdfr", Choices: "never", "any-change", "broadening"
```

---


packages/aws-cdk/test/cli/telemetry/sanitation.test.ts
```diff
@@ -76,12 +76,35 @@ describe(sanitizeCommandLineArguments, () => {
     const argv = {
       _: ['deploy'],
       STACKS: ['MyStack'],
-      ['require-approval']: 'broadening',
       ['build-exclude']: ['something'],
     };
     expect(sanitizeCommandLineArguments(argv)).toEqual({
       path: ['deploy', '$STACK1'],
-      parameters: { 'require-approval': '<redacted>', 'build-exclude': '<redacted>' },
+      parameters: { 'build-exclude': '<redacted>' },
```
iliapolo, line 110: How come?

kaizencc: i removed `require-approval` from the test 'non-boolean options are redacted' and added it as a separate test called 'enum options are allowed'

iliapolo (approved):

packages/aws-cdk/lib/cli/telemetry/sanitation.ts
```diff
@@ -34,7 +34,7 @@ export function sanitizeCommandLineArguments(argv: any): { path: string[]; param
     if (argv[argName] === undefined || (!globalOptions.includes(argName) && !commandOptions.includes(argName))) {
       continue;
     }
-    if (isNumberOrBoolean(argv[argName])) {
+    if (isNumberOrBoolean(argv[argName])) || isKnownEnumValue(argName, argv[argName], command, config)) {
```
kaizencc:
```suggestion
    if (isNumberOrBoolean(argv[argName]) || isKnownEnumValue(argName, argv[argName], command, config)) {
```

PR #713, opened 2025-07-14
feat: take all neighbors into account when computing resource digests

otaviomacedo:
The current definition of the digest of a resource computes a hash of its properties plus the hashes of its dependencies. 

But if we have a graph like:

        A --> B
        C --> D

where `B` and `D` are identical, the current algorithm will compute the same digest for them, even if `A` is different from `C` (and therefore have different digests). This is undesirable, as some cases are considered ambiguous, even if we could very well tell them apart. Real world example:

non-ambiguous

`Role1` and `Role2` tend to be identical, but they can nevertheless be told apart because at least the functions will have different properties. So, even if both roles are renamed, for example,  we can still provide a valid mapping.

This change improves the algorithm by also taking into account all neighbors of a resource for digest computation. It does this by doing a two pass over the graph: first navigating in topological order and computing the digests, then navigating in the opposite order and doing the same. So each resource has two digests, that are in turn, used to compute the final digest.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/context.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/context.ts
@@ -1,6 +1,7 @@
 import type { Environment } from '@aws-cdk/cx-api';
 import type { CloudFormationStack } from './cloudformation';
 import { ResourceLocation, ResourceMapping } from './cloudformation';
+import type { GraphDirection } from './digest';
 import { computeResourceDigests } from './digest';
 import { ToolkitError } from '../../toolkit/toolkit-error';
 import { equalSets } from '../../util/sets';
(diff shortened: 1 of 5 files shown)


rix0rrr:
So we now have two digests. Let's give them names (not the best ones, maybe you can come up with better ones).

* `downdigest` - the digest of a resource's properties and all of the resources it references.
* `updigest` - the digest of a resource's properties and all of the resources referencing it.

If we always use both digests to come up with the mapping digest (`mappingdigest = digest(down + up)`) we are sensitive to incoming references changing. Does that ever pose a problem?

Or would we be better off having a 2-step digest, where we map using the `downdigest` first, and then only involve the `updigest` if there is a collission?

otaviomacedo:
> Typo? A is different from C ?

Yes, thanks. Updated.

> If we always use both digests to come up with the mapping digest (`mappingdigest = digest(down + up)`) we are sensitive to incoming references changing. Does that ever pose a problem?

Right now, this is not a problem because we require the graphs to be isomorphic anyway. But I went ahead and implemented your suggestion, so it's ready when we need it.

rix0rrr (approved):

PR #714, opened 2025-07-14
feat: refactor options can be provided via the config file

otaviomacedo:
Allow the user to pass options via the config file:

```json
{
  "app": "node bin/main.js",
  "refactor": {
    "dryRun": true
  }
}
```

---

### DIFF
--- a/packages/aws-cdk/README.md
+++ b/packages/aws-cdk/README.md
@@ -1189,6 +1189,33 @@ locations for a given environment. Resource locations are in the format
 resource currently deployed, while the destination must refer to a location
 that is not already occupied by any resource.
 
+All command line options that are also available in the configuration file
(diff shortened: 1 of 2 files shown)


rix0rrr (commented):
I have two questions here:

* Can you give me an example of the use cases for this? I'm struggling to think why a user would choose to always run `cdk refactor` in dryrun mode.
* This seems like a one-off solution for refactoring. Should we not build it in such a way that it works for all commands?

otaviomacedo:
> I have two questions here:
> 
>     * Can you give me an example of the use cases for this? I'm struggling to think why a user would choose to always run `cdk refactor` in dryrun mode.
> 
>     * This seems like a one-off solution for refactoring. Should we not build it in such a way that it works for all commands?

Yeah, you're right. I'll just close it and revisit this feature if someone asks for it.

PR #718, opened 2025-07-15
feat(cli): `--no-version-reporting` equals `--no-telemetry` for cli telemetry and library metadata exclusion

kaizencc:
`--no-version-reporting` and `--no-telemetry` will disable cli telemetry and library metadata for that specific command execution. to disable cli telemetry permanently see options [here](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0732-cdk-cli-telemetry.md#opt-out)

---


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -180,7 +180,7 @@ export class CliIoHost implements IIoHost {
       });
     }
     // TODO: uncomment this at launch
-    // if (canCollectTelemetry(context)) {
+    // if (canCollectTelemetry(args, context)) {
```
kaizencc, line 184: `canCollectTelemetry` is not in use in the pre-launch phase where we are going to launch telemetry to local file (since the opt-in is the `--telemetry-file` itself, irregardless of our opt-out options). it is in use in `cdk cli-telemetry --status`, however.

packages/aws-cdk/test/cli/telemetry/collect-telemetry.test.ts
```diff
@@ -8,27 +8,31 @@ describe(canCollectTelemetry, () => {
...
     context.set('cli-telemetry', true);
-    expect(canCollectTelemetry(context)).toBeTruthy();
+    expect(canCollectTelemetry({}, context)).toBeTruthy();
 
     context.set('cli-telemetry', 'true');
-    expect(canCollectTelemetry(context)).toBeTruthy();
+    expect(canCollectTelemetry({}, context)).toBeTruthy();
+  });
+
+  test('returns false if no-version-reporting is set', async () => {
```
kaizencc, line 48: `--no-version-reporting` and `--no-telemetry` are aliased -- a yargs feature. yargs will turn `--no-telemetry` into `{ "version-reporting": false }` (i.e. it won't be its own option in the yargs object) which is why there isn't another test for `returns false if no-telemetry is set.

iliapolo (approved):

packages/aws-cdk/lib/cxapp/cloud-executable.ts
```diff
@@ -140,6 +140,8 @@ export class CloudExecutable implements ICloudAssemblySource {
     } finally {
       await synthSpan.end({ error });
     }
+
+    throw new ToolkitError('Unreachable Code');
```
kaizencc:
```suggestion
```

PR #727, opened 2025-07-16
fix(integ-runner): command to reproduce snapshot checkout does not always work

mrgrain:
During the update workflow, when checking out the previous snapshot failed, we print a git command to allow the user to reproduce potential issues. This command doesn't work as printed, as it assumes it is run from the same directory the test case file is in. However in most situations a users will have run `integ-runner` from a parent directory.

It would print something like this, assuming you are in the same directory as the test file:
```console
git checkout $(git merge-base HEAD main) -- integ.bucket-deployment-deployed-bucket.js.snapshot
```

The change in this PR updates `git` operations to use the `-C` flag to specify the correct directory instead of changing the working directory for the shell execution. Also changes the base the directory the snapshot is in, instead of `this.directory`. This results in the same dir, but is more localized to the feature and does not rely on the fact that `this.directory` and `this.snapshotDir` happen to share the same parent.

So now, we print a working command, where ever you are:
```console
git -C path/to/tests checkout $(git -C path/to/tests merge-base HEAD main) -- integ.bucket-deployment-deployed-bucket.js.snapshot
```

I could have just changed the printed command, but decided that consistency between what we run and what we show is important. This also simplifies the code because we don't have to manually maintain the logger output anymore.

---

### DIFF
--- a/packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
+++ b/packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
@@ -9,9 +9,9 @@ import * as workerpool from 'workerpool';
 import type { IntegRunnerOptions } from './runner-base';
 import { IntegRunner } from './runner-base';
 import * as logger from '../logger';
-import { chunks, exec, promiseWithResolvers } from '../utils';
+import { chunks, exec, execWithSubShell, promiseWithResolvers, renderCommand } from '../utils';
 import type { DestructiveChange, AssertionResults, AssertionResult } from '../workers/common';
-import { DiagnosticReason, formatAssertionResults } from '../workers/common';
+import { DiagnosticReason, formatAssertionResults, formatError } from '../workers/common';
 
 export interface CommonOptions {
   /**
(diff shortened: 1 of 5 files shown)


packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
```diff
@@ -135,28 +139,24 @@ export class IntegTestRunner extends IntegRunner {
...
 
+      const checkoutCommand = [...git, 'checkout', [...git, 'merge-base', 'HEAD', baseBranch], '--', relativeSnapshotDir];
       try {
-        const base = exec(['git', 'merge-base', 'HEAD', baseBranch], {
-          cwd,
-        });
-        exec(['git', 'checkout', base, '--', relativeSnapshotDir], {
-          cwd,
-        });
+        execWithSubShell(checkoutCommand);
```
iliapolo, line 152: Wouldn't doing this:

```ts
exec(renderCommand(checkoutCommand), { shell: true })
```

Achieve the same result, without needing the `execWithSubShell` function? Might fail on windows though, just wondering if you considered/tested that.

rix0rrr (commented):
We definitely don't need to include `-C` on the `git merge-base`.

Wouldn't it be easier to do `path.relpath(testFile)` to come up with the path of the file?

packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
mrgrain: I considered but didn't want to change the way this is executed, precisely because of Windows. This really is just a more condensed way of writing the same code we had before in away that can also be rendered as a string.

mrgrain: Also `shell: true` tends to be flagged up from a security perspective. It's not really an issue for client side tooling, but I'd like to avoid if we can.

iliapolo (approved):
Dequeued to give a chance to resolve comments from @rix0rrr

rix0rrr (approved):

PR #731, opened 2025-07-23
fix(cli): running `--telemetry-file` without the `--unstable` option creates a telemetry file

iliapolo:
This feature is currently unstable, so it shouldn't apply at all if invoked without the `--unstable` flag. 

Right now, it creates a telemetry file with an event indicating the aforementioned failure.

---

### DIFF
--- a/packages/aws-cdk/lib/cli/cli.ts
+++ b/packages/aws-cdk/lib/cli/cli.ts
@@ -99,6 +99,10 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
     caBundlePath: configuration.settings.get(['caBundlePath']),
   });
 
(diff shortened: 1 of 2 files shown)


packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -209,10 +213,6 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
       throw new ToolkitError('You must either specify a list of Stacks or the `--all` argument');
     }
 
-    if (args['telemetry-file'] && !configuration.settings.get(['unstable']).includes('telemetry')) {
```
iliapolo, line 212: This is too late in the process because telemetry has already been initialized, and the error thrown is then considered as a telemetry event.

We should really structure the unstable protection mechanism better.

mrgrain (approved):

PR #736, opened 2025-07-23
fix(toolkit-lib): throw AuthenticationError instead of ExpiredToken

kaizencc:
We own the `AuthenticationError` while `ExpiredToken` comes directly from the SDKs.

---


mrgrain (changes requested):
Can we use with cause instead please? It's maintains higher fidelity and produces a nicer output to the user.

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk-provider.ts
```diff
@@ -182,7 +182,7 @@ export class SdkProvider {
       return { sdk, didAssumeRole: true };
     } catch (err: any) {
       if (err.name === 'ExpiredToken') {
-        throw err;
+        throw AuthenticationError.withCause('Assuming role failed with ExpiredToken', err);
```
kaizencc:
```suggestion
        throw AuthenticationError.withCause('Assuming role failed: ExpiredToken', err);
```

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk-provider.ts
```diff
@@ -182,7 +182,7 @@ export class SdkProvider {
       return { sdk, didAssumeRole: true };
     } catch (err: any) {
       if (err.name === 'ExpiredToken') {
-        throw err;
+        throw AuthenticationError.withCause('Assuming role failed: ExpiredToken', err);
```
kaizencc, line 185: we have some tests asserting that ExpiredToken is in the name of the error

mrgrain (approved):

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk-provider.ts
mrgrain: Strange, but okay.

PR #738, opened 2025-07-24
feat: modify a single feature flag configuration

vivian12300:
This change allows a user to modify a single feature flag value, view the difference in CloudFormation templates, and accept or reject the change. 


packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -0,0 +1,242 @@
...
+  if (options.FLAGNAME && !options.set) {
+    await displayFlags(flagData, ioHelper, String(options.FLAGNAME));
+    return;
+  }
+
+  if (options.all && !options.set) {
+    await displayFlags(flagData, ioHelper, undefined, true);
+    return;
+  }
+  if (options.set && options.FLAGNAME) {
```
otaviomacedo: For completeness, we need to check for all edge cases, and do something about them, even if it's to give an error message explaining that the combination of options is not valid:

- `options.set && !options.FLAGNAME`: error (what are you trying to set?)
- `options.FLAGNAME && options.all`: error (mutually exclusive)
- `options.FLAGNAME && !options.all && !options.set`: maybe an error, or maybe we want to show only that specific flag?
- `options.set && options.all`: are you planning to handle this case somehow?

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -115,6 +115,18 @@ export async function makeConfig(): Promise<CliConfig> {
...
+          name: 'FLAGNAME',
+          variadic: true,
+        },
+        options: {
+          value: { type: 'string', desc: 'The value the user would like to set the feature flag configuration to', requiresArg: true },
+          set: { type: 'boolean', desc: 'Signifies the user would like to modify their feature flag configuration', requiresArg: false },
+          all: { type: 'boolean', desc: 'Modify or view all feature flags', requiresArg: false },
+          unconfigured: { type: 'boolean', desc: 'Modify unconfigured feature flags', requiresArg: false },
+          recommended: { type: 'boolean', desc: 'Change flags to recommended states', requiresArg: false },
+          default: { type: 'boolean', desc: 'Change flags to default state', requiresArg: false },
```
otaviomacedo: You're not doing anything with these options right now, but users will see them in the help and think they do something.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -0,0 +1,242 @@
...
+  const cdkJson = JSON.parse(cdkJsonContent);
+
+  const boolValue = value!.toLowerCase() === 'true';
+  cdkJson.context[flagName] = boolValue;
+
+  await ioHelper.defaults.info(`Setting flag '${flagName}' to: ${boolValue}`);
+  await fs.writeFile(cdkJsonPath, JSON.stringify(cdkJson, null, 2), 'utf-8');
+}
+
+function formatTable(headers: string[], rows: string[][]): string {
```
otaviomacedo, line 233: We already have a `formatTable` function in `@aws-cdk/cloudformation-diff`. See if you can reuse it.

otaviomacedo (changes requested):

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -0,0 +1,263 @@
...
+import type { FlagsOptions } from '../cli/user-input';
+
+export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, options: FlagsOptions, toolkit: Toolkit) {
+  if (options.FLAGNAME && options.all) {
+    await ioHelper.defaults.info('Error: Cannot use both --all and a specific flag name. Please use either --all to show all flags or specify a single flag name.');
+    return;
+  }
+
+  if (options.set && options.all) {
+    await ioHelper.defaults.info('Error: --set is currently only compatible with a flag name. Please specify which flag you want to set.');
```
otaviomacedo:
```suggestion
    await ioHelper.defaults.error('Error: --set is currently only compatible with a flag name. Please specify which flag you want to set.');
```

otaviomacedo (changes requested):

.projenrc.ts
```diff
@@ -854,7 +851,7 @@ const toolkitLib = configureProject(
       '@smithy/util-stream',
       '@types/fs-extra',
       '@types/split2',
-      `aws-cdk-lib@${AWS_CDK_LIB_VERSION}`,
```
otaviomacedo, line 857: We should now use the latest version.

otaviomacedo (approved):
Conditionally approved on my two comments.

.projenrc.ts
vivian12300: I think changing it back to 'aws-cdk-lib' will use the latest version.

PR #747, opened 2025-07-28
feat(cli): count resources/stacks with `UNKNOWN` drift status as `unchecked`

### LINKED ISSUE #740: drift: handle new CloudFormation drift detection "UNKNOWN" status
### Describe the feature

CloudFormation is introducing breaking changes to drift detection APIs that will affect CDK CLI's drift detection functionality. Action required by August 14, 2025.

## Background
CloudFormation team announced changes to drift detection behavior where:
- Resources that fail drift detection (due to permissions/throttling) will return "UNKNOWN" status instead of failing the operation
- Stack drift status will be "UNKNOWN" under certain conditions
- New `DriftStatusReason` field will provide error details

### CFN API Changes:
1. DetectStackResourceDrift API [4] will return resource drift information when CloudFormation is unable to perform drift detection on the resource. The returned information will have a drift status of “UNKNOWN” and a reason under “DriftStatusReason”. Previously, CloudFormation would have failed the DetectStackResourceDrift operation.
2. DescribeStackDriftDetectionStatus API [5] will return a stack drift status of “UNKNOWN” if CloudFormation is unable to perform drift detection on least one resource in the stack AND all other resources are "IN_SYNC". However, if any successfully checked resource is "MODIFIED", the stack drift status will be "DRIFTED", regardless of any incomplete checks.
3. DescribeStackResourceDrifts [6] will include resources with an "UNKNOWN" status, unless otherwise specified. For resources with an "UNKNOWN" status, there will be a field "DriftStatusReason" to explain the reason for the status.

## Impact on CDK CLI
CDK CLI depends on CloudFormation drift detection APIs and needs to be updated to handle the new "UNKNOWN" status and associated fields.

### Use Case

n/a

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

n/a

n/a

iankhou:
Fixes #740

The `cdk drift` action lists `UNKNOWN` for resources/stacks when CloudFormation returns that value (effective August 14).

This follows a new change in the CloudFormation API. Previously, `UNKNOWN` was not available, and stacks/resources that fit such a description would have been reported as unchanged.

Brought line coverage for `@aws-cdk/toolkit-lib/lib/api/drift` to 100%.
Screenshot 2025-09-01 at 02 59 27

- [X] `UNKNOWN` support
- [X] Unit Test

---


mrgrain (changes requested):
I think from a DX perspective `unchecked` and `unknown` are pretty much the same. I propose you just piggy back on `unchecked`.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -98,12 +98,9 @@ export class DriftFormatter {
   public formatStackDrift(): DriftFormatterOutput {
     const formatterOutput = this.formatStackDriftChanges(this.buildLogicalToPathMap());
 
-    // we are only interested in actual drifts and always ignore the metadata resource
+    // we are only interested in actual drifts (and ignore the metadata resource)
     const actualDrifts = this.resourceDriftResults.filter(d =>
-      d.StackResourceDriftStatus === 'MODIFIED' ||
-      d.StackResourceDriftStatus === 'DELETED' ||
-      d.ResourceType === 'AWS::CDK::Metadata',
```
iankhou, line 105: This line was dead code, removed as it was wrong. We do NOT want to include metadata changes, but this line, if it worked, WOULD include them.

But metadata results are filtered out before we get to this point, anyway, so we can just remove the code.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -172,19 +173,18 @@ export class DriftFormatter {
       unchanged += this.printSectionFooter();
     }
 
-    // Process all unchecked resources
-    if (this.allStackResources) {
```
iankhou, line 176: This is a JS Map, so always truthy. Removed.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -114,6 +111,7 @@ export class DriftFormatter {
         numResourcesWithDrift: 0,
         numResourcesUnchecked: this.allStackResources.size - this.resourceDriftResults.length,
         stackHeader,
+        unchecked: formatterOutput.unchecked,
```
iankhou, line 115: Include unchecked resources even when the count of drifted resources is 0, for the user's awareness.

packages/@aws-cdk/toolkit-lib/test/api/drift/drift.test.ts
```diff
@@ -580,4 +661,149 @@ describe('formatStackDrift', () => {
...
+    expect(result.numResourcesWithDrift).toBe(0); // UNKNOWN resources do not count as drift
+    expect(result.summary).toContain('No drift detected');
+
+    // All UNKNOWN resources should be in the unchecked section
+    expect(result.unchecked).toBeDefined();
+    expect(result.unchecked).toContain('AWS::S3::Bucket');
+    expect(result.unchecked).toContain('Resource1');
+    expect(result.unchecked).toContain('AWS::IAM::Role');
+    expect(result.unchecked).toContain('Resource2');
+  });
```
iankhou, line 871: Here, we have 0 drifted resources, but have an UNKNOWN stack.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
mrgrain: Nice!

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
```diff
@@ -39,10 +39,34 @@ export async function detectStackDrift(
     );
   }
 
-  // Get the drift results
-  return cfn.describeStackResourceDrifts({
+  // Handle UNKNOWN stack drift status
+  if (driftStatus?.StackDriftStatus === 'UNKNOWN') {
+    await ioHelper.defaults.trace(
+      'Stack drift status is UNKNOWN. This may occur when CloudFormation is unable to detect drift for at least one resource and all other resources are IN_SYNC.\n' +
+      `Reason: ${driftStatus.DetectionStatusReason || 'No reason provided'}`,
```
mrgrain: todo : I see this a bunch of time. please abstract to a single code instance.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
mrgrain: question (blocking): Just to triple check:
- Can you point me to where metadata is filtered out?
- Do we have test for metadata being excluded?

mrgrain (changes requested):
Almost there. I'm applauding your spotting of dead code and reducing complexity. Great job!

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
iankhou: My bad. It's filtered out on [line 88](https://github.com/aws/aws-cdk-cli/pull/747/files#diff-c0c2a6fc1d3f29a4c163a60e61287b220b52c79f07ad4f7e54b954710bdb2509R88), but this is a different variable. Good catch. I've added tests for this as well.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -140,11 +139,8 @@ export class DriftFormatter {
   }
 
   /**
-   * Renders stack drift information to the given stream
+   * Renders stack drift information
    *
-   * @param driftResults - The stack resource drifts from CloudFormation
-   * @param allStackResources - A map of all stack resources
-   * @param verbose - Whether to output more verbose text (include undrifted resources)
```
iankhou, line 147: Removed params that are not present in the function signature.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -172,29 +168,29 @@ export class DriftFormatter {
...
       }
+      unchecked += this.printSectionFooter();
     }
 
-    // Process modified resources
-    const modifiedResources = drifts.filter(d => d.StackResourceDriftStatus === StackResourceDriftStatus.MODIFIED);
+    // Process modified resources (exclude AWS::CDK::Metadata)
+    const modifiedResources = drifts.filter(d =>
+      d.StackResourceDriftStatus === StackResourceDriftStatus.MODIFIED
+      && d.ResourceType !== 'AWS::CDK::Metadata');
```
iankhou, line 188: Ensure the count is correct, by excluding metadata resources here.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -172,29 +168,29 @@ export class DriftFormatter {
...
     if (modifiedResources.length > 0) {
       modified = this.printSectionHeader('Modified Resources');
 
       for (const drift of modifiedResources) {
         if (!drift.LogicalResourceId || !drift.ResourceType) continue;
-        if (modified === undefined) modified = '';
```
iankhou, line 197: Dead code. `modified` is defined a few lines above, `printSectionHeader` always returns a string.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -209,8 +205,10 @@ export class DriftFormatter {
       modified += this.printSectionFooter();
     }
 
-    // Process deleted resources
-    const deletedResources = drifts.filter(d => d.StackResourceDriftStatus === StackResourceDriftStatus.DELETED);
+    // Process deleted resources (exclude AWS::CDK::Metadata)
+    const deletedResources = drifts.filter(d =>
+      d.StackResourceDriftStatus === StackResourceDriftStatus.DELETED
+      && d.ResourceType !== 'AWS::CDK::Metadata');
```
iankhou, line 211: Ensure the count is correct, by excluding metadata resources here.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift.ts
iankhou: Done. Introduced `formatReason`.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -250,16 +248,6 @@ export class DriftFormatter {
     return `${normalizedPath} ${chalk.gray(logicalId)}`;
   }
 
-  private formatValue(value: any, colorFn: (str: string) => string): string {
```
iankhou, line 253: We don't actually need this. It was also never called on anything that isn't a string.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
```diff
@@ -268,16 +256,16 @@ export class DriftFormatter {
...
+  private formatTreeDiff(propertyPath: string, difference: Difference<string>, isLast: boolean): string {
     let result = format(' %s─ %s %s\n', isLast ? '└' : '├',
       difference.isAddition ? ADDITION :
         difference.isRemoval ? REMOVAL :
           UPDATE,
       propertyPath,
     );
     if (difference.isUpdate) {
-      result += format('     ├─ %s %s\n', REMOVAL, this.formatValue(difference.oldValue, chalk.red));
-      result += format('     └─ %s %s\n', ADDITION, this.formatValue(difference.newValue, chalk.green));
```
iankhou, line 280: `difference.isUpdate` checks that oldValue and newValue are both not `undefined`. If they're defined, they must be strings.

packages/@aws-cdk/toolkit-lib/lib/api/drift/drift-formatter.ts
mrgrain: You still caught the incorrect filter though! Good job!

mrgrain (approved):

PR #750, opened 2025-07-29
feat: modify multiple feature flag values to their recommended states

vivian12300:
Allows users to change multiple feature flags to their recommended states and displays differences in CloudFormation templates. 

---


packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -38,110 +71,70 @@ export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, o
...
+  }
+}
+
+async function prototypeChanges(
+  flagData: FeatureFlag[],
+  ioHelper: IoHelper,
+  flagNames: string[] | string,
+  toolkit: Toolkit,
+  recommended: boolean,
+  value?: boolean,
```
otaviomacedo: Consider creating an interface with these properties to improve readability, especially of the boolean values.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -150,12 +143,30 @@ async function prototypeChanges(
...
+    if (baseContextValues[flagNames] == value) {
+      await ioHelper.defaults.error('Flag is already set to the specified value. No changes needed.');
+      return false;
+    }
+    updateObj[flagNames] = value!;
+  } else {
+    if (recommended) {
+      for (const flagName of flagNames) {
+        const flag = flagData.find(f => f.name === flagName);
+        const boolValue = toBooleanValue(flag!.recommendedValue);
```
otaviomacedo: Are you sure `flag` is defined here? The flag name is free form set by the user, right?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -150,12 +143,30 @@ async function prototypeChanges(
     outdir: path.join(process.cwd(), 'original'),
   });
 
-  const cx = await toolkit.synth(source);
-  const assembly = cx.cloudAssembly;
-
   const updateObj: Record<string, boolean> = {};
-  updateObj[flagName] = boolValue;
+  if (typeof (flagNames) == 'string') {
```
otaviomacedo: [nitpicking] I guess this works, but is a bit unusual. Why not make `flagNames` of type `string[]` and check if the length is 1?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -175,58 +186,83 @@ async function prototypeChanges(
...
   }
+  let prototypeSuccess = false;
 
-  const originalDir = path.join(process.cwd(), 'original');
-  const tempDir = path.join(process.cwd(), 'temp');
+  if (recommended) {
+    prototypeSuccess = await prototypeChanges(flagData, ioHelper, flagsToSet, toolkit, true);
+  } else {
+    prototypeSuccess = await prototypeChanges(flagData, ioHelper, flagsToSet, toolkit, false);
+  }
```
otaviomacedo:
```suggestion
prototypeSuccess = await prototypeChanges(flagData, ioHelper, flagsToSet, toolkit, recommended);
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -261,3 +297,86 @@ function formatTable(headers: string[], rows: string[][]): string {
...
+    return;
+  }
+
+  const headers = ['Feature Flag Name', 'Recommended Value', 'User Value'];
+  const rows: string[][] = [];
+
+  const getFlagPriority = (flag: FeatureFlag): number => {
+    if (flag.userValue === undefined) {
+      return 3;
+    } else if (String(flag.userValue) === String(flag.recommendedValue)) {
```
otaviomacedo: This condition appears multiple times throughout the code. Consider extracting it to a function.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -150,12 +143,30 @@ async function prototypeChanges(
...
   });
 
-  const cx = await toolkit.synth(source);
-  const assembly = cx.cloudAssembly;
-
   const updateObj: Record<string, boolean> = {};
-  updateObj[flagName] = boolValue;
+  if (typeof (flagNames) == 'string') {
+    if (baseContextValues[flagNames] == value) {
+      await ioHelper.defaults.error('Flag is already set to the specified value. No changes needed.');
```
otaviomacedo: I don't think we need to give this feedback to the user. The expectation is that the operation is idempotent: applying it twice is not an error. If we don't need to do anything, just return quietly.

otaviomacedo (changes requested):

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -150,12 +162,38 @@ async function prototypeChanges(
...
+    } else {
+      for (const flagName of flagNames) {
+        const flag = flagData.find(f => f.name === flagName);
+        if (!flag) {
+          await ioHelper.defaults.error(`Flag ${flagName} not found.`);
+          return false;
+        }
+        updateObj[flagName] = String(flag.unconfiguredBehavesLike?.v2) === 'true';
+      }
+    }
```
otaviomacedo: Just to reduce duplication:

```suggestion
    for (const flagName of flagNames) {
      const flag = flagData.find(f => f.name === flagName);
      if (!flag) {
        await ioHelper.defaults.error(`Flag ${flagName} not found.`);
        return false;
      }
      const newValue = recommended 
        ? toBooleanValue(flag.recommendedValue)
        : String(flag.unconfiguredBehavesLike?.v2) === 'true'
      updateObj[flagName] = newValue;
    }
```

otaviomacedo (approved):

PR #753, opened 2025-08-01
feat: add retry mechanism for flaky cli integ tests

abidhasan-aws:
We have observed frequent pipeline build failures due to flaky integration tests that usually succeed on retries. Currently, these tests have to be retried manually for unblocking the pipelines. By enabling automated retries for these flaky tests, we can reduce intermittent pipeline failures and eliminate unnecessary manual intervention which then will only be required for test failures that persist even after retries.

The change adds a retry wrapper for specific CLI integration tests that have been identified as frequently failing but succeeding on retries. The goal is to identify the the frequently failing flaky test that require manual intervention and add retry mechanism before we investigate the tests individually.

Fixes : https://github.com/cdklabs/cdk-ops/issues/4176

### Top failing tests from last week
image

  ---


packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
```diff
@@ -163,6 +163,37 @@ export function withCDKMigrateFixture(language: string, block: (content: TestFix
...
 }
 
+/**
+ * Retry wrapper that executes a test callback up to maxAttempts times
+ *
+ * If any attempt succeeds, it returns immediately. If all attempts fail,
+ * it throws the last error encountered.
+ */
+
+export function withRetry<T>(
```
mrgrain:
```suggestion
export function withRetry<T extends TestContext>(
```

packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
```diff
@@ -163,6 +163,37 @@ export function withCDKMigrateFixture(language: string, block: (content: TestFix
...
+
+    for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+      try {
+        await callback(context);
+        return;
+      } catch (error) {
+        lastError = error;
+        if (attempt < maxAttempts) {
+          const logger = (context as any).log || console.log;
+          logger(`Attempt ${attempt}/${maxAttempts} failed: ${(error as Error).message}. Retrying...`);
```
mrgrain:
```suggestion
          context.log(`Attempt ${attempt}/${maxAttempts} failed: ${(error as Error).message}. Retrying...`);
```

packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
```diff
@@ -163,6 +163,36 @@ export function withCDKMigrateFixture(language: string, block: (content: TestFix
...
+    let lastError;
+
+    for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+      try {
+        await callback(context);
+        return;
+      } catch (error) {
+        lastError = error;
+        if (attempt < maxAttempts) {
+          context.log(`Attempt ${attempt}/${maxAttempts} failed: ${(error as Error).message}. Retrying...`);
```
vishaalmehrishi: > ${(error as Error).message

Is this going to log the full error detail? Might want to log the full error instead of the message

packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
```diff
@@ -163,6 +163,36 @@ export function withCDKMigrateFixture(language: string, block: (content: TestFix
...
+/**
+ * Retry wrapper that executes a test callback up to maxAttempts times
+ *
+ * If any attempt succeeds, it returns immediately. If all attempts fail,
+ * it throws the last error encountered.
+ */
+
+export function withRetry<T extends TestContext>(
+  callback: (context: T) => Promise<void>,
+  maxAttempts: number = 3,
```
vishaalmehrishi: Not significant, but I'm wondering if 2 is a better default.

If a test fails once, it could be flaky; if it fails twice, the likelihood of it being actually flaky reduces significantly IMO (unless we have a known problem, in which case it should fail more than twice anyway). The third attempt may not bring much value here.

abidhasan-aws: makes sense, addressed it.

vishaalmehrishi (approved):

mrgrain (approved):

PR #760, opened 2025-08-01
chore(cli): funnel

kaizencc:
A funnel is a combination sink class. It will be used later to record telemetry both to file _and_ send to endpoint. That version will be used in the situation where telemetry is turned on _and_ `--telemetry-file` is set.

---


packages/aws-cdk/lib/cli/telemetry/sink/file-endpoint-sink.ts
```diff
@@ -0,0 +1,28 @@
+import type { EndpointTelemetrySinkProps } from './endpoint-sink';
+import { EndpointTelemetrySink } from './endpoint-sink';
+import type { FileTelemetrySinkProps } from './file-sink';
+import { FileTelemetrySink } from './file-sink';
+import type { TelemetrySchema } from '../schema';
+
+export interface FileEndpointTelemetrySinkProps extends FileTelemetrySinkProps, EndpointTelemetrySinkProps {
+}
+
+export class FileEndpointTelemetrySink {
```
mrgrain: Just make this a MultiSinkSink (Waterworks? Pumphouse? TelemteryDistributionNetwork?) that takes in any number of pre-configured sinks and emits and flushes to all of them.

packages/aws-cdk/lib/cli/telemetry/sink/file-endpoint-sink.ts
```diff
@@ -0,0 +1,28 @@
...
+  private endpointSink: EndpointTelemetrySink;
+
+  constructor(props: FileEndpointTelemetrySinkProps) {
+    this.fileSink = new FileTelemetrySink(props);
+    this.endpointSink = new EndpointTelemetrySink(props);
+  }
+
+  public async emit(event: TelemetrySchema): Promise<void> {
+    await this.fileSink.emit(event);
+    await this.endpointSink.emit(event);
```
mrgrain: Why not handle them in parallel?

packages/aws-cdk/lib/cli/telemetry/sink/file-endpoint-sink.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  public async flush(): Promise<void> {
+    await this.fileSink.flush();
+    await this.endpointSink.flush();
```
mrgrain: Why not handle them in parallel?

mrgrain (changes requested):

packages/aws-cdk/lib/cli/telemetry/sink/funnel.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class Funnel {
+  private readonly sinks: ITelemetrySink[];
+
+  constructor(props: FunnelProps) {
+    this.sinks = props.sinks;
+  }
+
+  public async emit(event: TelemetrySchema): Promise<void> {
+    // There is a limited set of sinks
```
mrgrain: suggestion (non-blocking): You could enforce a hard limit of i.e. 5 in the constructor.

mrgrain (approved):
Looks good. Left a minor suggestion but please update the PR title and description to match the updated code. Also, I think this is a refactor.

PR #763, opened 2025-08-04
feat: add local template support with --from-path option

rohang9000:
This PR adds local template support through the new `--from-path` CLI option. Users can initialize CDK projects from local custom templates, copying the template as is without placeholder processing. 

The feature includes language auto-detection when templates contain a single language option and error handling for invalid paths. 3 new test cases covering basic local template functionality, language auto-detection, and error handling scenarios were added.

Fixes #

---


packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -29,6 +29,11 @@ export interface CliInitOptions {
    */
   readonly libVersion?: string;
 
+  /**
+   * Path to a local custom template directory
+   */
+  readonly fromPath?: string;
+
```
iankhou: Optional parameters should have a `@default` indication. Format is as follows:

If the default is a VALUE, like the number 5, indicate the value directly after the @default keyword:
```
@default 5
```

If the default references something, use a hyphen and capitalize:
```
@default - Detects color from the TTY status of the IoHost
```

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -145,23 +261,30 @@ export class InitTemplate {
...
-      ioHelper,
-      { targetDirectory, language, templateName: this.name },
-      {
-        substitutePlaceholdersIn: async (...fileNames: string[]) => {
-          for (const fileName of fileNames) {
-            const fullPath = path.join(targetDirectory, fileName);
-            const template = await fs.readFile(fullPath, { encoding: 'utf-8' });
-            await fs.writeFile(fullPath, expandPlaceholders(template, language, projectInfo));
-          }
+    const isCustomTemplate = this.description === 'Custom template from local path';
```
iankhou: I don't like the way this is checked, would be better to have a boolean or an enum to indicate what kind of template it is.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: See some examples here: https://github.com/aws/aws-cdk-cli/blob/%40aws-cdk/toolkit-lib%40v1.4.0/packages/%40aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L120-L131

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -184,11 +307,38 @@ export class InitTemplate {
...
+        continue;
+      } else if (file.match(/^.*\.hook\.(d.)?[^.]+$/)) {
+        // Ignore hook files
+        continue;
+      } else {
+        await fs.copy(fromFile, toFile);
+      }
+    }
+  }
+
```
iankhou: There's a lot of logic in here that ultimately does something relatively simple - copy files while ignoring a specific file. Look into [`fs-extra.copy()`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/copy.md), which allows you to implement a filter using your regex.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +71,116 @@ export async function cliInit(options: CliInitOptions) {
...
+      if (entry.isDirectory() && supportedLanguages.includes(entry.name)) {
+        const langDir = path.join(templatePath, entry.name);
+        try {
+          const files = await fs.readdir(langDir);
+          if (files.length > 0) {
+            result.push(entry.name);
+          }
+        } catch (e) {
+          // Skip directories we can't read
+          continue;
```
iankhou: This doesn't seem like the right thing to do, we should at least notify the user if this is the case, or perhaps take action, such as failing the init. It's generally not a good idea to swallow errors like this unless you know exactly why, and it's for a very good reason.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +71,116 @@ export async function cliInit(options: CliInitOptions) {
...
+  return language;
+}
+
+/**
+ * Get valid CDK language directories from a template path
+ * @param templatePath - Path to the template directory
+ * @returns Promise resolving to array of supported language names
+ */
+async function getLanguageDirectories(templatePath: string): Promise<string[]> {
+  const result: string[] = [];
```
iankhou: Call it `results` instead, since it potentially has multiple things in it. Or you could be more specific.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +71,116 @@ export async function cliInit(options: CliInitOptions) {
...
+
+  try {
+    const entries = await fs.readdir(templatePath, { withFileTypes: true });
+
+    for (const entry of entries) {
+      if (entry.isDirectory() && supportedLanguages.includes(entry.name)) {
+        const langDir = path.join(templatePath, entry.name);
+        try {
+          const files = await fs.readdir(langDir);
+          if (files.length > 0) {
```
iankhou: I thought we do some validation that there is at least one file of the language?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +71,116 @@ export async function cliInit(options: CliInitOptions) {
...
+ * @returns Promise resolving to array of supported language names
+ */
+async function getLanguageDirectories(templatePath: string): Promise<string[]> {
+  const result: string[] = [];
+  const supportedLanguages = ['typescript', 'javascript', 'python', 'java', 'csharp', 'fsharp', 'go'];
+
+  try {
+    const entries = await fs.readdir(templatePath, { withFileTypes: true });
+
+    for (const entry of entries) {
```
iankhou: This is synchronous, blocking code. We can improve here by promisifying it with something like Promise.all(), which will parallelize expensive calls that we are awaiting inside loops.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -520,6 +520,7 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
             canUseNetwork: undefined,
             generateOnly: args.generateOnly,
             libVersion: args.libVersion,
+            fromPath: args['from-path'],
```
iankhou, line 530: What happens if a user runs `cdk init` with both `--from-path` and `--template`?

iankhou (changes requested):
I left some comments about code efficiency and validations.

Also, about option exclusivity for the --from-path option.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: I don't think you need to implement the recursion here yourself. I took a closer look and this file already uses `fs-extra`, aliased as `fs`.

packages/aws-cdk/lib/cli/cli.ts
rohang9000: Currently, `--from-path` takes precedence and `--template` gets ignored, which might be confusing for users. Let me add validation to handle this conflict so when a user tries to use both --from-path and --template, the code throws a clear error message explaining that these options are mutually exclusive.

rohang9000: Also added error catching for --lib-version or --stack-name with --from-path since custom templates don't process version or stack-name placeholders.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -16,19 +16,53 @@ const camelCase = require('camelcase');
 const decamelize = require('decamelize');
 
 export interface CliInitOptions {
+  /**
+   * @default - Throws error requiring template specification
```
iankhou: Why does it throw an error if it's optional?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -16,19 +16,53 @@ const camelCase = require('camelcase');
...
+   */
   readonly canUseNetwork?: boolean;
+
+  /**
+   * @default false
+   */
   readonly generateOnly?: boolean;
+
+  /**
+   * @default - Process.cwd()
```
iankhou:
```suggestion
   * @default - process.cwd()
```

packages/aws-cdk/lib/cli/cli.ts
iankhou: We can catch it here with yargs:
https://yargs.js.org/docs/#api-reference-conflictsx-y

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -40,33 +74,36 @@ export async function cliInit(options: CliInitOptions) {
...
+  // Validate conflicting options
+  if (options.fromPath && options.type) {
+    throw new ToolkitError('Cannot specify both --from-path and template name. Use either --from-path for custom templates or specify a built-in template name.');
+  }
 
-  const template = (await availableInitTemplates()).find((t) => t.hasName(type!));
-  if (!template) {
-    await printAvailableTemplates(ioHelper, options.language);
-    throw new ToolkitError(`Unknown init template: ${type}`);
+  if (options.fromPath && options.libVersion) {
```
iankhou: Catch this at yargs instead

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -40,33 +74,36 @@ export async function cliInit(options: CliInitOptions) {
...
   const generateOnly = options.generateOnly ?? false;
   const workDir = options.workDir ?? process.cwd();
-  if (!options.type && !options.language) {
-    await printAvailableTemplates(ioHelper);
-    return;
-  }
 
-  const type = options.type || 'default'; // "default" is the default type (and maps to "app")
+  // Validate conflicting options
+  if (options.fromPath && options.type) {
```
iankhou: Catch this at yargs instead

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -40,33 +74,36 @@ export async function cliInit(options: CliInitOptions) {
...
+  if (options.fromPath && options.libVersion) {
+    throw new ToolkitError('Cannot specify --lib-version with --from-path. Custom templates do not process version placeholders.');
   }
-  if (!options.language && template.languages.length === 1) {
-    const language = template.languages[0];
-    await ioHelper.defaults.warn(
-      `No --language was provided, but '${type}' supports only '${language}', so defaulting to --language=${language}`,
-    );
+
+  if (options.fromPath && options.stackName) {
```
iankhou: Catch this at yargs instead

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +113,164 @@ export async function cliInit(options: CliInitOptions) {
...
+ */
+async function loadBuiltinTemplate(ioHelper: IoHelper, type?: string, language?: string): Promise<InitTemplate> {
+  if (!type && !language) {
+    await printAvailableTemplates(ioHelper, language);
+    throw new ToolkitError('No template specified. Please specify a template name.');
+  }
+
+  if (!type) {
+    await printAvailableTemplates(ioHelper, language);
+    throw new ToolkitError('No template specified. Please specify a template name.');
```
iankhou: Should these be the same error? Looks like you throw this error regardless of whether `language` is truthy, why check on line 145?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +113,164 @@ export async function cliInit(options: CliInitOptions) {
...
+        }
+      });
+
+    // eslint-disable-next-line @cdklabs/promiseall-no-unbounded-parallelism
+    const validationResults = await Promise.all(languageValidationPromises);
+    return validationResults.filter((languageName): languageName is string => languageName !== null);
+  } catch (error: any) {
+    if (error instanceof ToolkitError) {
+      throw error;
+    }
```
iankhou: Why do we perform this check here?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +113,164 @@ export async function cliInit(options: CliInitOptions) {
...
+        if (entry.isFile() && extensions.some(ext => entry.name.endsWith(ext))) {
+          return true;
+        } else if (entry.isDirectory()) {
+          dirsToCheck.push(path.join(currentDir, entry.name));
+        }
+      }
+    } catch (error: any) {
+      // Skip directories that can't be read (permissions, broken symlinks, etc.)
+      // but continue searching other directories
+      if (error.code === 'ENOENT' || error.code === 'EACCES' || error.code === 'ENOTDIR') {
```
iankhou: Why would we encounter ENOTDIR if we have a check for it on line 253?

iankhou: I also would not expect to see ENOENT or EACCES, should probably fail if there are not sufficient access rights.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +113,164 @@ export async function cliInit(options: CliInitOptions) {
...
+        continue;
+      }
+      // Re-throw unexpected errors
```
iankhou:
```suggestion
      // throw unexpected errors
```

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -340,29 +607,47 @@ async function initializeProject(
...
-  if (files.filter((f) => !f.startsWith('.')).length !== 0) {
-    throw new ToolkitError('`cdk init` cannot be run in a non-empty directory!');
+  try {
+    const files = await fs.readdir(workDir);
+    if (files.filter((f) => !f.startsWith('.')).length !== 0) {
+      throw new ToolkitError('`cdk init` cannot be run in a non-empty directory!');
+    }
+  } catch (e: any) {
+    if (e.code === 'ENOENT') {
+      await fs.mkdirp(workDir);
```
iankhou: Why create a directory here?

packages/aws-cdk/lib/cli/user-input.ts
```diff
@@ -1326,6 +1326,13 @@ export interface InitOptions {
    */
   readonly libVersion?: string;
 
+  /**
+   * Path to a local custom template directory
+   *
+   * @default - undefined
```
iankhou, line 1377:
```suggestion
   * @default undefined
```

packages/aws-cdk/lib/commands/init/init.ts
iankhou: Whatever follows `@default -` should describe what the default value is

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -16,19 +16,53 @@ const camelCase = require('camelcase');
...
 
   /**
    * Override the built-in CDK version
+   * @default - Uses built-in CDK version
    */
   readonly libVersion?: string;
 
+  /**
+   * Path to a local custom template directory
+   * @default - Uses built-in templates
```
iankhou: shouldn't it be `undefined` if it's using built-in templates?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -16,19 +16,53 @@ const camelCase = require('camelcase');
...
   readonly stackName?: string;
+
+  /**
+   * @default undefined
+   */
   readonly migrate?: boolean;
 
   /**
    * Override the built-in CDK version
+   * @default - Uses built-in CDK version
```
iankhou: Shouldn't it be `undefined` if using the built-in version?

packages/aws-cdk/lib/commands/init/init.ts
iankhou: See my earlier comment, use yargs `.conflicts`

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +113,164 @@ export async function cliInit(options: CliInitOptions) {
...
+ * @param ioHelper - IO helper for user interaction
+ * @param type - Template type name
+ * @default - Throws error requiring template specification
+ * @param language - Programming language filter
+ * @default - Uses all available languages for template filtering
+ * @returns Promise resolving to the loaded InitTemplate
+ */
+async function loadBuiltinTemplate(ioHelper: IoHelper, type?: string, language?: string): Promise<InitTemplate> {
+  if (!type && !language) {
+    await printAvailableTemplates(ioHelper, language);
```
iankhou: why pass language if it's falsy?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +113,164 @@ export async function cliInit(options: CliInitOptions) {
...
+
+/**
+ * Get valid CDK language directories from a template path
+ * @param templatePath - Path to the template directory
+ * @returns Promise resolving to array of supported language names
+ */
+async function getLanguageDirectories(templatePath: string): Promise<string[]> {
+  const cdkSupportedLanguages = ['typescript', 'javascript', 'python', 'java', 'csharp', 'fsharp', 'go'];
+  const languageExtensions: Record<string, string[]> = {
+    typescript: ['.ts'],
```
iankhou: TypeScript is a superset of JavaScript.

https://www.typescriptlang.org/docs/handbook/typescript-from-scratch.html

iankhou (changes requested):
Let's handle flag conflicts at yargs, more details in comments. Some questions about error-handling, other minor corrections.

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Yeah since its optional, it shouldn't throw an error if not provided. So will change the default value to undefined. Thought this tag should also describe default behavior that occurs when the parameter is not provided, now I get it.

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Yes the first check `if (!type && !language)` is redundant since both conditions throw the same error message. If `!type && !language` is true, then `!type` will also be true so the second check handles the same case. We only need `if (!type)` so updated that in code.

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Yeah, we throw an error regardless of if the error is already a ToolkitError. So removed this check and now all errors are wrapped with a TookKitError

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Yeah this shouldn't be checked here, my logic had it silently ignore issues so I fixed so it fails the operation like you said

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Removing this because workDir is always process.cwd() which always exists

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: In my updated code, since there's only `if (!type)`, the language parameter could still have a value so passing it to printAvailableTemplates is correct for filtering. In the original code with the `if (!type && !language)` condition though yeah passing language was not needed since it was falsy

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Oh right! I updated the TypeScript extensions list to include both .ts and .js files so we handle valid TypeScript templates that might include JavaScript files

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Removed, this was extraneous comment of removed section

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Set conflicts configuration correctly in cli-config.ts. Yargs should now be preventing users from using --from-path and --lib-version together at the parsing level, so don't need this logic here

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -517,11 +777,9 @@ async function loadInitVersions(): Promise<Versions> {
     'aws-cdk': versionNumber(),
   };
   for (const [key, value] of Object.entries(ret)) {
-    /* c8 ignore start */
```
kaizencc, line 543: why deleted?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -375,8 +635,8 @@ async function initializeGitRepository(ioHelper: IoHelper, workDir: string) {
     await execute(ioHelper, 'git', ['init'], { cwd: workDir });
     await execute(ioHelper, 'git', ['add', '.'], { cwd: workDir });
     await execute(ioHelper, 'git', ['commit', '--message="Initial commit"', '--no-gpg-sign'], { cwd: workDir });
-  } catch {
-    await ioHelper.defaults.warn('Unable to initialize git repository for your project.');
+  } catch (error: any) {
```
kaizencc: these changes, if you would like to add them, should be brought out into a separate `chore()` PR to limit diff changes on this [big] PR

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -329,6 +556,21 @@ export async function printAvailableTemplates(ioHelper: IoHelper, language?: str
...
+ * @param canUseNetwork - Whether network access is available
+ * @param generateOnly - Whether to only generate files without post-install steps
+ * @param workDir - Working directory for the project
+ * @param stackName - Name of the stack
+ * @default undefined
+ * @param migrate - Whether this is a migration project
+ * @default undefined
+ * @param cdkVersion - Version of the CDK to use
+ * @default - Uses built-in CDK version
+ */
```
kaizencc: whoah. this is a lot of documentation. is it helpful? `stackName = Name of the stack` is self-explanitory. Also, this function isn't exported, so it's not like we have customers interacting with this function trying to determine how to use them (that is usually when we care deeply about the docstring).

imo documenting this is unnecessary. do you disagree?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -270,32 +490,33 @@ interface ProjectInfo {
...
-      resolve(templates);
-    } catch {
-      resolve([]);
+  try {
+    const templatesDir = path.join(cliRootDir(), 'lib', 'init-templates');
+    const templateNames = await listDirectory(templatesDir);
+    const templatePromises = templateNames.map(templateName =>
+      InitTemplate.fromName(templatesDir, templateName),
+    );
+    // eslint-disable-next-line @cdklabs/promiseall-no-unbounded-parallelism
```
kaizencc: provide a justification for this eslint disabled rule, like we have elsewhere when we disable this

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -146,22 +346,30 @@ export class InitTemplate {
...
+        ioHelper,
+        { targetDirectory, language, templateName: this.name },
+        {
+          substitutePlaceholdersIn: async (...fileNames: string[]) => {
+            const fileProcessingPromises = fileNames.map(async (fileName) => {
+              const fullPath = path.join(targetDirectory, fileName);
+              const template = await fs.readFile(fullPath, { encoding: 'utf-8' });
+              await fs.writeFile(fullPath, expandPlaceholders(template, language, projectInfo));
+            });
+            // eslint-disable-next-line @cdklabs/promiseall-no-unbounded-parallelism
```
kaizencc: same here, provide justification

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -124,6 +320,10 @@ export class InitTemplate {
    *
    * @param language    - the language to instantiate this template with
    * @param targetDirectory - the directory where the template is to be instantiated into
+   * @param stackName - Name of the stack
+   * @default undefined
+   * @param libVersion - Version of the CDK library to use
+   * @default undefined
```
kaizencc: this is self-explanitory

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -88,26 +256,54 @@ function pythonExecutable() {
 }
 const INFO_DOT_JSON = 'info.json';
 
+interface TemplateInitInfo {
+  readonly description: string;
+  readonly aliases?: string[];
+}
+
+enum TemplateType {
+  BUILTIN = 'builtin',
```
kaizencc: nit: this should be `BUILT_IN`

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +102,150 @@ export async function cliInit(options: CliInitOptions) {
...
+        const languageDirectoryPath = path.join(templatePath, directoryEntry.name);
+        try {
+          const hasValidLanguageFiles = await hasLanguageFiles(languageDirectoryPath, languageExtensions[directoryEntry.name]);
+          return hasValidLanguageFiles ? directoryEntry.name : null;
+        } catch (error: any) {
+          throw new ToolkitError(`Cannot read language directory '${directoryEntry.name}': ${error.message}`);
+        }
+      });
+
+    // eslint-disable-next-line @cdklabs/promiseall-no-unbounded-parallelism
```
kaizencc: justify

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -76,8 +102,150 @@ export async function cliInit(options: CliInitOptions) {
...
+}
+
+/**
+ * Load a built-in template by name
+ * @param ioHelper - IO helper for user interaction
+ * @param type - Template type name
+ * @default undefined
+ * @param language - Programming language filter
+ * @default undefined
+ * @returns Promise resolving to the loaded InitTemplate
```
kaizencc: review your documentation and keep only the additions that you think add value -- what would not be immediately obvious from the function name and a quick skim of the code?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -16,19 +16,55 @@ const camelCase = require('camelcase');
...
 export interface CliInitOptions {
+  /**
+   * Template name to initialize
+   * @default undefined
+   */
   readonly type?: string;
+
+  /**
+   * Programming language for the project
+   * @default undefined
```
kaizencc: `@default undefined` at best doesn't help here, at worst is incorrect. this value is probably required if multiple languages are available, or the default is the single available language, right? to me, that information provides much more value

packages/aws-cdk/lib/cli/user-input.ts
kaizencc: nit: this wasn't done

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -395,6 +395,7 @@ export async function makeConfig(): Promise<CliConfig> {
           'list': { type: 'boolean', desc: 'List the available templates' },
           'generate-only': { type: 'boolean', default: false, desc: 'If true, only generates project files, without executing additional operations such as setting up a git repo, installing dependencies or compiling the project' },
           'lib-version': { type: 'string', alias: 'V', default: undefined, desc: 'The version of the CDK library (aws-cdk-lib) to initialize the project with. Defaults to the version that was current when this CLI was built.' },
+          'from-path': { type: 'string', desc: 'Path to a local custom template directory', requiresArg: true, conflicts: ['lib-version'] },
```
kaizencc: q: why does this conflict with lib-version? and if so, can we update the lib-version description to make it clear its only for built-in projects

packages/aws-cdk/test/commands/init.test.ts
```diff
@@ -242,6 +242,85 @@ describe('constructs version', () => {
...
+    expect(await fs.pathExists(path.join(projectDir, 'app.ts'))).toBeTruthy();
+
+    const packageJson = JSON.parse(await fs.readFile(path.join(projectDir, 'package.json'), 'utf8'));
+    expect(packageJson.name).toBe('my-custom-project');
+
+    const appTs = await fs.readFile(path.join(projectDir, 'app.ts'), 'utf8');
+    expect(appTs).toBe('console.log("Hello from custom template!");');
+  });
+
+  cliTest('custom template with single language auto-detects language', async (workDir) => {
```
kaizencc: test for custom template + multiple languages fails if language not provided

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -88,26 +256,54 @@ function pythonExecutable() {
...
     public readonly name: string,
     public readonly languages: string[],
-    initInfo: any,
+    initInfo: TemplateInitInfo | null,
+    templateType: TemplateType,
   ) {
-    this.description = initInfo.description;
-    for (const alias of initInfo.aliases || []) {
-      this.aliases.add(alias);
+    this.description = initInfo?.description ?? 'Custom template';
```
kaizencc: where is description being used?

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -88,26 +256,54 @@ function pythonExecutable() {
...
+  readonly description: string;
+  readonly aliases?: string[];
+}
+
+enum TemplateType {
+  BUILTIN = 'builtin',
+  CUSTOM = 'custom',
+}
+
 export class InitTemplate {
```
kaizencc, line 324: meta comment (as in, not your fault and not to be addressed in this PR): 

why is `InitTemplate` exported? and can we make breaking changes to it? I don't see it used anywhere besides `init.ts` and I think we should be able to make the following changes

- unexport
- make the constructor private (interact with `InitTemplate` only through `fromName` or `fromPath`
- make the constructor take in a property bag

@iankhou @rohang9000 what do you think? maybe i am missing something

kaizencc (commented):
i don't want to step on @iankhou's toes on the review but here are some thoughts, mostly surrounding non-essential portions.

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: I added the /* c8 ignore */ comments because at one point my test coverage was below the threshold limit so I needed it for my build to succeed. Now my build passes, so i deleted because I don't need to exclude these lines from the coverage calculation. 

Its hard to test those lines because they check if the aws-cdk-lib, constructs, or aws-cdk version values are missing from the built-in .init-version.json file, which is generated when the build is run. With /* c8 ignore */, those lines are ignored from being counted in coverage even though I'm not testing them elsewhere, which is why I originally added it.

Whether I readd it would depend on if, outside of build passing or not, we want the test coverage metric to consider lines like this that can't really be tested. Is this the standard that I should've maintained?

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Yeah I definitely agree about removing documentation for params that have self-explanatory names so I'll remove that. Also, if the standard is to only have detailed docstrings for exported functions that customers interact with directly (and need to know what each param means/defaults), then like you said that isn't the case for this internal function so I can remove the documentation.

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: I agree! Changed to:

/**
   * Programming language for the project
   * @default - Optional/auto-detected if template supports only one language, otherwise required
   */
  readonly language?: string;

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Thanks! Added justification to this eslint disable comment

packages/aws-cdk/lib/cli/user-input.ts
rohang9000: Done now

packages/aws-cdk/lib/cli/cli-config.ts
rohang9000: Fixed! Updated the lib-version description in cli-config.ts to clarify it's only for built-in templates:

'lib-version': { type: 'string', alias: 'V', default: undefined, desc: 'The version of the CDK library (aws-cdk-lib) to initialize built-in templates with. Defaults to the version that was current when this CLI was built.' }

This conflicts because the lib-version and from-path options are mutually exclusive by design. `lib-version` overrides CDK version for built-in templates and `from-path` uses custom templates where the CDK version is determined by the template itself.

packages/aws-cdk/test/commands/init.test.ts
rohang9000: Fixed this, added the missing test case for custom templates with multiple languages failing when no language is provided.

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: It's not used. It's only used for built-in templates that specify a description in their info.json file. In the existing code (main branch), the description is then used to show what each built-in template does when a user runs `cdk init --list`. This doesn't apply to custom templates since they don't have info.json files that define a description and also usage instructions for them are not displayed through `cdk init --list` so I can remove it. I just had it in there to keep custom templates properties consistent with built-in template properties so I had custom templates default to a hardcoded string i made.

Fixing the code so that custom template's don't use descriptions at all or have any kind of hardcoded description that I define either.

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: K I agree! language, ioHelper and type (from original code) are self-explanatory names and obvious what they do so I can remove the documentation for them here

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Got it! Reverted this back to the original code for this PR since fixing error logging and its not related to the local-template-support functionality I added. Will make these changes in a seperate fix PR

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: I agree! Since it's only used internally in init.ts, there's no need to expose it as part of the public API. Also, making the constructor private and forcing creation through InitTemplate.fromName() and InitTemplate.fromPath() makes more sense to me too. All of this would be a breaking change to the InitTemplate API though, so might need its own issue?

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Made these changes in a seperate branch off of main and created a PR for them below: 

https://github.com/aws/aws-cdk-cli/pull/778

packages/aws-cdk/lib/cli/user-input.ts
iankhou: I don't think it was done

iankhou (changes requested):
Missed one of my comments but after that, looks good

packages/aws-cdk/lib/cli/user-input.ts
rohang9000: Hey I fixed this but I because user-input.ts is a generated file it automatically generates the "@default - undefined" notation.

I looked into it and this is actually a pre-existing issue throughout the entire user-input.ts file. I found it's actually a because of the @cdklabs/typewriter library that generates the JSDoc comments. The typewriter library is hardcoded to automatically adds a dash before all default values, so we also get "@default - true" instead of "@default true" throughout the entire codebase. So it's not something that is just broken in this branch for the local template feature, but a bug that appears everywhere in our codebase (user-input.ts in the main branch also has the dashes). The fix would require us to update the typewriter library itself to generate the @default documentation without the dash, but that's library is outside of our codebase.

Should we maintain this format (by the typewriter) in the source files that I write too (init.ts, etc)? Or should the source files use a different notation for parameter documentation than the generated files?

packages/aws-cdk/lib/commands/init/init.ts
rohang9000: Yes, but if I removed the descriptions for them, they are throwing eslint errors that they need descriptions. I could do an eslint-disable here but i'd have to do if for both params so Ian and I agreed that it just makes more sense to keep the descriptions.

packages/aws-cdk/test/commands/init.test.ts
```diff
@@ -242,6 +242,336 @@ describe('constructs version', () => {
...
+
+    // Create template files (custom templates don't process placeholders)
+    await fs.writeFile(path.join(tsDir, 'package.json'), JSON.stringify({
+      name: 'my-custom-project',
+      version: '1.0.0',
+      dependencies: {
+        'aws-cdk-lib': '^2.0.0',
+        'constructs': '^10.0.0',
+      },
+    }, null, 2));
```
iankhou: Consider separating test data from test code. I see other places in your tests where you initialize this file as well, albeit slightly differently.

In general, we should try to limit the amount of test data and fixtures written inside the test, and instead keep them in a separate file or directory. That way, we can reuse them to expect consistent outcomes and maintain them in a centralized place.

packages/aws-cdk/test/commands/init.test.ts
iankhou, line 1: In the description for `--template-path`, you indicate that it's used with `--from-path`. It's unclear whether `--from-path` is required for a user to use `--template-path` (ex. could they omit from-path and use template-path to point directly to the template on their fs?).

If `--from-path` is required, there should be a test that ensures the command will fail when not provided. Of course, this will need to be adjusted once you add additional template sources. You can also use .check() to ensure that it is provided when the user uses `--template-path`.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -394,7 +394,9 @@ export async function makeConfig(): Promise<CliConfig> {
...
           'list': { type: 'boolean', desc: 'List the available templates' },
           'generate-only': { type: 'boolean', default: false, desc: 'If true, only generates project files, without executing additional operations such as setting up a git repo, installing dependencies or compiling the project' },
-          'lib-version': { type: 'string', alias: 'V', default: undefined, desc: 'The version of the CDK library (aws-cdk-lib) to initialize the project with. Defaults to the version that was current when this CLI was built.' },
+          'lib-version': { type: 'string', alias: 'V', default: undefined, desc: 'The version of the CDK library (aws-cdk-lib) to initialize built-in templates with. Defaults to the version that was current when this CLI was built.' },
+          'from-path': { type: 'string', desc: 'Path to a local custom template directory or multi-template repository', requiresArg: true, conflicts: ['lib-version'] },
+          'template-path': { type: 'string', desc: 'Path to a specific template within a multi-template repository (used with --from-path)', requiresArg: true },
```
iankhou: if it is used with --from-path, there should be a `.check()` to ensure that from-path is used.

iankhou (changes requested):
Do you think we need integration testing? i.e. @aws-cdk-testing

If not, why not?

packages/aws-cdk/test/commands/init.test.ts
rohang9000: Yes I agree, I changed the code to create centralized fixture functions in packages/aws-cdk/test/_fixtures/init-templates/template-helpers.ts and refactored inline template creation to use reusable functions like createSingleLanguageTemplate(), createMultiLanguageTemplate(), and createMultiTemplateRepository()

The fixture functions are parameterized to handle different scenarios so tests can still create variations of templates without duplicating the underlying template creation logic

packages/aws-cdk/test/commands/init.test.ts
rohang9000: Thanks, I implemented both suggestions. `--from-path` is required when using `--template-path` because `--template-path` specifies a relative path within a multi-template repository, not a direct filesystem path. I also added a .check() function in the CLI configuration that validates this dependency and throws an error if `--template-path` is used without `--from-path`. I also added a test case "fails when --template-path is used without --from-path" that verifies the command correctly fails when this validation is triggered

packages/aws-cdk/lib/cli/cli-config.ts
rohang9000: Added a .check() function to the `--template-path` parameter in the CLI configuration that validates `--from-path` is provided when `--template-path` is used. The check function throws an error with the message "The --template-path option requires --from-path to be specified" if it isn't met

packages/@aws-cdk/user-input-gen/lib/yargs-gen.ts
```diff
@@ -52,6 +52,7 @@ export async function renderYargs(config: CliConfig, helpers: CliHelpers): Promi
     disabledEsLintRules: [
       EsLintRules.MAX_LEN, // the default disabled rules result in 'Definition for rule 'prettier/prettier' was not found
       '@typescript-eslint/consistent-type-imports', // (ironically) typewriter does not support type imports
+      '@cdklabs/no-throw-default-error', // generated check functions throw proper Error objects
```
iankhou: Why do we need to disable this rule here?

packages/aws-cdk/lib/cli/cli-config.ts
iankhou: Found a better way to do this that doesn't require you to throw an error
https://yargs.js.org/docs/#api-reference-impliesx-y

packages/@aws-cdk/user-input-gen/lib/yargs-gen.ts
rohang9000: Yeah you're right, we don't need to disable that rule anymore. It was originally needed because the generated check functions were throwing new Error() objects, which violated the @cdklabs/no-throw-default-error rule. Since we switched from using check functions to implies, these errors are no longer thrown so the rule disable is unnecessary. Removed it from the code

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -398,7 +398,7 @@ export async function makeConfig(): Promise<CliConfig> {
           'from-path': { type: 'string', desc: 'Path to a local custom template directory or multi-template repository', requiresArg: true, conflicts: ['lib-version'] },
           'template-path': { type: 'string', desc: 'Path to a specific template within a multi-template repository (requires --from-path)', requiresArg: true },
         },
-        check: '(argv: any) => { if (argv.templatePath && !argv.fromPath) { throw new Error("--template-path requires --from-path to be specified"); } return true; }',
+        implies: { 'template-path': 'from-path' },
```
iankhou, line 407: shouldn't this change break the existing test for init? Why doesn't it?

and the test needs to be updated as well

rohang9000: Yes, you're right, the implies constraint prevented the existing test from failing because `yargs` automatically makes sure that `--from-path` is provided when `--template-path` is used, so the manual validation never gets triggered. 

I corrected my code by removing the redundant manual validation in `init.ts` since the implies constraint handles this at the CLI parsing level. I also removed the test that expected the manual validation error and added a new test to verify that the implies constraint works correctly.

The code should now rely on the constraint validation of yargs instead of duplicating the logic with the manual validation

packages/@aws-cdk/user-input-gen/lib/yargs-types.ts
```diff
@@ -7,6 +7,8 @@ interface YargsCommand {
 
 export interface CliAction extends YargsCommand {
   options?: { [optionName: string]: CliOption };
+  check?: string;
```
iankhou: do we still need this, since you removed .check()?

packages/@aws-cdk/user-input-gen/lib/yargs-gen.ts
```diff
@@ -109,7 +111,24 @@ function makeYargs(config: CliConfig, helpers: CliHelpers): Statement {
...
+
+    // Collect check functions to add at global level
+    if (commandFacts.check) {
+      checkFunctions.push(commandFacts.check);
+    }
+  }
+
+  // Add check functions at global level
+  for (const checkFunction of checkFunctions) {
+    yargsExpr = yargsExpr.callMethod('check', code.expr.directCode(checkFunction));
```
iankhou: do we still need this?

packages/aws-cdk/test/commands/init.test.ts
```diff
@@ -243,14 +243,24 @@ describe('constructs version', () => {
...
+      { name: 'test-template', languages: ['typescript'] },
+    ]);
+    const projectDir = path.join(workDir, 'my-project');
+    await fs.mkdirp(projectDir);
+
+    await cliInit({
       ioHelper,
-      type: undefined,
+      fromPath: repoDir,
+      templatePath: 'test-template',
```
iankhou: what are we testing for here? We can only know if .implies() works correctly if we omit `--from-path`. Your previous test description `fails when --template-path is used without --from-path` didn't need to be changed.

iankhou (changes requested):

packages/@aws-cdk/user-input-gen/lib/yargs-types.ts
rohang9000: Removed

packages/@aws-cdk/user-input-gen/lib/yargs-gen.ts
rohang9000: No, removed

packages/aws-cdk/test/commands/init.test.ts
rohang9000: Removed the test entirely because it was directly calling the method cliInit which had a manual if statement validation in it. This was wrong and the test should have been  for the CLI argument parsing itself, not the cliInit() function. This test was only correct before we switched to relying on yargs

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -394,8 +394,11 @@ export async function makeConfig(): Promise<CliConfig> {
...
           'list': { type: 'boolean', desc: 'List the available templates' },
           'generate-only': { type: 'boolean', default: false, desc: 'If true, only generates project files, without executing additional operations such as setting up a git repo, installing dependencies or compiling the project' },
-          'lib-version': { type: 'string', alias: 'V', default: undefined, desc: 'The version of the CDK library (aws-cdk-lib) to initialize the project with. Defaults to the version that was current when this CLI was built.' },
+          'lib-version': { type: 'string', alias: 'V', default: undefined, desc: 'The version of the CDK library (aws-cdk-lib) to initialize built-in templates with. Defaults to the version that was current when this CLI was built.' },
+          'from-path': { type: 'string', desc: 'Path to a local custom template directory or multi-template repository', requiresArg: true, conflicts: ['lib-version'] },
+          'template-path': { type: 'string', desc: 'Path to a specific template within a multi-template repository (requires --from-path)', requiresArg: true },
```
iankhou:
```suggestion
          'template-path': { type: 'string', desc: 'Path to a specific template within a multi-template repository', requiresArg: true },
```

Since we validate that this option requires --from-path, we don't need to add it to the description.

packages/aws-cdk/lib/cli/cli-config.ts
iankhou: I don't see a test for the `implies` logic anymore

iankhou (changes requested):
Code looks pretty good, but I would like to see some additional test coverage:
- make sure unstable flag works
- conflict between `lib-version` and `from-path`
- `template-path` implies `from-path`
- hook files are ignored in your init copy
- file permission failures (you can use fs.chmod)
- using relative vs absolute paths for options requiring file paths

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -154,16 +154,12 @@ async function loadLocalTemplate(fromPath: string, templatePath?: string): Promi
  * Load a built-in template by name
  */
 async function loadBuiltinTemplate(ioHelper: IoHelper, type?: string, language?: string): Promise<InitTemplate> {
-  if (!type) {
-    await printAvailableTemplates(ioHelper, language);
-    throw new ToolkitError('No template specified. Please specify a template name.');
-  }
-
-  const template = (await availableInitTemplates()).find((t) => t.hasName(type));
+  const templateType = type || 'default'; // "default" is the default type (and maps to "app")
```
iankhou, line 163: This changes the existing behavior no?

i.e. if a user runs only `cdk init`, it will give them a Typescript app, instead of printing templates options?

iankhou (changes requested):

iankhou (changes requested):
Let's add a test for `cdk init --language=[language]`, since we are adjusting logic surrounding that behavior.

ex. Currently, when I run `cdk init --language=typescript`, I get an initialization of the typescript CDK app.

iankhou (approved):

PR #774, opened 2025-08-06
feat: interactive experience for the `cdk flags` command

vivian12300:
This uses the 'enquirer' npm package to allow users to call `cdk flags -interactive`, see a list of Menu options, and configure their feature flags.

---


packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -126,6 +126,7 @@ export async function makeConfig(): Promise<CliConfig> {
           unconfigured: { type: 'boolean', desc: 'Modify unconfigured feature flags', requiresArg: false },
           recommended: { type: 'boolean', desc: 'Change flags to recommended states', requiresArg: false },
           default: { type: 'boolean', desc: 'Change flags to default state', requiresArg: false },
+          i: { type: 'boolean', desc: 'Interactive option for the flags command', requiresArg: false },
```
otaviomacedo: The convention is to have a main descriptive name, like `interactive`, and then one or more short aliases like `i`. Look at other commands to see how it's done with `yargs`.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -38,6 +48,52 @@ export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, o
     unconfigured: options.unconfigured,
   };
 
+  const interactiveOptions = [
```
otaviomacedo: I recommend creating an enum for these values and to dynamically compute the array from the enum:

```ts
// Just a name suggestion. Feel free to come up with a better one
enum FlagChoice {
  ALL = 'Set all flags to recommended values',
  RECOMMENDED = 'Set unconfigured flags to recommended values',
  ...
};

const interactiveOptions = Object.values(FlagChoice);
```

In the answer comparison, you can use the enum:

```ts
if (answer == FlagChoice.ALL) {
...
```

otaviomacedo (changes requested):

otaviomacedo (approved):

PR #776, opened 2025-08-06
feat: allow filtering `cdk flags` feature flags by substring

vivian12300:
This allows users to run `cdk flags` with a substring and all matching feature flags will be displayed to them. 

When a user runs `cdk flags #substring#` and there is only one matching flag, a description of the flag will be displayed. If there are multiple matching flags, all matching flags will be displayed in a table. 

If a user runs `cdk flags` with multiple substrings, all matching flags that contain any of the substrings will be displayed in a table.

---


packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -381,6 +359,44 @@ export async function displayFlags(params: FlagOperationsParams): Promise<void>
   await ioHelper.defaults.info(formattedTable);
 }
 
+export async function displayFlags(params: FlagOperationsParams): Promise<void> {
+  const { flagData, ioHelper, flagName, all } = params;
+
+  if (flagName && flagName.length > 0) {
+    const searchTerm = flagName[0];
+    const matchingFlags = flagData.filter(f => f.name.toLowerCase().includes(searchTerm.toLowerCase()));
```
otaviomacedo: We need to document this behavior. By the way, do we have any documentation for this command in the README?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -381,6 +359,44 @@ export async function displayFlags(params: FlagOperationsParams): Promise<void>
   await ioHelper.defaults.info(formattedTable);
 }
 
+export async function displayFlags(params: FlagOperationsParams): Promise<void> {
+  const { flagData, ioHelper, flagName, all } = params;
+
+  if (flagName && flagName.length > 0) {
+    const searchTerm = flagName[0];
```
otaviomacedo: Do we just ignore the other flag names?

vivian12300: The substring the user enters (like `s3`) is going to be flagName[0] so there aren't any other flag names in the array. 

When a user enters a substring and there are multiple matches, it will display a table with only those flags.

packages/aws-cdk/lib/commands/flag-operations.ts
vivian12300: No, not yet. I've been considering the RFC as the documentation. We talked about writing a blog post for this command but it is already pretty straightforward. I can update the RFC to show these changes.

otaviomacedo: Yes, but the filtering, specifically, is not obvious. First, there is no indication that you can filter at all, and the filtering could obey other rules, such as by prefix, regex, glob etc. The fact that it's by substring has to be made explicit somewhere.

vivian12300: I spoke to Joey and we will be writing official documentation for the command. I'll make sure to state it there.

kaizencc: i am with otavio here. at a minimum we need to document how to use `cdk flags` here: https://github.com/aws/aws-cdk-cli/tree/main/packages/aws-cdk. that can be in a separate PR if you'd like, but i don't think this PR can go in without documentation. that is, we want users to use this feature and provide feedback right?

packages/aws-cdk/test/commands/flag-operations.test.ts
```diff
@@ -180,6 +180,22 @@ describe('displayFlags', () => {
     expect(plainTextOutput).toContain('aws-cdk-lib');
     expect(plainTextOutput).toContain('different-module');
   });
+  test('displays single flag details when only one substring match is found', async () => {
```
kaizencc, line 184:
```suggestion

  test('displays single flag details when only one substring match is found', async () => {
```

packages/aws-cdk/lib/commands/flag-operations.ts
kaizencc: an additional requirement: an example of how to use this feature in the PR description

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -459,6 +437,44 @@ export async function displayFlags(params: FlagOperationsParams): Promise<void>
   await ioHelper.defaults.info(formattedTable);
 }
 
+export async function displayFlags(params: FlagOperationsParams): Promise<void> {
+  const { flagData, ioHelper, flagName, all } = params;
```
kaizencc, line 453: what is `all`? i would expect a description in `FlagOperationsParams` but its not there. I also don't know what the default value is supposed to be in the documentation.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -459,6 +437,44 @@ export async function displayFlags(params: FlagOperationsParams): Promise<void>
...
+
+  if (flagName && flagName.length > 0) {
+    const searchTerm = flagName[0];
+    const matchingFlags = flagData.filter(f => f.name.toLowerCase().includes(searchTerm.toLowerCase()));
+
+    if (matchingFlags.length === 0) {
+      await ioHelper.defaults.error('Flag not found.');
```
kaizencc:
```suggestion
      await ioHelper.defaults.error(`Flag matching "${searchTerm}" not found.`);
```

kaizencc: also need a unit test for this

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -459,6 +437,44 @@ export async function displayFlags(params: FlagOperationsParams): Promise<void>
...
+      await ioHelper.defaults.info(`Flag name: ${flag.name}`);
+      await ioHelper.defaults.info(`Description: ${flag.explanation}`);
+      await ioHelper.defaults.info(`Recommended value: ${flag.recommendedValue}`);
+      await ioHelper.defaults.info(`User value: ${flag.userValue}`);
+      return;
+    }
+
+    await ioHelper.defaults.info(`Found ${matchingFlags.length} flags matching "${searchTerm}":`);
+    await displayFlagTable(matchingFlags, ioHelper);
+    return;
```
kaizencc: unit test for multiple flags found

kaizencc (commented):
will you / do you have integ tests for `cdk flags` as well?

also, some general feedback:

i ran the following locally and this is what i saw:

```bash
../../aws-cdk-cli/packages/aws-cdk/bin/cdk flags --unstable=flags --all 
+-------------------+-------------------+------------+
| Feature Flag Name | Recommended Value | User Value |
+-------------------+-------------------+------------+
+-------------------+-------------------+------------+
```

absent of any documentation i don't really know how to proceed, or if there's a bug, because i definitely do have feature flags set.

i then ran:

```
../../aws-cdk-cli/packages/aws-cdk/bin/cdk flags --unstable=flags --set --recommended --all
✨  Synthesis time: 3.59s
Successfully synthesized to /mydir/original
✨  Synthesis time: 3.58s
Successfully synthesized to /mydir/temp
✨  Synthesis time: 0s
Stack TelemetryPocStack
There were no differences

✨ Number of stacks with differences: 0
Do you want to accept these changes? (y/n) y
Flag value(s) updated successfully.
```

It looks like potentially synthesis happened 3 times? and it asked if i wanted to accept changes although i do not see any changes after I selected yes and saw `Flag value(s) updated successfully`. Does that message show up regardless of if there are changes? And why can't i see a preview of the changes before I accept?

packages/aws-cdk/lib/commands/flag-operations.ts
rix0rrr: I think Otavio means that `flagName: string[]`, so what we do with other names besides the first?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -392,44 +392,20 @@ function formatTable(headers: string[], rows: string[][]): string {
...
-    if (!flag) {
-      await ioHelper.defaults.error('Flag not found.');
-      return;
-    }
-
-    await ioHelper.defaults.info(`Description: ${flag.explanation}`);
-    await ioHelper.defaults.info(`Recommended value: ${flag.recommendedValue}`);
-    await ioHelper.defaults.info(`User value: ${flag.userValue}`);
-    return;
+function getFlagPriority(flag: FeatureFlag): number {
```
rix0rrr: Nitpick: "priority" sounds like there is some inherent property to these things that makes one of them more important than the other. Whereas I think you mean something more like `sortOrder` or `sortKey` here, i.e. purely a visual thing?

packages/aws-cdk/lib/commands/flag-operations.ts
vivian12300: `flagName` is an array to account for cases when the user wants to modify multiple feature flags. In this case, the user is looking to view one flag, so the array would just have a length of 1 (containing the substring). If the user enters multiple flag names/substrings like `cdk flags #substring# #substring#`, it'll return a flag not found error.

rix0rrr (approved):

PR #788, opened 2025-08-12
fix(cli-integ): add retry for iam eventual consistency issue and  migration tests for java

abidhasan-aws:
### Background
The CDK pipelines have been experiencing intermittent failures due to flaky tests that typically pass on retry. This pull request addresses the investigation of the two most frequent failing tests.

image (1)

### AWS IAM Eventual Consistency Issue
Test: `docker-credential-cdk-assets can assume role and fetch ECR credentials`

Issue: Docker credential fetching fails with AccessDenied errors because newly created IAM roles and policies require time to propagate across AWS regions.

Fix: Implemented a 60-second retry mechanism for `fetchDockerLoginCredentials()` when encountering AccessDenied errors.

### CDK Migration Test Instability
Test: `cdk migrate java deploys successfully`

Issue: Java CDK migration tests fail sporadically due to Maven Central repository rate limiting errors & dependency resolution failure

Fix: Implemented full test retry logic as these transient network-related issues could not be reproduced in local environments.

### Impact
These changes should improve pipeline stability and reduce the need for manual intervention.

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-assets/cdk-assets-docker-credential.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-assets/cdk-assets-docker-credential.integtest.ts
@@ -84,7 +84,7 @@ async function testDockerCredential(fixture: TestFixture, credSource: DockerDoma
 
   await fixture.cdkAssets.makeCliAvailable();
   let output: string = '';
-  
+
   await retry(process.stdout, 'Getting docker credentials', retry.forSeconds(60), async () => {
     output = await fixture.shell(['docker-credential-cdk-assets', 'get'], {
       modEnv: {
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-assets/cdk-assets-docker-credential.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-assets/cdk-assets-docker-credential.integtest.ts
@@ -84,6 +84,7 @@ async function testDockerCredential(fixture: TestFixture, credSource: DockerDoma
 
   await fixture.cdkAssets.makeCliAvailable();
   let output: string = '';
+  
   await retry(process.stdout, 'Getting docker credentials', retry.forSeconds(60), async () => {
     output = await fixture.shell(['docker-credential-cdk-assets', 'get'], {
       modEnv: {
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-assets/cdk-assets-docker-credential.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-assets/cdk-assets-docker-credential.integtest.ts
@@ -5,7 +5,7 @@ import { GetCallerIdentityCommand } from '@aws-sdk/client-sts';
 // eslint-disable-next-line import/no-relative-packages
 import type { DockerDomainCredentialSource } from '../../../../../@aws-cdk/cdk-assets-lib/lib/private/docker-credentials';
 import type { TestFixture } from '../../../lib';
-import { integTest, withDefaultFixture, withRetry } from '../../../lib';
+import { integTest, withDefaultFixture, withRetry, retry } from '../../../lib';
 
 jest.setTimeout(2 * 60 * 60_000); // Includes the time to acquire locks, worst-case single-threaded runtime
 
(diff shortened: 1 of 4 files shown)


packages/cdk-assets/bin/docker-credential-cdk-assets.ts
```diff
@@ -39,9 +39,32 @@ async function main() {
 
   // Read the domain to fetch from stdin
   let endpoint = fs.readFileSync(0, { encoding: 'utf-8' }).trim();
-  const credentials = await fetchDockerLoginCredentials(new DefaultAwsClient(), config, endpoint);
-  // Write the credentials back to stdout
-  fs.writeFileSync(1, JSON.stringify(credentials));
+
+  const deadline = Date.now() + 60_000; // 60 seconds timeout
```
Abogical: Didn't know you could have _ in numbers. nice!

Abogical (approved):

packages/cdk-assets/bin/docker-credential-cdk-assets.ts
```diff
@@ -39,9 +39,32 @@ async function main() {
...
+      const credentials = await fetchDockerLoginCredentials(new DefaultAwsClient(), config, endpoint);
+      // Write the credentials back to stdout
+      fs.writeFileSync(1, JSON.stringify(credentials));
+      return;
+    } catch (e: any) {
+      lastError = e;
+      
+      // Retry on AccessDenied errors due to eventual consistency in AWS IAM.
+      // Newly created roles and policies may take time to propagate across regions.
+      if (e.name === 'AccessDenied') {
```
kumvprat: Would it be better to match with error type : Like class/interface type rather than the name ?

kumvprat (commented):
Overall could we have a wrapper method : withBackoffRetry or something like that with a max retry /timeout parameter ? 

This would allow us to wrap any such tests in that method and have a centralized place for all retry logic than per test

Could be in future that other tests might come up which need these retries

packages/cdk-assets/bin/docker-credential-cdk-assets.ts
mrgrain, line 1: This seems like a very heave-handed approach to what appears to be mainly an issue with our internal integration tests (since I am not aware of customer reports).

In any case, this change needs to be a separate PR as it is Cx facing.

cc @rix0rrr

packages/cdk-assets/bin/docker-credential-cdk-assets.ts
```diff
@@ -39,9 +39,32 @@ async function main() {
...
   // Read the domain to fetch from stdin
   let endpoint = fs.readFileSync(0, { encoding: 'utf-8' }).trim();
-  const credentials = await fetchDockerLoginCredentials(new DefaultAwsClient(), config, endpoint);
-  // Write the credentials back to stdout
-  fs.writeFileSync(1, JSON.stringify(credentials));
+
+  const deadline = Date.now() + 60_000; // 60 seconds timeout
+  let lastError: Error | undefined;
+
+  while (Date.now() < deadline) {
```
mrgrain: This probably can be implemented at the SDK level using retries:
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-util-retry/
https://github.com/smithy-lang/smithy-typescript/blob/main/packages/types/src/retry.ts

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cdk-assets/cdk-assets-docker-credential.integtest.ts
```diff
@@ -83,13 +83,17 @@ async function testDockerCredential(fixture: TestFixture, credSource: DockerDoma
...
-  const output = await fixture.shell(['docker-credential-cdk-assets', 'get'], {
-    modEnv: {
-      ...fixture.cdkShellEnv(),
-      CDK_DOCKER_CREDS_FILE: credsFilePath,
-    },
-    stdio: [fs.openSync(input, 'r')],
-    captureStderr: false,
+  let output: string = '';
+
+  await retry(process.stdout, 'Getting docker credentials', retry.forSeconds(60), async () => {
```
kumvprat, line 88: Should this be `retry` or `waitForAssumeRole` or even `retryOnMatchingErrors` ?

abidhasan-aws: It should not be `waitForAssumeRole`. 

We are retrying this because in this call, there is a credential fetching that needs a wait period for IAM eventual consistency. We could use `retryOnMatchingErrors`, but there is no clean way to catch error and retry based on that. The package just returns `exited with code 1` in case of errors. 

So we are retrying anyway and not based on errors, thus the generic `retry`.

kumvprat: Final thought should we add a sleep time to retry function ? Is that something that can be used like the wait and retry we do in `retryOnMatchingErrors` ? Would that add any value ?

abidhasan-aws: It already has a sleep time of 5 seconds.

kumvprat (approved):

mrgrain (approved):

PR #794, opened 2025-08-13
feat: add notice when users run `cdk flags` with an incompatible version of the CDK Library

vivian12300:
Returns an error message for users if they run the `cdk flags` command with an incompatible version of `aws-cdk-lib`.

---

### DIFF
--- a/packages/aws-cdk/lib/commands/flag-operations.ts
+++ b/packages/aws-cdk/lib/commands/flag-operations.ts
@@ -1,4 +1,5 @@
 import * as path from 'path';
+import { formatTable } from '@aws-cdk/cloudformation-diff';
 import type { FeatureFlag, Toolkit } from '@aws-cdk/toolkit-lib';
 import { CdkAppMultiContext, MemoryContext, DiffMethod } from '@aws-cdk/toolkit-lib';
 import * as chalk from 'chalk';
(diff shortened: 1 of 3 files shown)


packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -44,6 +45,12 @@ interface FlagOperationsParams {
 
 export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, options: FlagsOptions, toolkit: Toolkit) {
   flagData = flagData.filter(flag => !OBSOLETE_FLAGS.includes(flag.name));
+
+  if (flagData.length == 0) {
+    await ioHelper.defaults.error('The \'cdk flags\' command is not compatible with the AWS CDK library used by your application. Please upgrade to the latest version.');
+    return;
```
kaizencc, line 51: the message should say the minimum aws-cdk-lib that supports `cdk flags`.

kaizencc: just a suggestion, we could get the actual aws-cdk-lib version this way: https://github.com/aws/aws-cdk-cli/blob/conroy/gc-consistency/packages/aws-cdk/lib/cli/telemetry/library-version.ts and assert against that (since we know the minimum supported version).

i'm not sure if that's better than your current method but bringing it up for you to be aware / decide.

PR #795, opened 2025-08-13
fix(cli): deprecate "bootstrap-stack-name" in favor of "toolkit-stack-name" for gc command

### LINKED ISSUE #385: (garbage-collection): Option inconsistency bootstrapStackName
### Describe the bug

The gc options is the only option naming the toolit stack `--bootstrap-stack-name` instead of `--toolkit-stack-name.`

Even the error message tells me to use toolkit...

That should be changed to be consistent.

Maybe I totally missed a reason for this but it cost me an hour of debugging.

### Regression Issue

_No response_

### Expected Behavior

consistent naming og the toolik stack name option

### Current Behavior

inconsistent naming

### Reproduction Steps

call cdk gc with a custom toolkit stack name

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.1009.0

### Framework Version

_No response_

### Node.js Version

22

### OS

mac

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

kaizencc:
Fixes #385

Elsewhere we are consistently naming this property `toolkit-stack-name`, so this PR aligns `gc` with the rest of the CLI world. We may have better options here, like making `toolkit-stack-name` a global option, but this is an easy fix for now.

Does not break users by maintaining `bootstrap-stack-name` as an alias of `toolkit-stack-name` for `gc`. However `bootstrap-stack-name` is marked as deprecated and will be removed when we GA garbage collection.

---

### DIFF
--- a/packages/aws-cdk/lib/cli/cli-config.ts
+++ b/packages/aws-cdk/lib/cli/cli-config.ts
@@ -110,7 +110,8 @@ export async function makeConfig(): Promise<CliConfig> {
           'rollback-buffer-days': { type: 'number', desc: 'Delete assets that have been marked as isolated for this many days', default: 0 },
           'created-buffer-days': { type: 'number', desc: 'Never delete assets younger than this (in days)', default: 1 },
           'confirm': { type: 'boolean', desc: 'Confirm via manual prompt before deletion', default: true },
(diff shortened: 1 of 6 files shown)


mrgrain (changes requested):
Can we deprecate and/or warn on the old option so we can remove it as part of GA?

mrgrain (approved):

PR #796, opened 2025-08-13
fix(cli): flags command shows unnecessary flags, not all table rows have vertical lines

vivian12300:
When a user runs `cdk flags`, a table is displayed with information about their feature flags. This PR omits the intentionally unconfigured feature flags from the table and makes the table more visually appealing. 

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
@@ -1276,29 +1276,46 @@ export class Toolkit extends CloudAssemblySourceBuilder {
   /**
    * Retrieve feature flag information from the cloud assembly
    */
-
   public async flags(cx: ICloudAssemblySource): Promise<FeatureFlag[]> {
     this.requireUnstableFeature('flags');
 
     const ioHelper = asIoHelper(this.ioHost, 'flags');
     await using assembly = await assemblyFromSource(ioHelper, cx);
-    const artifacts = assembly.cloudAssembly.manifest.artifacts;
+    const artifacts = Object.values(assembly.cloudAssembly.manifest.artifacts!);
+    const featureFlagReports = artifacts.filter(a => a.type === ArtifactType.FEATURE_FLAG_REPORT);
 
-    return Object.values(artifacts!)
-      .filter(a => a.type === ArtifactType.FEATURE_FLAG_REPORT)
-      .flatMap(report => {
-        const properties = report.properties as FeatureFlagReportProperties;
-        const moduleName = properties.module;
+    const flags = featureFlagReports.flatMap(report => {
+      const properties = report.properties as FeatureFlagReportProperties;
+      const moduleName = properties.module;
 
-        return Object.entries(properties.flags).map(([flagName, flagInfo]) => ({
+      const flagsWithUnconfiguredBehavesLike = Object.entries(properties.flags)
+        .filter(([_, flagInfo]) => flagInfo.unconfiguredBehavesLike != undefined);
+
+      const shouldIncludeUnconfiguredBehavesLike = flagsWithUnconfiguredBehavesLike.length > 0;
+
+      return Object.entries(properties.flags).map(([flagName, flagInfo]) => {
+        const baseFlag = {
           module: moduleName,
           name: flagName,
           recommendedValue: flagInfo.recommendedValue,
           userValue: flagInfo.userValue ?? undefined,
           explanation: flagInfo.explanation ?? '',
-          unconfiguredBehavesLike: flagInfo.unconfiguredBehavesLike,
-        }));
+        };
+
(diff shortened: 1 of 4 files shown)


packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -415,9 +384,9 @@ function getFlagSortOrder(flag: FeatureFlag): number {
 }
 
 async function displayFlagTable(flags: FeatureFlag[], ioHelper: IoHelper): Promise<void> {
-  const headers = ['Feature Flag Name', 'Recommended Value', 'User Value'];
+  const filteredFlags = flags.filter(flag => flag.unconfiguredBehavesLike?.v2 !== flag.recommendedValue);
```
rix0rrr, line 397:
```suggestion
  const filteredFlags = flags.filter(flag => (flag.unconfiguredBehavesLike?.v2 ?? false) !== flag.recommendedValue);
```

Perhaps this normalization should be done in the toolkit, so consumers don't have to worry about it all the time.

rix0rrr (changes requested):
Small comment.

Also, these 2 changes should *probably* be different PRs (though I'm guilty of combining changes into a single PR all the time as well 😈)

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1276,29 +1276,46 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
           explanation: flagInfo.explanation ?? '',
-          unconfiguredBehavesLike: flagInfo.unconfiguredBehavesLike,
-        }));
+        };
+
+        if (shouldIncludeUnconfiguredBehavesLike) {
+          return {
+            ...baseFlag,
+            unconfiguredBehavesLike: {
+              v2: flagInfo.unconfiguredBehavesLike?.v2 === true ? true : false,
```
rix0rrr: This assumes the value is always a boolean, which is not necessarily true.

```suggestion
              v2: flagInfo.unconfiguredBehavesLike?.v2 ?? false,
```

Have a look at what the `??` operator does.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -176,9 +176,23 @@ export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, o
...
+ * @returns true if --default options can be run
+ */
+async function checkDefaultBehavior(params: FlagOperationsParams) {
+  const { flagData, ioHelper } = params;
+  if (flagData[0].unconfiguredBehavesLike) {
     await setMultipleFlags(params);
     return;
   }
+  await ioHelper.defaults.error('The --default options are not compatible with the AWS CDK library used by your application. Please upgrade to 2.212.0 or above.');
+  return;
```
rix0rrr: `return` at the end of a function?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -176,9 +176,23 @@ export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, o
...
+    await checkDefaultBehavior(params);
+  }
+}
+
+/**
+ * Checks if the `unconfiguredBehavesLike` field is populated
+ *
+ * @returns true if --default options can be run
+ */
+async function checkDefaultBehavior(params: FlagOperationsParams) {
```
rix0rrr: I'd rather you either have this return a `boolean` and put the `if` into the calling code, or rename this to something like `setMultipleFlagsIfSupported`.

`checkDefaultBehavior` doesn't scream to me "do a thing if some precondition is satisfied".

Also "check" is a very generic word you should probably avoid unless it is *extremely* clear what is being checked.

rix0rrr (changes requested):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1276,29 +1276,46 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
    * Retrieve feature flag information from the cloud assembly
    */
-
   public async flags(cx: ICloudAssemblySource): Promise<FeatureFlag[]> {
     this.requireUnstableFeature('flags');
 
     const ioHelper = asIoHelper(this.ioHost, 'flags');
     await using assembly = await assemblyFromSource(ioHelper, cx);
-    const artifacts = assembly.cloudAssembly.manifest.artifacts;
+    const artifacts = Object.values(assembly.cloudAssembly.manifest.artifacts!);
```
rix0rrr:
```suggestion
    const artifacts = Object.values(assembly.cloudAssembly.manifest.artifacts ?? {});
```

rix0rrr (approved):

PR #798, opened 2025-08-14
feat(cli): allow users to enable all feature flags that do not impact their application

vivian12300:
Allows users to run `cdk flags --safe` to set all unconfigured feature flags to their recommended states without changing their CloudFormation template. 

---


packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -127,6 +127,8 @@ export async function makeConfig(): Promise<CliConfig> {
           recommended: { type: 'boolean', desc: 'Change flags to recommended states', requiresArg: false },
           default: { type: 'boolean', desc: 'Change flags to default state', requiresArg: false },
           interactive: { type: 'boolean', alias: ['i'], desc: 'Interactive option for the flags command', requiresArg: false },
+          safe: { type: 'boolean', desc: 'Enable all feature flags that do not impact the user\'s application', requiresArg: false },
+          concurrency: { type: 'number', desc: 'Maximum number of simultaneous synths to execute.', default: 4, requiresArg: true },
```
rix0rrr: Give this a short alias as well.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -233,6 +243,207 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+
+async function setSafeFlags(params: FlagOperationsParams): Promise<void> {
+  const { flagData, toolkit, ioHelper, concurrency } = params;
+
+  const cdkJson = await JSON.parse(await fs.readFile(path.join(process.cwd(), 'cdk.json'), 'utf-8'));
+  const app = cdkJson.app;
+
+  const isUsingTsNode = app.includes('ts-node');
+  if (isUsingTsNode && !app.includes('-T') && !app.includes('--transpileOnly')) {
+    await ioHelper.defaults.info('You are currently running with ts-node. Adding --transpileOnly may make this operation faster.');
```
rix0rrr:
```suggestion
    await ioHelper.defaults.info('Repeated synths with ts-node will type-check the application on every synth. Add --transpileOnly to cdk.json\'s "app" command to make this operation faster.');
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -233,6 +243,207 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  }
+
+  const unconfiguredFlags = flagData.filter(flag =>
+    flag.userValue === undefined &&
+    isBooleanFlag(flag) &&
+    (flag.unconfiguredBehavesLike?.v2 !== flag.recommendedValue),
```
rix0rrr:
```suggestion
    ((flag.unconfiguredBehavesLike?.v2 ?? false) !== flag.recommendedValue),
```

All the more reason to get this from a single place.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -233,6 +243,207 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  }
+
+  return isolateUnsafeFlags(flags, baseContextValues, toolkit, app, allStacks, queue);
+}
+
+async function setSafeFlags(params: FlagOperationsParams): Promise<void> {
+  const { flagData, toolkit, ioHelper, concurrency } = params;
+
+  const cdkJson = await JSON.parse(await fs.readFile(path.join(process.cwd(), 'cdk.json'), 'utf-8'));
+  const app = cdkJson.app;
```
rix0rrr: Usually, the `app` may also come from the command line, being passed as `--app`. We should probably support that as well.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -233,6 +243,207 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
   }
 }
 
+async function testFlagSafety(
```
rix0rrr: I think as a rule, I'd like to you reverse the order of your function definitions in source code.

Right now you're writing your code like this:

```ts
function a() {
}

function b() {
  a();
}
```

I see what you're doing. You're doing "*b depends on a, so a has to come before b*".

This is true outside of a function definition in some execute-as-you environments. It's not true if the caller is itself a function, and it's not true in general in JavaScript. You can put `a` and `b` in any order, and it will be fine.

Given that, I'd rather you optimize for *order of understanding*.

Since humans read top-to-bottom, I first see `a`. I know need to think: "What is this for? What is it doing? Where is it used?". I have little context: all I have to go on is the name of the function. You didn't even put in any documentation (because apparently it is self-evident what this function is for? 😜)

On the other hand, if you put the functions in reverse order, show me `a()` first and then `b()`, I'll have a much easier time parsing: presumably `a()` is the "main function" of this module, or it will be very obvious what it's doing or why. `b()` is just there to help out `a()`, but by the time I get to `b()` I will already know how it's being called and what `a()` is expecting from it. 

It will make it much easier to read the code that way.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -127,6 +127,8 @@ export async function makeConfig(): Promise<CliConfig> {
           recommended: { type: 'boolean', desc: 'Change flags to recommended states', requiresArg: false },
           default: { type: 'boolean', desc: 'Change flags to default state', requiresArg: false },
           interactive: { type: 'boolean', alias: ['i'], desc: 'Interactive option for the flags command', requiresArg: false },
+          safe: { type: 'boolean', desc: 'Enable all feature flags that do not impact the user\'s application', requiresArg: false },
+          concurrency: { type: 'number', alias: ['conc'], desc: 'Maximum number of simultaneous synths to execute.', default: 4, requiresArg: true },
```
rix0rrr: The alias should be single letter.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -40,9 +41,14 @@ interface FlagOperationsParams {
...
   unconfigured?: boolean;
+  safe?: boolean;
+  concurrency?: number;
+
+  /** User provided --app option */
+  app?: string;
 }
 
-export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, options: FlagsOptions, toolkit: Toolkit) {
+export async function handleFlags(flagData: FeatureFlag[], ioHelper: IoHelper, options: FlagsOptions & { app?: string }, toolkit: Toolkit) {
```
rix0rrr: It's nicer to define a structure with its own name.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -40,9 +41,14 @@ interface FlagOperationsParams {
 
   /** User ran --unconfigured option */
   unconfigured?: boolean;
+  safe?: boolean;
+  concurrency?: number;
+
+  /** User provided --app option */
```
rix0rrr: Don't forget to explain the default behavior.

And the other new fields need docstrings too.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  }
+
+  const unconfiguredFlags = flagData.filter(flag =>
+    flag.userValue === undefined &&
+    isBooleanFlag(flag) &&
+    (flag.unconfiguredBehavesLike?.v2 !== flag.recommendedValue),
+  );
+
+  if (unconfiguredFlags.length === 0) {
+    await ioHelper.defaults.info('No unconfigured feature flags found.');
```
rix0rrr: Let's make this a positive message.

```suggestion
    await ioHelper.defaults.info('All feature flags are configured.');
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+
+  if (safeFlags.length > 0) {
+    await ioHelper.defaults.info('Safe flags that can be set without template changes:');
+    for (const flag of safeFlags) {
+      await ioHelper.defaults.info(`- ${flag.name} -> ${flag.recommendedValue}`);
+    }
+
+    await handleUserResponse(params, safeFlags.map(flag => flag.name));
+  } else {
+    await ioHelper.defaults.info('No flags can be safely set without causing template changes.');
```
rix0rrr:
```suggestion
    await ioHelper.defaults.info('No more flags can be set without causing template changes.');
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  const allStacks = baseAssembly.stacksRecursively;
+
+  const queue = new PQueue({ concurrency: concurrency });
+
+  const safeFlags = await batchTestFlags(unconfiguredFlags, baseContextValues, toolkit, app, allStacks, queue);
+
+  await fs.remove(path.join(process.cwd(), 'baseline'));
+
+  if (safeFlags.length > 0) {
+    await ioHelper.defaults.info('Safe flags that can be set without template changes:');
```
rix0rrr:
```suggestion
    await ioHelper.defaults.info('Flags that can be set without template changes:');
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+      await ioHelper.defaults.info(`- ${flag.name} -> ${flag.recommendedValue}`);
+    }
+
+    await handleUserResponse(params, safeFlags.map(flag => flag.name));
+  } else {
+    await ioHelper.defaults.info('No flags can be safely set without causing template changes.');
+  }
+}
+
+async function batchTestFlags(
```
rix0rrr: Docstrings, please :)

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  allStacks: any[],
+  queue: PQueue,
+): Promise<FeatureFlag[]> {
+  if (flags.length === 0) {
+    return [];
+  }
+
+  const allFlagsContext = { ...baseContextValues };
+  flags.forEach(flag => {
+    allFlagsContext[flag.name] = toBooleanValue(flag.recommendedValue);
```
rix0rrr: If we need to do type coercion, could we not have done that already in the toolkit lib where we read these flags?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+
+  const allSafe = await testBatch(allFlagsContext, toolkit, app, allStacks, 'batch-all');
+  if (allSafe) {
+    return flags;
+  }
+
+  return isolateUnsafeFlags(flags, baseContextValues, toolkit, app, allStacks, queue);
+}
+
+async function testBatch(
```
rix0rrr: Docstring please.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  contextValues: Record<string, any>,
+  toolkit: Toolkit,
+  app: string,
+  allStacks: any[],
```
rix0rrr:
```ts
{
  allStacks: any[],
//           ^^^^^ 😱
}
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+      method: DiffMethod.LocalFile(templatePath),
+      stacks: {
+        strategy: StackSelectionStrategy.PATTERN_MUST_MATCH_SINGLE,
+        patterns: [stack.hierarchicalId],
+      },
+    });
+
+    for (const stackDiff of Object.values(diff)) {
+      if (stackDiff.differenceCount > 0) {
+        await fs.remove(path.join(process.cwd(), outdir));
```
rix0rrr: You can use a `try/finally`, so you don't have to duplicate the `fs.remove` call.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  contextValues: Record<string, any>,
+  toolkit: Toolkit,
+  app: string,
+  allStacks: any[],
+  outdir: string,
+): Promise<boolean> {
+  const testContext = new MemoryContext(contextValues);
+  const testSource = await toolkit.fromCdkApp(app, {
+    contextStore: testContext,
+    outdir: path.join(process.cwd(), outdir),
```
rix0rrr: I'd rather have `outdir` be a full path to a temporary directory, rather than create a subdirectory in the working directory.

The working directory will probably be a user's source tree, let's try not to pollute it.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+    batch.forEach(flag => {
+      batchContext[flag.name] = toBooleanValue(flag.recommendedValue);
+    });
+
+    const isSafeBatch = await testBatch(
+      batchContext,
+      toolkit,
+      app,
+      allStacks,
+      `batch-${Date.now()}-${Math.random()}`,
```
rix0rrr: This name is not meaningful, so it could be decided by the `testBatch` function. No need to pass it here.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+    const batchContext = { ...contextValues };
+    batch.forEach(flag => {
+      batchContext[flag.name] = toBooleanValue(flag.recommendedValue);
+    });
+
+    const isSafeBatch = await testBatch(
+      batchContext,
+      toolkit,
+      app,
+      allStacks,
```
rix0rrr: There are a lot of values being passed around between all functions here. They could be shared on a class, so you don't have to pass them. 

Classes don't necessarily have to represent "things". It is perfectly valid (and useful!) for a class to present an "operation", where the operation is split into multiple functions and some of the state that all functions need to access are members of the class. For example:

```ts
class DetermineSafeFlags {
  private readonly allStacks: Stack[];
  private readonly appCommand: string;
  private readonly toolkit: Toolkit;
  private readonly originalContext: Record<string, unknown>;

  public determine(): Record<string, unknown> { 
    // ...
  }

  private helper1() { 
   // ..
  }
 
  private helper2() {
    // ...
   }
}
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  }
+
+  void queue.add(() => processBatch(flags, baseContextValues));
+
+  await queue.onIdle();
+
+  return safeFlags;
+}
+
+async function testFlagSafety(
```
rix0rrr: This function looks a lot like `batchTestFlags` if I'm seeing it correctly, except it's for a single flag instead of a batch?

How about a batch of size 1? Won't that do the same?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -239,6 +253,208 @@ async function setFlag(params: FlagOperationsParams, interactive?: boolean) {
...
+  app: string,
+  allStacks: any[],
+): Promise<boolean> {
+  const testContext = new MemoryContext(baseContextValues);
+  const newValue = toBooleanValue(flag.recommendedValue);
+  await testContext.update({ [flag.name]: newValue });
+
+  const testSource = await toolkit.fromCdkApp(app, {
+    contextStore: testContext,
+    outdir: path.join(process.cwd(), `test-${flag.name}`),
```
rix0rrr: Looks like this directory is not being cleaned up?

rix0rrr (changes requested):

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -126,7 +126,9 @@ export async function makeConfig(): Promise<CliConfig> {
           unconfigured: { type: 'boolean', desc: 'Modify unconfigured feature flags', requiresArg: false },
           recommended: { type: 'boolean', desc: 'Change flags to recommended states', requiresArg: false },
           default: { type: 'boolean', desc: 'Change flags to default state', requiresArg: false },
-          interactive: { type: 'boolean', alias: ['i'], desc: 'Interactive option for the flags command', requiresArg: false },
+          interactive: { type: 'boolean', alias: ['i'], desc: 'Interactive option for the flags command' },
+          safe: { type: 'boolean', desc: 'Enable all feature flags that do not impact the user\'s application', requiresArg: false },
+          concurrency: { type: 'number', alias: ['t'], desc: 'Maximum number of simultaneous synths to execute.', default: 4, requiresArg: true },
```
rix0rrr: Why `t`? That letter is not in "concurrency" at all :). Did you copy this from somewhere where we also use `t` for concurrency?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -40,477 +47,684 @@ interface FlagOperationsParams {
...
+
+export class DetermineSafeFlags {
+  private readonly params: FlagOperationsParams;
+
+  constructor(flagData: FeatureFlag[], ioHelper: IoHelper, options: HandleFlagsOptions, toolkit: Toolkit) {
+    this.params = {
+      flagData: flagData.filter(flag => !OBSOLETE_FLAGS.includes(flag.name)),
+      toolkit,
+      ioHelper,
+      recommended: options.recommended,
```
rix0rrr: Couldn't the members of this class be:

```ts
private readonly flags: FeatureFlag[];
private readonly toolkit: Toolkit;
private readonly ioHelper: IoHelper;
private readonly options: Options;
```

Rather than combine everything into a single struct?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -40,477 +47,684 @@ interface FlagOperationsParams {
...
-  if (options.interactive) {
-    const prompt = new Select({
-      name: 'option',
-      message: 'Menu',
-      choices: interactiveOptions,
-    });
+  /**
+   * Main entry point for handling flag operations based on user options.
+   */
+  public async handleFlags(): Promise<void> {
```
rix0rrr: "handle" is another extremely generic word. Can you be more specific?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -40,477 +47,684 @@ interface FlagOperationsParams {
...
+   * Tests all flags together first, then isolates unsafe flags using binary search if needed.
+   *
+   * @returns array of flags that can be set to recommended values without template changes
+   */
+  private async batchTestFlags(
+    flags: FeatureFlag[],
+    baseContextValues: Record<string, any>,
+    app: string,
+    allStacks: CloudFormationStackArtifact[],
+    queue: PQueue,
```
rix0rrr: If you still need to pass this many function arguments (and apparently these cannot be converted to class members) did we choose the scope of the class narrowly enough?

rix0rrr:
This PR has picked up changes from `main` in an undesirable way that's making it hard for me to review, and I now wonder if this PR and the other one you had open are both changing how the toolkit returns flags, and whether that's not going to cause problems. 

Merge it with `main` first (by hand) and then I'll review.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -467,7 +467,8 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           unstableFeatures: configuration.settings.get(['unstable']),
         });
         const flagsData = await toolkit.flags(cloudExecutable);
-        return handleFlags(flagsData, ioHelper, args, toolkit);
+        const handler = new DetermineSafeFlags(flagsData, ioHelper, args, toolkit);
+        return handler.executeFlagOperations();
```
rix0rrr: "execute flag operations" is still extremely generic. Why not just

```suggestion
        return handler.determineSafeFlags();
```

packages/aws-cdk/lib/cli/cli-type-registry.json
```diff
@@ -383,8 +383,21 @@
...
+        },
+        "safe": {
+          "type": "boolean",
+          "desc": "Enable all feature flags that do not impact the user's application",
           "requiresArg": false
+        },
+        "concurrency": {
+          "type": "number",
+          "alias": [
+            "t"
```
rix0rrr: Still a `t` instead of a `j`?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -41,456 +44,643 @@ interface FlagOperationsParams {
...
+
+export class DetermineSafeFlags {
+  private readonly params: FlagOperationsParams;
+  private readonly flags: FeatureFlag[];
+  private readonly toolkit: Toolkit;
+  private readonly ioHelper: IoHelper;
+  private app!: string;
+  private baseContextValues!: Record<string, any>;
+  private allStacks!: CloudFormationStackArtifact[];
+  private queue!: PQueue;
```
rix0rrr: These `!`s are dangerous.

Can you not initialize these in the constructor? (If not, perhaps the scope of the class is still too large)

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -41,456 +44,643 @@ interface FlagOperationsParams {
...
-        default: true,
-        unconfigured: true,
-      };
-      await setMultipleFlagsIfSupported(params);
-    } else if (answer == FlagsMenuOptions.MODIFY_SPECIFIC_FLAG) {
-      await setFlag(params, true);
-    } else if (answer == FlagsMenuOptions.EXIT) {
+    const interactiveOptions = Object.values(FlagsMenuOptions);
+
+    if (this.params.interactive == true) {
```
rix0rrr: Oh I see. If this class does *ALL* flag operations, then the name is wrong, isn't it? `DetermineSafeFlags` is not really an appropriate name for a class that does `AllFlagOperations`.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -41,456 +44,643 @@ interface FlagOperationsParams {
...
+      const prompt = new Select({
+        name: 'option',
+        message: 'Menu',
+        choices: interactiveOptions,
+      });
+
+      const answer = await prompt.run();
+      if (answer == FlagsMenuOptions.ALL_TO_RECOMMENDED) {
```
rix0rrr: Make this a `switch`, it's nicer.

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -41,456 +44,643 @@ interface FlagOperationsParams {
...
+        this.params.recommended = true;
```
rix0rrr: This is passing values to the inner function by side effect. I don't like it.

packages/aws-cdk/lib/commands/flag-operations.ts
rix0rrr: Personally I would keep the interactive menu out of the class. The class can provide helpers though.

How about structuring it something like this:

```ts
class FlagReport { } 
class StackFlagOperations { }


function flagsMenu() {
  const ops = new StackFlagOperations(/* whatever is necessary here to make the below work */);

  const answer = await prompt();
  let flagsToSet;
  switch (answer) {
    case FlagsMenuOptions.ALL_TO_RECOMMENDED:
      flagsToSet = report.allRecommended();
      break;
    case FlagsMenuOptions. UNCONFIGURED_TO_RECOMMENDED:
      flagsToSet = report.unconfiguredRecommended();
      break;
    case FlagsMenuOptions.UNCONFIGURED_TO_DEFAULT:
      flagsToSet = report.unconfiguredDefault(); // unconfiguredDefault can throw it if needs to
      break;
    case FlagsMenuOptions. MODIFY_SPECIFIC_FLAG:
      flagsToSet = await promptForSpecificFlag();
      break;
  }

  if (params.safe) {
    flagsToSet = await ops.restrictToSafeFlags(flagsToSet);
  }

  await ops.set(flagsToSet);
}
```

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -41,456 +44,643 @@ interface FlagOperationsParams {
...
     }
-    return;
-  }
 
-  if (options.FLAGNAME && options.all) {
-    await ioHelper.defaults.error('Error: Cannot use both --all and a specific flag name. Please use either --all to show all flags or specify a single flag name.');
-    return;
-  }
+    if (this.params.safe) {
+      await this.setSafeFlags();
```
rix0rrr: Wait we did all the sets already in the big `if/else` chain above. Isn't this too late?

packages/aws-cdk/lib/commands/flag-operations.ts
```diff
@@ -41,456 +44,643 @@ interface FlagOperationsParams {
...
+
+  /**
+   * Sets a single flag value, either interactively or from provided parameters.
+   */
+  private async setFlag(interactive?: boolean) {
+    const { flagName } = this.params;
+    let updatedFlagName = flagName;
+    let updatedValue = this.params.value;
+
+    if (interactive) {
```
rix0rrr: But really, the "interactive mode" code should be a completely different set of functions. Interactive mode and noninteractive mode can't be interleaved like this:

```ts
if (interactive) {
  // A
} else {
  // B
}
```

Is very confusing.

Much easier instead, a single `if` distinguishes between the 2 modes of operation, which do their own modal input validation and finally forward to the same "actually do it" function:

```ts
function entryPoint(params) {
  if (interactive) { 
    await doInteractiveMenu(params);
  } else {
    await doCliOperation(params);
  }
}

function doInteractiveMenu() {
  // Interactive programs usually keep on running until the user explicitly quits them.
  // Also put a try/catch here to show errors to the user without quitting the program.
  while (true) {
    try {
      const answer = await prompt();
  
      switch (answer) {
        case SINGLE_FLAG:
          await doInteractiveSingleFlag();
          break;
       
        // ...

        case STOP:
          return;
        
    } catch(e) {
      // Print error and go around again
      cliHost.error(e);
    }
  }
}  

function doInteractiveSingleFlag() { 
  // Another loop so it's easy if the user wants to do multiple single flags one after the other
  while (true) {
    const answer = await promptSingleFlag();

    if (!answer) {
      // User explicitly escapes the loop by picking nothing, exit, etc.
      break;
    }
    
    await setSingleFlag(answer.flagName, answer.flagValue);
  }
}

function doCliOperation() {
  if (options.set) {
    await setSingleFlag(options.flagName, options.flagValue);
  } else {
    throw new Error('Invalid combination of options');
  }
}

function setSingleFlag(...) {
   // the actual functionality, potentially with some validations
   if (badStuff) { 
     throw new CliError('oh no');
   }
}
```

rix0rrr (changes requested):

packages/aws-cdk/lib/commands/flags/interactive-handler.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+
+  /** Checks if user value matches recommended value */
+  private isUserValueEqualToRecommended(flag: FeatureFlag): boolean {
+    return String(flag.userValue) === String(flag.recommendedValue);
+  }
+
+  /** Main interactive mode handler that shows menu and processes user selection */
+  async handleInteractiveMode(): Promise<FlagOperationsParams | null> {
```
rix0rrr, line 30: "Handler" and "handle" ?

I think some more specific terminology we could use is "prompt user for command".

rix0rrr (commented):
I like it! There's probably some more simplicity to be had, but this is a lot easier for us to continue work on. Cheers!

packages/aws-cdk/lib/commands/flags/operations.ts
```diff
@@ -0,0 +1,480 @@
...
+  async execute(params: FlagOperationsParams): Promise<void> {
+    if (params.set) {
+      if (params.FLAGNAME && params.value) {
+        await this.setFlag(params);
+      } else {
+        await this.setMultipleFlags(params);
+      }
+    } else {
+      await this.displayFlags(params);
+    }
```
rix0rrr, line 42: I think this `if` should probably be in the router.

packages/aws-cdk/lib/commands/flags/operations.ts
```diff
@@ -0,0 +1,480 @@
...
+    } else {
+      return this.flags
+        .filter(flag => flag.userValue === undefined)
+        .filter(flag => this.isBooleanFlag(flag))
+        .map(flag => flag.name);
+    }
+  }
+
+  /** Sets flags that don't cause template changes */
+  async setSafeFlags(params: FlagOperationsParams): Promise<void> {
```
rix0rrr, line 101: Separate out "determining what flags are safe" from "setting flags".

packages/aws-cdk/lib/commands/flags/operations.ts
```diff
@@ -0,0 +1,480 @@
...
+          ? flag.recommendedValue as boolean
+          : String(flag.unconfiguredBehavesLike?.v2) === 'true';
+        updateObj[flagName] = newValue;
+      }
+    }
+    return updateObj;
+  }
+
+  /** Prompts user for confirmation and applies changes if accepted */
+  private async handleUserResponse(flagNames: string[], params: FlagOperationsParams): Promise<void> {
```
rix0rrr, line 328: Wouldn't this be more appropriate inside the router or the FlagHandler?

packages/aws-cdk/lib/commands/flags/operations.ts
```diff
@@ -0,0 +1,480 @@
...
+      recommended === 'false';
+  }
+
+  /** Checks if the user's current value matches the recommended value */
+  private isUserValueEqualToRecommended(flag: FeatureFlag): boolean {
+    return String(flag.userValue) === String(flag.recommendedValue);
+  }
+
+  /** Shows helpful usage examples and available command options */
+  async displayHelpMessage(): Promise<void> {
```
rix0rrr, line 477: Wait, is this still used? You should get this for free when doing `cdk flags --help`.

rix0rrr (approved):
Some small notes. I'll give you provisional approval. Have a last consideration for some of the things I said, otherwise go ahead and merge it.

packages/aws-cdk/lib/cli/cli-type-registry.json
vivian12300: I saw that `j` is actually used for the global `json` option so I decided to switch to `n` (which users can associate with number).

PR #807, opened 2025-08-19
feat(cli): add git repository support with --from-git-url for single template repositories

rohang9000:
Added --from-git-url option to pull a custom template from a Git repository. This option supports scenarios where repository contains a single template which is represented by language directories for it at the repository root. The option auto-detects language for single-language template (one language directory at repo root) and requires --language flag for a multi-language template (multiple CDK supported language directories at repo root). This new option is gated with --unstable=init flag. Wrote tests for this functionality.

Fixes #

---

### DIFF
--- a/packages/@aws-cdk/user-input-gen/lib/yargs-gen.ts
+++ b/packages/@aws-cdk/user-input-gen/lib/yargs-gen.ts
@@ -84,8 +84,6 @@ function makeYargs(config: CliConfig, helpers: CliHelpers): Statement {
   // we must compute global options first, as they are not part of an argument to a command call
   yargsExpr = makeOptions(yargsExpr, config.globalOptions, helpers);
 
-  let checkFunctions: string[] = [];
-
   for (const command of Object.keys(config.commands)) {
     const commandFacts = config.commands[command];
     const commandArg = commandFacts.arg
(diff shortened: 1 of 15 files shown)


packages/@aws-cdk/user-input-gen/lib/yargs-gen.ts
```diff
@@ -109,6 +109,13 @@ function makeYargs(config: CliConfig, helpers: CliHelpers): Statement {
       commandCallArgs.push(optionsExpr);
     }
 
+    // Add implies calls if present
```
kaizencc, line 112: this needs to be tested within the `user-input-gen` package.

packages/aws-cdk/lib/cli/parse-command-line-arguments.ts
```diff
@@ -823,7 +823,27 @@ export function parseCommandLineArguments(args: Array<string>): any {
...
+          type: 'string',
+          desc: 'Git repository URL to clone and use as template source',
+          requiresArg: true,
+          conflicts: ['lib-version', 'from-path'],
+        })
+        .option('template-path', {
+          default: undefined,
+          type: 'string',
+          desc: 'Path to a specific template within a multi-template repository (requires --from-path)',
+          requiresArg: true,
```
kaizencc, line 846: i am expecting to see `implies` here but i do not.

kaizencc: would also like to see a test or some confirmation that you cannot supply `template-path` without `from-path`.

kaizencc (commented):
there are a lot of changes here and i suspect that is because it is based on #763. if so, you should use that branch as the base branch so that the diff of this PR doesn't overlap with the diff there.

will there be readme updates or documentation to show how to use this feature?

PR #811, opened 2025-08-21
fix(bootstrap): disallow AssumeRole with ExternalId by default

rix0rrr:
By default, CDK Bootstrap roles are not designed to be deputized.

(Deputized means that you give an external entity access to assume roles on your behalf. They will supply an ExternalId to avoid [Confused Deputy attacks](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html))

If a deputy system (i.e., a system that assumes IAM Roles on behalf of its tenants) is using CDK and its policies are not configured carefully, it can be tricked into assuming its own CDK roles.

Because CDK Roles are not intended to be used in this way, we are adding a default security control that will make this misconfiguration less likely: AssumeRole calls with ExternalIds will be denied by default.

What if I do want to use ExternalIds?
-------------------------------------

If you are currently passing `ExternalId`s in an `AssumeRole` call to CDK bootstrap roles *inside your own trusted organization* (expecting the ExternalId to be present but ignored), this protection can be disabled by calling:

```
$ cdk bootstrap --no-deny-external-id
```

If you want to give permissions for other organizations to assume your CDK bootstrap roles in a deputized way, customize the bootstrap template and add a proper `ExternalId` condition.

---


packages/@aws-cdk-testing/cli-integ/resources/bootstrap-templates/session-tags.all-roles-deny-all.yaml
```diff
@@ -56,6 +56,11 @@ Parameters:
     Description: Describe the provenance of the resources in this bootstrap
       stack. Change this when you customize the template. To prevent accidents,
       the CDK CLI will not overwrite bootstrap stacks with a different variant.
+  DenyExternalId:
+    Type: String
+    Default: 'true'
+    AllowedValues: ['true', 'false']
+    Description: Just here to make the test pass.
```
iankhou: Can we add a non-placeholder description here? I see good descriptions in other files like packages/aws-cdk/lib/cli/parse-command-line-arguments.ts

packages/@aws-cdk-testing/cli-integ/resources/bootstrap-templates/session-tags.deploy-role-deny-sqs.yaml
```diff
@@ -56,6 +56,11 @@ Parameters:
     Description: Describe the provenance of the resources in this bootstrap
       stack. Change this when you customize the template. To prevent accidents,
       the CDK CLI will not overwrite bootstrap stacks with a different variant.
+  DenyExternalId:
+    Type: String
+    Default: 'true'
+    AllowedValues: ['true', 'false']
+    Description: Just here to make the test pass.
```
iankhou: Same comment as above for using a non-placeholder description.

packages/aws-cdk/lib/cli/user-input.ts
```diff
@@ -461,6 +461,13 @@ export interface BootstrapOptions {
    */
   readonly publicAccessBlockConfiguration?: boolean;
 
+  /**
+   * Block AssumeRole access to all boostrapped roles if an ExternalId is provided (enabled by default)
```
iankhou, line 465:
```suggestion
   * Block AssumeRole access to all bootstrapped roles if an ExternalId is provided (enabled by default)
```

iankhou (approved):
Just a typo and some boilerplate left behind. Also, this is a breaking change, so PR title should reflect that right?

rix0rrr:
What do you mean this is a breaking change? W.r.t existing customized templates you mean?

packages/@aws-cdk-testing/cli-integ/resources/bootstrap-templates/session-tags.all-roles-deny-all.yaml
rix0rrr: But what would I put here?

packages/@aws-cdk/toolkit-lib/lib/api/bootstrap/bootstrap-environment.ts
```diff
@@ -213,27 +213,37 @@ export class Bootstrapper {
...
+
+    // Conditionally set these parameters: only set these parameters if they are accepted by the template.
+    // If we pass them unconditionally, older customized templates that don't know about these
+    // parameters yet will fail to deploy.
+    if (params.denyExternalId !== undefined) {
+      if (!templateParameters.includes('DenyExternalId')) {
+        throw new ToolkitError('The selected bootstrap template does not accept the DenyExternalId parameter');
+      }
+      bootstrapTemplateParameters.DenyExternalId = `${params.denyExternalId}`;
+    }
```
mrgrain, line 241: Q: Should this just be a warning?
Q: Should we apply this to EVERY paramter?

rix0rrr: A1: Perhaps. But we're making a clear distinction between passing the value and not passing it. If you pass the value, I don't think we'd want to drop it with only a warning.

A2: Perhaps. Hasn't caused a problem so far, so I'm not going to be holier than the pope without any customer pressure.

PR #814, opened 2025-08-22
feat(cli): support shorthand aliases for languages

### LINKED ISSUE #748: `cdk init`: should support shorthand aliases for languages
### Describe the feature

Today:

```
cdk init app -l typescript
cdk init app -l javascript
cdk init app -l csharp
cdk init app -l fsharp
cdk init app -l python
```

### Use Case

Very annoying to type

### Proposed Solution

Future:

```
cdk init app -l ts
cdk init app -l js
cdk init app -l cs
cdk init app -l fs
cdk init app -l py
```

### Other Information

`java` and `go` excluded because already short. I guess I am proposing we also accept the typically file ending as a language alias.

### Acknowledgements

### CDK version used

any

any

ren-yamanashi:
Fixes #748 

## Description Changes

Extended the `--language` option values for the `cdk init` and `cdk migrate` commands to support shorthand aliases for language like `ts` and `js`.

---


packages/aws-cdk/lib/cli/parse-command-line-arguments.ts
```diff
@@ -846,7 +847,7 @@ export function parseCommandLineArguments(args: Array<string>): any {
           type: 'string',
           alias: 'l',
           desc: 'The language to be used for the new project (default can be configured in ~/.cdk.json)',
-          choices: ['csharp', 'fsharp', 'go', 'java', 'javascript', 'python', 'typescript'],
+          choices: Object.keys(LANGUAGES)
```
ren-yamanashi, line 850: While it's conceivable to implement a solution using yargs' coerce method to transform option values (like ts -> typescript), this approach was not taken because it would have a destructive impact on error message content when invalid input is provided.

see: https://github.com/yargs/yargs/issues/1379

iankhou (changes requested):
Nice improvement for usability. Can we add a negative test case for invalid languages/aliases for init and migrate?

ren-yamanashi:
> Nice improvement for usability. Can we add a negative test case for invalid languages/aliases for init and migrate?

@iankhou 

Thanks for comments

I've added some negative test cases as suggested.

However, I discovered that my previous implementation approach was flawed - I was manually modifying generated code that should remain untouched.

So, I've made substantial changes throughout the codebase, not just in the test files.

iankhou (approved):

ren-yamanashi:
@iankhou 

Thank you for the approval!

The build was failing, so I've made the necessary fixes.  
Could you please approve it again?

---

commits:

https://github.com/aws/aws-cdk-cli/pull/814/commits/da364b178bfad91009717b02e4eb035269c7a2a6

https://github.com/aws/aws-cdk-cli/pull/814/commits/6f884a65f5eb1479741616db18014682be7ec880

PR #816, opened 2025-08-26
feat(cli): add --no-wait option to destroy command

perrin4869:
The motivation behind this option is to make deleting test stacks via the CI more efficient
In particular, stacks which contain lambdas attached to a VPC can take over 20 minutes to remove. This feature will improve CI run times significantly in those cases.
In cases where a stack has dependencies, this option will not be applied

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
@@ -14,8 +14,7 @@ export interface DestroyOptions {
   readonly roleArn?: string;
 
   /**
-   * Whether or not to wait for the stack to finish deleting
-   * This functionality does not work for stacks which have dependent stacks
+   * Skip waiting for the stack to finish deleting
    */
-  readonly noWait?: boolean;
+  readonly skipAwaitDeletion?: boolean;
 }
--- a/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
@@ -12,4 +12,10 @@ export interface DestroyOptions {
    * The arn of the IAM role to use for the stack destroy operation
    */
   readonly roleArn?: string;
+
+  /**
+   * Whether or not to wait for the stack to finish deleting
+   * This functionality does not work for stacks which have dependent stacks
+   */
+  readonly noWait?: boolean;
 }
--- a/packages/@aws-cdk/cloud-assembly-schema/lib/integ-tests/commands/destroy.ts
+++ b/packages/@aws-cdk/cloud-assembly-schema/lib/integ-tests/commands/destroy.ts
@@ -17,4 +17,12 @@ export interface DestroyOptions extends DefaultCdkOptions {
    * @default false
    */
   readonly exclusively?: boolean;
+
+  /**
+   * Whether or not to wait for the stack to finish deleting
+   * This functionality does not work for stacks which have dependent stacks
+   *
+   * @default false
+   */
+  readonly noWait?: boolean;
 }
--- a/packages/@aws-cdk/cloud-assembly-schema/lib/integ-tests/commands/destroy.ts
+++ b/packages/@aws-cdk/cloud-assembly-schema/lib/integ-tests/commands/destroy.ts
@@ -19,10 +19,9 @@ export interface DestroyOptions extends DefaultCdkOptions {
   readonly exclusively?: boolean;
 
   /**
-   * Whether or not to wait for the stack to finish deleting
-   * This functionality does not work for stacks which have dependent stacks
+   * Skip waiting for the stack to finish deleting
    *
    * @default false
    */
-  readonly noWait?: boolean;
+  readonly skipAwaitDeletion?: boolean;
 }
--- a/packages/@aws-cdk/cloud-assembly-schema/schema/integ.schema.json
+++ b/packages/@aws-cdk/cloud-assembly-schema/schema/integ.schema.json
@@ -347,8 +347,8 @@
                     "default": false,
                     "type": "boolean"
                 },
-                "noWait": {
-                    "description": "Whether or not to wait for the stack to finish deleting\nThis functionality does not work for stacks which have dependent stacks",
+                "skipAwaitDeletion": {
+                    "description": "Skip waiting for the stack to finish deleting",
                     "default": false,
                     "type": "boolean"
                 },
--- a/packages/@aws-cdk/cloud-assembly-schema/schema/integ.schema.json
+++ b/packages/@aws-cdk/cloud-assembly-schema/schema/integ.schema.json
@@ -347,6 +347,11 @@
                     "default": false,
                     "type": "boolean"
                 },
+                "noWait": {
+                    "description": "Whether or not to wait for the stack to finish deleting\nThis functionality does not work for stacks which have dependent stacks",
+                    "default": false,
+                    "type": "boolean"
+                },
                 "stacks": {
                     "description": "List of stacks to deploy\n\nRequried if `all` is not set (Default - [])",
                     "type": "array",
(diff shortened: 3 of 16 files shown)


iankhou:
Have you considered using async `destroy` for your CI use case? And if so, why not?

https://github.com/aws/aws-cdk-cli/blob/512894626b657ff86165be549e002cfd56bef2e6/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L972

I have concerns about exposing this as an option in a user-facing CLI, since it's not really meant for a CI system. What happens when destroy is run with --no-wait and a stack fails to be destroyed? This can cause serious issues when redeploying in a CI system.

Users who use it improperly without fully understanding the implications of not waiting might also run into race conditions or unexpectedly failing future deployments due to undeleted resources.

perrin4869:
> Have you considered using async destroy for your CI use case?
Well, even if I find a way to call that function, I would have no way to wait until it has sent the delete command without waiting for it to finish deleting?

> What happens when destroy is run with --no-wait and a stack fails to be destroyed?
Anyone using this option would need to regularly monitor their test stacks to ensure they were deleted and/or have confidence that deletion is not going to fail regularly. Since it is not enabled by default, anyone who finds out about this option would need to have sought it out by themselves

iankhou:
> > Have you considered using async destroy for your CI use case?
> 
> Well, even if I find a way to call that function, I would have no way to wait until it has sent the delete command without waiting for it to finish deleting?

Apologies, sent the wrong link. Should be here in our Programmatic Toolkit:
https://github.com/aws/aws-cdk-cli/blob/%40aws-cdk/toolkit-lib%40v1.6.0/packages/%40aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L1152

API reference: https://docs.aws.amazon.com/cdk/api/toolkit-lib/Package/toolkit-lib/Class/Toolkit/

perrin4869:
hm... as far as I can tell, both the toolkit and the CLI end up calling [`destroyStack`](https://github.com/aws/aws-cdk-cli/blob/5fc1b13b5977c45c054eee715b7ad41c472cddf6/packages/%40aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts#L705), which has no way to disable the waiting functionality.

The CLI calls it from [here](https://github.com/aws/aws-cdk-cli/blob/512894626b657ff86165be549e002cfd56bef2e6/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L992), the toolkit from [here](https://github.com/aws/aws-cdk-cli/blob/ad1a7ceff5e4895ba9b092938e74a42c4b7bae07/packages/%40aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L1194)

perrin4869:
My immediate motivation is to reduce CI run times and speed up deployments, but maybe someone would like to use this flag outside of a CI environment too, maybe because they want to fire the stack delete operation and monitor progress on the aws console?

packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
```diff
@@ -12,4 +12,10 @@ export interface DestroyOptions {
    * The arn of the IAM role to use for the stack destroy operation
    */
   readonly roleArn?: string;
+
+  /**
+   * Whether or not to wait for the stack to finish deleting
```
iankhou: Whether or not to wait...

When set to `true`, we do not wait, and when set to `false`, we do wait.

So this should say something like:

"*Skip* waiting for the stack to finish deleting"

to also match the style of descriptions in the other options.

iankhou (changes requested):
From our discussion, I see the value of adding the option. 

Currently, you would need to implement an async task in the Toolkit or use a hack in the CLI "like &" to skip waiting for stack deletion.

I would like to see a clear message in the output, though, to inform the user that subsequent actions may be performed before stacks are fully deleted.

iankhou:
> My immediate motivation is to reduce CI run times and speed up deployments, but maybe someone would like to use this flag outside of a CI environment too, maybe because they want to fire the stack delete operation and monitor progress on the aws console?

Specifically, I think using the CloudFormation CLI is more appropriate for these use cases. The `cdk destroy` command is ultimately an interface on top of `CFn`, and waits for completion for DX. Skipping the wait is a hacky thing to do to begin with, and introduces a lot of risk.

perrin4869:
Initially I considered this feature to only support cases where we are deleting a single stack. That would remove the need to have a `hasDependents` function. Would that make this feature more feasible?
I would ideally prefer to delete the stack using cdk destroy directly, without relying on the aws-cli cloudformation `delete-stack` command, just to be safe in case additional functionality is added in the future.
It's true that this is bespoke of my usecase, but it's also been an issue to many other people, for example: https://github.com/serverless/serverless/issues/5008 has had 143 thumbs up, and I expect other people are also deploying test stacks as part of their CI pipeline...

iankhou:
> Initially I considered this feature to only support cases where we are deleting a single stack. That would remove the need to have a `hasDependents` function. Would that make this feature more feasible? I would ideally prefer to delete the stack using cdk destroy directly, without relying on the aws-cli cloudformation `delete-stack` command, just to be safe in case additional functionality is added in the future.

What do you mean by additional functionality? Like functionality to the `aws-cli`? It seems like this would be a breaking change on CloudFormation's part. CloudFormation could also change their SDK to break us as well. Implementing a flag here to skip waiting for deletion doesn't remove the possibility of the feature being broken by an errant change in CloudFormation.

If we were to consider this flag, it would need to be a full feature that works on stacks with dependents, and thus a much bigger implementation.

However, `cdk destroy` is supposed to manage the entire lifecycle of stack destruction. Completing the command before we finish deleting the stack isn't just a feature on the command, it would fundamentally change what it does.


aws-cli cloudformation's `delete-stack` can be used to shortcut this functionality, but we do not want it in the CDK CLI. You can use `cdk ls` to get the inputs you need for it.

perrin4869:
> What do you mean by additional functionality?

By additional functionality, I mean that `cdk destroy` could for example, run additional cleanup of deployed assets as part of the `cdk destroy` command, and by bypassing `cdk destroy` and jumping straight to `aws cloudformation delete-stack`, this cleanup would not be executed

> The only implementation we can accept here is one that correctly takes inter-stack dependencies into account
well I tried to do that with the `hasDependents` function. If you can point me in the codebase where I can find the relevant code I'll definitely try to refactor this PR

mrgrain:
> `cdk destroy` could for example, run additional cleanup of deployed assets as part of the `cdk destroy` command, and by bypassing `cdk destroy` and jumping straight to `aws cloudformation delete-stack`, this cleanup would not be executed

Same problem applies to this feature though.  What would the expected behavior of `cdk destroy --no-wait-for-completion` be, if destroy starts doing other things? Like @iankhou said, `cdk destroy` is the end to end experience of destroy and only waiting till the end will tell you if it worked. If you want a quick fire-and-forget style delete `aws cloudformation delete-stack` is the way to go.

iankhou:
As we discussed above, this really isn't a feature we are looking to support, since the destroy command is meant to complete the lifecycle of stacks.

Instead of patching `aws-cdk`, consider directly using the AWS CLI to destroy your stacks (instead of using `cdk destroy`), using the output of `cdk ls`. This will prevent you from having to apply the patch each time you upgrade `aws-cdk`.

perrin4869:
Since changing the scope of the `cdk destroy` command seems to be a no-starter, how about a new command, `cdk destroy-without-waiting`? Lol

mrgrain:
> Since changing the scope of the `cdk destroy` command seems to be a no-starter, how about a new command, `cdk destroy-without-waiting`? Lol

Our recommendation remains to use the AWS CLI. There are no current plans to do anything fancy after destroy has completed. If that ever changes, we can revisit this feature requests. Please open an issue first next time, so we can have this discussion before you spend time implementing it.

perrin4869:
To be fair, for a simple change such as this, I thought I might as well start with a PR, to get something concrete to talk about off the go. Most of the heavy lifting was writing the `hasDependents` function, which I did after I got the initial go-ahead, so...

I understand that this feature can have weird edge-cases, which is why I am getting pushback, but on the other hand, I think those might be handled here, and this feature could see some real usage, so I wasn't sure which way it would go...

Of course my concern is that I start using the AWS CLI and somehow end up not being aware of a CDK change in the meantime that ends up coming back to bite me...

PR #819, opened 2025-08-28
feat(cli): auto-detect language for single-language custom templates

rohang9000:
Updated `cdk init` command to auto-detect language for single-language custom templates that are pulled from filesystem using `--from-path` so users do not need to specify `--language` in this case.

Fixes #

---


packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -336,10 +354,33 @@ export class InitTemplate {
...
+
+    // Auto-detect single language templates
+    if (languages.length === 0) {
+      const entries = await fs.readdir(basePath, { withFileTypes: true });
+      const languageDirs = entries.filter(entry =>
+        entry.isDirectory() &&
+        ['typescript', 'javascript', 'python', 'java', 'csharp', 'fsharp', 'go'].includes(entry.name),
+      );
+
+      if (languageDirs.length === 1) {
```
iankhou, line 350: Something is off with the indentation here, and it is slightly confusing. These should be at the same level, and consider using `else if`.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -336,10 +354,33 @@ export class InitTemplate {
...
-    const languages = await getLanguageDirectories(basePath);
+    let actualBasePath = basePath;
+    let languages = await getLanguageDirectories(basePath);
+
+    // Auto-detect single language templates
+    if (languages.length === 0) {
+      const entries = await fs.readdir(basePath, { withFileTypes: true });
+      const languageDirs = entries.filter(entry =>
+        entry.isDirectory() &&
+        ['typescript', 'javascript', 'python', 'java', 'csharp', 'fsharp', 'go'].includes(entry.name),
```
iankhou: No need to name these, iterate the keys from `languageExtensions` earlier in the file.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -298,6 +298,24 @@ async function hasLanguageFiles(directoryPath: string, extensions: string[]): Pr
...
+ */
+function getLanguageExtensions(language: string): string[] {
+  const languageExtensions: Record<string, string[]> = {
+    typescript: ['.ts', '.js'],
+    javascript: ['.js'],
+    python: ['.py'],
+    java: ['.java'],
+    csharp: ['.cs'],
+    fsharp: ['.fs'],
+    go: ['.go'],
```
iankhou: We already have languageExtensions in this file... https://github.com/aws/aws-cdk-cli/pull/819/files#diff-8a5cd4c3a707b56e2e21dd1333dd816bcc07a507effa6c3ad91d0d04381a39d6R238-R245

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -336,10 +354,33 @@ export class InitTemplate {
...
+      );
+
+      if (languageDirs.length === 1) {
+        // Validate that the language directory contains appropriate files
+        const langDir = languageDirs[0].name;
+        const langDirPath = path.join(basePath, langDir);
+        const hasValidFiles = await hasLanguageFiles(langDirPath, getLanguageExtensions(langDir));
+
+        if (hasValidFiles) {
+          actualBasePath = path.join(basePath, langDir);
```
iankhou: Two variables here have the same value—as long as `hasLanguageFiles` is a pure function (which it should be, if it's not, it should be corrected).

`langDirPath` and `actualBasePath`.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -336,10 +354,33 @@ export class InitTemplate {
       throw new ToolkitError(`Template path does not exist: ${basePath}`);
     }
 
-    const languages = await getLanguageDirectories(basePath);
+    let actualBasePath = basePath;
+    let languages = await getLanguageDirectories(basePath);
+
+    // Auto-detect single language templates
+    if (languages.length === 0) {
+      const entries = await fs.readdir(basePath, { withFileTypes: true });
```
iankhou: We are performing `fs.readDir` here twice with the same parameters. `getLanguageDirectories` calls it in the same way. Try to reuse the result instead of calling it again so soon. See https://github.com/aws/aws-cdk-cli/blob/5fc1b13b5977c45c054eee715b7ad41c472cddf6/packages/aws-cdk/lib/commands/init/init.ts#L236.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -336,10 +354,33 @@ export class InitTemplate {
...
+
+      if (languageDirs.length === 1) {
+        // Validate that the language directory contains appropriate files
+        const langDir = languageDirs[0].name;
+        const langDirPath = path.join(basePath, langDir);
+        const hasValidFiles = await hasLanguageFiles(langDirPath, getLanguageExtensions(langDir));
+
+        if (hasValidFiles) {
+          actualBasePath = path.join(basePath, langDir);
+          languages = [langDir];
```
iankhou: Looks like a silent failure if `hasValidFiles` is falsy. Shouldn't we fail or at least log a warning in this case?

iankhou (changes requested):
Please add unit tests for these changes. Consider separating the `basePath` and `actualBasePath` logic into a separate function, or adding some comments to explain it and its relationship with `language`—it's not straightforward at a glance.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: Misread, this code is fine.

iankhou:
Addressed my own comments and added a lot of unit testing to cover both new and old code.

Removed repeated code and consolidated the source of truth for supported CDK languages in a new `LanguageInfo` interface.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: Addressed this by implementing my suggestion. Moved language extensions to a new interface `LanguageInfo`.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: Moved language extensions to new interface `LanguageInfo`, which is the new, consolidated source of truth.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: Addressed this by consolidating the two variables.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: Changed the return signature of `getLanguageDirectories` to include `entries`, so that we can reuse it here and speed up the code.

packages/aws-cdk/lib/commands/init/init.ts
iankhou: Removed dead code.

iankhou:
[init.ts.html](https://github.com/user-attachments/files/22733286/init.ts.html)

otaviomacedo (approved):

iankhou (approved):

PR #820, opened 2025-08-28
feat(cli): git repository support for custom init templates

rohang9000:
Rohan:
> Updated code for `cdk init` command to handle custom templates from Git repository sources. This is compatible with the existing `--language` and `--template-path` flags. The `--from-git-url` flag was added to clone a Git URL that is passed in by the user, clones the repository, extracts the specified template files and initializes new project with them. The `cdk init` functionality of creating new local Git repo to track changes, making initial commit, and installing project dependencies is maintained.

Added additional tests and implemented changes I had requested. There were 40 commits before, and many merge conflicts from https://github.com/aws/aws-cdk-cli/pull/819.

Unit test coverage for `init.ts` is now 97.65%. Tested by hand for many cases as well.

---

### DIFF
--- a/packages/@aws-cdk/user-input-gen/lib/yargs-types.ts
+++ b/packages/@aws-cdk/user-input-gen/lib/yargs-types.ts
@@ -8,6 +8,7 @@ interface YargsCommand {
 export interface CliAction extends YargsCommand {
   options?: { [optionName: string]: CliOption };
   implies?: { [key: string]: string };
+  check?: (argv: any) => boolean | undefined;
 }
 
 interface YargsArg {
--- a/packages/aws-cdk/lib/cli/cli-config.ts
+++ b/packages/aws-cdk/lib/cli/cli-config.ts
@@ -8,6 +8,13 @@ import { getLanguageAlias } from '../commands/language';
 
 export const YARGS_HELPERS = new CliHelpers('./util/yargs-helpers');
 
+interface InitCommandArgs {
+  'template-path'?: string;
+  'from-path'?: string;
+  'from-git-url'?: string;
+  [key: string]: unknown;
+}
+
 /**
  * Source of truth for all CDK CLI commands. `user-input-gen` translates this into:
  *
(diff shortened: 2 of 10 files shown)


packages/@aws-cdk/user-input-gen/lib/yargs-types.ts
```diff
@@ -8,6 +8,7 @@ interface YargsCommand {
 export interface CliAction extends YargsCommand {
   options?: { [optionName: string]: CliOption };
   implies?: { [key: string]: string };
+  // check?: string;
```
iankhou: Comment left behind

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -401,10 +401,11 @@ export async function makeConfig(): Promise<CliConfig> {
...
           'generate-only': { type: 'boolean', default: false, desc: 'If true, only generates project files, without executing additional operations such as setting up a git repo, installing dependencies or compiling the project' },
           'lib-version': { type: 'string', alias: 'V', default: undefined, desc: 'The version of the CDK library (aws-cdk-lib) to initialize built-in templates with. Defaults to the version that was current when this CLI was built.' },
-          'from-path': { type: 'string', desc: 'Path to a local custom template directory or multi-template repository', requiresArg: true, conflicts: ['lib-version'] },
+          'from-path': { type: 'string', desc: 'Path to a local custom template directory or multi-template repository', requiresArg: true, conflicts: ['lib-version', 'from-git-url'] },
+          'from-git-url': { type: 'string', desc: 'Git repository URL to clone custom template from', requiresArg: true, conflicts: ['lib-version', 'from-path'] },
           'template-path': { type: 'string', desc: 'Path to a specific template within a multi-template repository', requiresArg: true },
         },
-        implies: { 'template-path': 'from-path' },
```
iankhou, line 410: This should be replaced by a check that ensures that `from-path` or `from-git-url` was also used.

packages/aws-cdk/lib/commands/init/init.ts
```diff
@@ -84,16 +91,27 @@ export async function cliInit(options: CliInitOptions) {
   const generateOnly = options.generateOnly ?? false;
   const workDir = options.workDir ?? process.cwd();
 
+  // Validate that if template-path is provided, exactly one custom template source must be provided
+  if (options.templatePath && !options.fromPath && !options.fromGitUrl) {
+    throw new ToolkitError('--template-path can only be used with --from-path or --from-git-url');
```
iankhou: why not use a check in yargs instead? In general, try to catch problems as early as you can.

iankhou (changes requested):
Probably will need to wait until https://github.com/aws/aws-cdk-cli/pull/819 is merged.

PR #821, opened 2025-08-28
fix: notices key in cdk.json is not respected

### LINKED ISSUE #724: CLI: notice suppression via `cdk.json` is not working
### Describe the bug

I'm supressing notices per documentation: https://docs.aws.amazon.com/cdk/v2/guide/ref-cli-cmd-notices.html

I have the following in my `cdk.json` file:

```json
{
  // ... snip

  "notices": false,
  "//": "~~ Generated by projen. To modify, edit .projenrc.js and run \"npx projen\"."
}
```

Yet I still notice spam in my term:

```
$> cdk ls
```

⏬ 

```plain
NOTICES         (What's this? https://github.com/aws/aws-cdk/wiki/CLI-Notices)

34892   CDK CLI will collect telemetry data on command usage starting at version 2.1100.0 (unless opted out)

        Overview: We do not collect customer content and we anonymize the
                  telemetry we do collect. See the attached issue for more
                  information on what data is collected, why, and how to
                  opt-out. Telemetry will NOT be collected for any CDK CLI
                  version prior to version 2.1100.0 - regardless of
                  opt-in/out.

        Affected versions: cli: ^2.0.0

        More information at: https://github.com/aws/aws-cdk/issues/34892


If you don’t want to see a notice anymore, use "cdk acknowledge <id>". For example, "cdk acknowledge 34892".
```

### Regression Issue

N/A

### Expected Behavior

Do not show notices.

### Current Behavior

Shows notices.

### Reproduction Steps

N/A

### Possible Solution

N/A

N/A

### CDK CLI Version

2.1019.2 (build c29855e)

### Framework Version

^2

### Node.js Version

v22.12.0

### OS

macOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

iankhou:
Fixes #724

## Description
Setting `"notices": false` in cdk.json doesn't suppress notices.

The root cause is that we handle logic to determine whether we should print notices at *yargs*. Yargs had a default value determined by a helper function (`YARGS_HELPERS.shouldDisplayNotices()`) that returned true when not in CI, or when in a safe CI.

This meant that any setting in our configuration would be overwritten by the output of this helper, because it looks like it's coming from the command line!

## Solution

Centralize logic for notices in `cli.ts`.

Made the notices option in yargs `undefined` by default. Moved the logic from `YARGS_HELPERS.shouldDisplayNotices()` into the `cli.ts`. This has several implications. 
- Setting `notices` in config now functions as expected; it will still not override the command line setting
- If a CDK app is running in a CI environment where it would not be safe to print notices
  - NEW BEHAVIOR: Notices WILL be printed if `notices: true` or `notices: [truthy value]` is set in config
  - Notices WILL be printed if flag `--notices` is passed, and NOT printed if `--no-notices` / `--notices=false` is passed at the command line (was already the case)

**Caveat: `"notices": "false"` in the config will suppress notices, but other "truthy" values will cause them to be printed.**

### Testing
Added unit tests to verify the hierarchy of behavior:
1. Flags in CLI (`--notices` or `--no-notices`)
2. Setting in configuration (`"notices": false` in cdk.json)
3. Print notices if not in CI or in CI where it is safe

---


packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -109,7 +111,34 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
+  let shouldDisplayNotices: boolean;
+  if (argv.notices !== undefined) {
+    // CLI argument takes precedence
+    shouldDisplayNotices = argv.notices;
+  } else {
+    // Fall back to configuration file setting, then autodetection
+    const configNotices = configuration.settings.get(['notices']);
+    if (configNotices !== undefined) {
+      // Consider string "false" to be falsy in this context
+      shouldDisplayNotices = configNotices !== "false" && Boolean(configNotices);
```
iankhou: Handle "false" as falsy.

Handle `""`, `0`, `null`, and `undefined` as falsy.

packages/aws-cdk/lib/cli/util/yargs-helpers.ts
```diff
@@ -48,16 +46,3 @@ export function browserForPlatform(): string {
...
- *
- * If the user didn't supply either `--notices` or `--no-notices`, we do
- * autodetection. The autodetection currently is: do write notices if we are
- * not on CI, or are on a CI system where we know that writing to stderr is
- * safe. We fail "closed"; that is, we decide to NOT print for unknown CI
- * systems, even though technically we maybe could.
- */
-export function shouldDisplayNotices(): boolean {
-  return !isCI() || Boolean(ciSystemIsStdErrSafe());
-}
```
iankhou, line 63: Moved this logic to cli code.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/synth/cdk-synth-telemetry-with-errors.integtest.ts
```diff
@@ -32,7 +32,6 @@ integTest(
               json: false,
               debug: false,
               staging: true,
-              notices: true,
```
iankhou, line 35: Remove this, as this option isn't passed by default in command line anymore.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/synth/cdk-synth-telemetry.integtest.ts
```diff
@@ -24,7 +24,6 @@ integTest(
               json: false,
               debug: false,
               staging: true,
-              notices: true,
```
iankhou, line 27: Remove this, as this option isn't passed by default in command line anymore.

rix0rrr (approved):

PR #826, opened 2025-08-29
feat(cli): `--yes` option to accept all questions

### LINKED ISSUE #732: (cli): `--yes` flag
### Describe the feature

A new global CLI option `--yes` that will cause all confirmation prompts (yes/no/always/never) to be answered positively.

### Use Case

I know what I am doing, I don't need any safety warnings. Using the flag indicates me accepting the risk of not being prompted.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

n/a

n/a

ren-yamanashi:
Adds a new `--yes` cli option so that when provided, commands can continue without confirmation. Any confirmation prompts will be accepted. Other user input will use a default value if available. The CLI might still prompt for input in situations where there is no default answer.

Fixes #732 

---

### DIFF
--- a/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
+++ b/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
@@ -81,6 +81,14 @@ export interface CliIoHostProps {
    * @default StackActivityProgress.BAR
    */
   readonly stackProgress?: StackActivityProgress;
+
+  /**
+   * Whether the CLI is running in non-interactive mode.
+   * When true, operation will proceed without confirmation.
+   *
+   * @default false
+   */
+  readonly nonInteractive?: boolean;
 }
 
 /**
@@ -160,6 +168,8 @@ export class CliIoHost implements IIoHost {
   private corkedCounter = 0;
   private readonly corkedLoggingBuffer: IoMessage<unknown>[] = [];
 
+  private readonly nonInteractive: boolean;
+
   public telemetry?: TelemetrySession;
 
   private constructor(props: CliIoHostProps = {}) {
@@ -170,6 +180,7 @@ export class CliIoHost implements IIoHost {
     this.requireDeployApproval = props.requireDeployApproval ?? RequireApproval.BROADENING;
 
     this.stackProgress = props.stackProgress ?? StackActivityProgress.BAR;
+    this.nonInteractive = props.nonInteractive ?? false;
   }
 
   public async startTelemetry(args: any, context: Context, _proxyAgent?: Agent) {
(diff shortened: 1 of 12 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -423,6 +434,11 @@ export class CliIoHost implements IIoHost {
         throw new ToolkitError(`${motivation}, but concurrency is greater than 1 so we are unable to get a confirmation from the user`);
       }
 
+      // In non-interactive mode, always return success (true).
+      if (this.nonInteractive) {
+        return true;
+      }
+
```
ren-yamanashi: Q. It should return `msg.defaultResponse` instead of `true`?

mrgrain: Yes, we need to return `true` for `isConfirmationPrompt`, `msg.defaultResponse` if there is one, and continue otherwise.

I would also like us to still print the question and answer and indicating that `--yes` was used.

We should also only do this after the `this.skipApprovalStep(msg)` check. On the other hand, `--yes` should make scenarios work where we don't have a TTY or are using concurrency. So we will need to restructure the checks.

mrgrain (changes requested):
Two existing uses of promptly that should be changed over as well as part of this:

https://github.com/aws/aws-cdk-cli/blob/5bb90cb387a5e29072244d27751b6c6b5da48fa7/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L980

https://github.com/aws/aws-cdk-cli/blob/5bb90cb387a5e29072244d27751b6c6b5da48fa7/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L2121

Refactor should be a confirm:

https://github.com/aws/aws-cdk-cli/blob/5bb90cb387a5e29072244d27751b6c6b5da48fa7/packages/%40aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L1193-L1196

ren-yamanashi:
@mrgrain 

I apologize for my slow response...

And Thanks for adding the some fix commit 🙏 

I'm not sure how to implement the `@TODO` comment section (below). What kind of changes should I make?

- https://github.com/aws/aws-cdk-cli/pull/826/files#diff-4c33b270e425ae2c3cb6ad81c6d79ac1130802223c7af2588bd9674274cfa493R441
- https://github.com/aws/aws-cdk-cli/pull/826/files#diff-4c33b270e425ae2c3cb6ad81c6d79ac1130802223c7af2588bd9674274cfa493R447

mrgrain (approved):

PR #838, opened 2025-09-09
fix(cli): text not readable in light terminals

### LINKED ISSUE #751: cli: unreadable text due to forced bright colors on light terminals
### Describe the bug

The color of most text written to the console is forced to be light gray, even on terminals with a white background. The contrast is very low, and it is barely readable.

https://github.com/aws/aws-cdk/issues/5242 (from 2019) seems to fix the problem, but it is present in the current version of the CLI.

Example:

---

Image

---

### Regression Issue

_No response_

### Expected Behavior

Text should be readable.

### Current Behavior

Text is very hard to read due to very low contrast.

### Reproduction Steps

Run `cdk init` on a terminal with white background.

### Possible Solution

A possible solution is to detect the background color with [`DECAC`](https://www.invisible-island.net/xterm/ctlseqs/ctlseqs.html#:~:text=Assign%20Color%20(DECAC)). For example, writing `\e]11;?\a` to a terminal with white background generates a response like `\e]11;rgb:ffff/ffff/ffff\a`, and `\e]11;rgb:0000/0000/0000\a` on terminals with black background.

If the background can't be determined, the CLI should use default color (non-bright variants), instead of forcing the the bright ones.

_No response_

aws-cdk-lib@2.206.0

### AWS CDK CLI version

2.1022.0

### Node.js Version

v22.14.0

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

ayosec:
With `chalk.white`, the text is hard to read on terminals with light backgrounds. `chalk.reset` works as expected in both dark and light background.

Fixes #751

---

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/lib/format.ts
+++ b/packages/@aws-cdk/cloudformation-diff/lib/format.ts
@@ -98,7 +98,7 @@ export class Formatter {
   }
 
   public print(fmt: string, ...args: any[]) {
-    this.stream.write(chalk.white(format(fmt, ...args)) + '\n');
+    this.stream.write(chalk.reset(format(fmt, ...args)) + '\n');
   }
 
   public warning(fmt: string, ...args: any[]) {
@@ -204,7 +204,7 @@ export class Formatter {
     if (diff.isRemoval) {
       return REMOVAL;
     }
-    return chalk.white('[?]');
+    return chalk.reset('[?]');
   }
 
   /**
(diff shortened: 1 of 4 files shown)


mrgrain (approved):

mrgrain:
Looks good, thank you! @ayosec I'm taking care if the PR from here. No need to interact with it anymore, especially don't update the branch.

PR #846, opened 2025-09-12
fix(cli): stack filter positional arg is not being respected

otaviomacedo:
At some point, the positional argument was lost, and there were no tests for it.

Add it to `cli-config.ts`.

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/refactoring-multiple-envs/refactoring.js
+++ b/packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/refactoring-multiple-envs/refactoring.js
@@ -1,0 +1,26 @@
+const cdk = require('aws-cdk-lib');
+const s3 = require('aws-cdk-lib/aws-s3');
+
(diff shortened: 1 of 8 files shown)


packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/refactoring-multiple-envs/refactoring.js
```diff
@@ -0,0 +1,26 @@
+const cdk = require('aws-cdk-lib');
```
otaviomacedo, line 1: Creating a new app, to avoid cluttering the existing one, used for other refactoring scenarios.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -493,6 +493,10 @@ export async function makeConfig(): Promise<CliConfig> {
             desc: 'Whether to do the refactor without asking for confirmation',
           },
         },
+        arg: {
```
otaviomacedo, line 496: This is the actual fix. The rest follows from it.

mrgrain (approved):

PR #850, opened 2025-09-17
feat(cli): gc can ignore user-verified unauthorized non-cdk stacks

### LINKED ISSUE #640: (garbage-collection): add flag when running garbage collector to ignore non-authorized stacks
### Describe the feature

add some kind of flag to the cdk gc command (for example --skip-unauthorized-stacks) so that when checking the stacks (GetTemplateSummary) it ignores those that the user does not have access instead of failing the command

### Use Case

I was very excited for this new feature, but when I tried running it, I got the following error:
```
npx cdk gc --unstable=gc --rollback-buffer-days 5
 ⏳  Garbage Collecting environment aws://<ACCOUNT>/eu-west-1...
Error refreshing stacks: AccessDenied: User: arn:aws:sts::<ACCOUNT>:assumed-role/<ROLE> is not authorized 
to perform: cloudformation:GetTemplateSummary on 
resource: arn:aws:cloudformation:eu-west-1:<ACCOUNT>:stack/<ORGANIZATION-STACK> 
with an explicit deny in a service control policy
```
(newlines added to improve readability)

Where <ORGANIZATION-STACK> is a stack used my my company to do initial setup of AWS accounts.

### Proposed Solution

Instead of automatically failing, have some way to ignore stacks in error. Right now I know of no way to skip it

### Other Information

_No response_

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.171.1

macOS Sonoma 14.7

dgandhi62:
Fixes #640 

The gc command currently fails when encountering AccessDenied errors on non-CDK stacks during garbage collection. This PR adds a new option `--unauth-native-cfn-stacks-to-skip` to handle this situation. 

Users can specify multiple stack names or glob patterns to skip when unauthorized access errors occur. When stacks matching the provided patterns encounter access denied errors, the command shows a confirmation prompt asking whether to skip these stacks and continue with garbage collection. The default response is 'no' to ensure users explicitly confirm the action. It remains the user's responsibility to ensure that the stacks entered are non-cdk stacks. 

The 'no' default was chosen over 'yes' to prevent unintended behavior until a future `--yes` flag is implemented for the cdk-cli. While a shorter alias was considered for the option name, it was rejected to avoid user confusion.

---


packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -178,6 +178,18 @@ interface GarbageCollectorProps {
    * @default true
    */
   readonly confirm?: boolean;
+
+  /**
```
kaizencc, line 182: missing documentation for both of these

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -97,9 +120,16 @@ function bootstrapFilter(parameters?: ParameterDeclaration[], qualifier?: string
           splitBootstrapVersion[2] != qualifier);
 }
 
-export async function refreshStacks(cfn: ICloudFormationClient, ioHelper: IoHelper, activeAssets: ActiveAssetCache, qualifier?: string) {
+export async function refreshStacks(
```
kaizencc: i think its time for refreshStacks to take in a property bag

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1918,6 +1918,18 @@ export interface GarbageCollectionOptions {
    * @default false
    */
   readonly confirm?: boolean;
+
```
kaizencc, line 1925: both missing documentation

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -178,6 +178,18 @@ interface GarbageCollectorProps {
...
+   *
+   * @default []
+   */
+  readonly ignoreStacks?: string[];
```
kaizencc:
```suggestion
  readonly ignoreNonCdkStackNames?: string[];
```

kaizencc: we need to make it clear that garbage collection is unsafe if we skip CFN stacks created by CDK. to use this property safely, the customer must know beforehand that the stack in question was not created via CDK.

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -178,6 +178,18 @@ interface GarbageCollectorProps {
...
+
+  /**
+   *
+   * @default false
+   */
+  readonly skipUnauthorizedStacks?: boolean;
```
kaizencc: in fact, i am not comfortable with this property at all. we cannot supply a way for a customer to unknowningly purge assets from an unauthorized CDK stack. since this is a tricky and uncommon situation, i would like us to make sure the user interface is explicit about which stacks we can safely ignore.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -487,6 +487,8 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           createdBufferDays: args['created-buffer-days'],
           bootstrapStackName: args.toolkitStackName ?? args.bootstrapStackName,
           confirm: args.confirm,
+          ignoreStacks: args['ignore-stacks'],
```
kaizencc: missing a cli integ test for this to ensure that this is actually exposed via the cli. integ test should deploy 2 stacks with an asset each, one stack with a prefix, and then verify that gc with ignore stacks actually _does_ delete the asset of the stack with prefix.

kaizencc (changes requested):
also missing a readme entry. this definitely needs it because it is a very tricky property to add and without the proper documentation, users will use it incorrectly and accidentally delete important assets.

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
dgandhi62: I've added warnings to the user and mentioned which stacks are being ignored. It isn't garbage collecting on the unauth stacks, instead, just skipping them and not throwing an error

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
dgandhi62: Have made the change

packages/aws-cdk/lib/cli/cdk-toolkit.ts
dgandhi62: Have made the change, thank you

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
dgandhi62: Updated in a separate pr - #855

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
dgandhi62: Have updated

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -178,6 +178,20 @@ interface GarbageCollectorProps {
    * @default true
    */
   readonly confirm?: boolean;
+
+  /**
+   * Stack names to ignore during garbage collection
+   *
+   * @default []
+   */
+  readonly ignoreNonCdkStacks?: string[];
```
mrgrain: The way this is implemented we are skipping _any_ stack, not just CDK stacks. Like @kaizencc I'm not convinced this is a good idea in the first place, but at leat the option need to be named correctly.

mrgrain (dismissed):
option names

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
dgandhi62: I've updated my pr to make it more explicit that only non-cdk stacks should be entered by the user. Have changed the option name accordingly. Thank you for the feedback

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -178,6 +178,14 @@ interface GarbageCollectorProps {
...
    */
   readonly confirm?: boolean;
+
+  /**
+   * Non-CDK stack names or glob patterns to skip when encountering unauthorized access errors during garbage collection.
+   * You must explicitly specify non-CDK stack names - CDK stacks will be rejected to prevent accidental asset deletion.
+   *
+   * @default undefined
+   */
+  readonly skipUnauthorizedStacksWhenNonCdk?: string[];
```
kaizencc: we need to workshop this name. how about:

```ts
{
  readonly nonCdkStacksToSkip?: string[];
}
```

or 

```ts
{
  readonly nativeCloudFormationStacksToSkip?: string[];
}
```

It's very important to get this right. @mrgrain do you have ideas?

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -178,6 +178,14 @@ interface GarbageCollectorProps {
    * @default true
    */
   readonly confirm?: boolean;
+
+  /**
+   * Non-CDK stack names or glob patterns to skip when encountering unauthorized access errors during garbage collection.
+   * You must explicitly specify non-CDK stack names - CDK stacks will be rejected to prevent accidental asset deletion.
```
kaizencc: "CDK stacks will be rejected to prevent accidental asset deletion"

this reads like we are checking for the user whether or not stacks supplied here are CDK stacks.

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
```diff
@@ -178,6 +178,14 @@ interface GarbageCollectorProps {
...
+   *
+   * @default undefined
```
kaizencc, line 186:
```suggestion
   * @default []
```

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -56,20 +80,40 @@ async function fetchAllStackTemplates(cfn: ICloudFormationClient, ioHelper: IoHe
...
       });
 
       templates.push((template.TemplateBody ?? '') + JSON.stringify(summary?.Parameters));
+    } catch (error: any) {
+      // Check if this is a CloudFormation access denied error
+      if (error.name === 'AccessDenied') {
+        if (shouldSkipStack(stack, skipUnauthorizedStacksWhenNonCdk)) {
+          await ioHelper.defaults.warn(
+            `Skipping unauthorized stack '${stack}' as specified in --skip-unauthorized-stacks-when-noncdk`,
+          );
```
kaizencc: instead of printing every time we find an unauthorized stack, why don't we store these skipped stacks in memory and then, after we are done parsing stacks and _before we delete_ we prompt the user with something like "skipped these stacks that YOU specified as not CDK stacks. if they are CDK stacks their assets will be deleted!!! are you sure? y/n/t"

y/n/t is yes/no/trust, which we should have an example of elsewhere in garbage collection since it prompts the user elsewhere too.

kaizencc (changes requested):
also definitely needs a readme update :)

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
dgandhi62: Ah! I missed this. I was trying to add a check for cdk stacks earlier but it was buggy since you technically can't do it if the stack is unauthorized. Forgot to edit this documentation. Will make the change

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
dgandhi62: Shouldn't it being an unauth stack be part of the name somehow? As per the original feature request. How about `unauthNonCdkStacksToSkip`?

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
dgandhi62: [**skipUnauthorizedStacksWhenNonCdk: arrayFromYargs(args['skip-unauthorized-stacks-when-noncdk'])**](https://github.com/aws/aws-cdk-cli/pull/850/files#diff-be9d125a91b3869bf888755d6d10cd596d19397a54f3a3bce60dd4c8ee2c0406R490)

arrayFromYargs returns undefined for an empty input. Would [] be technically valid documentation as a default?
https://github.com/aws/aws-cdk-cli/blob/2f0cfc496ac4e11b43b9ae33d6f44105423bdc04/packages/aws-cdk/lib/cli/cli.ts#L623

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
dgandhi62: This makes sense. Working on it now

dgandhi62: Latest commit adds this functionality. Will make another commit with a README once the option name is finalized

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
kaizencc: fair point :)

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/garbage-collector.ts
kaizencc: i think, while the original user had an unauthorized stack they wanted to skip, the real importance of this property is that it must must must not be a cdk stack. my preference is still `nativeCloudFormationStacksToSkip`. Please follow up with @mrgrain to make a determination here.

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -11,6 +12,9 @@ export class ActiveAssetCache {
   }
 
   public contains(asset: string): boolean {
+    // To reduce computation if asset is null or undefined
+    if (!asset) return false;
+
```
kaizencc: this change seems unrelated? its also wrong -- asset cannot be null or undefined as it is defined as a string. what it is currently doing is saying that if `asset == ''`, return `false`. fine, but document that rather than saying asset is null or undefined.

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -31,12 +52,78 @@ async function paginateSdkCall(cb: (nextToken?: string) => Promise<string | unde
...
+      level: 'info',
+      code: 'CDK_TOOLKIT_I9210',
+      message,
+      data: {
+        stacks: unauthorizedStacks,
+        count: unauthorizedStacks.length,
+        responseDescription: '[y]es/[n]o',
+      },
+      defaultResponse: 'y',
+    });
```
kaizencc: this needs to be defined message in private/messages.ts. see other places where we are calling `requestResponse`.

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -31,12 +52,78 @@ async function paginateSdkCall(cb: (nextToken?: string) => Promise<string | unde
...
+
+  try {
+    const stackList = unauthorizedStacks.join(', ');
+    const message = `Found ${unauthorizedStacks.length} unauthorized stack(s): ${stackList}\nDo you want to skip all these stacks?`;
+
+    // Ask user if they want to proceed
+    const response = await ioHelper.requestResponse({
+      time: new Date(),
+      level: 'info',
+      code: 'CDK_TOOLKIT_I9210',
```
kaizencc: this is definitely the wrong code... you need to create your own in messages.ts

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -31,12 +52,78 @@ async function paginateSdkCall(cb: (nextToken?: string) => Promise<string | unde
...
+
+  if (isCI) {
+    throw new ToolkitError(
+      `Found ${unauthorizedStacks.length} unauthorized stack(s) in CI environment. ` +
+      'Set CDK_GC_AUTO_APPROVE_UNAUTHORIZED=true to auto-approve or configure --skip-unauthorized-stacks-when-noncdk.',
+    );
+  }
+
+  try {
+    const stackList = unauthorizedStacks.join(', ');
```
kaizencc:
```suggestion
    const stackList = unauthorizedStacks.join(',\n');
```

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -31,12 +52,78 @@ async function paginateSdkCall(cb: (nextToken?: string) => Promise<string | unde
...
+               process.env.GITLAB_CI === 'true' ||
+               process.env.JENKINS_URL !== undefined ||
+               process.env.BUILDKITE === 'true';
+
+  if (isCI) {
+    throw new ToolkitError(
+      `Found ${unauthorizedStacks.length} unauthorized stack(s) in CI environment. ` +
+      'Set CDK_GC_AUTO_APPROVE_UNAUTHORIZED=true to auto-approve or configure --skip-unauthorized-stacks-when-noncdk.',
+    );
+  }
```
kaizencc: none of this is necessary. our ioHost handles this sort of thing for us. basically, we send all output to a global IoHost instance (in the cli, its called `CliIoHost`) and that IoHost decides where (or if) we send information. in the case of a ci environment and `CliIoHost`, we just accept the default.

in this case, you've defined the default as `yes`. is that what we want to do? (i think that's ok, but think about it)

kaizencc (changes requested):
few more comments. this one is tricky!

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
dgandhi62: That makes sense! Yes, this wasn't related to the problem at hand. Just a general optimization that made sense to me when I was looking at the garbage collector class. I did !asset instead just as a general coverage measure. But it makes sense that it would only be empty strings.

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
dgandhi62: The issue is that github kept failing the test if I didn't add this check. Is this expected?

packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
```diff
@@ -31,12 +52,78 @@ async function paginateSdkCall(cb: (nextToken?: string) => Promise<string | unde
...
+async function handleUnauthorizedStacks(unauthorizedStacks: string[], ioHelper: IoHelper): Promise<void> {
+  if (unauthorizedStacks.length === 0) {
+    return;
+  }
+
+  // Auto-approve in CI environments
+  if (process.env.CDK_GC_AUTO_APPROVE_UNAUTHORIZED === 'true') {
+    await ioHelper.defaults.info(`Auto-approving ${unauthorizedStacks.length} unauthorized stack(s) in CI mode`);
+    return;
+  }
```
mrgrain: We don't usually want per command behaviour like this. Instead you are looking for something like this: https://github.com/aws/aws-cdk-cli/pull/826

Saying that, @kaizencc is right: If your test doesn't work in CI than something else is not right in your testing setup. Broadly speaking:

- If you need to add test specific code to production code, something is not right.
- If your production code is hard to test, something is not right.

Typically you are aiming for solutions that don't require test specific code and are easy to test.

PR #893, opened 2025-09-24
feat(cli): display warning when --role-arn is used with gc command

dgandhi62:
The gc command currently ignores the --role-arn parameter and always uses the user's default credentials, regardless of whether a role ARN is specified. This can potentially lead to silent failures, permission issues, as well as a bad user experience due to confusion. 

This pr adds a check to throw a warning when the --role-arn parameter is used in the gc command. This helps improve user experience and avoid mismanaged expectations.

A consideration to throw an error was made. However, this could be a breaking change for existing ci/cd pipeline users. Therefore, this idea was discarded

---

### DIFF
--- a/packages/aws-cdk/lib/cli/cli.ts
+++ b/packages/aws-cdk/lib/cli/cli.ts
@@ -475,6 +475,11 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
         if (args.bootstrapStackName) {
           await ioHost.defaults.warn('--bootstrap-stack-name is deprecated and will be removed when gc is GA. Use --toolkit-stack-name.');
         }
+        // roleArn is defined for when cloudformation is invoked
+        // This conflicts with direct sdk calls existing in the gc command to s3 and ecr
+        if (args.roleArn) {
+          await ioHost.defaults.warn('The --role-arn option is not supported for the gc command and will be ignored.');
+        }
         return cli.garbageCollect(args.ENVIRONMENTS, {
           action: args.action,
           type: args.type,
--- a/packages/aws-cdk/test/cli/cli.test.ts
+++ b/packages/aws-cdk/test/cli/cli.test.ts
@@ -437,19 +437,27 @@ describe('notices configuration tests', () => {
 });
 
 describe('gc command tests', () => {
+  let originalCliIoHostInstance: any;
+
   beforeEach(() => {
     jest.clearAllMocks();
+    originalCliIoHostInstance = CliIoHost.instance;
+  });
+
+  afterEach(() => {
+    CliIoHost.instance = originalCliIoHostInstance;
   });
 
   test('should warn when --role-arn is used with gc command', async () => {
     const gcSpy = jest.spyOn(cdkToolkitModule.CdkToolkit.prototype, 'garbageCollect').mockResolvedValue();
-    
-    // Make exec use our TestIoHost and adds properties to TestIoHost to match ClioHosr
+
+    // Make exec use our TestIoHost and adds properties to TestIoHost to match CliIoHost
     const warnSpy = jest.fn();
-    (ioHost as any).defaults = { warn: warnSpy, debug: jest.fn() };
+    (ioHost as any).defaults = { warn: warnSpy, debug: jest.fn(), result: jest.fn() };
     (ioHost as any).asIoHelper = () => ioHelper;
+    (ioHost as any).logLevel = 'info';
     jest.spyOn(CliIoHost, 'instance').mockReturnValue(ioHost as any);
-    
+
     const mockConfig = {
       loadConfigFiles: jest.fn().mockResolvedValue(undefined),
       settings: {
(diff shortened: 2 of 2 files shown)


dgandhi62:
> Why not implement `roleArn` functionality instead?

In the command documentation, roleArn is defined for when we invoke Cloudformation. gc seems to have some direct calls to s3/Ecr. I think there is scope for implementing this functionality but it might need a bigger discussion on the design implications. I can investigate this more and make an update later for this if needed, this should work as a primary fix meanwhile. Thanks for the feedback

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -477,6 +477,9 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
         if (!configuration.settings.get(['unstable']).includes('gc')) {
           throw new ToolkitError('Unstable feature use: \'gc\' is unstable. It must be opted in via \'--unstable\', e.g. \'cdk gc --unstable=gc\'');
         }
+        if (args.roleArn) {
+          throw new ToolkitError('The --role-arn option is not supported for the gc command');
```
kaizencc: can you explain in documentation why this is the case?

kaizencc (commented):
what about a test?

packages/aws-cdk/lib/cli/cli.ts
dgandhi62: Added the reasoning and the test. Thank you

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -477,6 +477,11 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
         if (!configuration.settings.get(['unstable']).includes('gc')) {
           throw new ToolkitError('Unstable feature use: \'gc\' is unstable. It must be opted in via \'--unstable\', e.g. \'cdk gc --unstable=gc\'');
         }
+        // roleArn is defined for when cloudformation is invoked
+        // This conflicts with direct sdk calls existing in the gc command to s3 and ecr
+        if (args.roleArn) {
+          throw new ToolkitError('The --role-arn option is not supported for the gc command');
```
kaizencc: i thought we discussed that this was a breaking change? exit criteria for this is either:

a) change to warning -> my preference
b) argument for why an error makes more sense here, even though its breaking

If we do go with a breaking change, then we'd have to make sure in PR description that we mean to do this. Can also solicit opinions from others :).

dgandhi62: The biggest reason I suggest throwing an error in this case as opposed to a warning is that existing users may not notice the warning when using ci/cd pipelines. In such a situation, we would still need a reliable way to ensure that the user is informed. 

Furthermore, allowing this functionality (even with a warning) could be potentially hazardous if misused. For example, the user might want to intentionally provide limited permissions for gc. However, if the default credentials are used, unintended resources could be garbage collected. Another example i can think of is if a user might be using both a prod and dev account. They might try to apply gc for prod particularly and if they miss the warning, they might end up deleting dev resources, or vice versa. 

The main usage of warnings in `cli.ts` I see have been more for deprecated options - https://github.com/aws/aws-cdk-cli/blob/09c0061172faf811ab94d9319a169c5ee5513dee/packages/aws-cdk/lib/cli/cli.ts#L481

However, this feature in its current state is incompatible, yet allowed. Examples of previous usage of incompatibility and / or outright functionality we want to not allow can be seen in the code - https://github.com/aws/aws-cdk-cli/blob/09c0061172faf811ab94d9319a169c5ee5513dee/packages/aws-cdk/lib/cli/cli.ts#L331

https://github.com/aws/aws-cdk-cli/blob/09c0061172faf811ab94d9319a169c5ee5513dee/packages/aws-cdk/lib/cli/cli.ts#L390

https://github.com/aws/aws-cdk-cli/blob/09c0061172faf811ab94d9319a169c5ee5513dee/packages/aws-cdk/lib/cli/cli.ts#L538

kaizencc: > The biggest reason I suggest throwing an error in this case as opposed to a warning is that existing users may not notice the warning when using ci/cd pipelines. In such a situation, we would still need a reliable way to ensure that the user is informed.

This i have a hard disagree with. You are saying in ci/cd pipelines, someone with an existing setup that _works_ should be "informed", but by informed we mean breaking their pipelines. I think such a customer with a working setup would be more upset that they were broken.

> The biggest reason I suggest throwing an error in this case as opposed to a warning is that existing users may not notice the warning when using ci/cd pipelines. In such a situation, we would still need a reliable way to ensure that the user is informed.

I also disagree. This boils down to whether or not its a big problem that `--role-arn` is accepted but unused. I think earlier we talked about how this is mostly a cosmetic thing -- it was confusing why `--role-arn` didn't affect your command. Which I agree with, but I don't think its more problematic than that. It doesn't harm or cause `cdk gc` to be misused -- `gc` is clearly a env-scoped command and users of the command should be aware of that.

From my perspective, this is a courtesy warning of hey, `--role-arn` isn't being used for this command. I don't think we should throw an error for it.

dgandhi62: After some thought. I've updated the functionality to throw a warning instead. This makes sense, I didn't realize the environment scoping was explicit.

kaizencc (approved):

PR #905, opened 2025-10-06
chore(cli-lib-alpha): stop releasing deprecated package

mrgrain:
Stop releasing the deprecated `@aws-cdk/cli-lib-alpha` package. This change removes the package from the code base and integ tests. If we ever need to release it again, we can revert this PR.

I tried to just disable releasing, but then integ tests fail and I think removing it all in one PR is just easier if we need to go back to it.

---


rix0rrr (approved):

projenrc/cdk-cli-integ-tests.ts
```diff
@@ -334,9 +334,6 @@ export class CdkCliIntegTestsWorkflow extends Component {
         // assumptions about the availability of source packages.
         IS_CANARY: 'true',
         CI: 'true',
-        // This is necessary because the new versioning of @aws-cdk/cli-lib-alpha
-        // matches the CLI and not the framework.
-        CLI_LIB_VERSION_MIRRORS_CLI: 'true',
```
kaizencc, line 339: this change could be the reason why we don't pick up the right version anymore

PR #920, opened 2025-10-28
docs(aws-cdk): technical and conceptual flowcharts for deploy

9pace:
Publish diagrams that illustrate the CDK deployment process at two levels of detail. The first diagram provides a conceptual overview suitable for understanding the high-level flow, while the second diagram shows the technical implementation details for contributors working on the CDK codebase.

---


mrgrain (changes requested):
@9pace Love this 😍 

Only thing is that this is a user facing README  and I don't think this level of detail is needed or useful for them. In particular the part where the chart documents internal function calls. 

Can you find a different place for it?

9pace:
I think that in the future we may want to add more docs to a docs/ directory. I think that this information is useful to community and would be useful somewhere in the repo.

packages/aws-cdk/README.md
```diff
@@ -567,6 +567,140 @@ and might have breaking changes in the future.
...
+    c7 -->|"No"| c11
+    
+    %% Asset Processing Phase
+    c11 --> c13
+    c13 --> c14
+    c14 --> c19
+    
+    %% Deployment Method Selection
+    c19 -->|"Yes: Fast Deploy"| c20
+    c19 -->|"No: Standard Deploy"| c21
```
mrgrain: I think we can skip this in the base flow description

packages/aws-cdk/README.md
```diff
@@ -567,6 +567,140 @@ and might have breaking changes in the future.
...
 > *: `Fn::GetAtt` is only partially supported. Refer to [this implementation](https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/api/cloudformation/evaluate-cloudformation-template.ts#L256-L266) for supported resources and attributes.
 
+
+##### Deploy flowchart
+
+This flowchart provides a high-level overview of the deployment architecture and key decision points.
+For technical implementation details (function calls, file locations), see [docs/deploy-architecture.md](./docs/deploy-architecture.md).
+
+```mermaid
+graph TD
```
mrgrain: can you simplify the phases even further? For a high-level diagram I think this is still too much detail. The phases are good, maybe they can be a single box with some bullet point details.

packages/aws-cdk/README.md
9pace: Done

packages/aws-cdk/README.md
9pace: Simplified it. If you think there is relevant detail missing that should be included in bullet points let me know, keeping in mind you can always see the more detailed version if necessary.

mrgrain: Thanks, I like it! And agreed that more detail is available later.

Happy to approve content wise. But one thing for you to consider: Does the legend still add any value?

mrgrain (changes requested):
Last thing: The README on npm won't render mermaid diagrams (as far as I know). So the README version needs to be an image. See other diagrams already referenced in the README.

packages/aws-cdk/README.md
9pace: True. I feel like the colors are fairly intuitive to group purpose of a couple of blocks. Removed the legend

mrgrain (approved):

PR #926, opened 2025-11-04
feat(toolkit-lib): network detector

kaizencc:
this PR implements a connectivity detector that performs a HEAD request to the notices endpoint, with a timeout of 1 hour. before making any subsequent network calls, (including GET notices and PUT telemetry) we check to see if we have already determined the network to be disconnected. in those cases, we skip subsequent calls.

---


packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
```diff
@@ -44,6 +45,12 @@ export class WebsiteNoticeDataSource implements NoticeDataSource {
   }
 
   async fetch(): Promise<Notice[]> {
+    // Check connectivity before attempting network request
+    const hasConnectivity = await NetworkDetector.hasConnectivity(this.agent);
+    if (!hasConnectivity) {
+      throw new ToolkitError('No internet connectivity detected');
```
mrgrain, line 52: Is throwing the right thing here? Is that error caught elsewhere? Asking because Notices should just silently fail. A comment might help.

kaizencc: this is the right thing to do here. we are throwing errors in `web-data-source` on failures and expecting to swallow them elsewhere (which we do)

packages/aws-cdk/test/cli/telemetry/sink/endpoint-sink.test.ts
```diff
@@ -312,4 +324,20 @@ describe('EndpointTelemetrySink', () => {
       expect.stringContaining('Telemetry Error: POST example.com/telemetry:'),
     );
   });
+
+  test('skips request when no connectivity detected', async () => {
```
kaizencc, line 327: this is the true result: we do not ping the telemetry endpoint in environments without internet access

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
otaviomacedo: [non-blocking] Since this pattern will be very common (get the result, check whether it's true and throw an error if not), we could also have a method that takes a callback and does this for you:

```ts
return NetworkDetector.ifConnected(async() => {...});
```

otaviomacedo (approved):

packages/@aws-cdk/toolkit-lib/lib/api/notices/notices.ts
```diff
@@ -155,7 +155,16 @@ export class Notices {
    * @throws on failure to refresh the data source
    */
   public async refresh(options: NoticesRefreshOptions = {}) {
-    const innerDataSource = options.dataSource ?? new WebsiteNoticeDataSource(this.ioHelper, this.httpOptions);
+    await this.ioHelper.notify({
+      message: `notices refresh starting, ${JSON.stringify(options)}`,
+      time: new Date(Date.now()),
+      level: 'info',
+      data: undefined,
+    });
```
mrgrain: I don't think this should happen
```suggestion
```

mrgrain (changes requested):
see inline

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
```diff
@@ -28,6 +30,14 @@ export class WebsiteNoticeDataSourceProps {
...
    */
   readonly agent?: https.Agent;
+
+  /**
+   * Whether or not we want to skip the check for if we have already determined we are in
+   * a network-less environment. Forces WebsiteNoticeDataSource to make a network call.
+   *
+   * @default false
+   */
+  readonly skipNetworkCache?: boolean;
```
mrgrain: ?

```suggestion
  readonly checkNetwork?: boolean;
```

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
```diff
@@ -37,13 +47,29 @@ export class WebsiteNoticeDataSource implements NoticeDataSource {
...
     this.url = props.url ?? 'https://cli.cdk.dev-tools.aws.dev/notices.json';
+    this.skipNetworkCache = props.skipNetworkCache;
   }
 
   async fetch(): Promise<Notice[]> {
+    if (!this.skipNetworkCache) {
+      await this.ioHelper.notify({
+        message: `website data source fetch starting, ${this.agent !== undefined}}`,
+        time: new Date(Date.now()),
+        level: 'info',
```
mrgrain:
```suggestion
        level: 'debug',
```

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
```diff
@@ -37,13 +47,29 @@ export class WebsiteNoticeDataSource implements NoticeDataSource {
...
+  private readonly skipNetworkCache?: boolean;
 
   constructor(private readonly ioHelper: IoHelper, props: WebsiteNoticeDataSourceProps = {}) {
     this.agent = props.agent;
     this.url = props.url ?? 'https://cli.cdk.dev-tools.aws.dev/notices.json';
+    this.skipNetworkCache = props.skipNetworkCache;
   }
 
   async fetch(): Promise<Notice[]> {
+    if (!this.skipNetworkCache) {
```
mrgrain: why is this only for not skip network cache?

If you move this after the network check, it will always message. But do we even need this?

packages/@aws-cdk/toolkit-lib/lib/index.ts
```diff
@@ -17,3 +17,6 @@ export * from './api/cloud-assembly';
 export * from './api/io';
 export * from './api/tags';
 export * from './api/plugin';
+
+// Utilities
+export { NetworkDetector } from './util/network-detector';
```
mrgrain: should this be a public package?

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
mrgrain: what is this even used for?

mrgrain: Oh I see. I like positive booleans. So something like `checkNetwork`.

kaizencc: not used, will be removed

packages/@aws-cdk/toolkit-lib/lib/index.ts
kaizencc: will be reverted

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
kaizencc: not used will delete

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
kaizencc: this whole thing is gonna be deleted

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
kaizencc: not used, will delete

packages/@aws-cdk/toolkit-lib/lib/api/notices/notices.ts
kaizencc: also will be deleted

packages/@aws-cdk/toolkit-lib/lib/index.ts
kaizencc: actually i do need this so endpoint-sink can access network detector in aws-cdk. but ive updated how its imported to mirror that of notices

packages/@aws-cdk/toolkit-lib/lib/api/notices/web-data-source.ts
kaizencc: this is a good thought and i have considered this. im not entirely against it, but i feel like my more (naive) approach works more intuitively even if it reuses the same pattern. we can always refactor in the future if it turns out that `ifConnected` is a cleaner API

mrgrain (approved):

PR #927, opened 2025-11-05
fix(cli): requests get throttled when there is a large number of stacks

otaviomacedo:
The CLI is making an unlimited number of concurrent `GetTemplate` requests. When there is a large number of stacks, CloudFormation may throttle the requests.

Limit concurrency to 20.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/index.ts
@@ -15,6 +15,7 @@ import { ResourceLocation, ResourceMapping } from './cloudformation';
 import { hashObject } from './digest';
 import type { MappingGroup } from '../../actions';
 import { ToolkitError } from '../../toolkit/toolkit-error';
+import { pLimit } from '../../util/concurrency';
 
 export * from './exclude';
 export * from './context';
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/refactoring/index.ts
```diff
@@ -144,8 +145,10 @@ export async function getDeployedStacks(
     };
   };
 
+  const limit = pLimit(20);
```
rix0rrr, line 148: Still seems like a high number?

Are the retry settings configured thoroughly enough for the CFN client?

otaviomacedo: This is the current retry strategy for the CFN client:

```ts
new ConfiguredRetryStrategy(11, (attempt: number) => 1000 * (2 ** attempt))
```

I ran an experiment with this strategy, against an account with 342 templates. I didn't start getting throttling errors until concurrency = 47. So, 20 might be playing it too safely, actually.

Just as a sanity check, without any retry strategy, even with concurrency = 2, I would sometimes get a throttling error.

rix0rrr: SGTM, thanks!

rix0rrr (approved):

PR #929, opened 2025-11-05
feat(integ-runner): add --strict flag to throw error on missing tests

abidhasan-aws:
## Description of the change
This PR adds a new flag `--strict` to integ-runner. When enabled, this flag enforces that all specified tests are found and can be run.

With `strict` = `true`, integ-runner will throw an error when any specified tests are not found.

## Reason for this change
For the [Integration Test Workflow](https://github.com/aws/aws-cdk/blob/main/.github/workflows/integration-test-deployment.yml), we want to throw an error if some specified tests are not found. This ensures that all tests that have been changed in a PR are actually run in the workflow.

Currently, integ-runner does not throw an error when tests are not found. We can use the `--strict` flag in the workflow for our use case.

## Test
Added unit tests to verify the functionality of the new flag.

---


mrgrain (approved):
LGTM

PR #933, opened 2025-11-13
fix: CloudFormation timeout if $AWS_ENDPOINT_URL_S3 is set during `cdk deploy`

rix0rrr:
The CLI asks the SDK for the URL to an S3 bucket, in order to pass it to CloudFormation. CloudFormation will then attempt to contact S3 on that URL in order to download the template.

A feature of the the SDK is to respect the `$AWS_ENDPOINT_URL_S3` environment variable, which can be used to override the S3 endpoint that the SDK will hit; you might use this if you have private VPC endpoints for a number of AWS services.

The problem arises that `$AWS_ENDPOINT_URL_S3` also affects the URL that the CDK CLI passes to CloudFormation. This will most likely be an endpoint that is not routable for CloudFormation like `https://vpce-xxx.s3.us-east-1.vpce.amazonaws.com`, and CloudFormation will time out waiting for the template download.

To get around this, we will temporarily unset `$AWS_ENDPOINT_URL_S3` for the duration of calling the SDK to provide us with a URL.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/template-body-parameter.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/template-body-parameter.ts
@@ -99,9 +99,21 @@ export async function makeBodyParameter(
  *
  * Replaces environment placeholders (which this field may contain),
  * and reformats s3://.../... urls into S3 REST URLs (which CloudFormation
- * expects)
+ * expects).
+ *
+ * We need to return the official region- and partition-specific URL for AWS S3
+ * here, so we use the SDK's information about endpoints. At the same time, the
+ * SDK allows overriding this URL by setting an environment variable
+ * (specifically $AWS_ENDPOINT_URL_S3) but we want to *not* honor that, because
+ * there's a 99.9% chance this URL will not be routable from AWS CloudFormation.
+ *
+ * To allow for the off chance that someone is running this tool against a
+ * custom build of CloudFormation that does need a specific S3 endpoint passed
+ * to it, we'll introduce a new environment variable that we'll respect instead:
+ *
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/template-body-parameter.ts
```diff
@@ -99,9 +99,21 @@ export async function makeBodyParameter(
...
+ * here, so we use the SDK's information about endpoints. At the same time, the
+ * SDK allows overriding this URL by setting an environment variable
+ * (specifically $AWS_ENDPOINT_URL_S3) but we want to *not* honor that, because
+ * there's a 99.9% chance this URL will not be routable from AWS CloudFormation.
+ *
+ * To allow for the off chance that someone is running this tool against a
+ * custom build of CloudFormation that does need a specific S3 endpoint passed
+ * to it, we'll introduce a new environment variable that we'll respect instead:
+ *
+ *  AWS_ENDPOINT_URL_S3_FOR_CLOUDFORMATION
```
mrgrain, line 114: (blocking): I'm leaning against the proliferation of introducing undocumented env vars doing things. I'm guessing this was added to preempt new issues?  Let me reason this through:
- Currently when `AWS_ENDPOINT_URL_S3` is set it will break deployment
- This PR fixes that
- If someone deliberately used `AWS_ENDPOINT_URL_S3` to do _something_ unknown that we do not know about, they will be broken now.
- How do they find the new env var? I say they won't and they will open an issue anyway.

I vote we only address this if someone has this problem.

mrgrain: Discussed offline. This might help use cases like OpenStack and LocalStack.

mrgrain (approved):

Issue #624, opened 2025-06-17
Bootstrap: `--region` flag gets ignored by `cdk bootstrap`


garysassano:
### Describe the bug

The CDK CLI silently ignores the `--region` flag during bootstrapping. It doesn't throw an error or warning, just proceeds as if the flag wasn't provided.

```sh
❯ cdk bootstrap --region ap-southeast-1
 ⏳  Bootstrapping environment aws://XXXXXXXXXXXX/ap-southeast-2...
Trusted accounts for deployment: (none)
Trusted accounts for lookup: (none)
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

See above.

### Current Behavior

See above.

### Reproduction Steps

See above.

### CDK CLI Version

2.1018.0

### Node.js Version

22.16.0

### OS

Ubuntu 24.04

### Language

TypeScript

pahud:
Hi @garysassano 

After my investigation, the CLI is **silently ignoring** the global `--region` flag in multiple scenarios:

**Problem Scenarios:**

1. **Silent ignore with global region only:**
   bash
  cdk bootstrap --region ap-southeast-1  # Ignores --region, uses default


2. **Silent ignore with conflicting regions:**
   bash
  cdk bootstrap aws://123/us-east-1 --region ap-southeast-1  # Should error but doesn't



**Root Cause:**
The [bootstrap method](https://github.com/aws/aws-cdk-cli/blob/da1259648ae861aef8b4f828ed85a2b8000722cd/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L1042) completely ignores global CLI options. Unlike other commands that properly validate conflicting options (like
--execute vs --method), bootstrap has no validation for region conflicts.

**Expected Behavior:**
- When only --region is specified: Use that region for bootstrap
- When both --region and explicit environment are specified with different regions: **Throw an error**
- When both match: Work normally


**The Fix Should:**
1. Add region conflict validation that throws a ToolkitError when --region conflicts with explicit environment regions
2. Use global --region as fallback when no explicit environment is specified
3. Provide clear error messages about the conflict
4. If we don't accept --region, the CLI should throw the error.

This is definitely a bug that needs fixing - the silent ignore behavior is confusing and potentially dangerous. I am making it 1 p1 and bringing it up to the team for review.

iankhou:
[...]

PR #949, opened 2025-11-18
fix(integ-runner): region is not passed to CDK App

rix0rrr:
When using the `toolkit-lib` deployment engine, the selected region was not being passed to the CDK app.

In the old engine, we could set the `AWS_REGION` environment variable and the CDK CLI would respect that, but the toolkit lib was passing that environment variable directly to the CDK app which was not respecting that. Fix that by threading the region through as an explicit parameter and adding an override parameter to the `AwsCliCompatible` base credentials: we don't necessarily want to read this only from the INI files.

Additionally: add a validation to make sure that the CDK app respects the region we are requesting. CDK integ tests should either not pass an environment, or read `$CDK_DEFAULT_REGION` to determine the current region. It was too easy to just override the region with a hardcoded one, but that breaks the integ runner's attempts to isolate tests across regions. Guard against that by validating.

Also in this PR: disable the update workflow. I've lost 2 hours debugging why my deployment was failing with nonsenical values and the answer was: the checked-in snapshot was nonsensical, and the update workflow will first deploy the checked-in snapshot and only then deploy the current snapshot. Good idea, bad execution: if existing snapshots are not region-agnostic they are guaranteed to fail. We should probably resynthesize them from the previous git commit instead. Whatever it is, what we do right now doesn't work so I'm disabling it.

Fixes #

---

### DIFF
--- a/packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
+++ b/packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
@@ -1,8 +1,9 @@
 import * as path from 'node:path';
 import type { DeployOptions, ICdk, ListOptions, SynthFastOptions, SynthOptions, WatchEvents } from '@aws-cdk/cdk-cli-wrapper';
 import type { DefaultCdkOptions, DestroyOptions } from '@aws-cdk/cloud-assembly-schema/lib/integ-tests';
-import type { DeploymentMethod, ICloudAssemblySource, IIoHost, IoMessage, IoRequest, NonInteractiveIoHostProps, StackSelector } from '@aws-cdk/toolkit-lib';
-import { ExpandStackSelection, MemoryContext, NonInteractiveIoHost, StackSelectionStrategy, Toolkit } from '@aws-cdk/toolkit-lib';
+import { UNKNOWN_REGION } from '@aws-cdk/cx-api';
+import type { DeploymentMethod, ICloudAssemblySource, IIoHost, IoMessage, IoRequest, IReadableCloudAssembly, NonInteractiveIoHostProps, StackSelector } from '@aws-cdk/toolkit-lib';
+import { BaseCredentials, ExpandStackSelection, MemoryContext, NonInteractiveIoHost, StackSelectionStrategy, Toolkit } from '@aws-cdk/toolkit-lib';
 import * as chalk from 'chalk';
 import * as fs from 'fs-extra';
 
@@ -27,6 +28,11 @@ export interface ToolkitLibEngineOptions {
    * @default false
    */
   readonly showOutput?: boolean;
+
+  /**
+   * The region the CDK app should synthesize itself for
+   */
+  readonly region: string;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -256,6 +268,37 @@ export class ToolkitLibRunnerEngine implements ICdk {
...
+   *   region: process.env.CDK_DEFAULT_REGION,
+   * }
+   * ```
+   *
+   * This catches that misconfiguration.
+   */
+  private async validateRegion(asm: IReadableCloudAssembly) {
+    for (const stack of asm.cloudAssembly.stacksRecursively) {
+      if (stack.environment.region !== this.options.region && stack.environment.region !== UNKNOWN_REGION) {
+        throw new Error(`Stack ${stack.displayName} synthesizes for region ${stack.environment.region}, even though ${this.options.region} was requested. Please configure \`{ env: { region: process.env.CDK_DEFAULT_REGION, account: process.env.CDK_DEFAULT_ACCOUNT } }\`, or use no env at all. Do not hardcode a region or account.`);
```
Abogical: Won't this fail for cross-region integ tests?

garysassano:
I  was told this is not a bug just a couple of months ago: https://github.com/aws/aws-cdk-cli/issues/624#issuecomment-3260341445

packages/@aws-cdk/integ-runner/lib/cli.ts
```diff
@@ -100,7 +100,7 @@ export function parseCliArgs(args: string[] = []) {
     clean: argv.clean as boolean,
     force: argv.force as boolean,
     dryRun: argv['dry-run'] as boolean,
-    disableUpdateWorkflow: argv['disable-update-workflow'] as boolean,
+    disableUpdateWorkflow: true,
```
mrgrain: Let's make this a separate PR please, so we can have a separate discussion and a separate changelog entry.

mrgrain (changes requested):

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
rix0rrr: Yes it will. In fact cross-region integ tests won't work properly with an external system that imposes a region at all.

What do you suggest instead?

packages/@aws-cdk/integ-runner/lib/cli.ts
rix0rrr: Split off to here: https://github.com/aws/aws-cdk-cli/pull/951

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
Abogical: For now, we can throw a warning instead of an error so we don't make integ test workflows fail unnecessarily.

packages/@aws-cdk/integ-runner/lib/cli.ts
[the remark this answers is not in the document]
Abogical: We can change the default to true, but not always set it to true so it can be used in the future.

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
rix0rrr: Mmkay. My personal life philosophy is: no one looks at warnings. So that will do literally diddly squat.

But okay, I'll make it a warning.

mrgrain (approved):

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
Abogical: I share that philosophy too. Issue is that we currently have cross-region integ tests and this will block PR merges in aws-cdk whenever we update cross-region integ test snapshots.

PR #970, opened 2025-12-01
feat: show early validation errors on deploy

otaviomacedo:
CloudFormation has [recently launched early validation][1]. When creating the change set, CloudFormation validates a few things such as whether the deploy would try to create resources with existing physical IDs. More validations to come. But the specific information about the error (other than it's an early validation error) is not returned by the `DescribeChangeSet` API. The consumer is supposed to call a new API, `DescribeEvents` to get that information. 

Introduce a new class,`EarlyValidationReporter` that checks whether such an error occurred and call the `DescribeEvents` API. The bootstrap stack was also updated to include permission to this new API. 

If the role being used for deployment doesn't have permission to call `DescribeEvents`, the toolkit will warn the user that it cannot show the proper error message, and suggest a re-bootstrap.

[1]: https://aws.amazon.com/about-aws/whats-new/2025/11/cloudformation-dev-test-cycle-validation-troubleshooting/

---


packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
```diff
@@ -0,0 +1,60 @@
...
+import type { ValidationReporter } from './cfn-api';
+import { ToolkitError } from '../../toolkit/toolkit-error';
+import type { SDK } from '../aws-auth/sdk';
+import type { EnvironmentResources } from '../environment/index';
+
+/**
+ * Reports early validation failures from CloudFormation ChangeSets. To determine what the actual error is
+ * it needs to call the `DescribeEvents` API, and therefore the deployment role in the bootstrap bucket
+ * must have permissions to call that API. This permission was introduced in version 30 of the bootstrap template.
+ * If the bootstrap stack is older than that, a special error message is thrown to instruct the user to
```
rix0rrr: Let's do a warning, not a throw.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Checks whether the ChangeSet failed early validation, and if so, throw an error. Otherwise, do nothing.
+   * @param description - the ChangeSet description, resulting from the call to CloudFormation's DescribeChangeSet API
+   * @param changeSetName - the name of the ChangeSet
+   * @param stackName - the name of the stack
+   */
+  public async check(description: DescribeChangeSetCommandOutput, changeSetName: string, stackName: string) {
+    if (description.Status === ChangeSetStatus.FAILED && description.StatusReason?.includes('AWS::EarlyValidation')) {
+      await this.checkBootstrapVersion();
+      const eventsOutput = await this.sdk.cloudFormation().describeEvents({
```
rix0rrr: This is making an assumption that we'll always find the event in the first page, without pagination.

Is that true? At least worth writing down the assumption.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
```diff
@@ -0,0 +1,60 @@
...
+        ChangeSetName: changeSetName,
+        StackName: stackName,
+      });
+
+      const failures = (eventsOutput.OperationEvents ?? [])
+        .filter((event) => event.ValidationStatus === ValidationStatus.FAILED)
+        .map((event) => `  - ${event.ValidationStatusReason} (at ${event.ValidationPath})`)
+        .join('\n');
```
rix0rrr: This might include events from multiple deployments, as long as they fit on one page.

We should only check events from the last deployment (perhaps even locate them based on an identifier, if that's feasible at all)

packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
```diff
@@ -0,0 +1,60 @@
...
+    let bootstrapVersion: number | undefined = undefined;
+    try {
+      // Try to get the bootstrap version
+      bootstrapVersion = (await this.environmentResources.lookupToolkit()).version;
+    } catch (e) {
+      // But if we can't, keep going. Maybe we can still succeed.
+    }
+    if (bootstrapVersion != null && bootstrapVersion < 30) {
+      const env = `aws://${environment.account}/${environment.region}`;
+      throw new ToolkitError(
```
rix0rrr: Warning please

rix0rrr (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
otaviomacedo: The toolkit deletes the old change set before creating a new one. And when a change set is deleted, all events are also deleted. So when a new one is created, and we are looking at its description here, it will contain only the events for this deployment.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -121,10 +125,17 @@ export async function waitForChangeSet(
       return description;
     }
 
-    // eslint-disable-next-line @stylistic/max-len
-    throw new ToolkitError(
-      `Failed to create ChangeSet ${changeSetName} on ${stackName}: ${description.Status || 'NO_STATUS'}, ${description.StatusReason || 'no reason provided'}`,
-    );
+    if (description.Status === ChangeSetStatus.FAILED && description.StatusReason?.includes('AWS::EarlyValidation')) {
+      await validationReporter?.report(changeSetName, stackName);
+      return description;
```
rix0rrr: Why not make this an error, containing all early validation messages, like the "regular" fail case?

packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  public async report(changeSetName: string, stackName: string) {
+    let operationEvents: OperationEvent[] = [];
+    try {
+      operationEvents = await this.getFailedEvents(stackName, changeSetName);
+    } catch (error) {
+      const message =
+        'While creating the change set, CloudFormation detected errors in the generated templates,' +
+        ' but the deployment role does not have permissions to call the DescribeEvents API to retrieve details about these errors.\n' +
```
rix0rrr: At least include the original `error` in this message. This makes a guess as to what the error is.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
rix0rrr: Oh, `report` will maybe throw an error? This is confusing to me.

Also, if `report` doesn't throw an error because looking up fails, this now doesn't throw at all. That doesn't seem to be what we want.

rix0rrr: Why not this style:

```ts
if (isEarlyValidationFailure) {
  const details = reporter?.fetchDetails();
  if (details) {
    throw new ToolkitError(`Early validation failed with ${details}`);
  }
}
throw new Error(`Failed to create changset...`);
```

packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
rix0rrr: How about something like:

> The template cannot be deployed because of early validation errors, but retrieving more details about those
> errors failed (${error}). Make sure you have permissions to call the DescribeEvents API, or re-bootstrap
> your environment with the latest version of the CLI (need at least version 30, current version ${currentVersion ?? 'unknown'}).

packages/@aws-cdk/toolkit-lib/lib/api/deployments/early-validation.ts
rix0rrr: Ahh I see. Nice.

rix0rrr (approved):

PR #973, opened 2025-12-04
fix(cli): hotswap ECS flags result in `SerializationException: STRING_VALUE cannot be converted to Integer`

dysosmus:
The PR https://github.com/aws/aws-cdk-cli/pull/417 introduced extra hotswap flags. However, when used the flags yield serialisation errors like bellow because the expected type is a number not a string.

```
failed: SerializationException: STRING_VALUE cannot be converted to Integer
```

This PR changes hotswap-ecs-minimum-healthy-percent, hotswap-ecs-maximum-healthy-percent, and hotswap-ecs-stabilization-timeout-seconds from string to expected type number type.

---


mrgrain (changes requested):
Build is failing and really we should have a test case here demonstrating the error is fixed.

mrgrain (approved):
Agreed, thanks for the fix!

PR #983, opened 2025-12-07
feat(cli): concurrent asset builds

### LINKED ISSUE #226: cli: allow to parallelize asset build
### Describe the feature

Currently in `cdk deploy`, assets are built serially one by one, making it slow when there are many container image assets.

The current behavior assumes that asset build is CPU-bound and will not get benefit from parallelism:

https://github.com/aws/aws-cdk/blob/bbdd42c8f45916d5c6945f3429916f6199d2ec66/packages/aws-cdk/lib/cdk-toolkit.ts#L559-L563

However, in our use cases, our assets are not fully cpu-bound and will actually get faster when parallelized. That is why I want to configure the concurrency parameter by CLI arguments.

### Use Case

When a CDK app contains many docker images whose build process are not fully CPU-bound, we can make cdk deploy faster by parallelizing the asset build processes.

You can see the outcome by the following extreme example:

```dockerfile
# docker/Dockerfile
FROM nginx
ARG DUMMY_ARG
RUN echo ${DUMMY_ARG}
# simulating IO-bound image
RUN sleep 10
```

And this CDK code:

```ts
// stack.ts
import { DockerImageCode, DockerImageFunction } from 'aws-cdk-lib/aws-lambda';

export class SlowDockerParallelTestStack extends cdk.Stack {
    super(scope, id, props);

    Array(5)
      .fill(0)
      .map((_, i) => {
        new DockerImageFunction(this, `Function${i}`, {
          code: DockerImageCode.fromImageAsset('./docker', {
            buildArgs: {
              DUMMY_ARG: `${i}_v1`,
            },
          }),
        });
      });
  }
}
```

When you run cdk deploy, you can see the five images are built serially, requiring more than 50 seconds to finish deployment.

Currently you can configure the parallelism parameter by directly editing `node_modules/aws-cdk/lib/index.js`, search for `{"stack":concurrency,"asset-build":1` and replace it with `{"stack":concurrency,"asset-build":5`.

After updateing the parameter, run cdk deploy again (make sure to change `DUMMY_ARG` to invalidate caches), and all the images are built concurrently. It now takes about 10 seconds.

### Proposed Solution

Expose CLI argument like `asset-build-concurrency`. Default is 1, and set the value here.

https://github.com/aws/aws-cdk/blob/bbdd42c8f45916d5c6945f3429916f6199d2ec66/packages/aws-cdk/lib/cdk-toolkit.ts#L559-L563

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.174.1

macOS

go-to-k:
Fixes #226

This PR adds the `--asset-build-concurrency` option to `cdk deploy` to parallelize asset build.

```sh
cdk deploy --asset-build-concurrency 4
```

---


packages/aws-cdk/lib/legacy/configuration.ts
```diff
@@ -307,6 +307,7 @@ function commandLineArgumentsToSettings(argv: Arguments): Settings {
     rollback: argv.rollback,
     notices: argv.notices,
     assetParallelism: argv['asset-parallelism'],
+    assetBuildConcurrency: argv['asset-build-concurrency'],
```
go-to-k, line 310:
```suggestion
```
Maybe we don't need this?

xrn:
@otaviomacedo @mrgrain it would be extremely useful to have this feature

mrgrain (approved):

PR #984, opened 2025-12-10
feat(cli): warn of non-existent stacks in `cdk destroy`

### LINKED ISSUE #32836: cli: CDK destroy pretends no stack are found in the main cloud assembly
### Describe the bug

Hello,

Since the version 2.175.0 of the CLI we can no longer run cdk destroy.

```
Run cdk destroy --force true 'Dev*/*'

No stack found in the main cloud assembly. Use "list" to print manifest
```

While cdk list does yield some result:
```
`cdk list` 
  
LocalStage/NetworkStack (LocalStage-NetworkStack)
DevStage/NetworkStack (DevStage-NetworkStack)
```
The problem is not related to the wildcard nor the specific stack naming.
(Same result with `cdk destroy --all` or `cdk destroy 'DevStage/NetworkStack'`

One specificity that might explain the issue, we are using the cdk.Stage class to organize our stacks, but we are not using the CodePipeline constructs (our CI/CD runs outside of AWS and calls the cdk commands using the wildcards)

We didn't try without this code structuration.

### Regression Issue

- [X] Select this option if this issue appears to be a regression.

2.174.1

### Expected Behavior

Similar to 2.174.1 and before, it is expected to run the stack destruction for the given stacks (or all if --all is specified)

### Current Behavior

the `cdk destroy` commands yield an error:
`No stack found in the main cloud assembly. Use "list" to print manifest`

### Reproduction Steps

```
import * as s3 from 'aws-cdk-lib/aws-s3';
import {Construct} from 'constructs';

const devEnv = {account: 'XXXXXXXXXXXX', region: 'eu-central-1'};
const app = new cdk.App();

export class DevStage extends cdk.Stage {
  constructor(scope: Construct, id: string, props: cdk.StageProps) {
    super(scope, id, props);

    const networkStack = new NetworkStack(this, 'NetworkStack', {
      env: props.env
    });
  }
}

export class NetworkStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props: cdk.StackProps) {
    super(scope, id, props);

    new s3.Bucket(this, `MyBucket`, {
      autoDeleteObjects: true,
      removalPolicy: cdk.RemovalPolicy.DESTROY
    });
  }
}

new DevStage(app, 'DevStage', {env: devEnv});
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.175.0

### Framework Version

_No response_

### Node.js Version

20.18.1

### OS

Ubuntu 24.04

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

go-to-k:
Fixes https://github.com/aws/aws-cdk/issues/32836, https://github.com/aws/aws-cdk/issues/32545, https://github.com/aws/aws-cdk/issues/27179, https://github.com/aws/aws-cdk/issues/22240

## Reason for this change

This PR implements the feature that warns users when non-existent stacks are specified in `cdk destroy`.

* It does not display the message `Are you sure you want to delete:` if there is no matching stack.
* Even if the stack does not exist, `cdk destroy` will not fail, it will just print a warning.

For examples (that have `Stacka`, `StackA`, `StackX`):

destroy2
<br>
destroy3
<br>
destroy4

## Difference from previous PR

The [previous PR](https://github.com/aws/aws-cdk/pull/32636) was reverted in https://github.com/aws/aws-cdk/pull/32839 due to a regression with only nested stage stacks. So this version addresses that regression with comprehensive tests.

## Description of changes

The original implementation added warnings for non-existent stacks in `cdk destroy`, but it failed when applications had only nested stage stacks (no top-level stacks). This happened because the code used `allTopLevel: true`, which only searched for stacks directly under the App, ignoring stacks within nested Stages.

Fixed the regression by changing `suggestStacks` method to use `DefaultSelection.AllStacks` instead of `allTopLevel: true`, ensuring the warning feature works for all stack configurations (top-level only, nested only, or both). Added regression tests to verify the fix for the nested stage scenario.

```diff
  private async suggestStacks(props: {
    selector: StackSelector;
    stacks: StackCollection;
    exclusively?: boolean;
  }) {
    const assembly = await this.assembly();
    const selectorWithoutPatterns: StackSelector = {
-      ...props.selector,
-      allTopLevel: true,
      patterns: [],
    };
    const stacksWithoutPatterns = await assembly.selectStacks(selectorWithoutPatterns, {
      extend: props.exclusively ? ExtendedStackSelection.None : ExtendedStackSelection.Downstream,
-      defaultBehavior: DefaultSelection.OnlySingle,
+      defaultBehavior: DefaultSelection.AllStacks,
    });
```

## Additional Information

When running `cdk destroy --all` or `cdk deploy --all` against a configuration with **no top-level stacks (nested stages only)**, the following error occurs.

However, this behavior existed **before this PR** and is unrelated to the changes made here. Since it is outside the scope of this PR, no fix has been implemented for this behavior.

FYI: I have submitted an [issue](https://github.com/aws/aws-cdk-cli/issues/1003) and a [PR](https://github.com/aws/aws-cdk-cli/pull/985) about this behavior.

```ts
const app = new cdk.App();

new MyStage(app, 'MyStage'); // This has CdkSampleStack
```

```
> cdk deploy
Since this app includes more than a single stack, specify which stacks to use (wildcards are supported) or specify `--all`
Stacks: MyStage/CdkSampleStack

> cdk deploy --all
...
No stack found in the main cloud assembly. Use "list" to print manifest

> cdk destroy --all
...
No stack found in the main cloud assembly. Use "list" to print manifest
```

---


mrgrain:
@go-to-k I am looking to get this in soon (hence the rebase), but first want to change `destroy` to use toolkit-lib code internally, so we can avoid duplicate and complicated code.

go-to-k:
@mrgrain Done. Destroy output now flows through toolkit-lib, so it's coded and consistent with deploy (adds `Synthesis time` / `Destroy time`, and the success line is `result` level). User-facing text is unchanged. (see the diff for the snapshots: packages/aws-cdk/test/commands/\__io_snapshots__/destroy/xx.ndjson.)

go-to-k:
@mrgrain Also restored the confirmation prompt's stack-name color to blue: https://github.com/aws/aws-cdk-cli/pull/984/changes/46261305c9b427d08ec1aeffc8f1d97e3cf7b254

It had turned red because toolkit-lib's destroy used `chalk.red`, whereas the CLI used `chalk.blue` before.

Before: 
color-diff

After:
after

P.S. recommit: https://github.com/aws/aws-cdk-cli/pull/984/changes/d6661874ae9b099b84f3aa29442662096e2e4291

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1593,18 +1604,28 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
 
-    const motivation = 'Destroying stacks is an irreversible action';
-    const question = `Are you sure you want to delete: ${chalk.red(stacks.hierarchicalIds.join(', '))}`;
-    const confirmed = await ioHelper.requestResponse(IO.CDK_TOOLKIT_I7010.req(question, { motivation }));
-    if (!confirmed) {
-      await ioHelper.notify(IO.CDK_TOOLKIT_E7010.msg('Aborted by user'));
+    if (stacks.stackCount === 0) {
+      await ioHelper.notify(IO.CDK_TOOLKIT_W7011.msg(
+        `No stacks match the name(s): ${chalk.red((selectStacks.patterns ?? []).join(', '))}`,
+      ));
```
go-to-k, line 1787: Note on the exit code: no-match (W7011) exits 0, while a declined confirmation exits 1 (#1667). This is deliberate and keeps the existing behavior. The old `selectStacksForDestroy` had a `// No validation` comment and returned an empty selection, so `cdk destroy <nonexistent>` already exited 0 (deploy throws `NoStacksMatched`, destroy never did). We keep that contract and add a warning in place of the old empty prompt.

exit

go-to-k:
Merged your PR and fixed conflicts.

Now: 

cancel

go-to-k:
@mrgrain 

Drops this branch's own toolkit-lib migration in favor of upstream #1686, keeps only the destroy warning feature.

go-to-k:
@mrgrain Could you please take a look at my changes?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1692,6 +1702,39 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+
+    for (const pattern of patterns) {
+      const matched = selected.stackArtifacts.some((stack) => picomatch.isMatch(stack.hierarchicalId, pattern));
+      if (matched) {
+        continue;
+      }
+
+      const closeMatches = allStacks.stackArtifacts
+        .filter((stack) => picomatch.isMatch(stack.hierarchicalId.toLowerCase(), pattern.toLowerCase()))
+        .map((stack) => stack.hierarchicalId);
```
mrgrain, line 1886: This function to match candidates (basically to this line) should live on `StackAssembly` class. It should not log any output but return a `string[]`.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1692,6 +1702,39 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      return;
+    }
+
+    const allStacks = await assembly.selectStacksV2(ALL_STACKS);
+
+    for (const pattern of patterns) {
+      const matched = selected.stackArtifacts.some((stack) => picomatch.isMatch(stack.hierarchicalId, pattern));
+      if (matched) {
+        continue;
+      }
```
mrgrain, line 1882: Can you explain what purpose this is serving? I just don't understand it. We are excluding any of the stacks that already match the pattern themselves? But why?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1620,12 +1621,21 @@ export class Toolkit extends CloudAssemblySourceBuilder {
     const ioHelper = asIoHelper(this.ioHost, action);
     const stacks = await assembly.selectStacksV2(selectStacks);
 
+    await this.suggestStacks(ioHelper, assembly, selectStacks, stacks);
```
mrgrain, line 1778: I think we should only call this if the original selector returned 0 results.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: yes that's great!

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1620,12 +1621,21 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+
     const ret: DestroyResult = {
       stacks: [],
     };
 
+    if (stacks.stackCount === 0) {
+      await ioHelper.notify(IO.CDK_TOOLKIT_W7011.msg(
```
mrgrain, line 1785: I think both messages can be folded into a single message that informs the user that no matching stacks were found. And if we have identified possible close matches ("candidates"), then we suggest them ("Did you mean...").

The Toolkit message should include a payload of the unmatched pattern and the identified matches as `string[] | undefined`.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
```diff
@@ -398,6 +398,15 @@ export const IO = {
...
   }),
 
+  CDK_TOOLKIT_W7010: make.warn({
+    code: 'CDK_TOOLKIT_W7010',
+    description: 'A provided stack name does not match any stack',
+  }),
+  CDK_TOOLKIT_W7011: make.warn({
+    code: 'CDK_TOOLKIT_W7011',
+    description: 'No stacks match the provided names, nothing to destroy',
+  }),
```
mrgrain, line 408: Like said elsewhere, I think a single message will be enough but it should carry a payload of the unmatched selector and idenfified matches.

mrgrain (changes requested):
Thanks for your continued work! FYI I want to let the changes to `destroy` bake a little bit longer to see if there's any issue from that.

But this PR is getting close to be ready now. Made some final comments on code organisation and messaging. We will also expect new tests in `packages/aws-cdk/test/commands/destroy.test.ts` and associated snapshots to record the exact output.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
go-to-k: This handles the case where several names are given and only some exist. E.g. `cdk destroy StackA Typo`: `StackA` is destroyed, `Typo` matches nothing. The `matched` check skips `StackA` and warns only about `Typo` ("Did you mean …?").

With "only warn when 0 stacks matched", a typo mixed in with valid names would slip through unnoticed, that's what I wanted to catch. That said, if you'd prefer a single message (only when nothing matched at all), I'm happy to fold it into one payload-carrying message. Which do you prefer?

mrgrain: Okay, thank you. I understand now what this is trying to achieve. Let me think about this for a bit. Some of my comments are now obviously contingent on this decision here.

go-to-k: Holding off until the direction above is settled, then I'll adjust accordingly.

go-to-k: @mrgrain How has it been since then?

go-to-k:
@mrgrain How has it been since then?

mrgrain (changes requested):
Right. I'd like to encapsulate the suggestion mechanism inside the `StackAssembly` class.

- We introduce a new `selectStacksV3` method that replaces the current `selectStacksV2`.
- This will get a new options bag as second argument. For now, we have one option `suggestPatternMatches: boolean // default: false`
- The return will change to an object: `stacks: StackCollection`
-When `suggestPatternMatches === true` the return value also includes `suggestions: Recod<pattern: string, string[]>` if the stack selector is a pattern selector and no stack was found for a given pattern.
- The existing `selectStacksV2` becomes a thin wrapper around `selectStacksV3` to maintain the current API. Let's not migrate existing usage just yet, we can do a follow up.
- In `_destroy` we start using `selectStacksV3`. We then use the information to craft messages. 

Start with this, then we can do another round. Sorry this took so long.

go-to-k:
@mrgrain 

okay, changed it.

mrgrain (approved):

mrgrain:
:shipit:

PR #985, opened 2025-12-10
fix(cli): deploy and destroy with `--all` option fail on apps with no top-level stacks

### LINKED ISSUE #1003: (cli): deploy and destroy with `--all` option fails on apps with no top-level stacks
### Describe the bug

When running `cdk destroy --all` or `cdk deploy --all` against a configuration with **no top-level stacks (nested stages only)**, the following error occurs. This PR fixes the behavior.

```ts
const app = new cdk.App();

new MyStage(app, 'MyStage'); // This has CdkSampleStack
```

```
> cdk deploy
Since this app includes more than a single stack, specify which stacks to use (wildcards are supported) or specify `--all`
Stacks: MyStage/CdkSampleStack

> cdk deploy --all # exits with status code 1 (error)
...
No stack found in the main cloud assembly. Use "list" to print manifest

> cdk destroy --all # exits with status code 1 (error)
...
No stack found in the main cloud assembly. Use "list" to print manifest
```

### Regression Issue

_No response_

### Expected Behavior

It can deploy/destroy all the stacks, or no errors occur, or a correct message / documentation is needed.

### Current Behavior

It doesn't deploy/destroy all the stacks, and it exits with status code 1 (error).

### Reproduction Steps

```ts
const app = new cdk.App();

new MyStage(app, 'MyStage'); // This has CdkSampleStack

export class MyStage extends Stage {
  constructor(scope: Construct, id: string, props?: StageProps) {
    super(scope, id, props);

    new CdkSampleStack(this, 'CdkSampleStack');
  }
}

```

```
> cdk deploy --all

> cdk destroy --all
```

### Possible Solution

- Make it deployable / destroyable
- Or any of the following approaches:
  - Prevent errors
  - Correct messages
  - Correct documentation

_No response_

### CDK CLI Version

2.1100.1

### Framework Version

_No response_

### Node.js Version

24

### OS

mac

### Language

TypeScript

### Language Version

_No response_

### Other information

# Is this a bug or by design?

## The error message by `cdk deploy/destroy` (without `--all`) on the app

> Since this app includes more than a single stack, specify which stacks to use (wildcards are supported) or specify `--all`

Since it says `specify --all` on the command, it may not be by design, or this message may not be appropriate.

## Code

The `--all` option says that `Deploy all available stacks` or `Deploy all stacks`.

see: https://github.com/aws/aws-cdk-cli/blob/cdk%40v2.1100.1/packages/aws-cdk/lib/cli/cli-config.ts#L141

see: https://github.com/aws/aws-cdk-cli/blob/cdk%40v2.1100.1/packages/%40aws-cdk/cloud-assembly-schema/lib/integ-tests/commands/common.ts#L34-L41

```
'all': { type: 'boolean', desc: 'Deploy all available stacks', default: false },
```

```ts
  /**
   * Deploy all stacks
   *
   * Requried if `stacks` is not set
   *
   * @default - false
   */
  readonly all?: boolean;
```

The `--all` is passed into `allTopLevel`.

see: https://github.com/aws/aws-cdk-cli/blob/cdk%40v2.1100.1/packages/aws-cdk/lib/cli/cli.ts#L244

```ts
    const selector: StackSelector = {
      allTopLevel: args.all,
      patterns: args.STACKS,
    };
```

But the `allTopLevel` says that `all stacks at the top level assembly`. Looking at the doc, it seems like this is by design.

see: https://github.com/aws/aws-cdk-cli/blob/cdk%40v2.1100.1/packages/aws-cdk/lib/cxapp/cloud-assembly.ts#L80

```ts
export interface StackSelector {
  /**
   * Whether all stacks at the top level assembly should
   * be selected and nothing else
   */
  allTopLevel?: boolean;
```

There also appears to be a test that verifies this.

see: https://github.com/aws/aws-cdk-cli/blob/cdk%40v2.1100.1/packages/aws-cdk/test/cxapp/cloud-assembly.test.ts#L7-L18

```ts
test('select all top level stacks in the presence of nested assemblies', async () => {
  // GIVEN
  const cxasm = await testNestedCloudAssembly();

  // WHEN
  const x = await cxasm.selectStacks({ allTopLevel: true, patterns: [] }, { defaultBehavior: DefaultSelection.AllStacks });

  // THEN
  expect(x.stackCount).toBe(2);
  expect(x.stackIds).toContain('witherrors');
  expect(x.stackIds).toContain('withouterrors');
});
```

## If it is by design

However, if it is by design that the Stage stacks cannot be deployed or deleted with `--all`, it is a bit confusing.

There is no documentation to understand that, and the error message says to specify `--all`. Furthermore, it seems like a bug that it terminates abnormally instead of normally.

Isn't some kind of action necessary?

go-to-k:
Fixes: https://github.com/aws/aws-cdk-cli/issues/1003

---


rix0rrr (changes requested):
I don't think this behavior is desirable or intuitive. 

The constraints are:

* `--all` was introduced before Stages existed. When Stages were introduced, it would continue to select stacks only in the top-level assembly.
* Stages are intended to be deployed using a pipeline in the same application. For that use case, you don't really want to select them using `cdk deploy`. 
* I think having a "sometimes `cdk deploy` deploys stacks in stages and sometimes not" kind of behavior will be extremely confusing.

I understand this might be desirable for your specific use case, but is your use case represenative?

As written on your issue, 

```
$ cdk deploy '**' 
```

Should solve your particular problem exactly, and it's not that much more to type than `--all`.

-------

Don't get me wrong: I'm not against making changes here altogether. But I think it should be a co-operation between the CDK app and the CLI. Perhaps a Stack gets a new option that interplays with a new selection mechanism:

```ts
new Stack(this, 'SomeStack', {
  deployUsingCli: WITH_ALL_FLAG | ONLY_IF_EXPLICITLY_SELECTED | NEVER,
});
```

And then we set appropriate defaults to retain existing behavior.

I'm also not opposed to changing the documentation to better reflect actual behavior.

But as-is, I think changing this behavior in the CLI only will get too confusing.

PR #988, opened 2025-12-11
feat(cli): stabilize `--telemetry-file` option

kaizencc:
`--telemetry-file` option no longer needs to be included with `--unstable=telemetry`.

---


packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -100,10 +100,6 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
     caBundlePath: configuration.settings.get(['caBundlePath']),
   });
 
-  if (argv['telemetry-file'] && !configuration.settings.get(['unstable']).includes('telemetry')) {
```
iliapolo, line 103: Will running with `--unstable=telemetry-file` still work?

kaizencc: yes, we're only deleting a check. but also i tested it

iliapolo (approved):

PR #997, opened 2025-12-16
fix(cli): drop invalid commands from telemetry

iliapolo:
We should only capture CLI sessions that were invoked with a valid subcommand.

---


packages/aws-cdk/test/cli/io-host/cli-io-host.test.ts
```diff
@@ -293,7 +303,7 @@ describe('CliIoHost', () => {
       telemetryIoHost = CliIoHost.instance({
         logLevel: 'trace',
       }, true);
-      await telemetryIoHost.startTelemetry({ '_': 'init', 'telemetry-file': telemetryFilePath }, new Context());
+      await telemetryIoHost.startTelemetry({ '_': ['init'], 'telemetry-file': telemetryFilePath }, new Context());
```
iliapolo, line 306: This was wrong before - just didn't impact test assertion

rix0rrr (approved):

PR #1020, opened 2026-01-11
feat(cli): add publish-assets command for asset publishing without deployment

### LINKED ISSUE #982: (cli): Add cdk publish command for asset publishing
### Describe the feature

Add a new cdk publish CLI command that uploads synthesized assets (container images, Lambda bundles, file assets, etc.) to their respective destinations without performing a deployment.

### Use Case

In CI/CD pipelines, the recommended best practice is to separate the build/publish phase from the deployment phase:

Synthesize (cdk synth) - Generate CloudFormation templates and asset manifests
Publish - Upload all assets (Docker images, Lambda ZIPs, S3 files) to their destinations
Deploy - Deploy stacks sequentially across stages/environments (without re-uploading assets)

Currently, cdk deploy combines asset publishing and deployment into a single operation. This creates challenges in enterprise CI/CD pipelines:

* Immutable artifacts: Assets should be published once and reused across multiple deployments (dev → staging → production)
* Separation of concerns: Build/publish should complete before any deployment begins
* Reduced deployment time: When deploying to multiple stages, assets only need to be uploaded once
* Atomic releases: All assets are verified to be published before any deployment starts
* Pipeline structure: Most CI/CD systems (GitHub Actions, GitLab CI, CodePipeline, Jenkins) work best with distinct publish and deploy stages

The only way to achieve this workflow today is by installing and using the cdk-assets package directly:

This approach has drawbacks:

* Requires an additional dependency
* Not well documented in the main CDK documentation
* The API and usage is less discoverable than built-in CLI commands
* Users must understand the internal asset manifest structure

### Proposed Solution

Add a cdk publish command to the CDK CLI.

### Other Information

* This aligns with the CDK Pipelines concept where assets are published before deployment
* The underlying functionality already exists in cdk-assets - this would expose it as a first-class CLI command

If this is implemented, it would be also beneficial to have an option in the cdk deploy command to skip publishing during the deploy step. 

### Acknowledgements

### CDK version used

2.x

all

go-to-k:
### `cdk publish-assets`

Publishes assets for the specified stack(s) without performing a deployment.

> [!IMPORTANT]
> `cdk publish-assets` is under development and therefore must be opted in via the `--unstable` flag: `cdk publish-assets --unstable=publish-assets`. `--unstable` indicates that the scope and API of feature might still change. Otherwise the feature is generally production ready and fully supported.

Publishes assets (such as Docker images and file assets) for the specified stack(s) to their respective destinations (ECR repositories, S3 buckets) without performing a deployment.

This is useful in CI/CD pipelines where you want to separate the asset build/publish phase from the deployment phase.

```console
$ # Publish assets for a single stack
$ cdk publish-assets MyStack --unstable=publish-assets

$ # Publish assets for all stacks
$ cdk publish-assets --all --unstable=publish-assets

$ # Force re-publish even if assets already exist
$ cdk publish-assets MyStack --unstable=publish-assets --force
```

---

Currently, `cdk deploy` combines asset publishing and stack deployment into a single operation. This makes it difficult to structure CI/CD pipelines where assets should be built and published once, then reused across multiple deployment stages (e.g. dev → staging → production). The solution today is to use the standalone `cdk-assets` CLI directly, which requires an additional dependency and likely not easily discoverable.

This PR adds `cdk publish-assets` as a first-class CLI command, making the same functionality available without extra dependencies and with the familiar CDK CLI experience (stack selectors, `--exclusively`, `--all`, `--concurrency`). 

On the implementation side, the inline `buildAsset` and `publishAsset` functions in the deploy path have been extracted into reusable `createBuildAssetFunction` and `createPublishAssetFunction` methods on the `Toolkit` class, so both `deploy` and `publish-assets` share the same asset handling logic.

Fixes #982

---


mrgrain (changes requested):
This needs to be implemented in `toolkit-lib` and on the `Toolkit` class. The CLI should then just forward the call. See Drift and Refactor.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -244,6 +244,21 @@ export async function makeConfig(): Promise<CliConfig> {
           },
         },
       },
+      'publish': {
```
mrgrain: not sure about the name yet

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -244,6 +244,20 @@ export async function makeConfig(): Promise<CliConfig> {
...
+          name: 'STACKS',
+          variadic: true,
+        },
+        options: {
+          'all': { type: 'boolean', desc: 'Publish assets for all available stacks', default: false },
+          'exclusively': { type: 'boolean', alias: 'e', desc: 'Only publish assets for requested stacks, don\'t include dependencies' },
+          'force': { type: 'boolean', alias: 'f', desc: 'Always publish assets, even if they are already published', default: false },
+          'asset-parallelism': { type: 'boolean', desc: 'Whether to build/publish assets in parallel' },
+          'concurrency': { type: 'number', desc: 'Maximum number of simultaneous asset publishing operations (dependency permitting) to execute.', default: 1, requiresArg: true },
+        },
```
mrgrain, line 260: Isn't this the same setting?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -488,6 +489,105 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+        asset: assetNode.asset,
+      });
+      await deployments.publishSingleAsset(assetNode.assetManifest, assetNode.asset, {
+        stack: assetNode.parentStack,
+        roleArn: options.roleArn,
+        stackName: assetNode.parentStack.stackName,
+        forcePublish: options.force,
+      });
+      await publishAssetSpan.end();
+    };
```
mrgrain: Lot's of duplication with `_deploy()` here. Let's find a way to reuse the same code.

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: I see. Do you have any good ideas?

publish, build, release, upload, push, prepare, etc...

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: The `deploy` command's settings already have that configurations. The `asset-parallelism` determines "whether to parallelize or not," and `concurrency` determines the number of parallel processes.
However, they certainly look similar. I removed `asset-parallelism`.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
go-to-k: Created `createBuildAssetFunction` and `createPublishAssetFunction`.

go-to-k:
@mrgrain 

I've updated them based on your feedback.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1856,6 +1870,41 @@ export interface RollbackOptions {
...
+   * @default 1
+   */
+  readonly concurrency?: number;
+
+  /**
+   * Role to pass to CloudFormation for deployment
+   *
+   * @default - Current role
+   */
+  readonly roleArn?: string;
```
mrgrain: what is this used for?

packages/aws-cdk/lib/cli/cli-config.ts
mrgrain: I think we now also have concurrent building: https://github.com/aws/aws-cdk-cli/pull/983

I'm leaning towards for a single flag that handles both, but open to your views. Either way, description and code need to reflect that.

packages/aws-cdk/lib/cli/cli-config.ts
mrgrain: `cdk publish-assets` ?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -790,6 +791,19 @@ export class CdkToolkit {
     }
   }
 
+  public async publish(options: PublishOptions): Promise<void> {
+    await this.toolkit.publish(this.props.cloudExecutable, {
```
mrgrain: Nice!

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -488,6 +489,75 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+    if (assetNodes.length === 0) {
+      await ioHelper.defaults.info(chalk.green('\n✨  All assets are already published\n'));
+      return {
+        publishedAssets: [],
+      };
+    }
+
+    const graphConcurrency: Concurrency = {
+      'stack': 1,
+      'asset-build': 1,
```
mrgrain: Should we not allow control all for all of these? (I guess `stack` doesn't matter here)

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -488,6 +489,75 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+
+    await workGraph.doParallel(graphConcurrency, {
+      deployStack: async () => {
+        // No-op: we're only publishing assets, not deploying
+      },
+      buildAsset: this.createBuildAssetFunction(ioHelper, deployments, options.roleArn),
+      publishAsset: this.createPublishAssetFunction(ioHelper, deployments, options.roleArn, options.force),
+    });
+
+    await ioHelper.defaults.info(chalk.green('\n✨  Assets published successfully\n'));
```
mrgrain: This should be a numbered message with a data payload of  publishedAssets

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -488,6 +489,75 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+
+    if (!options.force) {
+      await removePublishedAssetsFromWorkGraph(workGraph, deployments, options);
+    }
+
+    const assetNodes = Object.values(workGraph.nodes)
+      .filter((n): n is AssetPublishNode => n.type === 'asset-publish');
+
+    if (assetNodes.length === 0) {
+      await ioHelper.defaults.info(chalk.green('\n✨  All assets are already published\n'));
```
mrgrain: give it a unique code pls

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -488,6 +489,75 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      ioHelper,
+      true, // prebuild all assets
+    ).build(stacksAndTheirAssetManifests);
+
+    if (!options.force) {
+      await removePublishedAssetsFromWorkGraph(workGraph, deployments, options);
+    }
+
+    const assetNodes = Object.values(workGraph.nodes)
+      .filter((n): n is AssetPublishNode => n.type === 'asset-publish');
```
mrgrain, line 536: Between these we should have a payloaded info message that says something like "Publishing assets" and has the list of asset manifests as payload.

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: I haven't seen any two-word subcommands so far, is that okay?

mrgrain (changes requested):
Almost there, thank you.

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: Agreed on having a single flag that handles both building and publishing. A couple of things to consider:

1. In `deploy`, we have separate `--concurrency` (for stacks), `--asset-build-concurrency` (for builds), and `--asset-parallelism` (for publish). Is it okay that `publish` takes a different approach with a single `--concurrency` controlling both build and publish?

2. In `deploy`, `asset-publish` defaults to 8 while `asset-build` defaults to 1. Since we're unifying them, what should the default be? For now, I'll set to 1.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
go-to-k: `roleArn` is passed through to [`buildSingleAsset`](https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts#L552) and `publishSingleAsset` via `createBuildAssetFunction` (`packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts:L1485`), following the same pattern as `deploy`. However, looking at the implementation, `buildSingleAsset` and `publishSingleAsset` don't actually use `roleArn` internally. It's defined in the [`AssetOptions`](https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts#L209) interface but not referenced in the method body. I'll remove it.

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: Changed:
https://github.com/aws/aws-cdk-cli/pull/1020/changes/f41f437da8dab9c23912cfddc32ced667403ac47

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
go-to-k: Changed:
https://github.com/aws/aws-cdk-cli/pull/1020/changes/f41f437da8dab9c23912cfddc32ced667403ac47

```ts
    const graphConcurrency: Concurrency = {
      'stack': 1,
      'asset-build': options.concurrency ?? 1,
      'asset-publish': options.concurrency ?? 1,
    };
```

packages/aws-cdk/lib/cli/cdk-toolkit.ts
go-to-k: Removed: https://github.com/aws/aws-cdk-cli/pull/1020/changes/577a95d3f22e541c56b7a183896f6b16086d7460

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
go-to-k: Added it as 94xx

https://github.com/aws/aws-cdk-cli/pull/1020/changes/4addfb887845b37fc595f659d0c8ed7f940939e5

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
go-to-k: Called with a `PublishAssetsPayload`.

https://github.com/aws/aws-cdk-cli/pull/1020/changes/e52431150976e334bb95d691e749477c130cd052

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
go-to-k: Called with `PublishResultPayload`:
https://github.com/aws/aws-cdk-cli/pull/1020/changes/9359fb99ac18f8962c92e0d2ed2b53ff97b12b16

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: Found one `cdk cli-telemetry`. I'll change it to `publish-assets` for now.

mrgrain (changes requested):
I think this is the last request: Can you add two more integ tests plase:

- publish the asset, than deploys; the deploy needs to assert that it detects the already uploaded assets
- publish assets, publish again should detect and assert nothing to do, publish again with `--force` should detect the re-upload.

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: Tried to change to `publish-assets`.

https://github.com/aws/aws-cdk-cli/pull/1020/changes/9d16908f23a703bbc45c4899d050bfab86f51735

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: Is this okay?

packages/@aws-cdk/toolkit-lib/lib/payloads/publish-assets.ts
```diff
@@ -0,0 +1,16 @@
...
+   */
+  readonly assets: IManifestEntry[];
+}
+
+export interface PublishAssetsResultPayload {
+  /**
+   * The publish assets result
+   */
+  readonly result: PublishAssetsResult;
+}
```
mrgrain: (blocking) These seems to be the exact same payload in the end, because

```ts
export interface PublishAssetsResult {
  /**
   * List of assets that were published
   */
  readonly publishedAssets: IManifestEntry[];
}
```

So really we have a single payload with a list of assets:

```ts
export interface AssetsPayload {
  /**
   * List of assets
   */
  readonly assets: IManifestEntry[];
}
```

let's unify and use this in both places.

packages/aws-cdk/lib/cli/cli-config.ts
mrgrain: Nice! @go-to-k any thoughts on a higher default? Maybe 4?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1887,6 +1900,34 @@ export interface RollbackOptions {
   readonly validateBootstrapStackVersion?: boolean;
 }
 
+export interface PublishAssetsOptions {
```
mrgrain: suggestion: can we not just use the toolkit-lib one for this? We can move the `exclusively` check into `cli.ts` and build a shared conversion helper somewhere.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -441,6 +441,18 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           orphanLogicalIds: args.orphan,
         });
 
+      case 'publish-assets':
+        ioHost.currentAction = 'publish-assets';
+        if (!configuration.settings.get(['unstable']).includes('publish-assets')) {
+          throw new ToolkitError('UnstablePublishAssets', 'Unstable feature use: \'publish-assets\' is unstable. It must be opted in via \'--unstable\', e.g. \'cdk publish-assets --unstable=publish-assets\'');
+        }
```
mrgrain, line 448: blocking: if it's unstable here, it also needs to be unstable in toolkit-lib and we have a helper for that there `requireUnstableFeature`. So really you only need to add it there.

packages/aws-cdk/README.md
```diff
@@ -19,6 +19,7 @@ The AWS CDK Toolkit provides the `cdk` command-line interface that can be used t
 | [`cdk synth`](#cdk-synth)                 | Synthesize a CDK app to CloudFormation template(s)                                |
 | [`cdk diff`](#cdk-diff)                   | Diff stacks against current state                                                 |
 | [`cdk deploy`](#cdk-deploy)               | Deploy a stack into an AWS account                                                |
+| [`cdk publish-assets`](#cdk-publish-assets)| Publish assets for stack(s) without deploying (unstable)                         |
```
mrgrain: nit: re-align rest of table

mrgrain (changes requested):
Sorry, found a few more things in the latest iteration.

Also this: https://github.com/aws/aws-cdk-cli/pull/1020/changes#r3016913489

packages/aws-cdk/README.md
go-to-k: yes: https://github.com/aws/aws-cdk-cli/pull/1020/changes/1d83be9852c4598d5224d2b7110c11b9e4efc436

packages/@aws-cdk/toolkit-lib/lib/payloads/publish-assets.ts
go-to-k: okay, unified
https://github.com/aws/aws-cdk-cli/pull/1020/changes/afd8f3bf5b8b28f4ccc28d1a76eb9631d18a8074

packages/aws-cdk/lib/cli/cli-config.ts
go-to-k: Sounds good. I considered 8 to align with the publish default, but 4 is probably safer in case of CPU-intensive builds. Let's go with 4.

https://github.com/aws/aws-cdk-cli/pull/1020/changes/cc2600c8e659cfda5510a62a55df7b07ca794bde

packages/aws-cdk/lib/cli/cli.ts
go-to-k: I see. Called that:
https://github.com/aws/aws-cdk-cli/pull/1020/changes/302582811b1f28fe8288e8278480a708a4119b73

packages/aws-cdk/lib/cli/cdk-toolkit.ts
go-to-k: Tried to use the toolkit-lib's type instead of it and to create a conversion method.
https://github.com/aws/aws-cdk-cli/pull/1020/changes/c6d0df48de12e564cdd2d7eb0c9c2452850455ee

mrgrain (approved):

PR #1028, opened 2026-01-16
feat: add terminal width support to `StringWriteStream`

orien:
### Context

The diff formatting library expects streams to provide an optional `columns` property (via the [`FormatStream` interface](https://github.com/aws/aws-cdk-cli/blob/af43d431455b87913fd5487c2ff932363b3b5fc8/packages/%40aws-cdk/cloudformation-diff/lib/format.ts#L17-L19)) to enable terminal width-aware table formatting. StringWriteStream currently lacks this property, causing the diff formatter to format tables without width constraints, leading to overflow and poor rendering in different terminal environments.

### Changes

- Added `columns` getter to StringWriteStream that dynamically reads terminal width from `process.stdout.columns`
- Getter returns current terminal width on every access, allowing it to reflect runtime changes such as terminal resizing
- Returns `undefined` in non-TTY environments (CI/CD pipelines, redirected output)
- The formatter will automatically use this property when calling `formatTable()` for IAM changes, security group changes, and other table-formatted output
- Added comprehensive unit tests

### Consequences

The [`Formatter` class](https://github.com/aws/aws-cdk-cli/blob/main/packages/@aws-cdk/cloudformation-diff/lib/format.ts) will automatically use this property when calling [`formatTable()`](https://github.com/aws/aws-cdk-cli/blob/af43d431455b87913fd5487c2ff932363b3b5fc8/packages/%40aws-cdk/cloudformation-diff/lib/format-table.ts#L10) for [IAM changes](https://github.com/aws/aws-cdk-cli/blob/af43d431455b87913fd5487c2ff932363b3b5fc8/packages/%40aws-cdk/cloudformation-diff/lib/format.ts#L409-L429), [security group changes](https://github.com/aws/aws-cdk-cli/blob/af43d431455b87913fd5487c2ff932363b3b5fc8/packages/%40aws-cdk/cloudformation-diff/lib/format.ts#L433-L440), and other table-formatted output

### Considerations

The `columns` getter reads from `process.stdout.columns` on every access, ensuring it always reflects the current terminal width. This design allows the stream to respond to terminal resize events without requiring additional event listeners or manual updates. The diff formatter will automatically pick up this property through the existing `FormatStream` interface.

Now, there's no guarantee that people will output the text collected in the `StringWriteStream` to stdout, but that is _my_ use case. 🤷 

---


packages/@aws-cdk/toolkit-lib/lib/api/streams.ts
```diff
@@ -8,8 +8,16 @@ import { Writable } from 'node:stream';
...
+  /**
+   * Terminal width in columns.
+   * Used by formatTable() to apply width constraints and prevent table overflow.
+   */
+  public columns: number | undefined;
+
   constructor() {
     super();
+    // Read terminal width from stdout to enable proper table formatting
+    this.columns = process.stdout.columns;
```
rix0rrr: Should probably be a getter, because the column count can change over time.

rix0rrr (changes requested):
Thanks. Does this mean we need to implement a specific interface that advertises columns? Who gets to take advantage of this?

rix0rrr (approved):

PR #1030, opened 2026-01-19
feat(cli): add --[no-]color flag to control output colorization

AndyBodnar:
This addresses issue aws/aws-cdk-cli#1031.

When calling the CDK CLI programmatically (e.g., from a subprocess), colors are automatically disabled because stdout isn't a TTY. This makes it difficult to capture colored output for tools that want to preserve formatting.

I've added a `--color` flag to indicate color output is desired. The `--no-color` flag takes priority if both are specified.

I also fixed a subtle bug where the existing `--no-color` flag wasn't actually being used - the color disabling was happening at module load time before arguments were even parsed. Now both flags work correctly and respect the priority order: `--no-color` > `--color` > `TTY` detection.

Tested locally and added unit tests covering all the new behavior.

---


mrgrain (dismissed):
see comment

rix0rrr (approved):

PR #1117, opened 2026-02-04
refactor: consolidate IO message to telemetry event logic

rix0rrr:
The logic to convert I/O messages to telemetry events used to be a 3-step one:

1. First, check if the I/O message is in a list of special messages
2. Second, look at the I/O message again in a `switch` statement to come up with an event type.
3. Finally, copy some data like `duration` and `error` from the message.

In this change we consolidate all steps into a single function:

```ts
function eventFromMessage(msg: IoMessage<unknown>): TelemetryEvent | undefined;
```

If the function returns an event, it's fully formed and ready to send. Otherwise the message does need to lead to a telemetry event. This removes the ability for two lists of special messages to get out of sync, and it moves the logic that deals with building a full event message into a single place which purely deals with the conversion. Also use the `IoMessageMaker<>.is()` function to assert the type of a message so we can't accidentally get a wrong type cast.

---

### DIFF
--- a/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
+++ b/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
@@ -10,10 +10,8 @@ import type { IoHelper, ActivityPrinterProps, IActivityPrinter } from '../../../
 import { asIoHelper, IO, isMessageRelevantForLevel, CurrentActivityPrinter, HistoryActivityPrinter } from '../../../lib/api-private';
 import { StackActivityProgress } from '../../commands/deploy';
 import { canCollectTelemetry } from '../telemetry/collect-telemetry';
-import type { EventResult } from '../telemetry/messages';
-import { CLI_PRIVATE_IO, CLI_TELEMETRY_CODES } from '../telemetry/messages';
-import type { EventType } from '../telemetry/schema';
-import { TelemetrySession } from '../telemetry/session';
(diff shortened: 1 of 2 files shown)


iankhou (approved):

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -618,19 +615,23 @@ function isNoticesMessage(msg: IoMessage<unknown>): msg is IoMessage<void> {
   return IO.CDK_TOOLKIT_I0100.is(msg) || IO.CDK_TOOLKIT_W0101.is(msg) || IO.CDK_TOOLKIT_E0101.is(msg) || IO.CDK_TOOLKIT_I0101.is(msg);
 }
 
-function isTelemetryMessage(msg: IoMessage<unknown>): msg is IoMessage<EventResult> {
-  return CLI_TELEMETRY_CODES.some((c) => c.is(msg));
-}
+function eventFromMessage(msg: IoMessage<unknown>): TelemetryEvent | undefined {
```
kaizencc, line 618: sure, but this ends up creating a behavior change that was intentional before -- if we add a new telemetry code for a new message, we will have to remember to modify this function, otherwise we just don't get the message sent.

previously, this was alleviated by use of a `switch`, intentionally throwing an error in this case. 

i see the value in doing everything in one go, but not in removing the const `CLI_TELEMETRY_CODES` and the subsequent `switch`

PR #1124, opened 2026-02-09
fix: collect metrics to help analyze performance problems

rix0rrr:
This PR changes telemetry to help diagnose and address performance problems.

The effective changes are as follows:

- We are seeing slow synth times.
  - Log the total number of assemblies and stacks to correlate and segment synth times and stack counts (is the relationship linear/nonlinear, are slow synth times due to huge applications, ...)
  - Log the (guessed) implementation language, to see if synth times behave differently between languages, if startup times make a lot of difference, etc.
- We are concerned about the time needed for agents to produce a successful CDK application.
  - Try to determine whether we are currently being run by an agent so we can segment on that.
  - Log the counts of warnings and errors, to see if deployment failure is or can be preceded/prevented by better error reporting at synth time.
- The library version we were collecting was a rough guess, do a better one once we have accurate information.
- Count the number of resources in a stack to correlate and segment deployment times by resource count. 

For consistency, the same counters changes are being made to `toolkit-lib`, even though those counters aren't being consumed by anything.

---


packages/aws-cdk/lib/cli/util/guess-agent.ts
```diff
@@ -0,0 +1,34 @@
+
+/**
+ * Guess whether we're being executed by an AI agent
+ *
+ * It's hard for us to say `false` for sure, so we only respond
+ * with `yes` or `don't know`.
+ */
+export function guessAgent(): true | undefined {
+  if ((process.env.AWS_EXECUTION_ENV ?? '').toLocaleLowerCase().includes('amazonq')) {
```
kaizencc: wonder if we should pre-empt searching for 'kiro' as a string also. we are not really protected in the case where an AI agent changes their 'signature'; it will just manifest as a drop in AI usage for us

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -155,3 +219,7 @@ function isAbortedError(error?: ErrorDetails) {
   }
   return false;
 }
+
+function mutable<A extends object>(x: A): { -readonly [k in keyof A]: A[k] } {
+  return x;
+}
```
kaizencc, line 225: are we *gasp* playing god here and turning immutable objects mutable?

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -27,6 +27,22 @@ export interface TelemetryEvent {
   readonly eventType: EventType;
   readonly duration: number;
   readonly error?: ErrorDetails;
+  counters?: Record<string, number>;
```
kaizencc: nit: why not readonly?

packages/@aws-cdk/toolkit-lib/lib/api/io/private/span.ts
```diff
@@ -161,26 +193,50 @@ class MessageSpan<S extends object, E extends SpanEnd> implements IMessageSpan<E
...
+    this.openTimers.clear();
+
+    const endInput = parseArgs<SpanEndArguments<E>>(x, y);
     const endMsg = endInput.message ?? util.format(this.timingMsgTemplate, this.definition.name, duration.asSec);
     const endPayload = endInput.payload;
 
     await this.notify(this.definition.end.msg(
       endMsg, {
         duration: duration.asMs,
+        ...(Object.keys(this.counters).length > 0 ? { counters: this.counters } : {}),
```
kaizencc, line 217: nit, doesn't matter: is there a reason you didn't want to send `counters: {}`?

packages/@aws-cdk/toolkit-lib/lib/api/io/private/span.ts
```diff
@@ -23,29 +27,35 @@ export interface SpanDefinition<S extends object, E extends SpanEnd> {
...
+ * `SpanEnd` represents fields that are added by the underlying `end` function.
+ *
+ * - In principle, the intersection of the underlying type and the SpanEnd keys
+ *   are made optional, the rest is the same as the underlying type.
+ * - If the expected type is subsumed by the SpanEnd type, then either void or
+ *   their intersection with all fields optional.
+ */
+type SpanEndArguments<T> = keyof T extends keyof SpanEnd
+  ? (Pick<Partial<SpanEnd>, keyof T & keyof SpanEnd> | void)
+  : Optional<T, keyof T & keyof SpanEnd>;
```
kaizencc, line 48: i read your comments but this was still greek to me. This is what Kiro said, if true, it was a lot simpler to understand:

```
In short: if the span type already covers everything SpanEnd needs, ending is 
no-arg. If the span type has extra fields beyond SpanEnd, you're 
forced to supply them when calling end().
```

packages/aws-cdk/test/cli/telemetry/session.test.ts
```diff
@@ -137,6 +137,54 @@ describe('TelemetrySession', () => {
...
+
+    // THEN
+    expect(clientEmitSpy).toHaveBeenCalledWith(expect.objectContaining({
+      project: expect.objectContaining({
+        language: 'basic',
+      }),
+    }));
+  });
+
+  test('attack agent', async () => {
```
kaizencc: ⚔️💥🛡️🗡️

oh do you mean attach :)

packages/aws-cdk/lib/cli/util/guess-agent.ts
```diff
@@ -0,0 +1,34 @@
+
+/**
+ * Guess whether we're being executed by an AI agent
+ *
+ * It's hard for us to say `false` for sure, so we only respond
+ * with `yes` or `don't know`.
```
kaizencc, line 6: It's hard for us to say 'yes' for sure too, since anyone can add an env variable like this.

packages/aws-cdk/lib/cli/util/guess-agent.ts
rix0rrr: And then we will detect it and fix it. But sure, I can add "kiro".

packages/aws-cdk/lib/cli/telemetry/session.ts
rix0rrr: Yes. There are good reasons for it.

* https://rust-unofficial.github.io/patterns/idioms/temporary-mutability.html
* https://clojure.org/reference/transients

packages/@aws-cdk/toolkit-lib/lib/api/io/private/span.ts
rix0rrr: Yeah, I like sending as little as possible.

packages/aws-cdk/lib/cli/telemetry/session.ts
rix0rrr: Sure.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/span.ts
rix0rrr: Kiro is not entirely correct. You *can* still, but don't *have to* override the implicit values with optionals.

packages/aws-cdk/lib/cli/util/guess-agent.ts
rix0rrr: True. That's why the name is `guess`

packages/aws-cdk/lib/cli/telemetry/session.ts
kaizencc: I've only ever seen this as re-defining the entire variable

kaizencc (approved):

PR #1131, opened 2026-02-10
fix(cli): glob patterns ignored in chokidar v4

### LINKED ISSUE #1146: (cli): watch not working on MacOS ^2.1100.1
### Describe the bug

After globally updating aws-cdk to `v2.1104.0`, `cdk watch` (and `cdk deploy --watch`) no longer trigger a deploy when TypeScript files change. No "watch is observing..." messages are printed.

### Regression Issue

2.1034.0

### Expected Behavior

CDK watches project files and redeploys when a file is modified. 

### Current Behavior

Files are not watched and nothing is triggered when files are modified. 

### Reproduction Steps

Update global aws-cdk:
```
npm install -g aws-cdk@2.1104.0
```

Call watch or deploy --watch:
```
cdk watch
```

Touch the stack file to trigger a deploy:
```
touch lib/*-stack.ts
```

### Possible Solution

The only fix was downgrading to `v2.1034.0`  

```
npm install -g aws-cdk@2.1034.0
```

Projects are TypeScript, scaffolded with `cdk init app --language=typescript`. Nothing special, just a few lambdas, S3 and CloudFront. MacOS Sequoia 15.7.3.

Same result with node `v24.13.0` and `v20.20.0`

Downgrading the global aws-cdk to version `2.1034.0` (`npm install -g aws-cdk@2.1034.0`) made search work again. 

These initial output watch messages never print with the recent aws-cdk versions. 

```
'watch' is observing directory 'lib' for changes
'watch' is observing directory 'lambda' for changes
'watch' is observing directory 'bin' for changes
```

### CDK CLI Version

2.1104.0

### Framework Version

_No response_

### Node.js Version

24.13.0 and 20.20.0

### OS

MacOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

kaizencc:
Fixes aws/aws-cdk-cli#1146.

In CDK CLI v2.1103.0, we released aws/aws-cdk-cli#1059 which migrated our chokidar dependency from v3 to v4. We missed a breaking change in this migration: [glob is no longer supported](https://dev.to/43081j/migrating-from-chokidar-3x-to-4x-5ab5). This causes `cdk watch` in CDK CLI v2.1102.0 - v2.1105.0 to work improperly with `include` or `exclude` parameters that utilize globs -- including the default configuration that is `include: ['**']`. Using the default `include` causes `cdk watch` to not watch any files at all.

Since CDK CLI exposes an `include` and `exclude` property that users can set in their own `cdk.json` configuration, and we support globs, we must continue to do so. Therefore, this PR migrates globs into the new format that chokidar v4 understands - a `includes` path adn an `ignored` function.

This code was written with the help of AI, including all tests, and also inspired by similar updates to other projects like this [one](https://github.com/11ty/eleventy/blob/deb6dbe62f95545aba80ff4d7bafe5e3d0b72f08/src/Util/GlobStripper.js).

---


packages/aws-cdk/lib/cli/convert-globs.ts
```diff
@@ -0,0 +1,105 @@
...
+  watchPaths: string[];
+  ignored: (path: string, stats?: any) => boolean;
+}
+
+/**
+ * Extracts base directory from a glob pattern for watching
+ */
+function extractWatchPath(pattern: string): string {
+  if (pattern === '**') return '.';
+  if (pattern.startsWith('**/')) return '.';
```
iankhou, line 11: Why not do: `if (pattern.startsWith('**')) return '.';` to cover both cases?

iankhou (approved):

PR #1159, opened 2026-02-19
refactor(cli-integ): convert telemetry tests to matrix job

mrgrain:
This is a follow-up to the recent integration test workflow refactoring efforts (#1151, #1155).

The telemetry test suite was previously implemented as a special-cased job that ran before all other integration tests. Every matrix job in the workflow depended on telemetry tests completing first, which created an unnecessary bottleneck in the CI pipeline. This sequential dependency added significant overhead since all other test suites had to wait for telemetry tests to complete before starting, even though there was no actual dependency between them.

This change converts the telemetry tests into a regular matrix job, making them consistent with how all other test suites are configured. The telemetry job now runs in parallel with other test suites after the prepare job completes, rather than blocking them. This should reduce overall CI time by removing the artificial sequencing constraint.

The unused `logUploadSteps` helper function has also been removed as part of this cleanup, since all jobs now use the inline log upload steps that were already present in the matrix job template.

---


iliapolo (approved):
PR body needs an update since we don't disable telemetry.

PR #1160, opened 2026-02-19
feat(toolkit-lib): notices can now target CDK app language

mrgrain:
This change enables CLI notices to target users based on the programming language of their CDK app.

The language is automatically detected by examining project files in the cdk.json directory (e.g. `package.json` with TypeScript indicators, `requirements.txt`, `pom.xml`, `.csproj`, `go.mod`). Detection is lazy and only runs when the fetched notice data actually contains `language-*` components, so there is no performance impact for existing notices.

A notice targeting TypeScript users would use a component like `{ name: "language-typescript", version: "*" }`. The "Affected versions" line renders this as `TypeScript apps` rather than exposing the internal component name. Notice text can use `{resolve:LANGUAGE}` placeholders which resolve to properly cased names like "TypeScript" or ".NET".

If the language cannot be determined, language-specific notices are simply not shown.

---


packages/@aws-cdk/toolkit-lib/lib/api/notices/filter.ts
```diff
@@ -11,8 +12,15 @@ function normalizeComponents(xs: Array<Component | Component[]>): Component[][]
   return xs.map(x => Array.isArray(x) ? x : [x]);
 }
 
+function renderComponent(c: Component): string {
+  if (c.name.startsWith('language-')) {
```
rix0rrr:
```suggestion
  if (c.name.startsWith('language:')) {
```

?

packages/@aws-cdk/toolkit-lib/lib/api/notices/filter.ts
```diff
@@ -55,6 +63,22 @@ export interface NoticesFilterFilterOptions {
...
+   */
+  readonly projectDir?: string;
+
+  /**
+   * The guessed language of the CDK app.
+   * If provided, projectDir is ignored for language detection.
+   *
+   * @default - auto-detected from projectDir when needed
+   */
+  readonly language?: string;
```
rix0rrr: I would do one or the other here.

Either, the caller is responsible for detecting a language and passing it in, OR this object always autodetects.

My preference would be the first (pass responsibility upwards to callers).

rix0rrr (approved):
Conditionally approved

packages/@aws-cdk/toolkit-lib/lib/api/notices/filter.ts
mrgrain: fixed!

rix0rrr (approved):

PR #1181, opened 2026-02-26
first test on fixing locks for integration tests with lookups

### LINKED ISSUE #1178: (integ-runner): CLI concurrency issue when running integration tests with lookups
### Describe the bug

When running integration tests, where the tested Constructs contains lookups the CLI aborts with an error about other CLIs running in parallel, altough there are none.

This happens also when only a single worker is running, and the tests are run with

Everything works fine as soon as the lookup is removed.

### Regression Issue

_No response_

### Expected Behavior

Integration tests with lookups should just work like others.

### Current Behavior

```sh
npm run integ-runner -- --parallel-regions eu-central-1 --max-workers 1 --language typescript --update-on-failed
```

```txt
 FAILED     integ.cdk-repro.test-ApiGatewayCompleteDeploymentTest/DefaultTest (undefined/eu-central-1) 3.675s
      Integration test failed: ToolkitError: Other CLIs (PID=26620) are currently reading from /Users/dummyuuser/git/cdk-repro/test/cdk-integ.out.integ.cdk-repro.test.ts.snapshot. Invoke the CLI in sequence, or use '--output' to synth into different directories.
```

### Reproduction Steps

I have created a small repo at https://github.com/webratz/cdk-repro that contains a full example of just the issue, so its easier to reproduce

### Possible Solution

_No response_

I have tried different variants and settings but was unable to work around this behavior

### CDK CLI Version

2.1107.0

### Framework Version

2.240.0

### Node.js Version

24.13.1

### OS

MacOs 15.7

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

asieferlinger-enmacc:
Fixes #1178

Very first test that seems to work. I have too little understanding how the locking _should_ work though, so not sure if thats useful or might even break existing functionality

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
+++ b/packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
@@ -103,14 +103,21 @@ export class ToolkitLibRunnerEngine implements ICdk {
     });
 
     try {
-      await using lock = await this.toolkit.synth(cx, {
+      // Explicitly manage the lock lifecycle instead of using 'await using'
+      // to ensure it's properly released before any subsequent synth operations
+      const lock = await this.toolkit.synth(cx, {
         validateStacks: false,
         stacks: {
           strategy: StackSelectionStrategy.ALL_STACKS,
           failOnEmpty: false,
         },
       });
-      await this.validateRegion(lock);
+      try {
+        await this.validateRegion(lock);
+      } finally {
+        // Always dispose the lock to avoid conflicts with subsequent synth operations
+        await lock.dispose();
+      }
     } catch (e: any) {
       if (e.message.includes('Missing context keys')) {
         // @TODO - silently ignore missing context
@@ -140,15 +147,26 @@ export class ToolkitLibRunnerEngine implements ICdk {
    */
   public async deploy(options: DeployOptions) {
     const cx = await this.cx(options);
-    await this.toolkit.deploy(cx, {
-      roleArn: options.roleArn,
-      traceLogs: options.traceLogs,
-      stacks: this.stackSelector(options),
-      deploymentMethod: {
-        method: 'change-set',
-      },
-      outputsFile: options.outputsFile ? path.join(this.options.workingDirectory, options.outputsFile) : undefined,
-    });
+    try {
+      await this.toolkit.deploy(cx, {
+        roleArn: options.roleArn,
+        traceLogs: options.traceLogs,
+        stacks: this.stackSelector(options),
+        deploymentMethod: {
+          method: 'change-set',
+        },
+        outputsFile: options.outputsFile ? path.join(this.options.workingDirectory, options.outputsFile) : undefined,
+      });
+    } finally {
+      // Always dispose the cloud assembly source to release any locks
+      if (this.isDisposableCloudAssemblySource(cx)) {
+        await cx.dispose();
+      }
+
+      if (options.output) {
+        this.cleanupReadLocks(path.join(this.options.workingDirectory, options.output));
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -103,14 +103,21 @@ export class ToolkitLibRunnerEngine implements ICdk {
     });
 
     try {
-      await using lock = await this.toolkit.synth(cx, {
+      // Explicitly manage the lock lifecycle instead of using 'await using'
+      // to ensure it's properly released before any subsequent synth operations
+      const lock = await this.toolkit.synth(cx, {
```
rix0rrr, line 108: I don't understand this change.

It seems to me that the new code should do the same thing as the old code. What does this change?

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -140,15 +147,26 @@ export class ToolkitLibRunnerEngine implements ICdk {
    */
   public async deploy(options: DeployOptions) {
     const cx = await this.cx(options);
```
rix0rrr, line 149: Shouldn't this be `await using` then?

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -140,15 +147,26 @@ export class ToolkitLibRunnerEngine implements ICdk {
...
+        },
+        outputsFile: options.outputsFile ? path.join(this.options.workingDirectory, options.outputsFile) : undefined,
+      });
+    } finally {
+      // Always dispose the cloud assembly source to release any locks
+      if (this.isDisposableCloudAssemblySource(cx)) {
+        await cx.dispose();
+      }
+
+      if (options.output) {
```
rix0rrr, line 166: Why is this necessary?

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -241,6 +259,25 @@ export class ToolkitLibRunnerEngine implements ICdk {
...
+      if (entry.startsWith(prefix) && entry.endsWith('.lock')) {
+        fs.removeSync(path.join(directory, entry));
+      }
+    }
+  }
+
+  private isDisposableCloudAssemblySource(
+    source: ICloudAssemblySource,
+  ): source is ICloudAssemblySource & { dispose: () => Promise<void> } {
+    return typeof (source as { dispose?: unknown }).dispose === 'function';
```
rix0rrr, line 278: Shouldn't this be called `isDisposable`?

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -241,6 +259,25 @@ export class ToolkitLibRunnerEngine implements ICdk {
     };
   }
 
+  private cleanupReadLocks(directory: string): void {
```
rix0rrr, line 262: It seems like you're saying "the read locks aren't cleaned up properly, so I will do it myself in a couple of additional places".

Did we figure out already why they weren't being cleaned up?

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/source-builder.ts
```diff
@@ -410,6 +417,25 @@ export abstract class CloudAssemblySourceBuilder {
         const readLock = await new RWLock(directory).acquireRead();
         try {
           const asm = await assemblyFromDirectory(directory, services.ioHelper, props.loadAssemblyOptions);
+          if (props.holdLock === false) {
```
rix0rrr, line 420: Why do we not want a read lock? It's there for a reason.

PR #1198, opened 2026-03-05
feat: display CloudFormation Guard Hook failures

jkelley-godaddy:
The [AWS Guard Hook](https://docs.aws.amazon.com/cloudformation-cli/latest/hooks-userguide/guard-hooks.html) currently provides a truncated output during a failure to the Hook Status Reason field of a CloudFormation Stack Event like the following:

```
Hook failed with message: Template failed validation, the following rule(s) failed: AWS_S3_Bucket_AccessControl. Full output was written to s3://test-guard-logs-bucket/cfn-guard-validate-report/AWS--S3--Bucket-AwsS3Bucket/1234567890123.json
```

This is due to a CloudFormation size limit on the Hook Status Reason which may not be able to contain the full output of error messages that Guard creates.

This output informs the user of the name of the rule that failed (`AWS_S3_Bucket_AccessControl` in the example) but not the details written by the Guard Hook (ie. what resource property value caused the failure).  This requires users to fetch the indicated S3 object outside of CDK which increases iteration time when troubleshooting failures. 

In order to improve developer ergonomics and velocity, this update automatically attempts to display the failure details from the new Hook Annotations API via [HookResultCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/cloudformation/command/GetHookResultCommand/).

Annotations work for all types of CloudFormation Hooks [including Guard, Control Tower, Lambda, and Custom Hooks](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Annotation.html).

Messages are normalized to trim leading/tailing spaces, replace whitespace with single spaces, and truncate output beyond 400 characters to better fit in the terminal.

Example of updated CDK output:

```
ExampleStack
ExampleStack: deploying... [1/2]
ExampleStack: creating CloudFormation changeset...
2:45:05 PM | CREATE_FAILED        | AWS::S3::Bucket            | AwsS3Bucket
The following hook(s) failed: [Private::Guard::TestHook] : NonCompliant Rules:

[AWS_S3_Bucket_AccessControl]
• Check was not compliant as property [/Resources/NonCompliantBucket/Properties/AccessControl[L:0,C:91]] existed.
Remediation: AccessControl is deprecated

❌  ExampleStack failed: ToolkitError: The stack named ExampleStack failed creation, it may need to be manually deleted from the AWS console: ROLLBACK_COMPLETE: The following hook(s) failed: [Private::Guard::TestHook]
```

---


packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -186,6 +194,104 @@ export class StackActivityMonitor {
...
+        this.extractMessagesFromCheck(nestedCheck, lines);
+      });
+    }
+  }
+
+  /**
+   * Parses Guard Hook output JSON and formats it into a human-readable string
+   */
+  private parseGuardHookOutput(jsonOutput: any): string {
+    const lines: string[] = ['NonCompliant Rules:', ''];
```
otaviomacedo: Is it guaranteed that there is at least one item with at least one non-compliant report? Otherwise, it will print this header and no list, which will be confusing.

mrgrain (changes requested):
Can you also add an integration test please.

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
jkelley-godaddy: Good question!  For a definitive answer I will defer to the CloudFormation Service Team as the The [AWS Guard Hook](https://docs.aws.amazon.com/cloudformation-cli/latest/hooks-userguide/guard-hooks.html) itself appears to have the lambda wrapper code [here](https://github.com/aws-cloudformation/cloudformation-guard/blob/main/guard-lambda/src/main.rs) to run [CloudFormation Guard](https://github.com/aws-cloudformation/cloudformation-guard).

Personally I don't know much Rust, but I believe the hook is running the equivalent of `cfn-guard validate --output-format json --structured --show-summary none` and saving the output to S3.  Whenever there is one or more "not compliant" rules:
- Top level key `status` is always `FAIL`
- Top level key `not_compliant` always contains one or more items
- Exit code is non-zero

This appears to be what triggers the hook itself to fail (and fail stack deployment if hook is configured to `FAIL` mode), so in short: `not_compliant` should always have one or more items if we are printing a failure message due to a Guard Hook failure.

I removed an extraneous check that tested if `not_compliant` was missing or empty.

mrgrain (changes requested):
Thanks for adding the test! This is pretty good already, but we'd prefer if you used a helper stack to setup the hook.

That should be fairly straight forward: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.CfnGuardHook.html

The test case would then become:

- deploy helper stack
- deploy invalid stack and see it fail

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -186,6 +194,96 @@ export class StackActivityMonitor {
...
+   * Pattern: "...Full output was written to s3://bucket-name/path/to/file.json"
+   */
+  private extractS3Location(hookStatusReason: string): { bucket: string; key: string } | undefined {
+    const s3UrlPattern = /Full output was written to s3:\/\/([^/]+)\/(.+)$/;
+    const match = hookStatusReason.match(s3UrlPattern);
+    if (match) {
+      return { bucket: match[1], key: match[2] };
+    }
+    return undefined;
+  }
```
mrgrain: This fills me with pain. Can we not use the [GetHookResult](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_GetHookResult.html) API with the `HookInvocationId` from the event?

jkelley-godaddy: Can you explain a bit more? 

I'm not sure what we would do with the `HookInvocationId` - if you are asking if it can be used as the S3 object key it cannot - that is an entirely different string value generated at runtime. From observation the s3 URI is in the pattern:
`s3://[Guard Logs Bucket Name]/cfn-guard-validate-report/[CloudFormation Resource Type]-[CloudFormation Resource Logical ID]/[Numeric string].json`
For example:
`s3://test-guard-logs-bucket/cfn-guard-validate-report/AWS--S3--Bucket-MyBucket/1234567890123.json`
Whereas `HookInvocationId` is a GUID like:
`12345689-1234-1234-1243-123456789012`

As far as I know the only way to tell that this is a Guard Hook failure is to look for this specific pattern in the Hook Status Reason string, and the only way to get the full error is from the S3 object URI which is only shown in this string.

I agree this is very painful - if the error details were in the status message there would be no need for this PR.  From what I understand there are character limits on the Hook Status Reason in CloudFormation, which is why the CloudFormation Service Team designed the Guard Hook to log the output to S3, but the AWS Service Team would have more information.

mrgrain: Right, sorry. After I deploy the `guard-hook-setup` stack, I am deploying `guard-hook-test`. This deployment will fail. As part of this failure the CDK gets stack events. I can inspect them using the AWS CLI:

```
aws cloudformation describe-stack-events --stack-name guard-hook-test
```

This returns something like this:

```
        {
            "EventId": "NonCompliantBucket-e7abcb7e-808f-418a-8df9-7c2b366483f1",
            "StackName": "guard-hook-test",
            "LogicalResourceId": "NonCompliantBucket",
            "ResourceType": "AWS::S3::Bucket",
            "ResourceStatus": "CREATE_IN_PROGRESS",
            "HookType": "Private::Guard::TestHook",
            "HookStatus": "HOOK_COMPLETE_FAILED",
            "HookStatusReason": "Hook failed with message: Template failed validation, the following rule(s) failed: AWS_S3_Bucket_AccessControl. Full output was written to s3://<bucket>/cfn-guard-validate-report/AWS--S3--Bucket-NonCompliantBucket/1774457354610.json",
            "HookInvocationPoint": "PRE_PROVISION",
            "HookInvocationId": "6dbedd85-c808-45b7-ad63-3c717d137a32",
            "HookFailureMode": "FAIL"
        },
```

Specifically this:

```
"HookInvocationId": "6dbedd85-c808-45b7-ad63-3c717d137a32",
```

Which I can then use to call 


```
aws cloudformation get-hook-result --hook-result-id 6dbedd85-c808-45b7-ad63-3c717d137a32
```

Which returns something like this:

```
{
    "HookResultId": "6dbedd85-c808-45b7-ad63-3c717d137a32",
    "InvocationPoint": "PRE_PROVISION",
    "FailureMode": "FAIL",
    "TypeName": "Private::Guard::TestHook",
    "OriginalTypeName": "AWS::Hooks::GuardHook",
    "Status": "HOOK_COMPLETE_FAILED",
    "HookStatusReason": "Template failed validation, the following rule(s) failed: AWS_S3_Bucket_AccessControl. Full output was written to s3://<bucket>/cfn-guard-validate-report/AWS--S3--Bucket-NonCompliantBucket/1774457354610.json",
    "Target": {
        "TargetType": "RESOURCE",
        "TargetTypeName": "AWS::S3::Bucket",
        "TargetId": "NonCompliantBucket",
        "Action": "CREATE"
    },
    "Annotations": [
        {
            "AnnotationName": "AWS_S3_Bucket_AccessControl",
            "Status": "FAILED",
            "StatusMessage": "Check was not compliant as property [/Resources/NonCompliantBucket/Properties/AccessControl[L:0,C:91]] existed.",
            "RemediationMessage": "\n            AccessControl is deprecated\n        "
        }
    ]
}
```

While `Annotations` is not the exact same output as the S3 output, it seems to have all information we need.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/deploy/cdk-deploy-guard-hook-failure-displays-s3-output.integtest.ts
```diff
@@ -0,0 +1,31 @@
...
+      expect(deployOutput).toContain('NonCompliant Rules:');
+      expect(deployOutput).toContain('[AWS_S3_Bucket_AccessControl]');
+      expect(deployOutput).toContain('• AccessControl is deprecated');
+      expect(deployOutput).toContain(`Full output was written to s3://${logsBucket}`);
+    } finally {
+      // Destroy the test stack if it exists (may be in ROLLBACK_COMPLETE state)
+      await fixture.cdkDestroy('guard-hook-test', { allowErrExit: true });
+      // Destroy the setup stack, which removes the Guard Hook, S3 buckets, and IAM role
+      await fixture.cdkDestroy('guard-hook-setup');
+    }
```
mrgrain: there's no need to remove them, it will happen automatically

mrgrain: (i will fix it)

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
jkelley-godaddy: Thank you for the clarification.  That's a great suggestion.

I know the `Annotations` API is new - I'm unsure if it has the same limitations as the HookStatusReason and other fields.  If there is no array limit then this is a superior solution.  The logs bucket is optional while Annotations should be available regardless of configuration, and this would also replace the brittle regex check for the hook status message.  Would also remove the need to create or add permissions to the logs bucket, making this more universal and simpler.

Just want to ensure there aren't any cases where we would still need to go to the bucket for any data.

I've asked the CloudFormation Service Team to confirm if there are any limits on Annotations - if not we will proceed with making these changes.

jkelley-godaddy: @mrgrain - good call, I've updated this PR to use the Annotations API which makes this feature work with all hooks (not just Guard) and removes the brittle S3 lookup logic.

I've rebased on main and updated all unit tests / integration tests to ensure they pass locally.

One note - since this API is new it requires the `cloudformation:GetHookResult` permission for the bootstrap `CloudFormationExecutionRole` which was not present before.  Obviously this requires a version bump to the bootstrap template.  Users on an older bootstrap version will get a warning message and the original HookStatusReason will be displayed.

Let me know if you want us to change how this works or if this is acceptable.  Thank you for your assistance and guidance!

mrgrain (approved):

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -231,6 +233,55 @@ export class StackActivityMonitor {
...
+        }
+        lines.push('');
+      }
+      return lines.join('\n').trimEnd();
+    } catch (e) {
+      const errorMessage = e instanceof Error ? e.message : String(e);
+      await this.ioHelper.defaults.warn(
+        util.format('Failed to fetch Guard Hook details for invocation %s: %s', hookInvocationId, errorMessage),
+      );
+      return undefined;
```
mrgrain, line 315: One small final request: Can we detect a permission issue here and print a warning that tells users to upgrade bootstrap. We already do this in the `EarlyValidationReporter`, so should be straight forward to adapt that code.

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
jkelley-godaddy: I've added this functionality and a unit test however I'm unable to do any manual testing as some tests outside of this branch are failing since I merged the latest main.  This prevents me from making a local build for testing.

I confirmed I'm also getting these failures on a fresh git clone of the main branch so not sure what is causing that, here's the first one: 
```
      Summary of all failing tests
       FAIL  test/actions/bootstrap.test.ts
        ● bootstrap › bootstrap parameters › bootstrap with additional parameters

          TypeError: Object not disposable.

            13 |   /**
            14 |    * AWS object (used by contextprovider)
          > 15 |    * @deprecated context should be moved to the toolkit itself
               |                                                  ^
            16 |    */
            17 |   readonly services: ToolkitServices;
            18 |

            at Object.<anonymous>.__addDisposableResource (lib/api/cloud-assembly/private/context-aware-source.ts:15:50)
            at ContextAwareCloudAssemblySource.produce (lib/api/cloud-assembly/private/context-aware-source.ts:72:30)
            at assemblyFromSource (lib/toolkit/private/index.ts:33:54)
            at BootstrapEnvironments.envProvider (lib/actions/bootstrap/index.ts:30:30)
            at Toolkit.bootstrap (lib/toolkit/toolkit.ts:262:35)
            at Object.<anonymous> (test/actions/bootstrap.test.ts:295:7)
```

I'm hoping the changes of the last commit and unit test are sufficient to get this across the finish line however if not any assistance is greatly appreciated.

Thank you again for your help!

mrgrain (approved):

PR #1230, opened 2026-03-17
fix: hotswap is not discoverable via telemetry

kaizencc:
Adds a new `HOTSWAP` telemetry event type that fires when hotswap deployments are attempted, tracking success/failure state and resource change counts. The motivation behind this is to track hotswap efficiency.

For example, 

```json
{
  "event": {
    "state": "FAILED",
    "eventType": "HOTSWAP",
    "duration": 456,
    "error": {
      "name": "UnknownError"
    },
    "counters": {
      "hotswapped": 0,
      "hotswappableChanges": 1,
      "nonHotswappableChanges": 0
    }
  }
}
```

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
@@ -112,6 +112,12 @@ export async function tryHotswapDeployment(
 
   await hotswapSpan.end(result);
 
+  // This is a hotswap error that was caught during the hotswapDeployment function
+  // It is thrown after the hotswap span ends so it can be reported to telemetry
+  if (result.error) {
+    throw result.error;
+  }
+
   if (result?.hotswapped === true) {
     return {
       type: 'did-deploy-stack',
@@ -188,14 +194,20 @@ async function hotswapDeployment(
   }
 
   // apply the short-circuitable changes
-  await applyAllHotswapOperations(sdk, ioSpan, hotswappable);
+  let error: Error | undefined;
+  try {
+    await applyAllHotswapOperations(sdk, ioSpan, hotswappable);
+  } catch (e: any) {
+    error = e;
+  }
 
   return {
     stack,
     mode: hotswapMode,
-    hotswapped: true,
+    hotswapped: !error,
     hotswappableChanges,
     nonHotswappableChanges,
+    error,
   };
 }
 
--- a/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
+++ b/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
@@ -645,13 +645,12 @@ function eventFromMessage(msg: IoMessage<unknown>): TelemetryEvent | undefined {
 }
 
 function hotswapToEventResult(result: HotswapResult): TelemetryEvent {
-  const data = result as any;
   return {
     eventType: 'HOTSWAP' as const,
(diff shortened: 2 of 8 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -636,3 +643,20 @@ function eventFromMessage(msg: IoMessage<unknown>): TelemetryEvent | undefined {
     };
   }
 }
+
+function hotswapToEventResult(result: HotswapResult): TelemetryEvent {
```
kaizencc, line 647: callout in case you disagree:

a hotswap event that results in `hotswapped: false` (as in, it falls back to regular OR does nothing), will result in `state: SUCCEEDED`. my opinion is that `state` is for situations where errors are reported. The way we will differentiate hotswap happening or not is the `hotswap` counter.

ShadowCat567: This sounds reasonable to me, since the deployment technically did not fail.

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
```diff
@@ -188,14 +194,20 @@ async function hotswapDeployment(
...
+    await applyAllHotswapOperations(sdk, ioSpan, hotswappable);
+  } catch (e: any) {
+    error = e;
+  }
 
   return {
     stack,
     mode: hotswapMode,
-    hotswapped: true,
+    hotswapped: !error,
```
ShadowCat567, line 207: This is a bit confusing to me. 
From what I understand, `hotswapped: true` means that we successfully executed hotswap. However `hotswap: false` and `error` being defined means we tried to hotswap but ran into some SDK issue so hotswap deployment was failed and `hotswap: false` and `error: undefined` means we did some fallback behavior. 
So does that mean is `hotswap` set to `false` by default? And the only way it can get set to `true` is if it passes through here?

ShadowCat567 (commented):
I have a question about hotswap event registration to make sure I am understanding how it works.

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
kaizencc: hotswap behavior when `nonHotswappableChanges` are detected is either 1) fall back to full CFN deployment or 2) ignore `nonHotswappableChanges` and hotswap what can be hotswapped.

Behavior 1) happens before this line:

```ts
if (hotswapMode === 'fall-back') {
    if (nonHotswappableChanges.length > 0) {
      return {
        stack,
        mode: hotswapMode,
        hotswapped: false,
        hotswappableChanges,
        nonHotswappableChanges,
      };
    }
  }
```

So this is behavior 2). Previously, if an SDK error is encountered we do not catch and actually this span never ends. This now matters because the `end` command is what sends telemetry. So I have decided that in this scenario, we say `hotswapped: false` because although we attempted hotswap, it was unsuccessful. If there's no error, we return `hotswapped: true` like before.

---

All that is to say, I am preserving old behavior and also your explanation of what is happening is also right. 

> So does that mean is hotswap set to false by default? And the only way it can get set to true is if it passes through here?

Yes. the only way we set `hotswap: true` is if `applyAllHotswapOperations` is invoked and does not return an error

ShadowCat567 (approved):

PR #1232, opened 2026-03-18
fix(integ-runner): support per-test-case AWS profile override via cdkCommandOptions

### LINKED ISSUE #1231: (integ-runner): IntegTestRunner ignores profile from testcase cdkCommandOptions.
### Describe the bug

Profiles specified in IntegTests like so:

```
new IntegTest(someApp, "sometest", {
...
  cdkCommandOptions: {
    deploy: {
      args: {
        profile: "some-profile",
      },
...
```

seems to be ignored during deploy. I believe it happens here:
https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/integ-runner/lib/runner/integ-test-runner.ts#L249

where `actualTestCase.cdkCommandOptions.deploy.args.profile` is completely disregarded.

This is a problem that has occurred before, and I believe it was fixed somewhere around 2.194.2, but I can't seem to find the corresponding issue right now, maybe there never was one. I know the fix was commited the exact. same day I pinpointed the problem back then.

### Regression Issue

2.194.2

### Expected Behavior

Profile from cdkCommandOptions should be used if specified.

### Current Behavior

Default profile (or from --profiles on command line) is used.

### Reproduction Steps

- Specify test with cdkCommandOptions.deploy.args.profile
- Run test with some other default profile
- Check cloudtrail for example for ExecuteChangeSet and not the userIdentity

### Possible Solution

```
            profile: actualTestCase.cdkCommandOptions?.deploy?.args?.profile ?? this.profile,
```

We use this to test deployment with different sets of permissions.

### CDK CLI Version

2.1027.0 (build 5fc1b13)

### Framework Version

_No response_

### Node.js Version

v22.22.1

### OS

All

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

sai-ray:
Fixes #1231

When an integration test specifies a profile via `cdkCommandOptions.deploy.args.profile`:

```typescript
new IntegTest(app, 'my-test', {
  testCases: [myStack],
  cdkCommandOptions: {
    deploy: {
      args: { profile: 'deploy-profile' },
    },
  },
});
```

The profile is silently ignored. The `ToolkitLibRunnerEngine` creates a single `Toolkit` instance at construction time with credentials locked to the runner-level profile (from `--profiles` CLI flag or machine default). The engine's `deploy()` and `destroy()` methods receive the correct profile in their options but never use it.

This capability existed implicitly in the old `cli-wrapper` engine (removed in #1108) which spawned separate CLI processes per action, each with its own `--profile` flag.

#### Changes

Single file: `packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts`

- Extracted `createToolkit(profile, region)` helper from the constructor
- Added `toolkitFor(options)` that returns a new `Toolkit` when the per-action profile differs from the constructor profile, or reuses the existing one when they match
- Updated `deploy()`, `destroy()`, and `cx()` to use the appropriate toolkit

#### Testing

Added 3 tests in `packages/@aws-cdk/integ-runner/test/engines/toolkit-lib.test.ts`:

- Deploy with override profile creates new Toolkit with correct credentials
- Destroy with override profile creates new Toolkit with correct credentials
- Deploy with same profile reuses existing Toolkit (no perf regression)

All 149 existing tests continue to pass.

### Checklist

---


packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -194,23 +216,25 @@ export class ToolkitLibRunnerEngine implements ICdk {
...
+      return tk.fromAssemblyDirectory(potentialCxPath);
     }
 
     let outdir;
     if (options.output) {
       outdir = path.join(this.options.workingDirectory, options.output);
     }
 
-    return this.toolkit.fromCdkApp(options.app, {
+    return tk.fromCdkApp(options.app, {
```
iankhou: Why make this change here? We still have `this.toolkit`, as before.

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -59,31 +59,51 @@ export class ToolkitLibRunnerEngine implements ICdk {
...
+   * Returns the appropriate Toolkit for the given options.
+   * If the options specify a profile different from the engine's default,
+   * a new Toolkit is created with the overridden profile.
+   */
+  private toolkitFor(options: { profile?: string }): Toolkit {
+    if (options.profile && options.profile !== this.options.profile) {
+      return this.createToolkit(options.profile, this.options.region);
+    }
+    return this.toolkit;
   }
```
iankhou: hmm... seems prescriptive, given it only changes one aspect of the toolkit. It's also kind of strange that we hit a fallback to the same Toolkit. The function should either always return the same Toolkit (albeit mutated), or always return a new Toolkit. Here we might return the old one or a new one, depending on options.

iankhou (changes requested):

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
sai-ray: Fair, I was being overly cautious thinking lookups during synthesis might need the overridden credentials, but realized that context comes from the snapshot so it doesn't actually matter here. Reverted.

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
sai-ray: This was just to reuse the toolkit when profiles matched as I thought construction might be expensive. Didn't realize it's just object allocation, so the optimization saved nothing while making the code harder to follow. Now deploy/destroy always create new one.

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
```diff
@@ -139,8 +147,9 @@ export class ToolkitLibRunnerEngine implements ICdk {
    * Deploys the CDK app
    */
   public async deploy(options: DeployOptions) {
+    const toolkit = this.createToolkit(options.profile ?? this.options.profile, this.options.region);
```
mrgrain, line 151: For completeness, we should do this for all actions then.

mrgrain (changes requested):
I'm okay with the direction of this. There's a bit of an issue here that `cdkCommandOptions` and `integ-runner --profile` might conflict with each other. 

We are also planning to deprecate the `cdkCommandOptions` option feature in the near future (integ-tests-alpha is still experimental).

packages/@aws-cdk/integ-runner/lib/engines/toolkit-lib.ts
sai-ray: added the pattern to list() and watch() as well. Skipped synth() since SynthOptions doesn't carry a profile field.

mrgrain (approved):

iankhou (approved):

PR #1243, opened 2026-03-23
fix: CLI ignores root cause of CDK App termination

rix0rrr:
This is the companion PR to https://github.com/aws/aws-cdk/pull/37294, to have the CLI read the synthesis failure root cause from the CDK app output.

Rather than introduce a new error type, the error code is stored on `AssemblyError.synthErrorCode` instead (if available).

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/exec.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/exec.ts
@@ -1,8 +1,8 @@
 import * as child_process from 'child_process';
+import { readFileSync } from 'fs';
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import split = require('split2');
 import { AssemblyError } from '../../../toolkit/toolkit-error';
-import { readFileSync } from 'fs';
 
 type EventPublisher = (event: 'open' | 'data_stdout' | 'data_stderr' | 'close', line: string) => void;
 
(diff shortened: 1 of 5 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/exec.ts
```diff
@@ -60,19 +64,55 @@ export async function execInChildProcess(commandAndArgs: string, options: ExecOp
...
+
+        let error = AssemblyError.withCause(`${commandAndArgs}: Subprocess exited with error ${code}`, cause);
+
+        // Search for an error code, and throw that if we have it
+        if (options.errorCodeFile) {
+          const contents = tryReadFile(options.errorCodeFile);
+          if (contents) {
+            const errorInStdErr = contents
+              .split('\n')
+              .find(c => stdErrString.includes(`${SYNTH_ERROR_CODE_MARKERS[0]}${c}${SYNTH_ERROR_CODE_MARKERS[1]}`));
```
kaizencc, line 91: confirming that i double checked that we mean to only supply one possible error code in this file

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit-error.ts
```diff
@@ -62,9 +62,9 @@ export class ToolkitError extends Error {
 
   constructor(message: string, type: string = 'toolkit', cause?: unknown) {
     super(message);
+    this.name = this.constructor.name;
```
kaizencc, line 65: why this over `new.target.name` deleted below? same behavior is expected right

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit-error.ts
```diff
@@ -120,12 +120,25 @@ export class AssemblyError extends ToolkitError {
...
+
+  /**
+   * The synthesis error code
+   */
+  public get synthErrorCode(): string | undefined {
+    return this._synthErrorCode;
+  }
+
+  public attachSynthesisErrorCode(synthesisErrorCode: string) {
+    this._synthErrorCode = synthesisErrorCode;
```
kaizencc, line 140: do we wanna block multiple attachments?

```ts
if (this._synthErrorCode) {
 throw new Error('Synthesis error code already attached');
}
```

kaizencc (approved):
just nits. approving and up to you whether you want to address before merging

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/exec.ts
rix0rrr: Not true though. It can contain multiple errors.

PR #1244, opened 2026-03-23
fix: error codes are not collected

rix0rrr:
In order to start doing something about the most common errors people run into, we need to know the actual error codes. This PR changes the error codes we send to the server.

We are making sure to:

- only collect error *codes*, not messages; and
- only collect errors from AWS sources (this toolkit, the CDK construct library or the SDK/AWS services).

In line with our telemetry policy.

(Depends on https://github.com/aws/aws-cdk-cli/pull/1243)

---


packages/aws-cdk/test/cli/telemetry/error.test.ts
```diff
@@ -2,9 +2,9 @@ import { AuthenticationError } from '@aws-cdk/toolkit-lib';
 import { cdkCliErrorName } from '../../../lib/cli/telemetry/error';
 
 test('returns known error names', () => {
-  expect(cdkCliErrorName(AuthenticationError.name)).toEqual(AuthenticationError.name);
+  expect(cdkCliErrorName(new AuthenticationError('Oh no'))).toEqual(AuthenticationError.name);
 });
 
 test('returns UnknownError for unknown error names', () => {
-  expect(cdkCliErrorName('ExpiredToken')).toEqual('UnknownError');
+  expect(cdkCliErrorName(new Error('ExpiredToken: damn'))).toEqual('UnknownError');
```
kaizencc: nah we gotta do better than damn, there are kids watching!

packages/aws-cdk/lib/cli/telemetry/schema.ts
```diff
@@ -51,16 +51,8 @@ interface Duration {
...
-  TOOLKIT_ERROR = 'ToolkitError',
-  AUTHENTICATION_ERROR = 'AuthenticationError',
-  ASSEMBLY_ERROR = 'AssemblyError',
-  CONTEXT_PROVIDER_ERROR = 'ContextProviderError',
-  UNKNOWN_ERROR = 'UnknownError',
-}
-
 export interface ErrorDetails {
-  readonly name: ErrorName;
+  readonly name: string;
```
kaizencc, line 55: Server side change required

packages/aws-cdk/lib/cli/telemetry/error.ts
```diff
@@ -1,14 +1,29 @@
...
-  // contain text that we have no control over so it is safer to not send it.
-  if (!isKnownErrorName(name)) {
-    return ErrorName.UNKNOWN_ERROR;
+/**
+ * Return the transmitted error code for this error object
+ *
+ * We are taking care to only transmit errors that originate from AWS systems
+ * (this toolkit itself, the CDK construct library, the AWS SDK, AWS services).
+ */
+export function cdkCliErrorName(err: Error): string {
```
kaizencc, line 20: The order here makes it seem like we want `ServiceException` if that exists (and i agree). But what if we receive a `ToolkitError` that wraps `ServiceException`? We'll receive the `ToolkitError` instead, and that may be more generic. At any rate, I don't see any testing that confirms or denys this assumption

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -93,7 +93,7 @@ export class TelemetrySession {
     process.on('SIGINT', async () => {
       try {
         await this.end({
-          name: ErrorName.TOOLKIT_ERROR,
+          name: 'UserInterrupted',
```
kaizencc: nit: think this should be a constant living in `error.ts`

kaizencc (commented):
some nits and questions, and of course a server side change is needed

packages/aws-cdk/lib/cli/telemetry/schema.ts
rix0rrr: No. It's typed as `string` on the server.

mrgrain (approved):

PR #1246, opened 2026-03-23
fix(cli): `--require-approval any-change` only asks for confirmation on broadening changes

### LINKED ISSUE #1236: (cli): --require-approval any-change silently skips approval for non-broadening changes
### Describe the bug

`cdk deploy --require-approval any-change` no longer prompts for approval on non-broadening changes. It behaves the same as `--require-approval broadening`, only prompting for broadening permission changes.

Per the [docs](https://docs.aws.amazon.com/cdk/v2/guide/ref-cli-cmd-deploy.html), `any-change` should require "manual approval for any change to the stack."

Two issues in `CliIoHost`:

1. **`requireDeployApproval` is not wired from CLI args to `CliIoHost`** — In `packages/aws-cdk/lib/cli/cli.ts`, `CliIoHost.instance()` is created without `requireDeployApproval`, so it defaults to `RequireApproval.BROADENING`. The `--require-approval` value is only passed to `CdkToolkit.deploy()` (line 411), but never to `CliIoHost`.

2. **Missing `return` in `skipApprovalStep`** — In `packages/aws-cdk/lib/cli/io-host/cli-io-host.ts` line 380, the early-return guard is missing `return`:
```typescript
if (!(msg.code && approvalToolkitCodes.includes(msg.code))) {
    false;  // BUG: should be `return false;`
}
```

### Regression Issue

2.1019.2

### Expected Behavior

When running `cdk deploy --require-approval any-change`, the CLI should prompt for confirmation before deploying ANY stack changes (security or not), as documented.

### Current Behavior

The CLI only prompts for broadening security changes, silently deploying all other changes without approval. The `any-change` option behaves identically to `broadening`.

### Reproduction Steps

```bash
# Make any non-IAM change to a CDK stack (e.g., change a tag, add a resource)
cdk deploy --require-approval any-change <stack-name>
# Expected: prompt "Do you wish to deploy these changes (y/n)?"
# Actual: deploys without prompting
```

### Possible Solution

1. In `cli.ts`, set `ioHost.requireDeployApproval` after configuration is parsed:
```typescript
const requireApproval = configuration.settings.get(['requireApproval']);
if (requireApproval) {
    ioHost.requireDeployApproval = requireApproval;
}
```

2. In `cli-io-host.ts`, add the missing `return`:
```typescript
if (!(msg.code && approvalToolkitCodes.includes(msg.code))) {
    return false;
}
```

The regression was introduced during the `CliIoHost` refactor that replaced direct `promptly.confirm()` calls with the `ioHost.requestResponse()` abstraction. In the previous version (2.1019.2), `askUserConfirmation` directly checked `process.stdin.isTTY` and used `promptly.confirm()`, which worked correctly.

### CDK CLI Version

2.1100.3 (build 0aa3e98)

### Framework Version

aws-cdk-lib 2.201.0

### Node.js Version

v24.11.0

### OS

macOS (Darwin 25.3.0)

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

iliapolo:
Fixes https://github.com/aws/aws-cdk-cli/issues/1236

### Problem

The value of `--require-approval` is not being passed to the `CliIoHost`, which defaults it to `RequireApproval.BROADENING`.

### Solution

The correct `requireApproval` value is already being passed to `deploy`, so we just need to propagate it into the `CliIoHost` via the setter. This isn't ideal because it prevents concurrent deployments with different `requireApproval` values, but the CLI can't do that anyway so its not an actual issue. 

### Checklist

---

### DIFF
--- a/packages/aws-cdk/lib/cli/cdk-toolkit.ts
+++ b/packages/aws-cdk/lib/cli/cdk-toolkit.ts
@@ -504,13 +504,13 @@ export class CdkToolkit {
         });
         const securityDiff = formatter.formatSecurityDiff();
         if (requiresApproval(requireApproval, securityDiff.permissionChangeType)) {
-          const motivation = requireApproval === RequireApproval.ANYCHANGE 
-            ? `"--require-approval" is set to '${RequireApproval.ANYCHANGE}'` 
+          const motivation = requireApproval === RequireApproval.ANYCHANGE
+            ? `"--require-approval" is set to '${RequireApproval.ANYCHANGE}'`
             : '"--require-approval" is enabled and stack includes security-sensitive updates';
           await this.ioHost.asIoHelper().defaults.info(securityDiff.formattedDiff);
 
           // the ioHost uses this internally to determine if a confirmation
-          // is actually needed, so it needs the same value we have here. 
+          // is actually needed, so it needs the same value we have here.
           // ideally this would threaded into the request instead of it being an instance field.
           this.ioHost.requireDeployApproval = requireApproval;
 
--- a/packages/aws-cdk/lib/cli/cdk-toolkit.ts
+++ b/packages/aws-cdk/lib/cli/cdk-toolkit.ts
@@ -504,8 +504,16 @@ export class CdkToolkit {
         });
         const securityDiff = formatter.formatSecurityDiff();
         if (requiresApproval(requireApproval, securityDiff.permissionChangeType)) {
-          const motivation = '"--require-approval" is enabled and stack includes security-sensitive updates';
+          const motivation = requireApproval === RequireApproval.ANYCHANGE 
+            ? `"--require-approval" is set to '${RequireApproval.ANYCHANGE}'` 
+            : '"--require-approval" is enabled and stack includes security-sensitive updates';
           await this.ioHost.asIoHelper().defaults.info(securityDiff.formattedDiff);
+
+          // the ioHost uses this internally to determine if a confirmation
+          // is actually needed, so it needs the same value we have here. 
+          // ideally this would threaded into the request instead of it being an instance field.
+          this.ioHost.requireDeployApproval = requireApproval;
+
           await askUserConfirmation(
             this.ioHost,
             IO.CDK_TOOLKIT_I5060.req(`${motivation}: 'Do you wish to deploy these changes'`, {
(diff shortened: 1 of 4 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -377,7 +377,7 @@ export class CliIoHost implements IIoHost {
   private skipApprovalStep(msg: IoRequest<any, any>): boolean {
     const approvalToolkitCodes = ['CDK_TOOLKIT_I5060'];
     if (!(msg.code && approvalToolkitCodes.includes(msg.code))) {
-      false;
```
iliapolo, line 380: This is something the original issue creator mention and indeed seems like a bug. Looks like we are missing some tests here but I'd like to not delay this fix for it.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -504,8 +504,16 @@ export class CdkToolkit {
...
-          const motivation = '"--require-approval" is enabled and stack includes security-sensitive updates';
+          const motivation = requireApproval === RequireApproval.ANYCHANGE
+            ? `"--require-approval" is set to '${RequireApproval.ANYCHANGE}'`
+            : '"--require-approval" is enabled and stack includes security-sensitive updates';
           await this.ioHost.asIoHelper().defaults.info(securityDiff.formattedDiff);
+
+          // the ioHost uses this internally to determine if a confirmation
+          // is actually needed, so it needs the same value we have here.
+          // ideally this would threaded into the request instead of it being an instance field.
+          this.ioHost.requireDeployApproval = requireApproval;
```
iliapolo, line 515: See https://github.com/aws/aws-cdk-cli/blob/ad90be6af5fe78e2b2c8b2aae289bfef3ef8bbd0/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts#L383-L393

kaizencc: this is fine and im approving. but i feel like passing the `requireApproval` property to the iohost should happen when we define `requireApproval` on line 428 rather than here. 

```ts
const requireApproval = options.requireApproval ?? RequireApproval.BROADENING;
this.ioHost.requireDeployApproval = requireApproval;
```

and then maybe that can be a helper function to make those actions atomic.

kaizencc (approved):

packages/aws-cdk/lib/cli/cdk-toolkit.ts
iliapolo: Makes sense.

PR #1247, opened 2026-03-23
chore: colocate requireApproval properties

kaizencc:
Fast follow for https://github.com/aws/aws-cdk-cli/pull/1246. Existing tests are sufficient.

---

### DIFF
--- a/packages/aws-cdk/lib/cli/cdk-toolkit.ts
+++ b/packages/aws-cdk/lib/cli/cdk-toolkit.ts
@@ -426,6 +426,7 @@ export class CdkToolkit {
     });
 
     const requireApproval = options.requireApproval ?? RequireApproval.BROADENING;
+    this.ioHost.requireDeployApproval = requireApproval;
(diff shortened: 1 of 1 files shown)


packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -426,6 +426,7 @@ export class CdkToolkit {
     });
 
     const requireApproval = options.requireApproval ?? RequireApproval.BROADENING;
+    this.ioHost.requireDeployApproval = requireApproval;
```
kaizencc, line 431:
```suggestion
    // the ioHost uses this internally to determine if a confirmation
    // is actually needed, so it needs the same value we determine here.
    const requireApproval = options.requireApproval ?? RequireApproval.BROADENING;
    this.ioHost.requireDeployApproval = requireApproval;
```

iliapolo (approved):

PR #1253, opened 2026-03-25
fix: CloudFormation deployment errors are not classified

rix0rrr:
In order to get a good view on deployment errors that we should try to
catch at synthesis time, capture errors from the CloudFormation event
log.

We only capture error codes, and only those that follow a particular
string format in the error message response. Examples of messages
we will catch:

```
Resource handler returned message: "Exceeded attempts to wait" (RequestToken: ***, HandlerErrorCode: NotStabilized)
Provided private key cannot be used for Sign in with Apple. (Service: AWSCognitoIdentityProviderService; Status Code: 400; Error Code: InvalidParameterException; Request ID: ***; Proxy: null)
```

Not all CloudFormation resources hold to this pattern, which is
unfortunate, but the newer ones certainly should. We include the
resource type (as long as it starts with `AWS::`) so that we'll at least
know the resource that caused the problem, if not the actual error code.

---


packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -254,14 +297,34 @@ export class StackActivityMonitor {
...
+
+/**
+ * Extract an error code from the given stack event.
+ *
+ * Always contains the services, and includes the handler error code if available.
+ */
+export function extractErrorCode(event: StackEvent): string {
+  // Get the resource type; if it is non-AWS then we are done.
+  const resourceTypeParts = (event.ResourceType ?? '').split('::');
+  if (resourceTypeParts[0] !== 'AWS') {
```
iliapolo: This will also skip our own custom resources. I'm good with that to start off with - just making sure it was intentional.

rix0rrr: Yeah we can get those later

iliapolo (approved):

PR #1262, opened 2026-03-26
fix: error annotations are not collected

rix0rrr:
We collect an "synth failed because of error annotations" error code, but the codes for the error annotations themselves are not collected.

Ultimately, these are more important than the "there were errors" code, so collect these as well. They are counters, because there can be more than one.

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1459,13 +1459,20 @@ function zeroTime(): ElapsedTime {
   return { asMs: 0, asSec: 0 };
 }
 
-function countAssemblyResults(span: IMessageSpan<any>, assembly: cxapi.CloudAssembly) {
+export function countAssemblyResults(span: IMessageSpan<any>, assembly: cxapi.CloudAssembly) {
```
kaizencc: wont this become part of the public contract of toolkit lib?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1474,3 +1481,5 @@ function countAssemblyResults(span: IMessageSpan<any>, assembly: cxapi.CloudAsse
 function sum(xs: number[]) {
   return xs.reduce((a, b) => a + b, 0);
 }
+
+const ANNOTATION_ERROR_CODE_TYPE = 'aws:cdk:error-code';
```
kaizencc: this error code is coming from here right:

https://github.com/aws/aws-cdk/blob/fa21e623484ad1ed191e542e93b16811a8ef08b6/packages/aws-cdk-lib/core/lib/annotations.ts#L172

we should make it very clear in the docs at least that these are connected and they cannot be unilaterally changed

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1459,13 +1459,20 @@ function zeroTime(): ElapsedTime {
...
   span.incCounter('stacks', stacksRecursively.length);
   span.incCounter('assemblies', asmCount(assembly));
   span.incCounter('errorAnns', sum(stacksRecursively.map(s => s.messages.filter(m => m.level === SynthesisMessageLevel.ERROR).length)));
   span.incCounter('warnings', sum(stacksRecursively.map(s => s.messages.filter(m => m.level === SynthesisMessageLevel.WARNING).length)));
 
+  const annotationErrorCodes = stacksRecursively
+    .flatMap(s => Object.values(s.metadata)
+      .flatMap(ms => ms.filter(m => m.type === ANNOTATION_ERROR_CODE_TYPE)));
+  for (const annotationErrorCode of annotationErrorCodes) {
+    span.incCounter(`errorAnn:${annotationErrorCode.data}`);
```
kaizencc: i dont love this... this seems like a clever way of storing info in the counter name. at the very least we should enforce exactly what properties of `data` we want to store. what if `annotationErrorCode.data` information expands in the future?

kaizencc: i looked and found `data = id` right now. but we don't enforce that `id` is not user data (although we say its an error code, so it maybe feels like common sense) and we don't enforce that `data` must only be `id`. i think its only a matter of time before we misuse `_addTrackableError` and get data we dont want to track.

kaizencc (changes requested):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
rix0rrr: Yeah, and it shouldn't. It won't even compile like this anyway.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
rix0rrr: I feel that should be obvious, but sure I'll add a comment.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
rix0rrr: What are you saying here? 

Are you worried about this data not coming from the library? Or something else?

rix0rrr: Will this address your concerns? https://github.com/aws/aws-cdk/pull/37381

packages/@aws-cdk/toolkit-lib/lib/toolkit/private/count-assembly-results.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function sum(xs: number[]) {
+  return xs.reduce((a, b) => a + b, 0);
+}
+
+/**
+ * Well-known and agreed-upon value between aws-cdk-lib and the toolkit
+ *
+ * Do not change, obviously.
```
kaizencc, line 31: you jest but i like

kaizencc (approved):

Issue #1123, opened 2026-02-06
(cli): Support a least-privilege diff-only IAM role for `cdk diff`


garysassano:
### Describe the feature

There is currently no way to run `cdk diff` using a read-only or least-privilege IAM role when using the AWS CDK CLI.

Even though `cdk diff` is a non-deploying operation, the CLI requires permissions equivalent to deployment roles (including assuming deploy and asset publishing roles). This makes it difficult to use `cdk diff` safely in CI environments.

I am proposing support in the CDK CLI for a diff-only IAM role that allows secure execution of `cdk diff` without granting full deployment capabilities.

### Use Case

In CI pipelines, I want to run `cdk diff` automatically to provide visibility into infrastructure changes without introducing a manual approval gate.

Using a role with deployment-level permissions in CI is risky: if the workflow is modified or compromised, the role could be abused to deploy or modify infrastructure.

Today, there is no secure alternative:
- `cdk diff` creates a CloudFormation change set
- This requires assuming the deploy role
- For stacks with assets, the CLI also attempts to assume file-publishing and image-publishing roles

As a result, it is currently impossible to run `cdk diff` with a read-only or least-privilege role using the CDK CLI, unless one builds a custom diff workflow outside the CLI.

### Proposed Solution

Add first-class support in the CDK CLI for a diff-only role with limited permissions.

Conceptually, this role would:
- Allow creation of CloudFormation change sets
- Explicitly deny `cloudformation:ExecuteChangeSet`
- Allow asset-related actions required for diffing:
  - S3 uploads for `Code.fromAsset`
  - Optional asset handling only when relevant
- Have no permissions to deploy, update, or delete stacks

Possible implementation approaches:
- Bootstrap an optional `diff-role` alongside existing toolkit roles
- Allow explicitly specifying a diff role via CLI flags (e.g. `cdk diff --role-arn ...`)
- Automatically select a diff role when running `cdk diff`

This would preserve accurate change-set–based diffs while significantly improving security for CI usage.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.1105.0

### Environment details (OS name and version, etc.)

Ubuntu 24.04

PR #1271, opened 2026-03-27
feat(cli): add `cdk deploy --method=execute-change-set` for two-step deployment workflows

mrgrain:
### Problem

CDK supports creating change sets without executing them via `--method=prepare-change-set`, but there is no way to execute those change sets through CDK afterwards. Users who want a review-then-deploy workflow — common in regulated environments and CI/CD pipelines with manual approval gates — have to drop down to raw CloudFormation API calls or the AWS Console to execute the change set. This breaks the CDK abstraction and loses CDK's progress monitoring, approval prompts, and error handling.

### Use case

Teams that require human review of infrastructure changes before deployment need a complete workflow within CDK:

```console
$ cdk deploy MyStack --method=prepare-change-set
# → Change set created, stack in REVIEW_IN_PROGRESS

# Review in AWS Console, run compliance checks, get approval...

$ cdk deploy MyStack --method=execute-change-set
# → Executes the change set with CDK's progress monitoring
```

This is especially valuable for:
- **Regulated environments** where infrastructure changes require explicit approval before execution
- **CI/CD pipelines** with separate "plan" and "apply" stages (similar to Terraform's workflow)
- **Team workflows** where one person prepares changes and another approves and executes them

### What this PR does

- Adds `--method=execute-change-set` to `cdk deploy`
- Defaults the change set name to `cdk-deploy-change-set` (matching `prepare-change-set`), so the simplest workflow needs no extra flags
- Shows an accurate diff based on the actual change set during the approval prompt
- Rejects options that only apply to change set creation (`--force`, `--parameters`, `--import-existing-resources`, `--revert-drift`) with clear errors
- Requires exactly one stack name
- Exposes `ExecuteChangeSetDeployment` as part of the public `DeploymentMethod` type in `@aws-cdk/toolkit-lib`

---

### Checklist

---

### DIFF
--- a/packages/aws-cdk/README.md
+++ b/packages/aws-cdk/README.md
@@ -409,6 +409,11 @@ be deployed and then executes it. This behavior can be controlled with the
 - `--method=prepare-change-set`: create the change set but don't execute it.
   This is useful if you have external tools that will inspect the change set or
   you have an approval process for change sets.
+- `--method=execute-change-set`: execute a previously created change set. This
+  bypasses change set creation and asset publishing entirely. Requires exactly
+  one stack name. Defaults to the `cdk-deploy-change-set` change set name if
+  `--change-set-name` is not provided. This is useful for two-step deployment
+  workflows where you first review a change set and then execute it.
 - `--method=direct`: do not create a change set but apply the change immediately.
   This is typically a bit faster than creating a change set, but it loses
   the progress information.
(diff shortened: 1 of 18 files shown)


packages/aws-cdk/README.md
```diff
@@ -428,8 +433,23 @@ set to make it easier to later execute:
...
+To review a change set before executing it, use a two-step workflow:
+
+```console
+$ # Step 1: Create the change set without executing it
+$ cdk deploy MyStack --method=prepare-change-set
+
+$ # Step 2: Review the change set (e.g., in the AWS Console or via CLI)
+
+$ # Step 3: Execute the change set
+$ cdk deploy MyStack --method=execute-change-set
```
otaviomacedo, line 445: Is the change set name mandatory?

mrgrain: @otaviomacedo it's optional, see next line: 

> A custom change set name can be provided with `--change-set-name` in both steps.

otaviomacedo (approved):

garysassano:
This is a nice addition, but it doesn't address the core issue raised in #1123. Ideally, there should be a more restrictive role that can prepare the changeset without having permission to also execute it.

PR #1273, opened 2026-03-28
fix(cli): use change set for deploy approval diff when using change-set method

mrgrain:
When deploying with `--method=change-set`, the approval diff shown to the user was inaccurate. It used a template-only comparison — the same approach as `cdk diff --method=template` — which overpredicts resource replacements. Any property change that *could* cause a replacement was always reported as one, even when CloudFormation would not actually replace the resource.

Meanwhile, `cdk diff --method=change-set` already solved this by creating a real CloudFormation change set and reading the accurate replacement information from it. But `cdk deploy` never used this approach for its approval diff, so users were making deployment approval decisions based on potentially wrong information.

This change fixes the disconnect by making `cdk deploy --method=change-set` use a real change set for the approval diff.

### How it works

The deploy flow changes from a single-phase to a two-phase approach:

```
Before:
  ┌──────────┐    ┌──────────────┐    ┌──────────────────┐    ┌─────────┐
  │ Synth    │───▶│ Template     │───▶│ Approval prompt  │───▶│ Create  │
  │          │    │ diff (local) │    │ (inaccurate diff)│    │ + Exec  │
  └──────────┘    └──────────────┘    └──────────────────┘    │ CS      │
                                                              └─────────┘

After:
  ┌──────────┐    ┌──────────────┐    ┌──────────────────┐    ┌─────────┐
  │ Synth    │───▶│ Create CS    │───▶│ Approval prompt  │───▶│ Execute │
  │          │    │ (no execute) │    │ (accurate diff)  │    │ same CS │
  └──────────┘    └──────────────┘    └──────────────────┘    └─────────┘
```

The change set is created upfront without executing it. Its description is fed to the diff formatter for an accurate approval diff. After the user approves, the same change set is executed — no extra change set is created.

If the change set is empty (no changes), the deploy exits early. If the user rejects the approval, the change set is cleaned up. On retries after a rollback, a fresh change set is created since the original one is gone.

### Checklist

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/shell.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/shell.ts
@@ -68,9 +68,22 @@ export async function shell(command: string[], options: ShellOptions = {}): Prom
 
           // now write the input with a slight delay to ensure
           // the child process has already started reading.
-          setTimeout(() => {
+          const sendInput = () => {
             child.writeStdin(interaction.input + (interaction.end ?? os.EOL));
-          }, 500);
+          };
+
+          if (interaction.beforeInput) {
+            void interaction.beforeInput()
+              .catch((err) => {
+                writeToOutputs(`\n[Prompt: ${interaction.prompt.toString()}] beforeInput hook failed!\n`);
+                writeToOutputs(`${err}\n\n`);
+              })
+              .finally(() => {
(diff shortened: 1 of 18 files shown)


mrgrain:
# Nested Stack deployment output

## Before

image

## After (no change)

image

rix0rrr (approved):

PR #1310, opened 2026-04-03
feat(toolkit-lib): Implement Hotswap with CCAPIs for broader resource coverage and improve intrinsic function resolution for `Get::Attr`

ShadowCat567:
The primary change in this PR is the introduction of hotswap with CCAPIs. This allows us to support resource hotswapping for new resource types without having to write custom implementations for each of them. 

Other changes in this PR: 
- **Improvements to resolution of Fn::GetAttr**: previously we could only resolve the CloudFormation intrinsic function, `Fn::GetAttr` for a small number of resources (11). This stems from us having hardcoded implementations for each of those resources. Now if a resource is not one we have a hardcoded attribute resolution for, we try to get it from the deployed Cloudformation template. If we cannot get it from the deployed Cloudformation template, we attempt to resolve it using its live state through CCAPIs. Only if we cannot resolve it with CCAPIs, do we throw an error that we cannot resolve that attribute.
- **Lambda Function hotswapping takes advantage of exponential backoff**: waiters in AWS SDK V3 support exponential back off and we were not taking advantage of that while hotswapping Lambda functions, hotswapping Lambda functions that involve Docker or VPCs should be a bit faster. 
- **Parallelize hotswapping with nested stacks**: Previously when hotswapping nested stacks we would do it sequentially, now nested stack hotswapping is done in parallel so it should be a bit faster

Resource types that are now supported by hotswap through CCAPIs:
- `AWS::ApiGateway::RestApi`
- `AWS::ApiGateway::Deployment`
- `AWS::ApiGateway::Method`
- `AWS::ApiGatewayV2::Api`
- `AWS::ApiGatewayV2::Integration`
- `AWS::Bedrock::Agent`
- `AWS::Events::Rule`
- `AWS::DynamoDB::Table`
- `AWS::DynamoDB::GlobalTable`
- `AWS::SQS::Queue`
- `AWS::CloudWatch::Alarm`
- `AWS::CloudWatch::CompositeAlarm`
- `AWS::CloudWatch::Dashboard`
- `AWS::StepFunctions::StateMachine` - previously supported by custom SDK implementation
- `AWS::BedrockAgentCore::Runtime` - previously supported by custom SDK implementation

---


packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
```diff
@@ -100,6 +100,12 @@ export interface EvaluateCloudFormationTemplateProps {
   readonly nestedStacks?: {
     [nestedStackLogicalId: string]: NestedStackTemplates;
   };
+  /**
+   * The processed template from CloudFormation with all intrinsic functions
+   * resolved to their concrete values from the last successful deployment.
```
rix0rrr: That is not what "processed" is, unfortunately, so I hope you're not basing too much functionality off of this assumption.

[For troubleshooting, use the processed template. If a template doesn't reference macros, the original and processed templates are identical.](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros-troubleshoot-processed-template.html)

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
```diff
@@ -144,6 +151,10 @@ export class EvaluateCloudFormationTemplate {
 
     // CloudFormation Exports lookup to be able to resolve Fn::ImportValue intrinsics in template
     this.lookupExport = new LazyLookupExport(this.sdk);
+
+    // The processed template has all intrinsics resolved by CloudFormation from the last deployment.
+    // Used as a fallback when local evaluation of Fn::GetAtt fails for unsupported resource types.
+    this.processedTemplate = props.processedTemplate ?? {};
```
rix0rrr: The order should be 

* `{ Ref }` -- get from CFN's `DescribeStackResources`
* `{ GetAtt }` -- either get from a custom handler, otherwise from CCAPI, otherwise fail

We *could* potentially get a `{ GetAtt }` from the template, if it is an input property that is set. But if we can parallellize all the necessary `{ GetAtt }`s I don't see why this should add a lot of time, and the extremely fast path is very speculative and rare so probably not worth it.

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
```diff
@@ -444,29 +456,51 @@ export class EvaluateCloudFormationTemplate {
...
-      account: this.account,
-      resourceType: resourceTypeArnPart,
-      resourceName: physicalId!,
-    });
+
+    // Last resort: query the live resource via Cloud Control API.
+    // This will fail for resource types with compound primaryIdentifiers because
+    // CloudFormation's PhysicalResourceId only contains a single value.
+    try {
+      const cloudControl = this.sdk.cloudControl();
```
rix0rrr: Do we really want a new SDK client for every single invocation? I imagine there will be a bunch.

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
```diff
@@ -444,29 +456,51 @@ export class EvaluateCloudFormationTemplate {
...
-      service,
-      region: this.region,
-      account: this.account,
-      resourceType: resourceTypeArnPart,
-      resourceName: physicalId!,
-    });
+
+    // Last resort: query the live resource via Cloud Control API.
+    // This will fail for resource types with compound primaryIdentifiers because
+    // CloudFormation's PhysicalResourceId only contains a single value.
```
rix0rrr: So... introduce a mapper! 

```ts
type CfnToCcapiIdentifierMapper = (cfnPhysicalId: string) => string;
```

Have a table of these, with `(x) => x` as a default for resources that don't occur in the table.

Aaannndd... it's probably fine for 90% of resources? 

> [For compound primary identifiers (that is, one that consists of multiple resource properties strung together), to specify the primary identifier as a string, list the property values in the order they are specified in the primary identifier definition, separated by |.](https://docs.aws.amazon.com/cloudcontrolapi/latest/APIReference/API_GetResource.html)

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
```diff
@@ -444,29 +456,51 @@ export class EvaluateCloudFormationTemplate {
...
+      const response = await cloudControl.getResource({
+        TypeName: resource.ResourceType!,
+        Identifier: physicalId!,
+      });
+      const props = JSON.parse(response.ResourceDescription?.Properties ?? '{}');
+      if (attribute in props) {
+        return props[attribute];
+      }
+    } catch {
+      // Cloud Control lookup failed — fall through to the error below
```
rix0rrr, line 521: Give a more descriptive error here, include the CCAPI one perhaps. How otherwise will we know what to fix?

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/cloud-control-resource.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  return ret;
+}
+
+/**
+ * Resolves the Cloud Control API identifier for a resource.
+ *
+ * CCAPI resources with compound primary identifiers need their identifiers to be
+ * built by joining each component with "|". CloudFormation's PhysicalResourceId
+ * only returns a single value, which doesn't work for compound keys.
```
rix0rrr, line 111: 9 times out of 10 the PhysicalResourceId will already be that string with components separated by `|`.

It should be only for certain legacy resources that they aren't (singular in CFN, compound in CCAPI).

In fact, [that list is here](https://github.com/cdklabs/awscdk-service-spec/tree/main/sources/CloudFormationRefOverrides). I would say we add a dependency on the service spec, and at build time extract that information and dump it into a JSON file that we use to decide what to do here (or that's the list we must have custom identifier mappers for, as mentioned before).

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
```diff
@@ -314,10 +314,11 @@ async function waitForLambdasPropertiesUpdateToFinish(
 
   // if the function is deployed in a VPC or if it is a container image function
   // then the update will take much longer and we can wait longer between checks
-  // otherwise, the update will be quick, so a 1-second delay is fine
-  const delaySeconds = functionIsInVpcOrUsesDockerForCode ? 5 : 1;
+  // otherwise, the update will be quick. This is reflected in the delay times we
+  // set for exponential backoff while waiting
+  const minDelaySeconds = functionIsInVpcOrUsesDockerForCode ? 3 : 1, maxDelaySeconds = functionIsInVpcOrUsesDockerForCode ? 5 : 3;
```
rix0rrr: This is not a readable way of writing this (at least imo).

More code but more readable:

```ts
let minDelaySeconds;
let maxDelaySeconds;
if (functionIsInVpcOrUsesDockerForCode) {
  minDelaySeconds = 3;
  maxDelaySeconds = 5;
} else {
  minDelaySeconds = 1;
  maxDelaySeconds = 3;
}
```

Btw the difference between these number is so small that we might as well not bother.

The nice thing about exponential backoff is that it will automatically find its own stable point. Do not try to outsmart it, do not go "oh but what if the update finished between poll 13 and 14, then the user is waiting 30 seconds for nothing, better make the poll shorter so they don't have to wait..."

If you've already been waiting 30 seconds, chances are high you'll be waiting 30 seconds longer. Either things are fast and your polling intervals don't matter, or the service is busted and quicker polling won't make (much of) a difference and just degrades the service for everyone.

----

Relatedly, I was a little confused while reading this:

> Lambda Function hotswapping takes advantage of exponential backoff: waiters in AWS SDK V3 support exponential back off and we were not taking advantage of that while hotswapping Lambda functions, hotswapping Lambda functions that involve Docker or VPCs should be a bit faster.

Exponential backoff usually makes things slower (for good reason!). So why would the benefit of using it be that we are faster for Docker or VPCs?

rix0rrr (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
```diff
@@ -60,10 +60,8 @@ const RESOURCE_DETECTORS: { [key: string]: HotswapDetector } = {
   'AWS::AppSync::GraphQLSchema': isHotswappableAppSyncChange,
   'AWS::AppSync::ApiKey': isHotswappableAppSyncChange,
 
-  'AWS::BedrockAgentCore::Runtime': isHotswappableBedrockAgentCoreRuntimeChange,
```
rix0rrr, line 63: Should we not be deleting these functions now then?

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/cloud-control-resource.ts
```diff
@@ -0,0 +1,172 @@
...
+        resourceType,
+        physicalName: identifier,
+        metadata: evaluateCfnTemplate.metadataFor(logicalId),
+      }],
+    },
+    hotswappable: true,
+    service: 'cloudcontrol',
+    apply: async () => {
+      const cloudControl = sdk.cloudControl();
+      const currentResource = await cloudControl.getResource({
```
rix0rrr: I'm in doubt about this.

Upstairs, we already decided we had a "change" that's hotswappable. That means we already look at a "current state" and a "new state", and made a calculation based on the diff between them.

*(It might be that that state is inaccurate, but that's a different problem!)*

If we already have a current state that we calculated a diff on, it seems dangerous to me to fetch *another* current state, and do another diff calculation based on that.

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
ShadowCat567: I'm a little unclear about what your question is, are you asking:
1. Since AgentCore Runtime and StepFunctions are now handled by CCAPI, should we also be deleting their SDK hotswap implementations?
2. Why are you moving these resource types to use CCAPIs when they have perfectly good SDK hotswap implementations?
3. (or if you didn't notice AgentCore Runtime/Stepfunctions were switched to CCAPI support) Are you removing AgentCore Runtime/StepFunctions from hotswap capabilities?

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
ShadowCat567: The main thing I wanted to get out of the processed template was information about Custom resources. Most other things I can get from CCAPIs (it would also be nice to get information about resources not supported by CCAPIs from there as well).

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
ShadowCat567: This mostly comes from the Cloudformation team looking at our Update function implementation and complaining about it, I would like to keep exponential backoff for VPC/Docker (it might make them marginally faster). Don't mind leaving non-Docker/VPC functions alone with their linear back off.

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
ShadowCat567: no, this should maybe get passed in by the caller (?), let me see where I can move it where it is more reusable

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
rix0rrr: Oh I meant, since we are now implementing these using CCAPI, we can remove the custom implementations. No?

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
```diff
@@ -316,6 +324,24 @@ export class EvaluateCloudFormationTemplate {
...
+    if (resourceType in RESOURCE_TYPE_PRIMARY_IDENTIFIERS) {
+      const primaryProps = RESOURCE_TYPE_PRIMARY_IDENTIFIERS[resourceType];
+      const parts: string[] = [];
+
+      for (const propName of primaryProps) {
+        const templateVal = this.getResourceProperty(logicalId, propName);
+        if (templateVal) {
+          parts.push(await this.evaluateCfnExpression(templateVal));
+        } else {
+          parts.push(physicalId);
```
rix0rrr, line 357: This is very much not guaranteed to be correct, right?

This is just praying?

If so, a couple of lines of comments would be nice to describe under what conditions this is reasonable to do. A doc comment on the function would also be nice.

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
```diff
@@ -314,10 +314,20 @@ async function waitForLambdasPropertiesUpdateToFinish(
 
   // if the function is deployed in a VPC or if it is a container image function
   // then the update will take much longer and we can wait longer between checks
-  // otherwise, the update will be quick, so a 1-second delay is fine
-  const delaySeconds = functionIsInVpcOrUsesDockerForCode ? 5 : 1;
+  // otherwise, the update will be quick. This is reflected in the delay times we
+  // set for exponential backoff while waiting
+  let minDelaySeconds: number;
+  let maxDelaySeconds: number;
```
rix0rrr: I'm still not quite happy about this. It's not been proven to me that a plain exponential backoff isn't good enough to cover all cases. 

Here's the thing about exponential backoff: you're pretty much expected to never exceed the actual waiting time by a factor of 2 (think about why that's true).

- Does the actual update take 1s? You are guaranteed to see it in 2s.
- Does the actual update take 10s? You are guaranteed to see it in 20s.
- etc.

You can complain about "oh my god 10s wasted in the second case", but compared to the time you have to wait anyway, it's not that bad. 

And in fact, I ran a little simulation and it turns out to be ~1.5x the actual wait time on average, and gets better if we clamp the sleep on the top side (and slightly worse with jitter, which I admit I would not have predicted).

So I'd say we do the following:

```ts
await lambda.waitUntilFunctionUpdated(1, 10, {
```

And stop thinking about it.

----------

My little simulation script in case you want to play with the numbers to convince yourself:

```py
import random


def expo(target):
  current = 0
  sleep = 100
  while current < target:
    current += sleep
    sleep *= 2
  return current


def expo_jitter(target):
  current = 0
  sleep = 100
  while current < target:
    current += random.randint(0, sleep)
    sleep *= 2
  return current


def expo_jitter_clamped(target, max_sleep):
  current = 0
  sleep = 100
  while current < target:
    current += min(max_sleep, random.randint(0, sleep))
    sleep *= 2
  return current


def avg(xs):
  return sum(xs) / len(xs)


targets = [random.randint(1000, 60000) for _ in range(10000)]

print('expo', avg([expo(target) / target for target in targets]))
print('expo_jitter', avg([expo_jitter(target) / target for target in targets]))
print('expo_jitter_clamped', avg([expo_jitter_clamped(target, 10000) / target for target in targets]))
```

ShadowCat567: this sounds reasonable to me, this would also decrease the complexity of our code so we don't have to check whether the function uses VPC or Docker anymore

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
ShadowCat567: it works most of the time, but there are some cases where the identifier we construct will not be correct (mostly cases where the attribute we need is not resolvable for some reason) - added a comment

rix0rrr (approved):

PR #1319, opened 2026-04-07
feat(yarn-cling): support yarn berry lockfiles using @yarnpkg/parsers

mrgrain:
yarn-cling previously only understood yarn classic (v1) lockfiles via the `@yarnpkg/lockfile` package. Repos that are migrating to yarn berry (v2/v3/v4) use a different lockfile format, which caused shrinkwrap generation to silently produce incomplete output since none of the dependencies were resolved from the lockfile.

This replaces `@yarnpkg/lockfile` with `@yarnpkg/parsers`, the official parser package from the yarn berry project. `@yarnpkg/parsers` handles both v1 and berry lockfile formats, so we no longer need two separate parser dependencies. For berry lockfiles, a conversion step maps the berry-specific fields (`resolution` → `resolved`, `checksum` → `integrity`) and `npm:` prefixed descriptor keys to the classic `YarnLock` interface used by the rest of the pipeline.

Additionally replaces the deprecated `fs.exists` usage with the existing `fileExists` helper based on `fs.stat`.

### Checklist

---


packages/@aws-cdk/yarn-cling/test/test-fixture/yarn.lock
```diff
@@ -1,3 +1,6 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
mrgrain, line 2: the new library uses the header comment to detect v1 lockfiles. I think this is a safe change to make in favor of dropping another dependency. Yarn Classic generated lockfiles will have the header.

rix0rrr (approved):

PR #1324, opened 2026-04-09
feat(cli): add `cdk orphan` command to detach resources from a stack

rix0rrr:
Adds a new CLI command that safely removes resources from a CloudFormation stack without deleting them, enabling resource type migrations (e.g. DynamoDB Table to GlobalTable).

`cdk orphan --path <ConstructPath>` will:
- Find all resources under the construct path via aws:cdk:path metadata
- Resolve {Ref} values via DescribeStackResources
- Resolve {Fn::GetAtt} values by injecting temporary stack Outputs
- Set DeletionPolicy: Retain on all matched resources
- Remove the resources from the stack
- Output an inline `cdk import` command with the resource mapping

Also adds inline JSON support for `--resource-mapping` in `cdk import`, and exposes `stackSdk()` on Deployments for read-only SDK access.

---


rix0rrr:
This is actually @LeeroyHannigan's change, I'm just turning it into a PR so I can leave comments more conveniently.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts
```diff
@@ -314,6 +314,15 @@ export class Deployments {
     return loadCurrentTemplate(stackArtifact, env.sdk);
   }
 
+  /**
+   * Get the CloudFormation SDK client for a stack's environment.
+   * Used by the orphaner to call DescribeStackResources.
+   */
+  public async stackSdk(stackArtifact: cxapi.CloudFormationStackArtifact) {
```
rix0rrr: Do we need this method? We probably should just inline this at the call site.

packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts
```diff
@@ -141,8 +141,10 @@ export class ResourceImporter {
   /**
    * Load the resources to import from a file
    */
-  public async loadResourceIdentifiers(available: ImportableResource[], filename: string): Promise<ImportMap> {
-    const contents = await fs.readJson(filename);
+  public async loadResourceIdentifiers(available: ImportableResource[], filenameOrJson: string): Promise<ImportMap> {
```
rix0rrr: Not a fan of implicitly overloading an argument like this. I'd rather be explicit, like this:

```ts
type ResourceIdentifiersSource = 
  | { type: 'file'; fileName: string }
  | { type: 'direct'; resourceIdentifiers: Record<string, ResourceIdentifierProperties> };


 public async loadResourceIdentifiers(available: ImportableResource[], source: ResourceIdentifiersSource): Promise<ImportMap> {
```


------

Or honestly, even simpler

```ts
 public async loadResourceIdentifiersFromFile(available: ImportableResource[], fileName: string): Promise<ImportMap> {
   const contents = /* load file */;
   return this.loadResourceIdentifiers(available, contents);
}

 public async loadResourceIdentifiers(available: ImportableResource[], identifiers: Record<string, ResourceIdentifierProperties>): Promise<ImportMap> {

```

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -934,6 +935,32 @@ export class CdkToolkit {
       });
   }
 
+  public async orphan(options: OrphanOptions) {
+    const stacks = await this.selectStacksForDeploy(options.selector, true, true, false);
```
rix0rrr: If the argument is one or more construct paths (which it should be) then the stack selection is implicit. No need for the user to pick the stack.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -934,6 +935,32 @@ export class CdkToolkit {
...
+
+    if (stacks.stackCount > 1) {
+      throw new ToolkitError(
+        'AmbiguousStackSelection',
+        `Stack selection is ambiguous, please choose a specific stack for orphan [${stacks.stackArtifacts.map((x) => x.id).join(', ')}]`,
+      );
+    }
+
+    const stack = stacks.stackArtifacts[0];
+    await this.ioHost.asIoHelper().defaults.info(chalk.bold(`Orphaning construct '${options.constructPath}' from ${stack.displayName}`));
```
rix0rrr: We should split this into "plan" and "execute" stages:

- First determine actually which constructs are going to be orphaned, by evaluating the construct path. And include what resources are going to have their references replaced with literals. It should probably be an error if there are 0. Then we print a proper report of what we're going to do to the user (not a guess at the result of their instruction). We should probably give them a chance to confirm as well.
- Only after they confirm do we do the actual work.

I would like to see that in the API in some way. For example:


```ts
class Orphaner {
  constructor(...) { }

  public async makePlan(...): Promise<OrphanPlan> { 
    // ...
  }
}

class OrphanPlan {
   // The properties below here are purely hypothetical to show the idea! I have not done enough
   // mental design to think about whether these are the best to expose.
   public readonly orphanedResoures: OrpanPlanResource[];
   public readonly affectedResources: OrpanPlanResource[];
   public readonly stacks: OrphanPlanStack[];

   public async execute() {
     // ...
   }
}


class Toolkit {

   public async orphan(...) { 

     // And then something like this
     const orphaner = new Orphaner(...):
     const plan = await orphaner.makePlan(...);
    
     await showPlanToUser(plan, this.ioHost);
     const yes = await this.ioHost.confirm(...);
     if (yes) {
       await plan.execute();
     }
   }
}
```

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -934,6 +935,32 @@ export class CdkToolkit {
...
+    const orphaner = new ResourceOrphaner({
+      deployments: this.props.deployments,
+      ioHelper: asIoHelper(this.ioHost, 'deploy'),
+    });
+
+    await orphaner.orphan({
+      stack,
+      constructPath: options.constructPath,
+      roleArn: options.roleArn,
+      toolkitStackName: this.toolkitStackName,
```
rix0rrr: Feels like both of these should be constructor arguments.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1891,6 +1918,13 @@ export interface RollbackOptions {
   readonly validateBootstrapStackVersion?: boolean;
 }
 
+export interface OrphanOptions {
+  selector: StackSelector;
+  constructPath: string | string[];
+  roleArn?: string;
+  toolkitStackName?: string;
```
rix0rrr: Customarily we make all of these `readonly`

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -452,6 +452,15 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           concurrency: args.concurrency,
         });
 
+      case 'orphan':
+        ioHost.currentAction = 'deploy';
```
rix0rrr: Well no. A new `action` 😉

packages/aws-cdk/lib/cli/parse-command-line-arguments.ts
```diff
@@ -669,6 +669,33 @@ export function parseCommandLineArguments(args: Array<string>): any {
...
+    .command('orphan [STACK]', 'Orphan a resource from a stack (detach without deleting)', (yargs: Argv) =>
+      yargs
+        .positional('STACK', {
+          type: 'string',
+          desc: 'The stack to orphan the resource from',
+        })
+        .option('path', {
+          type: 'array',
+          string: true,
+          desc: 'Construct path(s) of the resource to orphan (e.g. MyTable). Can be specified multiple times.',
```
rix0rrr: A construct path looks like `/MyStack/MyTable`.

rix0rrr: Doesn't the top of this file say "GENERATED, do not edit by hand"? I'm pretty sure it does.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+   * The stack to operate on.
+   */
+  readonly stack: cxapi.CloudFormationStackArtifact;
+
+  /**
+   * Construct path prefix(es). All resources whose `aws:cdk:path` metadata
+   * starts with `<StackName>/<path>/` will be orphaned together.
+   * Pass multiple paths to include related infrastructure.
+   */
+  readonly constructPath: string | string[];
```
rix0rrr: `string[]` subsumes `string`, so no need to have both.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+
+export interface ResourceOrphanerProps {
+  readonly deployments: Deployments;
+  readonly ioHelper: IoHelper;
+}
+
+/**
+ * Orphans all resources under a construct path from a CloudFormation stack.
+ *
+ * 1. Finds all resources whose aws:cdk:path matches the construct path
```
rix0rrr: We should detect if that metadata has been disabled, and fail the deployment if so. Are we doing that?

(Alternatively we discover this when we end up matching 0 resources, and then we go and look whether there are any `aws:cdk:path` metadatas anywhere and add a helpful note to the error message)

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    if (logicalIds.length === 0) {
+      throw new Error(`No resources found under construct path '${paths.join(', ')}' in stack '${stack.stackName}'`);
+    }
+
+    await this.ioHelper.defaults.info(`Found ${logicalIds.length} resource(s) to orphan: ${logicalIds.join(', ')}`);
+
+    // 4. Identify the primary resource (the one with the shortest path — typically the main resource)
+    const primaryId = this.findPrimaryResource(logicalIds, resources);
```
rix0rrr: There might be more than one since construct paths is a list.

```
cdk orphan /MyStack/Bucket /MyStack/Table
```

Which now is the "main" resource?

(Also I don't know that you need this)

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    const physicalName = this.extractPhysicalName(resources[primaryId]);
+
+    // 4. Prepare step 1 template: set RETAIN, replace Ref with physical IDs,
+    //    and inject temporary Outputs for each GetAtt so CloudFormation resolves them.
+    const templateStep1 = JSON.parse(JSON.stringify(currentTemplate));
```
rix0rrr: I know we don't have a great location for this right now, but I would like to concentrate the utilities for working with CloudFormation templates concentrated in some central location. So can we delegate the parsing and updating to some helper functions in a file that gets stuffed into a location that seems appropriate? Probably close to other functions that are already doing similar things.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    for (const id of logicalIds) {
+      const v = values.get(id);
+      if (v) {
+        this.replaceReferences(templateStep1, id, v);
+      }
+      templateStep1.Resources[id].DeletionPolicy = 'Retain';
+      templateStep1.Resources[id].UpdateReplacePolicy = 'Retain';
+    }
+
+    // 5. Deploy step 1 (sets RETAIN, replaces Refs with literals, adds temp outputs for GetAtts)
```
rix0rrr: > adds temp outputs for GetAtts

Sorry what now?

rix0rrr: Oh I see.

This is a clever trick but we should probably just be using Cloud Control API to get those attribute values. 

This is fine for now but once [this PR lands](https://github.com/aws/aws-cdk-cli/pull/1310) we should refactor this. Can you decouple the step of rewriting the template in order to retrieve the `Fn::GetAtt` values and make it independent of anything else, put it in a function, and call it first? We should then be able to replace it with a CCAPI-based method once we have that code in a good place, while the rest of this code won't have to change.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+      overrideTemplate: templateStep1,
+      usePreviousParameters: true,
+      forceDeployment: true,
+    });
+    if (step1Result.type !== 'did-deploy-stack') {
+      throw new Error(`Step 1 failed: unexpected result type '${step1Result.type}'`);
+    }
+
+    // 6. Read the resolved GetAtt values from the temporary outputs
+    await this.readResolvedGetAttValues(stack, logicalIds, currentTemplate, values);
```
rix0rrr: Do. Not. Like. In-place. Modifications.

Prefer to return values rather than mutating objects in place (golden rule that may be broken only for performance reasons).

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+
+    // 7. Prepare step 2 template: replace GetAtts with resolved literals
+    const templateStep2 = JSON.parse(JSON.stringify(templateStep1));
```
rix0rrr: What is the point of stringifying and then parsing? If we want a clone, there is an API for that (`structuredClone`).

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    for (const id of logicalIds) {
+      const v = values.get(id);
+      if (v) {
+        this.replaceReferences(templateStep2, id, v);
```
rix0rrr: Again, template manipulation should go into dedicated helper functions.

Maybe a 

```ts
class CloudFormationTemplate {
}
```

With functions on it to operate on it.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    // 8. Remove all orphaned resources
+    const templateStep3 = JSON.parse(JSON.stringify(templateStep2));
+    for (const id of logicalIds) {
+      delete templateStep3.Resources[id];
+      this.removeDependsOn(templateStep3, id);
+    }
+
+    // 9. Deploy step 3 (resources are orphaned)
+    await this.ioHelper.defaults.info('Step 3/3: Removing resources from stack...');
+    const step3Result = await this.deployments.deployStack({
```
rix0rrr: We should probably check whether any of this leads to resource replacement, and warn and confirm an additional time if so. We will need an actual change set to accurately check for replacement.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    if (step3Result.noOp) {
+      const blockers = this.findBlockingResources(templateStep3, logicalIds, templateStep2);
+      throw new Error(
+        'Step 3 was a no-op — resources were not removed from the stack. ' +
+        'Other resources still reference the orphaned resources' +
+        (blockers.length > 0
+          ? `: ${blockers.join(', ')}. Include them with additional --path flags.`
+          : '. Check the deployed template for remaining {Ref} or {Fn::GetAtt} references.'),
+      );
+    }
```
rix0rrr: Is this a result of something happening in practice?

Wouldn't the deployment have failed if there were remaining references? 

The error message doesn't make sense: it points to a mistake in our logic, not to a user mistake.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    await this.ioHelper.defaults.info(`✅ Construct '${constructPath}' orphaned from ${stack.stackName}`);
+    await this.ioHelper.defaults.info('');
+    if (physicalName) {
+      await this.ioHelper.defaults.info(`  Physical resource: ${physicalName}`);
+      await this.ioHelper.defaults.info('');
+    }
+    await this.ioHelper.defaults.info('Next steps:');
+    await this.ioHelper.defaults.info('  1. Update your code to use the new resource type');
+    if (mappingJson) {
+      await this.ioHelper.defaults.info(`  2. cdk import --resource-mapping '${mappingJson}' --force`);
```
rix0rrr: Why the `--force` here?

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+        'Step 3 was a no-op — resources were not removed from the stack. ' +
+        'Other resources still reference the orphaned resources' +
+        (blockers.length > 0
+          ? `: ${blockers.join(', ')}. Include them with additional --path flags.`
+          : '. Check the deployed template for remaining {Ref} or {Fn::GetAtt} references.'),
+      );
+    }
+
+    // 8. Output next steps
+    const identifier = primaryId && physicalName ? this.extractResourceIdentifier(resources[primaryId]) : undefined;
```
rix0rrr: This is starting to get a little too GlobalTable-specific here. There might be more than 1 relevant resource for example.

Can't we print all primary identifiers and let the user decide what to supply when running import?

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+
+  /**
+   * Find all resources whose aws:cdk:path starts with `<stackName>/<constructPath>/`.
+   */
+  private findResourcesByPath(resources: Record<string, any>, stackName: string, constructPath: string): string[] {
+    const prefix = `${stackName}/${constructPath}/`;
+    const ids: string[] = [];
+
+    for (const [id, resource] of Object.entries(resources)) {
+      const cdkPath = resource.Metadata?.['aws:cdk:path'] ?? '';
```
rix0rrr: This should be a constant defined somewhere in `@aws-cdk/cx-api`. Use that constant.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+   */
+  private extractResourceIdentifier(resource: any): Record<string, string> | undefined {
+    const props = resource.Properties ?? {};
+    if (props.TableName) return { TableName: props.TableName };
+    if (props.BucketName) return { BucketName: props.BucketName };
+    if (props.QueueName) return { QueueName: props.QueueName };
+    if (props.TopicName) return { TopicName: props.TopicName };
+    if (props.FunctionName) return { FunctionName: props.FunctionName };
+    if (props.RoleName) return { RoleName: props.RoleName };
+    return undefined;
```
rix0rrr: Oh my god no. We can get this information from CloudFormation I'm pretty sure. `cdk import` uses that. Check it out.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  /**
+   * Extract the physical name from resource properties.
+   */
+  private extractPhysicalName(resource: any): string | undefined {
```
rix0rrr: Again, not the way this is implemented. But there's a high chance you won't find the resource name in its properties. So we shouldn't rely on this.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    return props.TableName ?? props.BucketName ?? props.FunctionName ??
+      props.QueueName ?? props.TopicName ?? props.RoleName ?? props.Name;
+  }
+
+  /**
+   * Resolve Ref and GetAtt values from known resource type patterns.
+   */
+  /**
+   * Resolve all Ref and GetAtt values for orphaned resources.
+   *
```
rix0rrr: Orphaned comment.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    stack: cxapi.CloudFormationStackArtifact,
+    logicalIds: string[],
+    currentTemplate: any,
+    templateWithLiterals: any,
+  ): Promise<{ values: Map<string, { ref: string; attrs: Record<string, string> }>; tempOutputKeys: string[] }> {
+    const sdk = await this.deployments.stackSdk(stack);
+    const cfn = sdk.cloudFormation();
+
+    // Get physical resource IDs (Ref values) from CloudFormation
+    const describeResult = await cfn.describeStackResources({ StackName: stack.stackName });
```
rix0rrr: Is this paginated?

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    for (const section of ['Resources', 'Outputs', 'Conditions']) {
+      if (!template[section]) continue;
+      for (const [key, value] of Object.entries(template[section])) {
+        if (section === 'Resources' && key === logicalId) continue;
+        template[section][key] = this.replaceInObject(value, logicalId, values);
+      }
+    }
+  }
+
+  private replaceInObject(obj: any, logicalId: string, values: { ref: string; attrs: Record<string, string> }): any {
```
rix0rrr: Doesn't use `this` so doesn't need to be a method. Could be a helper function.

And in fact should be since it operates on a CFN template.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    const result: any = {};
+    for (const [key, value] of Object.entries(obj)) {
+      result[key] = this.replaceInObject(value, logicalId, values);
+    }
+    return result;
+  }
+
+  private removeDependsOn(template: any, logicalId: string): void {
```
rix0rrr: Doesn't use `this` so doesn't need to be a method. Could be a helper function.

And in fact should be since it operates on a CFN template.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
```diff
@@ -0,0 +1,458 @@
...
+      if (Array.isArray(res.DependsOn)) {
+        res.DependsOn = res.DependsOn.filter((dep: string) => dep !== logicalId);
+        if (res.DependsOn.length === 0) delete res.DependsOn;
+      } else if (res.DependsOn === logicalId) {
+        delete res.DependsOn;
+      }
+    }
+  }
+
+  private walkObject(obj: any, visitor: (value: any) => void): void {
```
rix0rrr: Doesn't use `this` so doesn't need to be a method. Could be a helper function.

And in fact should be since it operates on a CFN template.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -934,6 +935,32 @@ export class CdkToolkit {
       });
   }
 
+  public async orphan(options: OrphanOptions) {
```
rix0rrr, line 972: This feature needs to be `--unstable`. Check out other code to see how we do that.

rix0rrr (commented):
Great start! 

Initial round of comments on this.

This also needs an integ test (and to be frank probably more than 1 😉 ).

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: The TS target for toolkit-lib is es2022 so `structuredClone` isn't available without a cast. Kept `JSON.parse(JSON.stringify(...))` to stay consistent with the rest of the codebase (e.g. refactoring/digest.ts). Happy to switch if we bump the target.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
mrgrain, line 972: this should be implemented fully in `toolkit-lib`'s `Toolkit` class, with the CLI passing through. There are existing actions that follow that pattern.

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: nice catch, its not (truncates at 100). I used `listStackResources` which is paginated, instead. This meant I needed permissions on the CLI role, to run integ tests. LMK if thats not correct.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done, walkObject, replaceInObject, replaceReferences, removeDependsOn, findResourcesByPath, findBlockingResources, hasAnyCdkPathMetadata, and assertSafeDeployResult are all standalone functions in actions/orphan/private/helpers.ts. None of them are methods on the class.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done, walkObject, replaceInObject, replaceReferences, removeDependsOn, findResourcesByPath, findBlockingResources, hasAnyCdkPathMetadata, and assertSafeDeployResult are all standalone functions in actions/orphan/private/helpers.ts. None of them are methods on the class.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done, walkObject, replaceInObject, replaceReferences, removeDependsOn, findResourcesByPath, findBlockingResources, hasAnyCdkPathMetadata, and assertSafeDeployResult are all standalone functions in actions/orphan/private/helpers.ts. None of them are methods on the class.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: Done

LeeroyHannigan: Done, thanks.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Removed

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Kept them scoped to orphan for now since no other code needs them yet,  happy to move to a shared location if you'd prefer, or as a follow-up when other commands need similar utilities.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Removed the hardcoded `extractPhysicalName` (which checked props.TableName, props.BucketName, etc.) and replaced it with `PhysicalResourceId` from `ListStackResources`, combined with `ResourceIdentifierSummaries` from `GetTemplateSummary` to know which properties are the import identifiers.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Replaced it with `deployments.resourceIdentifierSummaries(stack)` which calls CloudFormation's `GetTemplateSummary` API

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done. I replaced the old `findPrimaryResource` + hardcoded `extractResourceIdentifier` with `getResourceIdentifiers` that uses `GetTemplateSummary` API and outputs all importable resources. Then filter to only primary construct resources (path matches `StackName/ConstructPath/Resource` exactly) to 
avoid noise from internal CDK resources like managed policies.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Removed, but: when I tested it earlier, cdk import failed with *"No resource updates or deletes are allowed on import operation"* because switching from Table to TableV2 also changes other resources (nested stack, policies, ESM). I had to add `--force` manually. Up to you how we handle it.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done. I rewrote the error message to acknowledge it's an unexpected situation, not a user mistake. Let me know if you want to remove it entirely.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done. All three deploy steps now go through `assertSafeDeployResult()` which checks the deploy result type. If CloudFormation detects a replacement ('replacement-requires-rollback'), we abort immediately with a clear error before any resources are affected. The existing deployStack flow already creates a changeset and inspects it for replacements before executing.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: All template manipulation is extracted into standalone helper functions in `actions/orphan/private/helpers.ts`. I didn't create a CloudFormationTemplate class or move them to a shared location, they're scoped to orphan since no other code needs them yet.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done. The old `readResolvedGetAttValues` mutated the values map in place. The new `resolveGetAttValues` returns a fresh `Map<string, ResolvedValues>`  no mutation of inputs.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done. The `GetAtt` resolution is isolated in `resolveGetAttValues()`  a self-contained method that owns the entire lifecycle (inject temp outputs → deploy → read outputs → return resolved values). The rest of `execute()` just receives the resolved values and doesn't know how they were obtained. 

Swapping in CCAPI later means replacing that one method.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: I looked for an existing shared location for CFN template utilities but didn't find one that fits. Happy to move these to a new shared module (e.g. api/cloudformation/template-helpers.ts) if you have a preferred location in mind, or we can do that as a follow-up.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done. Removed `findPrimaryResource` entirely. The resource mapping now includes all primary resources (one per construct path, matched by exact `StackName/ConstructPath/Resource` path). No single "main" resource, all are included in the import command output.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done. When 0 resources match, we call `hasAnyCdkPathMetadata()` to check if any resources in the stack have the metadata at all. If none do, the error includes: "(no resources in this stack have aws:cdk:path metadata — was it disabled?)". Unit test covers this case.

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
LeeroyHannigan: Done

packages/aws-cdk/lib/cli/parse-command-line-arguments.ts
LeeroyHannigan: Done

packages/aws-cdk/lib/cli/cli.ts
LeeroyHannigan: Done

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: Done

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: Done. `roleArn` and `toolkitStackName` are on `ResourceOrphanerProps` (the constructor), not on the `orphan()` method options.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: Done. Implemented exactly as described:

- `makePlan(stack, constructPaths)` → returns `OrphanPlan` with `stackName`, `orphanedResources` (each with `logicalId`, `resourceType`, `cdkPath`), and an `execute()` method. Read-only, no deployments.
- `Toolkit.orphan()` shows the plan, then confirms via `requestResponse` (skippable with `--force`).
- `plan.execute()` runs the 3 CloudFormation deployments, returns `OrphanResult` with `resourceMapping`.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: Done

packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts
LeeroyHannigan: Done. Split into `loadResourceIdentifiersFromFile(available, fileName)` which loads the file and calls `loadResourceIdentifiers(available, identifiers)`. The CLI layer decides which to call. Also added `--resource-mapping-inline` as a separate CLI option for passing JSON directly (used by cdk orphan output).

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts
LeeroyHannigan: Done. Removed `stackSdk()` from Deployments. The orphaner accesses the SDK via `deployments.envs.accessStackForReadOnlyStackOperations(stack)` directly.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
rix0rrr: Good call, I thought I check that but apparently didn't.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
```diff
@@ -0,0 +1,78 @@
...
+    const outputs = describeResponse.Stacks?.[0]?.Outputs ?? [];
+    const tableName = outputs.find((o) => o.OutputKey === 'TableName')?.OutputValue;
+    const functionName = outputs.find((o) => o.OutputKey === 'FunctionName')?.OutputValue;
+    expect(tableName).toBeDefined();
+    expect(functionName).toBeDefined();
+
+    // Put an item in the table before orphan
+    const itemFile = path.join(fixture.integTestDir, 'test-item.json');
+    await fs.writeFile(itemFile, JSON.stringify({ PK: { S: 'before-orphan' } }));
+    await fixture.shell([
```
rix0rrr: We are in a programming language, and we have an AWS SDK available to us (or we could without a lot of effort). I would prefer us using `@aws-sdk/client-dynamodb` directly.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
```diff
@@ -0,0 +1,78 @@
...
+    // Verify the output contains a resource mapping for import
+    expect(orphanOutput).toContain('resource-mapping-inline');
+    expect(orphanOutput).toContain('TableName');
+
+    // Verify the table is no longer in the stack template
+    const templateAfter = await fixture.aws.cloudFormation.send(
+      new GetTemplateCommand({ StackName: stackName }),
+    );
+    const templateBody = yaml.parse(templateAfter.TemplateBody!);
+    expect(templateBody.Resources).not.toHaveProperty('MyTable794EDED1');
```
rix0rrr: This would also silently succeed if the logical ID changed, for some reason. We should also check that the expected resource is present before we start orphaning.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Verify the Lambda still exists and its env vars have been replaced with literals
+    // (Ref -> physical table name, GetAtt -> physical ARN)
+    const lambdaResource = Object.values(templateBody.Resources).find(
```
rix0rrr: This is very hard to read, and if any of these steps fail the output will not be helpful to diagnose what the problem is.

I'd rather use a single structural assertion like:

```ts
expect(templateBody).toMatchObject({
  Resources: {
    MyFunction1234: {
     Properties: {
       ... 
     }
   }
  }
});
```

etc.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Verify the table still exists and data is intact
+    const getItemOutput = await fixture.shell([
+      'aws', 'dynamodb', 'get-item',
+      '--table-name', tableName!,
+      '--key', `file://${itemFile}`,
+      '--region', fixture.aws.region,
+      '--output', 'json',
+    ]);
+    expect(getItemOutput).toContain('before-orphan');
```
rix0rrr: We need to clean up that table otherwise it will leak.

Stick it in a `try/finally`, I really don't want to leak it even if the test fails.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/index.ts
```diff
@@ -0,0 +1,26 @@
+export interface OrphanOptions {
+  /**
+   * Construct path prefix(es) to orphan. Each path must be in the format
+   * `StackName/ConstructPath`, e.g. `MyStack/MyTable`.
+   *
+   * The stack is derived from the path — all paths must reference the same stack.
```
rix0rrr, line 6: Ah yeah? 😅 

Bummer. But acceptable for now I guess.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+    for (const res of stackResources) {
+      if (res.LogicalResourceId && res.PhysicalResourceId) {
+        physicalIds.set(res.LogicalResourceId, res.PhysicalResourceId);
+      }
+    }
+
+    // Build resolved values with Ref values
+    const values = new Map<string, ResolvedValues>();
+    for (const id of logicalIds) {
+      values.set(id, { ref: physicalIds.get(id) ?? id, attrs: {} });
```
rix0rrr: What is this fallback `?? id`? That is not correct.

The physical ID of a resource is not "well I guess it's the same as the logical ID then". There is an assumption here that we should not fake. Better explode if we can't satisfy it.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/private/helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+  return false;
+}
+
+import type { DeployStackResult, SuccessfulDeployStackResult } from '../../../api/deployments/deployment-result';
+import { ToolkitError } from '../../../toolkit/toolkit-error';
+
+/**
+ * Verify a deploy result is successful and did not trigger resource replacement.
+ */
+export function assertSafeDeployResult(result: DeployStackResult, step: string): asserts result is SuccessfulDeployStackResult {
```
rix0rrr: This function doesn't entirely make sense.

"Would not trigger resource replacement" -> well we're checking this only after already having done it, so if we would have a resource replacement then it's too late.

`replacement-requires-rollback` -> this status code means that you're not allowed to use `--no-rollback` with a deployment that requires replacement. Since you're never requesting no-rollback mode, this will never occur, and also the error message thrown here doesn't make sense.

The check for `did-deploy-stack` is correct, but if that's all this function does its name should be changed.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+      usePreviousParameters: true,
+      forceDeployment: true,
+    });
+    assertSafeDeployResult(step2Result, 'Step 2');
+
+    // Step 3: Remove orphaned resources
+    const templateStep3 = JSON.parse(JSON.stringify(templateStep2));
+    for (const id of logicalIds) {
+      delete templateStep3.Resources[id];
+      removeDependsOn(templateStep3, id);
```
rix0rrr: Removing DependsOn could have been done in the previous step already. Makes more sense to me to do all the "decoupling" in one step.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    // CloudFormation requires at least one resource in the template
+    if (Object.keys(templateStep3.Resources).length === 0) {
```
rix0rrr: Neat.

Helper function?

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+    for (const id of logicalIds) {
+      values.set(id, { ref: physicalIds.get(id) ?? id, attrs: {} });
+    }
+
+    // Resolve GetAtt values via temporary stack outputs.
+    // This is decoupled so it can be replaced with Cloud Control API later.
+    const attrValues = await this.resolveGetAttValues(stack, logicalIds, currentTemplate, values);
+
+    // Step 2: Replace all references with literals, remove temp outputs
+    const templateStep2 = JSON.parse(JSON.stringify(currentTemplate));
```
rix0rrr: Can we name `templateStep2` and `templateStep3` something more useful, about what it represents? I had to go to remember and check if `templateStep3` didn't have the bogus outputs anymore.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+  ): Promise<Map<string, ResolvedValues>> {
+    const getAttRefs = this.findGetAttReferences(currentTemplate, logicalIds);
+
+    // If there are no GetAtt references, skip the deploy — just return Ref values
+    if (getAttRefs.length === 0) {
+      return refValues;
+    }
+
+    // Inject temporary outputs and set RETAIN
+    const templateStep1 = JSON.parse(JSON.stringify(currentTemplate));
```
rix0rrr: Here as well, rename this variable.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+    if (!templateStep1.Outputs) {
+      templateStep1.Outputs = {};
+    }
+    for (const ref of getAttRefs) {
+      templateStep1.Outputs[ref.outputKey] = {
+        Value: { 'Fn::GetAtt': [ref.logicalId, ref.attr] },
+      };
+    }
+    for (const id of logicalIds) {
```
rix0rrr: This doesn't have to happen here, does it? It happens as part of step 2.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+      replaceReferences(templateStep1, id, refValues.get(id)!);
+      templateStep1.Resources[id].DeletionPolicy = 'Retain';
+      templateStep1.Resources[id].UpdateReplacePolicy = 'Retain';
+    }
+
+    await this.ioHelper.defaults.info('Step 1/3: Resolving attribute values...');
```
rix0rrr: Don't do this printing here. Print all the things that logically belong together in the same function.

If we print `step 2/3` and `step 3/3` in another function, then we should print `step 1/3` there as well.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+      deploymentMethod: { method: 'change-set', changeSetName: 'cdk-orphan-step1' },
+      overrideTemplate: templateStep1,
+      usePreviousParameters: true,
+      forceDeployment: true,
+    });
+    assertSafeDeployResult(step1Result, 'Step 1');
+
+    // Read resolved values from stack outputs
+    const env = await this.deployments.envs.accessStackForReadOnlyStackOperations(stack);
+    const cfn = env.sdk.cloudFormation();
```
rix0rrr: I've seen this happen too often. Grab this client once and reuse it?

If that's awkward, then the lifetime of the classes may be incorrect. It's perfectly fine for a class to represent "an operation", and that operation should be use the same CFN client all the way through.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+    const stackDesc = await cfn.describeStacks({ StackName: stack.stackName });
+
+    // Build result with both Ref and resolved GetAtt values
+    const result = new Map<string, ResolvedValues>();
+    for (const [id, rv] of refValues) {
+      result.set(id, { ref: rv.ref, attrs: { ...rv.attrs } });
+    }
```
rix0rrr: What is this doing? Just copying that map? Why has it been built in the calling function already, shouldn't it be fully constructed here?

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+
+    for (const output of stackDesc.Stacks?.[0]?.Outputs ?? []) {
+      if (!output.OutputKey || !output.OutputValue) continue;
+      const ref = getAttRefs.find(r => r.outputKey === output.OutputKey);
```
rix0rrr: If `getAttRefs` was a map or an object, we wouldn't need to search a list here.

Doubly-indexing maps is a bit brain breaking but perfectly legit and more concise.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
rix0rrr: Not convinced building this map here is correct, shouldn't this be done in the next function?

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+
+      for (const id of logicalIds) {
+        const resource = resources[id];
+        if (!resource) continue;
+
+        // Only include the primary resource for each construct path
+        // e.g. for path "MyTable", match "StackName/MyTable/Resource" exactly
+        const cdkPath = resource.Metadata?.[PATH_METADATA_KEY] ?? '';
+        const primaryPaths = constructPaths.map(p => `${stack.stackName}/${p}/Resource`);
+        if (!primaryPaths.includes(cdkPath)) continue;
```
rix0rrr: This smells to me. I can't be bothered to try to understand what this is doing exactly, but it doesn't feel right.

Why the interpolation with `stackName`? Why only exactly this L1 reosurce? Why `includes(cdkPath)` instead of `startsWith(cdkPath)`? 


What are we trying to do here?

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,346 @@
...
+            identifier[prop] = physicalIds.get(id)!;
+          }
+        }
+
+        if (Object.keys(identifier).length > 0) {
+          result[id] = identifier;
+        }
+      }
+    } catch {
+      await this.ioHelper.defaults.debug('Could not retrieve resource identifier summaries');
```
rix0rrr: Wait what?

There was an exception, we're not stopping, we're not even logging the except, we're just chugging along as if nothing happened? And *whispering* that "something might be off" at `debug` level?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1555,3 +1630,38 @@ async function synthAndMeasure(
...
+
+  for (const raw of paths) {
+    const p = raw.replace(/^\//, ''); // strip leading slash
+    const slashIdx = p.indexOf('/');
+    if (slashIdx < 0) {
+      throw new ToolkitError('InvalidConstructPath', `Construct path '${raw}' must be in the format StackName/ConstructPath (e.g. MyStack/MyTable)`);
+    }
+
+    const thisStack = p.substring(0, slashIdx);
+    const constructPath = p.substring(slashIdx + 1);
```
rix0rrr, line 1673: Unfortunately a construct ID !== a stack name.

"Construct ID" is a CDK concept, "Stack Name" is a CloudFormation concept.

Yes, **by default** the stack name of a Stack is the same as its Construct ID, but not necessarily.

```ts
new Stack(this, 'A', {    // <-- construct ID, in the path
  stackName: 'B'          // <-- Stack Name in CFN
});
```

The translation of construct path to stack name needs to go through the assembly. To be frank, I don't know the best way offhand (it should probably be somewhere in the manifest), but perhaps the magic black box can come up with an idea.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -934,6 +934,20 @@ export class CdkToolkit {
...
+    try {
+      await this.toolkit.orphan(this.props.cloudExecutable, {
+        constructPaths: options.constructPath,
+        roleArn: options.roleArn,
+        toolkitStackName: options.toolkitStackName,
+        force: options.force,
+      });
+    } catch (e) {
+      await this.ioHost.asIoHelper().defaults.error((e as Error).message);
+      throw e;
```
rix0rrr: Logging and rethrowing is usually an antipattern, leading to duplicated error messages in the output. Is this what we do in other places?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -934,6 +934,20 @@ export class CdkToolkit {
...
+    }
```
mrgrain:
```suggestion
    await this.toolkit.orphan(this.props.cloudExecutable, {
      constructPaths: options.constructPath,
      roleArn: options.roleArn,
      toolkitStackName: options.toolkitStackName,
      force: options.force,
    });
```

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
LeeroyHannigan: Wasn't sure how to easily import the SDK, but its done now, thanks.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: I was following the refactor implementation here: https://github.com/aws/aws-cdk-cli/blob/b96bccef9805f3c45c4ea97a3ac9506569f8b668/packages/aws-cdk/lib/cli/cdk-toolkit.ts#L1356

But I'll remove the try catch

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
LeeroyHannigan: Added `expect(templateBodyBefore.Resources).toHaveProperty('MyTable794EDED1')` before orphan

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
LeeroyHannigan: Single `toMatchObject` with `expect.objectContaining` and `expect.stringContaining`

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/cdk-orphan-detaches-resource.integtest.ts
LeeroyHannigan: Test body in `try`, `DeleteTableCommand` in `finally`

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: Throws `ToolkitError('OrphanMissingPhysicalId')` if physical ID is missing. Values map built entirely inside `resolveGetAttValues`

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/private/helpers.ts
LeeroyHannigan: Renamed to `assertDeploySucceeded`, only checks for `did-deploy-stack`. Removed the `replacement-requires-rollback` check

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: All decoupling (RETAIN + ref replacement + DependsOn removal) done in step 2

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: Done

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: renamed to: `resolveTemplate`, `decoupledTemplate`, `removalTemplate`

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: done

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: RETAIN moved to step 2. Step 1 only injects temp outputs.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: All three `Step X/3` messages printed from `execute()`

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: Created once in `execute()`, passed to `resolveGetAttValues`

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: Removed

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: Converted to map lookup

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: Removed the path filter entirely. Will need to discuss.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
LeeroyHannigan: Now `warn` with the actual error message instead of `debug` with a generic string. Added unit test.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
LeeroyHannigan: Done, now matches `stack.hierarchicalId` instead of using the path segment as a CloudFormation stack name pattern

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: Done

packages/aws-cdk/lib/cli/cdk-toolkit.ts
LeeroyHannigan: Done

packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml
```diff
@@ -688,6 +688,8 @@ Resources:
                   - sts:GetCallerIdentity
                   # `cdk import`
                   - cloudformation:GetTemplateSummary
+                  # `cdk orphan`
+                  - cloudformation:ListStackResources
```
rix0rrr: This needs an update to the bootstrap stack version as well (in 2 places)

LeeroyHannigan: Done, thanks

rix0rrr: I changed my mind. Can we change this to:

```
                  - cloudformation:List*
                  - cloudformation:Describe*
```

I'm tired of having to add every new read-only permissions one by one :(

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,310 @@
...
+    const physicalIds = new Map<string, string>();
+    for (const res of stackResources) {
+      if (res.LogicalResourceId && res.PhysicalResourceId) {
+        physicalIds.set(res.LogicalResourceId, res.PhysicalResourceId);
+      }
+    }
+
+    // Step 1/3: Resolve GetAtt attribute values via temporary stack outputs
+    await this.ioHelper.defaults.info('Step 1/3: Resolving attribute values...');
+    const resolvedValues = await this.resolveGetAttValues(stack, cfn, logicalIds, currentTemplate, physicalIds);
```
rix0rrr, line 131: I just realized there's something missing. 

A `{Fn::Sub}` has implicit interpolation inside its format string. It can look like this:

```ts
{ 
  "Fn::Sub": "before ${ImplicitRef} ${ImplicitAtt.AttrName} after"
}

{ 
  "Fn::Sub": [
    "before ${ImplicitRef} ${ImplicitAtt.AttrName} or ${Explicit} after",
    {
      "Explicit": { "Fn::GetAtt": ["Resource", "Attr"] }
    }
}
```

For the final case we don't have to do anything (`Resource.Attr`) because (I assume) it would already be found. But the 2 strings can contain implicit references:

* `${ImplicitRef}` inside format string - resolves the same as `{ "Ref": "ImplicitRef" }` would.
* `${ImplicitAtt.AttrName}` inside format string - resolves the same as `{ "Fn::GetAtt": ["ImplicitAtt", "AttrName"] }` would.

We need to discover these references in the discovery stage, and replace them in the replacement stage (be sure to add tests).

Note: the format string can occur in 2 forms, either inside an array or as a direct argument.

Be sure to add tests on these.

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
   IAM permissions to the `deploy-role`.
 
 
+### `cdk orphan`
+
+⚠️**CAUTION**⚠️: CDK Orphan is currently experimental and may have 
+breaking changes in the future. Make sure to use the `--unstable=orphan` flag 
+when using this command.
```
rix0rrr: Let's make this sound less scary.

```suggestion
> `cdk orphan` is currently experimental, meaning we reserve the right to change 
> options and flag names in the future. Pass the `--unstable=orphan` flag when using
> this command and be aware of this when using it in scripts.
```

PR #1327, opened 2026-04-10
fix: unnecessary 1-second delay on non-EC2 workstations

### LINKED ISSUE #627: CLI: 1 second delay when not running on an EC2 instance
### Describe the bug

Even when not running on an EC2 instance, the new CLI default credential provider will try to hit IMDS with a timeout of 1 second when it needs to figure out the region:

```
[11:52:10] Unable to retrieve AWS region from IMDS: Error: Error fetching metadata token: TimeoutError: Socket timed out without establishing a connection within 1000 ms
```

### Regression Issue

_No response_

### Expected Behavior

Don't try the EC2 credentials provider if we're not on an EC2 instance

### Current Behavior

It does do it.

### Reproduction Steps

Run the CDK CLI with `-vvv`

### Possible Solution

_No response_

_No response_

### CDK CLI Version

-

### Framework Version

_No response_

### Node.js Version

-

### OS

-

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

rix0rrr:
Because we query IMDS for credentials on every startup, which has a 1-second timeout to fail to connect to `169.254.169.254`, we incur a 1 second delay on regular workstations.

Do a check to see whether we are on an EC2 instance, and if not disable the IMDS lookup.

Fixes #627.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
@@ -116,12 +116,13 @@ export class AwsCliCompatible {
      */
 
     // Skip the IMDS credential provider if we're not on EC2, to avoid a
-    // 1-2 second timeout on non-EC2 machines. The SDK's remoteProvider
-    // already respects AWS_EC2_METADATA_DISABLED.
-    if (!isEc2Instance() && process.env.AWS_EC2_METADATA_DISABLED === undefined) {
+    // 1-2 second timeout on non-EC2 machines.
+    if (process.env.AWS_EC2_METADATA_DISABLED === undefined && !isEc2Instance()) {
       process.env.AWS_EC2_METADATA_DISABLED = 'true';
     }
 
+    // This chain respects the AWS_EC2_METADATA_DISABLED envvar, which can otherwise
+    // not be easily disabled using a property.
     const nodeProviderChain = fromNodeProviderChain({
       profile: envProfile,
       clientConfig,
--- a/packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
@@ -5,6 +5,7 @@ import { createCredentialChain, fromEnv, fromIni, fromNodeProviderChain } from '
 import { MetadataService } from '@aws-sdk/ec2-metadata-service';
 import { loadSharedConfigFiles } from '@smithy/shared-ini-file-loader';
 import type { RequestHandlerSettings } from './base-credentials';
+import { isEc2Instance } from './ec2-detection';
 import { makeCachingProvider } from './provider-caching';
 import type { ISdkLogger } from './sdk-logger';
 import { AuthenticationError } from '../../toolkit/toolkit-error';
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/ec2-detection.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+  cachedIsEc2 = detectEc2();
+  return cachedIsEc2;
+}
+
+function detectEc2(): boolean {
+  // If the user has explicitly told us about IMDS, respect that
+  if (process.env.AWS_EC2_METADATA_DISABLED !== undefined) {
+    return process.env.AWS_EC2_METADATA_DISABLED === 'false';
+  }
```
mrgrain: should this check be here? feels like `detectEc2` should always do the check, regardless of env var.

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
```diff
@@ -113,6 +114,14 @@ export class AwsCliCompatible {
      *
      * The NodeProviderChain is already cached.
      */
+
+    // Skip the IMDS credential provider if we're not on EC2, to avoid a
+    // 1-2 second timeout on non-EC2 machines. The SDK's remoteProvider
+    // already respects AWS_EC2_METADATA_DISABLED.
+    if (!isEc2Instance() && process.env.AWS_EC2_METADATA_DISABLED === undefined) {
+      process.env.AWS_EC2_METADATA_DISABLED = 'true';
+    }
```
mrgrain, line 122: Seems odd that we are setting the env var.

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
```diff
@@ -172,6 +181,10 @@ export class AwsCliCompatible {
    * @returns The region for the instance identity
    */
   private async regionFromMetadataService() {
+    if (!isEc2Instance()) {
+      return undefined;
+    }
```
mrgrain, line 187: How about this? 

```suggestion
    if (!shouldTryInstanceMetaData()) {
      return undefined;
    }
```

And `shouldTryInstanceMetaData` can do the env var check.

mrgrain (changes requested):
minor stuff

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
rix0rrr: This is the easiest way of disabling the IMDS lookup without messing with the internals of the default credential provider chain too much. It's either this, or fully recreate the default chain from scratch, and it's pretty intricate.

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/ec2-detection.ts
rix0rrr: Yeah agreed.

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/awscli-compatible.ts
rix0rrr: I like the `isEC2Instance` framing. I've moved the code around a bit.

mrgrain (approved):

PR #1342, opened 2026-04-10
chore: make builds faster by using NX build cache in workflows

mrgrain:
The NX cache was previously always skipped in CI (`NX_SKIP_NX_CACHE: true`), meaning every workflow run rebuilt everything from scratch. This adds `actions/cache@v5` to persist the `.nx/cache` directory across runs for the build, integ, and codecov workflows.

To ensure correctness, the cache is never used on the `main` branch or during merge queue builds. On those runs, `NX_SKIP_NX_CACHE` evaluates to `true` and the cache action is skipped entirely. On pull request builds, the cache is restored from a previous run and saved with the current commit SHA as key.

The shared cache configuration lives in `projenrc/nx-cache.ts` and is used by the codecov and integ workflow projen components directly. For the build workflow, a `JsonPatch` is necessary because `NX_SKIP_NX_CACHE` is hardcoded by `cdklabs-projen-project-types`.

### Checklist

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,9 @@ name: build
 on:
   pull_request: {}
   workflow_dispatch: {}
+  push:
+    branches:
+      - main
   merge_group: {}
 jobs:
   build:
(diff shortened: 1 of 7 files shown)


mrgrain:
NX really doesn't want us to use non paid cache.

PR #1368, opened 2026-04-13
fix(cli): add projen as devDependency to aws-cdk package

dgandhi62:
### Issue

The `upgrade-aws-cdk-lib_aws-cdk` workflow fails with `command not found: projen` (exit code 127).

### Root Cause

The workflow runs `yarn projen upgrade-aws-cdk-lib` with `working-directory: ./packages/aws-cdk`. Workspace packages can only resolve binaries from their own declared dependencies. The `configureProject` helper in `.projenrc.ts` removes `projen` from all workspace packages, assuming it gets hoisted from the monorepo root — but Yarn 4 doesn't hoist binaries that way.

### Fix

Stop removing `projen` from workspace packages in `configureProject`, and add it to the `consistentVersions` list to ensure all workspaces use the same version.

Source of truth - refer to section 1 of https://yarnpkg.com/advanced/rulebook

### Checklist


mrgrain (changes requested):
Nice find! Can we just get rid of the removal in the helper then? Feels cleaner.

.projenrc.ts
```diff
@@ -103,9 +103,11 @@ function configureProject<A extends pj.typescript.TypeScriptProject>(x: A): A {
...
   );
 
+  // Yarn 4 requires each workspace to declare its own dependencies for binary
+  // resolution. Ensure projen is available so `yarn projen <task>` works from
+  // any workspace directory (e.g. in CI workflows).
   if (x instanceof TypeScriptWorkspace) {
-    // Individual workspace packages shouldn't depend on "projen", it gets brought in at the monorepo root
-    x.deps.removeDependency('projen');
+    x.addDevDeps('projen');
   }
```
mrgrain: Pretty sure you can just remove this
```suggestion
```

mrgrain (changes requested):
can you also add projen to the `consistenVersions` list?

mrgrain (approved):

PR #1372, opened 2026-04-14
feat(cli): show diff and prompt for deploy on non-addition changes

LeeroyHannigan:
Currently, `cdk import` throws an error when it detects non-addition changes (updates/deletes) to existing resources, requiring `--force` to override. This is a poor experience for the `orphan → import` migration workflow, where hardcoded string literals in the deployed template naturally differ from the `Fn::GetAtt`/`Ref` intrinsics in the synth'd template.

Changes:
- `cdk import` no longer throws on non-addition changes. Instead it shows the projected diff and prompts the user to confirm.
- After a successful import with non-addition changes, prompts "*Finish with a cdk deploy now?*" to reconcile remaining drift.
- `--force` still works, skips the prompt and prints a reminder to deploy.

User-facing flow (e.g. `Table` v1 → `TableV2` migration):
1. `cdk orphan --path MyStack/MyTable` : detaches table, hardcodes refs
2. Edit code: `Table` → `TableV2`
3. `cdk import` : shows diff of hardcoded → intrinsic ref changes, asks to confirm, imports, then offers to deploy

No `--force` required. No manual `cdk deploy` step to remember.

Backwards compatible: If there are no non-addition changes, behavior is identical to before.

### Checklist
- [x] This change contains a major version upgrade for a dependency and I confirm all breaking changes are addressed

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts
@@ -6,7 +6,6 @@ import type { ResourceIdentifierSummary, ResourceToImport } from '@aws-sdk/clien
 import * as chalk from 'chalk';
 import * as fs from 'fs-extra';
 import type { DeploymentMethod } from '../../actions/deploy';
-import { ToolkitError } from '../../toolkit/toolkit-error';
 import type { Deployments } from '../deployments';
 import { assertIsSuccessfulDeployStackResult } from '../deployments';
 import { IO, type IoHelper } from '../io/private';
@@ -238,15 +237,9 @@ export class ResourceImporter {
     const nonAdditions = resourceChanges.filter(([_, dif]) => !dif.isAddition);
     const additions = resourceChanges.filter(([_, dif]) => dif.isAddition);
 
-    if (nonAdditions.length) {
+    if (nonAdditions.length && allowNonAdditions) {
       const offendingResources = nonAdditions.map(([logId, _]) => this.describeResource(logId));
-
-      if (allowNonAdditions) {
-        await this.ioHelper.defaults.warn(`Ignoring updated/deleted resources (--force): ${offendingResources.join(', ')}`);
-      } else {
-        throw new ToolkitError('ImportNonAdditionChanges', 'No resource updates or deletes are allowed on import operation. Make sure to resolve pending changes ' +
-          `to existing resources, before attempting an import. Updated/deleted resources: ${offendingResources.join(', ')} (--force to override)`);
-      }
+      await this.ioHelper.defaults.warn(`Ignoring updated/deleted resources (--force): ${offendingResources.join(', ')}`);
     }
 
     // Resources in the new template, that are not present in the current template, are a potential import candidates
@@ -257,6 +250,8 @@ export class ResourceImporter {
         resourceDefinition: addDefaultDeletionPolicy(this.stack.template?.Resources?.[logicalId] ?? {}),
       })),
       hasNonAdditions: nonAdditions.length > 0,
+      nonAdditionNames: nonAdditions.map(([logId, _]) => this.describeResource(logId)),
+      diff,
     };
   }
 
@@ -505,6 +500,8 @@ function addDefaultDeletionPolicy(resource: any): any {
 export interface DiscoverImportableResourcesResult {
   readonly additions: ImportableResource[];
   readonly hasNonAdditions: boolean;
+  readonly nonAdditionNames: string[];
+  readonly diff: cfnDiff.TemplateDiff;
 }
 
 export function removeNonImportResources(stack:cxapi.CloudFormationStackArtifact) {
--- a/packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts
@@ -6,6 +6,7 @@ import type { ResourceIdentifierSummary, ResourceToImport } from '@aws-sdk/clien
 import * as chalk from 'chalk';
 import * as fs from 'fs-extra';
 import type { DeploymentMethod } from '../../actions/deploy';
+import { DiffFormatter } from '../diff';
 import type { Deployments } from '../deployments';
 import { assertIsSuccessfulDeployStackResult } from '../deployments';
 import { IO, type IoHelper } from '../io/private';
@@ -251,7 +252,12 @@ export class ResourceImporter {
       })),
       hasNonAdditions: nonAdditions.length > 0,
       nonAdditionNames: nonAdditions.map(([logId, _]) => this.describeResource(logId)),
-      diff,
+      diffFormatter: new DiffFormatter({
+        templateInfo: {
+          oldTemplate: currentTemplate,
+          newTemplate: this.stack,
+        },
+      }),
     };
   }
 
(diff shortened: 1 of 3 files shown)


packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1003,7 +1004,30 @@ export class CdkToolkit {
...
+    const { additions, hasNonAdditions, diff } = await resourceImporter.discoverImportableResources(options.force);
+
+    // If there are non-addition changes (e.g. after orphan, hardcoded refs differ from Fn::GetAtt),
+    // warn the user and ask for confirmation unless --force was given.
+    if (hasNonAdditions && !options.force) {
+      const ioHelper = this.ioHost.asIoHelper();
+      await ioHelper.defaults.info(
+        `The following resources have pending updates that will be reconciled with a ${chalk.blueBright('cdk deploy')} after import:`,
+      );
+      const stream = new (await import('stream')).PassThrough();
```
rix0rrr: Can just be `import { PassThrough } from 'stream'` ?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1003,7 +1004,30 @@ export class CdkToolkit {
...
+      let diffOutput = '';
+      stream.on('data', (chunk: Buffer) => {
+        diffOutput += chunk.toString();
+      });
+      formatDifferences(stream, diff);
+      stream.end();
```
rix0rrr: Doesn't this need to be `await` ed?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1003,7 +1004,30 @@ export class CdkToolkit {
...
+      await ioHelper.defaults.info(diffOutput);
+      const confirmed = await ioHelper.requestResponse(IO.CDK_TOOLKIT_I7010.req('Perform import?', { motivation: 'Import will proceed with drift remaining' }));
```
rix0rrr:
```suggestion
      const confirmed = await ioHelper.requestResponse(IO.CDK_TOOLKIT_I7010.req('Perform import?', { motivation: 'Confirm import with pending drift' }));
```

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1046,22 +1070,41 @@ export class CdkToolkit {
...
     if (actualImport.importResources.length < additions.length) {
-      await this.ioHost.asIoHelper().defaults.info('');
       await this.ioHost.asIoHelper().defaults.warn(
         `Some resources were skipped. Run another ${chalk.blueBright('cdk import')} or a ${chalk.blueBright('cdk deploy')} to bring the stack up-to-date with your CDK app definition.`,
       );
     } else if (hasNonAdditions) {
-      await this.ioHost.asIoHelper().defaults.info('');
-      await this.ioHost.asIoHelper().defaults.warn(
-        `Your app has pending updates or deletes excluded from this import operation. Run a ${chalk.blueBright('cdk deploy')} to bring the stack up-to-date with your CDK app definition.`,
+      // After orphan→import, the deployed template still has hardcoded values that differ from
```
rix0rrr, line 1071: I'm not sure about all these instructions. It could probably be simpler, and `cdk deploy` doesn't actually "reconcile drift" so we should probably steer away from that terminology. But we haven't done super deep analysis on this yet.

Here's what you need to know about drift and the different places your infra state is kept:

```
                                                                                             
                   ┌──────────┐     diff    ┌─────────────────┐   drift   ┌─────────────────┐
                   │          │             │                 │           │                 │
    REALITY        │ CDK app  │◀───────────▶│ CloudFormation  │◀─────────▶│  Real resource  │
                   │          │             │                 │           │                 │
                   └──────────┘     cdk     └─────────────────┘   drift   └─────────────────┘
                                   deploy                       detection                    
                                                                                             
                                                                                             
                                                                                             
                                                                                             
                                                                                             
                                                                                             
                   ┌──────────┐             ┌─────────────────┐           ┌─────────────────┐
   WHAT USERS      │          │             │                 │           │                 │
    PROBABLY       │ CDK app  │             │ CloudFormation  │           │  Real resource  │
     EXPECT        │          │             │                 │           │                 │
                   └──────────┘             └─────────────────┘           └─────────────────┘
                         ▲                                                         ▲         
                         │                                                         │         
                         └─────────────────────────────────────────────────────────┘         
                                                                                             
                                                 "drift"                                     
                                                                                             
```

`cdk deploy` will update CloudFormation's view of the world, but if CloudFormation's view of the world was incorrect (due to drift), `cdk deploy` will not fix that drift.

Not quite sure what the best terminology here is, but I'm open to suggestions

rix0rrr: EDIT: although there is also `cdk deploy --revert-drift`, which is maybe the thing we should recommend?

And there is now also a `cdk drift` command we can recommend, instead of sending them to a CloudFormation website.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
mrgrain: We shouldn't do this manually, instead use the `DiffFormatter` class (which is probably misnamed and should be calls `DiffManager` or something).

packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts
```diff
@@ -257,6 +250,8 @@ export class ResourceImporter {
         resourceDefinition: addDefaultDeletionPolicy(this.stack.template?.Resources?.[logicalId] ?? {}),
       })),
       hasNonAdditions: nonAdditions.length > 0,
+      nonAdditionNames: nonAdditions.map(([logId, _]) => this.describeResource(logId)),
+      diff,
```
mrgrain: You can return a `DiffFormatter` instance here, to be used in `cdk-toolkit.ts`. A bit further up, where this method is currently calling `cfnDiff.fullDiff`, you can create the `DiffFormatter` which will also allow you to get access the raw `TemplateDiff` (which is what `cfnDiff.fullDiff` returns).

We might need some minor refactoring on the `DiffFormatter` to make it more suitable for this use case.

mrgrain (dismissed):

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1003,7 +1003,21 @@ export class CdkToolkit {
...
-    const { additions, hasNonAdditions } = await resourceImporter.discoverImportableResources(options.force);
+    const { additions, hasNonAdditions, diffFormatter } = await resourceImporter.discoverImportableResources(options.force);
+
+    // If there are non-addition changes (e.g. after orphan, hardcoded refs differ from Fn::GetAtt),
+    // warn the user and ask for confirmation unless --force was given.
+    if (hasNonAdditions && !options.force) {
+      const ioHelper = this.ioHost.asIoHelper();
+      const { formattedDiff } = diffFormatter.formatStackDiff();
+      await ioHelper.defaults.info(formattedDiff);
+      const confirmed = await ioHelper.requestResponse(IO.CDK_TOOLKIT_I7010.req('Perform import?', { motivation: 'Confirm import with pending drift' }));
```
rix0rrr, line 1017: This used to have a longer message that included something like "There are more changes in your CDK app than a single `cdk import` will resolve. After import, an additional `cdk deploy` will be required" (or something to that effect).

I think I'd like that better.

rix0rrr: Print that message before the diff I think.

rix0rrr (changes requested):
Almost there, I'd like slightly more elaborate wording and then it's good to go.

rix0rrr (approved):

PR #1373, opened 2026-04-14
chore: counts of non hotswappable resources

ShadowCat567:
Adds counts for resource types that are not hotswappable that were attempted to be hotswapped. 
We want to track this as a part of our initiative to improve the speed of hotswap deployments since having non hotswappable resources in a hotswap deployment with fallback slows down the hotswap deployment.
Payload would look like this:

```
{
  "event": {
    "state": "FAILED",
    "eventType": "HOTSWAP",
    "duration": 456,
    "error": {
      "name": "UnknownError"
    },
    "counters": {
      "hotswapped": 0,
      "hotswappableChanges": 1,
      "nonHotswappableChanges": 3,
      "hotswapFallback:AWS::Some::Resource:Id": 2,
      "hotswapFallback:AWS::Other::Resource:Name": 1,
    }
  }
}
``` 

---


rix0rrr (changes requested):
```
      "AWS::Some::Resource": 2,
      "AWS::Other::Resource": 1,
```

This list of counters goes into a huge grab bag. So just a resource type by itself is not going to be meaningful. Let's name it something like

```
hotswapfallback:AWS::Other::Resource
```

But also don't do that literally! I'm pretty sure we already emit counters with resource identifiers in there somewhere. I forget where, but I'm sure you can find it. Probably something about error tracking. Don't invent a new format, just follow what the other location is doing.

And also: I'm pretty sure the top-level property that is causing the fallback should be in there as well.

ShadowCat567:
> ```
>       "AWS::Some::Resource": 2,
>       "AWS::Other::Resource": 1,
> ```
> 
> This list of counters goes into a huge grab bag. So just a resource type by itself is not going to be meaningful. Let's name it something like
> 
> ```
> hotswapfallback:AWS::Other::Resource
> ```
> 
> But also don't do that literally! I'm pretty sure we already emit counters with resource identifiers in there somewhere. I forget where, but I'm sure you can find it. Probably something about error tracking. Don't invent a new format, just follow what the other location is doing.
> 
> And also: I'm pretty sure the top-level property that is causing the fallback should be in there as well.

It is error tracking, that does `errorAnn:[errorCode]` - so we would want to also use a `:` delimiter for hotswap resources

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -645,6 +645,18 @@ function eventFromMessage(msg: IoMessage<unknown>): TelemetryEvent | undefined {
 }
 
 function hotswapToEventResult(result: HotswapResult): TelemetryEvent {
+  const nonHotswappableResources: Record<string, number> = {};
+  for (const { subject } of result.nonHotswappableChanges) {
+    if ('resourceType' in subject) {
+      const keys = 'rejectedProperties' in subject && subject.rejectedProperties
+        ? subject.rejectedProperties.map(p => `hotswapFallback:${subject.resourceType}:${p}`)
```
rix0rrr: Let's use a different separator for the 2 pieces to make it easier to split.

```suggestion
        ? subject.rejectedProperties.map(p => `hotswapFallback:${subject.resourceType}#${p}`)
```

packages/aws-cdk/test/cli/io-host/cli-io-host.test.ts
```diff
@@ -502,6 +502,42 @@ describe('CliIoHost', () => {
...
+      expect(telemetryEmitSpy).toHaveBeenCalledWith(expect.objectContaining({
+        eventType: 'HOTSWAP',
+        duration: 456,
+        counters: {
+          'hotswapped': 1,
+          'hotswappableChanges': 2,
+          'nonHotswappableChanges': 3,
+          'hotswapFallback:someResource': 1,
+          'hotswapFallback:someOtherResource:Name': 1,
+          'hotswapFallback:someOtherResource:Id': 1,
```
rix0rrr:
```suggestion
          'hotswapFallback:someOtherResource#Name': 1,
          'hotswapFallback:someOtherResource#Id': 1,
```

rix0rrr (changes requested):

rix0rrr (approved):

PR #1374, opened 2026-04-14
chore: upgrade-aws-cdk-lib workflow fails with `command not found: projen`

dgandhi62:
The `upgrade-aws-cdk-lib_aws-cdk` workflow fails with `command not found: projen` (exit code 127).

This pr is an extension of #1368 , which added `projen` as a devDependency to all workspace packages. This does not resolve the issue because `nodeLinker: node-modules` still hoists the binary to the root.

### Fix
Patch the generated workflow to use `yarn run upgrade-aws-cdk-lib`, which resolves the script through Yarn's script execution context where PATH is correctly configured.

### Checklist
  - Release notes for the new version: N/A

### DIFF
--- a/.github/workflows/upgrade-aws-cdk-lib_aws-cdk.yml
+++ b/.github/workflows/upgrade-aws-cdk-lib_aws-cdk.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
       - name: Upgrade dependencies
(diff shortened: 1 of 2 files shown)


mrgrain (changes requested):
This doesn't fix the issue and the issue doesn't seem to match the error.

https://github.com/aws/aws-cdk-cli/actions/runs/24442226843/job/71409741387

PR #1378, opened 2026-04-16
feat: `cdk diagnose` displays the root cause of a past stack deployment failure

rix0rrr:
Useful when you use a CI/CD system to deploy stacks, but the failure reason is hard to parse.

This PR unifies and makes the error analysis and printing paths of `cdk deploy` accessible as a separate subcommand, `cdk diagnose`. This allows you to get the same interpretation of CloudFormation error reporting strategies, as well as source localization for a CI/CD deployment as is possible today for a deployment performed directly via the CLI.

In the future, we will further extend the diagnosis that the CLI performs to get you more information.

This is what errors look like now (both for `cdk deploy` and `cdk diagnose`):

```
    ❌ Stack cdktest-0648snacfpxl-diagnose-deploy-fail:
    Early validation failed for change set cdk-deploy-change-set:
     └─ cdktest-0648snacfpxl-diagnose-deploy-fail
         └─ BadPolicy  (AWS::IAM::Policy BadPolicy)
            🛑 Required property [PolicyName] not found (at /Resources/BadPolicy/Properties)
            Source Location: new DeployFailStack (/private/var/folders/w3/3m2f73xn1rq69wbbx9zw70s00000gq/T/cdk-integ-0648snacfpxl/app.js:29:5)
                             Object.<anonymous> (/private/var/folders/w3/3m2f73xn1rq69wbbx9zw70s00000gq/T/cdk-integ-0648snacfpxl/app.js:69:1)
                             Module._compile (node:internal/modules/cjs/loader:1761:14)
                             Module._extensions..js (node:internal/modules/cjs/loader:1893:10)
                             Module.load (node:internal/modules/cjs/loader:1481:32)
                             Module._load (node:internal/modules/cjs/loader:1300:12)
                             TracingChannel.traceSync (node:diagnostics_channel:328:14)
                             wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
                             Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
                             node:internal/main/run_main_module:33:47
```

Broad design of this code:

image

---


packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/deploy/cdk-import-existing-resources-error-message-includes-construct-paths.integtest.ts
```diff
@@ -25,7 +25,7 @@ integTest(
         allowErrExit: true,
       });
 
-      expect(stdErr).toContain('Import of existing resources failed');
+      expect(stdErr).toContain('needs a DeletionPolicy');
```
mrgrain, line 28: Why not keep both?

packages/@aws-cdk/toolkit-lib/lib/actions/diagnose/index.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export interface TracedResourceError {
+  /**
+   * The stack this resource error occurred in
+   *
+   * NOTE: This will be a stack ID (which is a full ARN including the unique identifier),
+   * not just a name.
+   */
+  readonly stackId: string;
```
mrgrain: Maybe `stackArn` then? Same for others. Not sure this is better than `id`.

```suggestion
   * NOTE: This will be a stack ID (which is a full ARN including the unique identifier),
   * not just a name.
   */
  readonly stackArn: string;
```

packages/@aws-cdk/toolkit-lib/lib/actions/diagnose/index.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Error code of the resource
+   */
+  readonly errorCode?: string;
+
+  /**
+   * Optionally a source trace
+   *
+   * (Not optional on purpose so we are not allowed to forget to call the code that should fill it)
```
mrgrain, line 86: Nice! Although this feels like a linter rules/TS config trying to come out 😱

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -589,6 +595,55 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+  /**
+   * Try to find the root causes for deployment failures of the given stacks.
+   *
+   * Both emits the diagnosis results over the IO host as they are coming in, as well
+   * as returns all of them as the result of the function.
+   *
+   * NOTE: The Cloud Assembly Source **should** be configured with `debug: true` to add
+   * the maximum number of diagnostics.
+   */
+  public async diagnose(cx: ICloudAssemblySource, options: DiagnoseOptions = {}): Promise<DiagnoseResult> {
```
mrgrain, line 610: Should this be unstable for now?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -589,6 +595,55 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+    const ioHelper = asIoHelper(this.ioHost, 'diagnose');
+    const selectStacks = stacksOpt(options);
+    await using assembly = await synthAndMeasure(ioHelper, cx, selectStacks);
+    const stackCollection = await assembly.selectStacksV2(selectStacks);
+    const envs = new EnvironmentAccess(await this.sdkProvider('diagnose'), options.toolkitStackName ?? DEFAULT_TOOLKIT_STACK_NAME, ioHelper);
+
+    // Do all stacks in parallel, for speed.
+    const limit = pLimit(options.concurrency ?? 10);
```
mrgrain: All or just 10, eh? 😉

packages/@aws-cdk/toolkit-lib/lib/util/type-brands.ts
```diff
@@ -43,4 +43,4 @@ export function createBranded<A extends Branded<any, any>>(value: TypeUnderlying
 }
 /* c8 ignore stop */
 
-type TypeUnderlyingBrand<A> = A extends Branded<infer T, any> ? T : never;
+type TypeUnderlyingBrand<A> = Omit<A, keyof Brand<any>>;
```
mrgrain, line 46: Seems risky to go from `never` to `any`... Can you help explain this?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -430,6 +431,24 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           ignoreNoStacks: args.ignoreNoStacks,
         });
 
+      case 'diagnose':
+        ioHost.currentAction = 'diagnose';
+
+        // Implicitly switch 'debug' mode to true, that is going to be most useful.
+        configuration.settings.temporarilyMutable((settings) => {
+          settings.set(['debug'], true);
+        });
```
mrgrain, line 442: nit: This seems nifty, but surely this isn't the first time we are forcing settings based on a command?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -84,6 +84,7 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
   const ioHelper = asIoHelper(ioHost, ioHost.currentAction as any);
 
   // Debug should always imply tracing
+  // (That's not what --debug means, --debug is intended for the CDK app -- huijbers@)
```
mrgrain: issue: Or maybe it now does mean that because it has been like this for a long time? What's the practical consequence of this comment?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -430,6 +431,24 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
+        configuration.settings.temporarilyMutable((settings) => {
+          settings.set(['debug'], true);
+        });
+
+        return cli.diagnose({
+          // Implicitly do all stacks if no pattern given
+          stacks: {
+            strategy: args.STACKS.length > 0 ? StackSelectionStrategy.PATTERN_MATCH : StackSelectionStrategy.ALL_STACKS,
+            patterns: args.STACKS,
+          },
```
mrgrain: chore: I am guessing we do this in multiple locations now and we can have a helper for it

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -430,6 +431,24 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           ignoreNoStacks: args.ignoreNoStacks,
         });
 
+      case 'diagnose':
```
mrgrain, line 435: q: should this be unstable?

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -806,6 +808,18 @@ export class CdkToolkit {
...
+  /**
+   * Diagnose errors
+   */
+  public async diagnose(options: DiagnoseOptions): Promise<number> {
+    const results = await this.toolkit.diagnose(this.props.cloudExecutable, options);
+
+    if (results.stacks.some(s => s.result.type !== 'no-problem')) {
+      return 1;
+    }
+    return 0;
```
mrgrain, line 856: praise: Nice!

todo: can we add a test for this so we don't regress here in future.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -316,9 +317,10 @@ export class CdkToolkit {
           removeNonImportResources(stack);
         }
 
-        const changeSet = (options.method !== 'template')
-          ? await this.tryCreateDiffChangeSet(stack, options, parameterMap, resourcesToImport, quiet)
-          : undefined;
+        let changeSet;
+        if (options.method !== 'template') {
+          changeSet = await this.tryCreateDiffChangeSet(stack, options, parameterMap, resourcesToImport, quiet);
+        }
```
mrgrain: chore: seems like a taste fix that is not backed-up by a linter/formatter rule. I.e. we will be super inconsistent on this. E.g. I prefer having `const` over the readability that a proper if conditional gives.

packages/aws-cdk/lib/cxapp/exec.ts
```diff
@@ -30,6 +30,10 @@ export async function execProgram(aws: SdkProvider, ioHelper: IoHelper, config:
   };
   await debugFn(format('context:', context));
 
+  if (params.env.CDK_DEBUG === 'true') {
+    await ioHelper.defaults.info('🔍 Synthesizing with debug information. This may take a bit longer. (--debug)');
```
mrgrain: nit: This message will also (mostly?) be printed on `cdk diagnose` so not sure if mentioning `--debug` makes sense.

```suggestion
    await ioHelper.defaults.info('🔍 Synthesizing with debug information. This may take a bit longer.');
```

packages/@aws-cdk/toolkit-lib/lib/api/settings.ts
```diff
@@ -53,6 +60,16 @@ export class Settings {
     return new Settings(this.settings, true);
   }
 
+  public temporarilyMutable(block: (x: Settings) => void) {
```
mrgrain, line 63: thought (non-blocking): IDK, if we just want this class to be mutable now why not make it mutable? Or call `freeze()` (or similar) at some point when we want this to become an immutable property bag.

packages/@aws-cdk/toolkit-lib/lib/api/index.ts
```diff
@@ -23,3 +23,5 @@ export * from './work-graph';
 export * from './tree';
 export * from './tags';
 export * from './drift';
+export * from '../actions/diagnose';
```
mrgrain: These already are public exports. No need here.

```suggestion
```

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/stack-selector.ts
```diff
@@ -71,7 +71,7 @@ export enum ExpandStackSelection {
  */
 export interface StackSelector {
   /**
-   * The behavior if if no selectors are provided.
+   * How the patterns should be interpreted.
```
mrgrain, line 74: praise: Ha! Nice catch!

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/stack-helpers.ts
```diff
@@ -65,6 +65,13 @@ export class CloudFormationStack {
   ) {
   }
 
+  public get underlyingStackObject(): Stack {
```
mrgrain: chore: Not entirely sure what that means... Maybe add a jsdoc?

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/resource-errors.ts
```diff
@@ -0,0 +1,183 @@
...
+  const reason = event.ResourceStatusReason ?? '';
+
+  const errorRe = /(?:HandlerErrorCode:|Error Code:) ([a-zA-Z0-9:-]+)/;
+  const handlerCode = reason.match(errorRe);
+
+  return `${resourceType}:${handlerCode ? handlerCode[1] : DeploymentErrorCodes.UNKNOWN_ERROR}`;
+}
+
+// Some custom resource types that the CDK standard library creates that we
+// would like to see it if they fail.
```
mrgrain, line 152: thought (non-blocking): What a shame we don't have a consistent naming pattern for these 🤦🏻

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-event-poller.ts
```diff
@@ -14,21 +16,80 @@ export interface StackEventPollerProps {
...
+}
+
+export interface IOldestEvent {
+  /**
+   * Whether polling should stop on seeing this event
+   */
+  shouldStop(event: ResourceEvent): 'stop-include' | 'stop-exclude' | 'continue';
+}
+
+export abstract class OldestEvent {
```
mrgrain: issue: The naming of this doesn't click with me.

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-event-poller.ts
```diff
@@ -14,21 +16,80 @@ export interface StackEventPollerProps {
...
+   * Stop when events are older than a given time
    */
-  readonly startTime?: number;
+  public static timestamp(startTime: number): IOldestEvent {
+    return {
+      shouldStop(event) {
+        return event.event.Timestamp!.valueOf() < startTime ? 'stop-exclude' : 'continue';
+      },
+    };
+  }
```
mrgrain: `fromTimestamp` or `withStartingPoint` ?


`OldestEvent.timestamp(...)` just doesn't speak to me at all.

packages/aws-cdk/README.md
```diff
@@ -158,6 +159,24 @@ The `quiet` option can be set in the `cdk.json` file.
 See the [AWS Documentation](https://docs.aws.amazon.com/cdk/latest/guide/apps.html#apps_cloud_assembly) to learn more about cloud assemblies.
 See the [CDK reference documentation](https://docs.aws.amazon.com/cdk/api/latest/docs/cloud-assembly-schema-readme.html) for details on the cloud assembly specification
 
+### `cdk diagnose`
+
+Looks at a failed stack deployment in CloudFormation, and prints the root cause
+reason with pointers to the CDK source code that caused the problem.
+
+This peforms the same diagnosis that `cdk deploy` does, but does it
+after-the-fact.  This can be useful to refresh your memory, ask a colleague to
```
mrgrain:
```suggestion
after-the-fact. This can be useful to refresh your memory, ask a colleague to
```

packages/aws-cdk/README.md
```diff
@@ -158,6 +159,24 @@ The `quiet` option can be set in the `cdk.json` file.
...
+help diagnose a deployment problem, or try to diagnose a deployment problem if
+your way of working dictates that you perform stack deployment via a CI/CD
+system (instead of directly using the CLI).
```
mrgrain:
```suggestion
system instead of directly using the CDK CLI.
```

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/deploy/cdk-import-existing-resources-error-message-includes-construct-paths.integtest.ts
rix0rrr: *shrug*. The message changed slightly. This is enough to confirm that the message is in the output.

mrgrain (changes requested):
Mostly minor stuff.

Bigger points for for me are:

- Should the new command be `--unstable` for now?
- The naming and jsdocs around `OldestEvent` don't click with me at all. Would love to see an iteration on the API to make the intent more clear.

packages/@aws-cdk/toolkit-lib/lib/actions/diagnose/index.ts
rix0rrr: I used to have branded types here but I also didn't want to expose those publicly, so this is the middle ground.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
rix0rrr: We can unstablify it. I think the API should be uncontroversial, but perhaps I'm too optimistic. Anything in particular you're unsure of?

packages/@aws-cdk/toolkit-lib/lib/util/type-brands.ts
rix0rrr: Sure.

The goal is, given `Branded<A>`, to find `A`. For whatever reason, the `A extends Branded<infer T>` never worked. I forget what it resolved to, but it was either `any` or `unknown`.

Since `Branded<A>` does `A & Something`, we can also try to remove the intersection. Googling told me that the way to de-intersect a type is to do `Omit<A, keyof Something>`. In our case the `Something` is a `Brand<B>`, but we don't care about the `B` because the keys are always the same, so `B=any`.

TL;DR:  the `never` and `any` have nothing to do with each other, they just both happened to occur in 2 type manipulation signatures trying to do the same thing in a different way.

packages/aws-cdk/lib/cli/cli.ts
rix0rrr: I think normally we just fail if you specify an illegal combination. But that seems unnecessarily burdensome.

packages/@aws-cdk/toolkit-lib/lib/api/settings.ts
rix0rrr: I don't mind either. Just having it mutable seems simpler, this felt like a smaller (therefore safer) change. But yeah, let's just go mutable.

packages/aws-cdk/lib/cli/cli.ts
rix0rrr: Practically I suppose we can't change anything. Or perhaps we can, motivation being that this will hardly ever be used non-interactively, and humans can cope with change.

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/resource-errors.ts
rix0rrr: Yes.

packages/@aws-cdk/toolkit-lib/lib/api/settings.ts
rix0rrr: Okay, turns out that some property bags being immutable is load-bearing. It is used for deciding where to write context (automatically to the right settings overlay). So I'm going to back out of the larger change and just keep this as is.

mrgrain (approved):

PR #1384, opened 2026-04-17
fix(toolkit-lib): cdk gc hangs when BackgroundStackRefresh.stop() races an in-flight refresh

### LINKED ISSUE #1385: (cli): cdk gc hangs indefinitely after completing because BackgroundStackRefresh reschedules itself after stop()
### Describe the bug

`cdk gc` completes its work but the process never exits. The CLI hangs silently after the last progress-printer line until killed. Triggered on any run whose total duration exceeds ~5 minutes.

Root cause is in `packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts`. `BackgroundStackRefresh.refresh()` unconditionally reschedules itself via `setTimeout` on every completion, and `stop()` only calls `clearTimeout` on the currently-tracked timer. If `stop()` fires while an awaited `refresh()` is in flight, `clearTimeout` has no effect on the in-flight call, and when that call resolves it installs a fresh timer that `stop()` has already passed over. The new timer pins Node's event loop indefinitely.

Confirmed present on `aws/aws-cdk-cli` `main` at commit `68789664`, lines 161-207 of the above file; the offending `setTimeout` is line 174.

### Regression Issue

_No response_

### Expected Behavior

`cdk gc` exits cleanly after completing its work, regardless of total runtime.

### Current Behavior

The process hangs silently. No further output unless `--verbose` is set (in which case `refreshStacks` debug messages appear every 5 min, then silence again). The only way out is SIGKILL. In CI this causes the job to hit its step timeout.

Representative log tail:

```
[99.62%] 5494 files scanned: 100 assets (0.16 GiB) tagged, 4792 assets (8.91 GiB) deleted.
[NaN%] 0 files scanned: 0 assets (0.00 MiB) tagged, 0 assets (0.00 MiB) deleted.
<hangs here until killed>
```

(The `[NaN%]` line is a separate cosmetic issue caused by an empty bootstrap ECR repo; not the subject of this report.)

### Reproduction Steps

**Code-level repro** (no AWS account required) - standalone unit test against `BackgroundStackRefresh` showing `stop()` cannot cancel an in-flight `refresh()`:

```ts
import { BackgroundStackRefresh, ActiveAssetCache } from '<path-to>/stack-refresh';

test('stop() does not cancel an in-flight refresh, leaking a timer', async () => {
  let resolveRefresh: () => void;
  const fakeCfn = {
    listStacks: () => new Promise(r => { resolveRefresh = () => r({ StackSummaries: [] }); }),
    getTemplateSummary: async () => ({ Parameters: [] }),
    getTemplate: async () => ({ TemplateBody: '' }),
  };
  const bg = new BackgroundStackRefresh({
    cfn: fakeCfn as any,
    ioHelper: { defaults: { debug: async () => {} } } as any,
    activeAssets: new ActiveAssetCache(),
  });

  jest.useFakeTimers();
  bg.start();
  jest.advanceTimersByTime(300_000);   // first refresh fires, now awaits listStacks
  bg.stop();                            // caller thinks it's done
  resolveRefresh!();                    // in-flight refresh completes
  await Promise.resolve();              // drain microtasks

  // BUG: stop() was called, yet a new timer is pending.
  expect(jest.getTimerCount()).toBe(0); // FAILS - one pending timer leaks
});
```

**Real-world repro** (for severity context):

```
cdk gc --unstable=gc aws://<redacted>/us-east-1 \
  --rollback-buffer-days=3 --created-buffer-days=1 \
  --action=full --type=all --confirm=false
```

Against a bootstrap S3 bucket large enough that total runtime exceeds 5 min. Observed 100% correlation between runtime > 5 min and hang across multiple runs in a CI pipeline.

### Possible Solution

Gate the reschedule on a stopped flag, and drain `queuedPromises` so any caller parked in `noOlderThan()` unblocks:

```diff
 export class BackgroundStackRefresh {
   private timeout?: NodeJS.Timeout;
+  private stopped = false;

   private async refresh() {
     const startTime = Date.now();
     await refreshStacks({ ... });
     this.justRefreshedStacks();
+    if (this.stopped) return;
     this.timeout = setTimeout(() => this.refresh(),
       Math.max(startTime + 300_000 - Date.now(), 0));
   }

   public stop() {
+    this.stopped = true;
     clearTimeout(this.timeout);
+    for (const p of this.queuedPromises.splice(0)) p(undefined);
   }
 }
```

Happy to submit the PR against `aws/aws-cdk-cli` with the above diff plus the regression test - confirm the approach and I'll send it.

Related: aws/aws-cdk#33441 (same class of bug, closed). The attempted fix in aws/aws-cdk#33466 is incomplete - adding `clearTimeout` inside `refresh()` before re-setting cannot cancel the already-executing `refresh()` either, so the rescheduling race still exists. That PR was never merged.

The gc CLI code now lives in `aws/aws-cdk-cli` (file path above). The `aws-cdk-cli` README directs user-facing issue reports here, which is why this is filed on `aws/aws-cdk`.

2.236.0

### AWS CDK CLI version

2.1111.0 (build 69089a7)

### Node.js Version

v22.22.2

### OS

ubuntu-24.04

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

rubenfonseca:
### Issue # (if applicable)

Fixes aws/aws-cdk-cli#1385

### Reason for this change

`cdk gc` hangs silently after completing its work on any run whose total duration exceeds ~5 minutes. The process never exits; CI pipelines hit step timeouts and users have to SIGKILL.

Root cause is in `BackgroundStackRefresh`: `refresh()` unconditionally reschedules itself via `setTimeout` on every completion, and `stop()` only calls `clearTimeout` on the tracked timer. If `stop()` fires while an awaited `refreshStacks()` is in flight, the `clearTimeout` call has no effect on the already-executing `refresh()` (the timer that launched it was consumed when it fired). When the in-flight `refresh()` finally resolves, it installs a fresh `setTimeout` that `stop()` has already passed over. That timer pins the Node event loop indefinitely, and re-fires every 5 minutes forever.

This also explains the previously reported stuck-at-NN.NN% / silent-hang patterns — see aws/aws-cdk#33441 and the unmerged-but-insufficient fix in aws/aws-cdk#33466 (which added a redundant `clearTimeout` inside `refresh()` but did not prevent the reschedule from winning the race with `stop()`).

### Description of changes

- Add a `stopped` flag on `BackgroundStackRefresh`. `stop()` sets it; `refresh()` checks it after `refreshStacks()` resolves and returns without rescheduling if `stop()` has been called. This closes the race without changing any other behavior.
- Add a regression test that drives a controllable `listStacks` mock to keep `refresh()` in flight, calls `stop()` during the await, resolves the mock, and asserts that no new `setTimeout` is installed. The test fails on the unpatched code and passes with the fix.

No public API changes, no behaviour change on the happy path.

### Describe any new or updated permissions being added

None.

- New unit test `stop() prevents rescheduling even if called during an in-flight refresh` — passes with the fix, fails without it (verified by reverting just the library-side change and re-running).
- All existing `BackgroundStackRefresh` tests still pass (6 passed, 0 failed in `garbage-collection.test.ts` within the `BackgroundStackRefresh` describe block).

### Checklist

- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts
@@ -147,6 +147,7 @@ export class BackgroundStackRefresh {
   private timeout?: NodeJS.Timeout;
   private lastRefreshTime: number;
   private queuedPromises: Array<(value: unknown) => void> = [];
+  private stopped = false;
 
   constructor(private readonly props: BackgroundStackRefreshProps) {
     this.lastRefreshTime = Date.now();
@@ -168,6 +169,13 @@ export class BackgroundStackRefresh {
       qualifier: this.props.qualifier,
     });
     this.justRefreshedStacks();
+
+    // If stop() was called while the awaited refreshStacks() above was in flight, do not
+    // reinstall a timer — clearTimeout() in stop() could not cancel the already-executing
+    // refresh() call, and scheduling a new one here would pin the event loop forever.
+    if (this.stopped) {
+      return;
+    }
 
     // If the last invocation of refreshStacks takes <5 minutes, the next invocation starts 5 minutes after the last one started.
     // If the last invocation of refreshStacks takes >5 minutes, the next invocation starts immediately.
@@ -203,6 +211,7 @@ export class BackgroundStackRefresh {
   }
 
   public stop() {
+    this.stopped = true;
     clearTimeout(this.timeout);
   }
 }
--- a/packages/@aws-cdk/toolkit-lib/test/api/garbage-collection/garbage-collection.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/garbage-collection/garbage-collection.test.ts
@@ -1008,6 +1008,37 @@ describe('BackgroundStackRefresh', () => {
 
     await expect(waitPromise).rejects.toThrow('refreshStacks took too long; the background thread likely threw an error');
   });
+
+  test('stop() prevents rescheduling even if called during an in-flight refresh', async () => {
+    // Make listStacks return a promise we control, so refresh() stays in flight
+    // while stop() is called.
+    let resolveListStacks: (value: any) => void = () => {
+    };
+    cfnClient.on(ListStacksCommand).callsFake(() => new Promise((resolve) => {
+      resolveListStacks = resolve;
+    }));
+
+    void backgroundRefresh.start();
+
+    // Fire the initial 5-minute timer; refresh() begins and suspends awaiting listStacks.
+    await jest.runOnlyPendingTimersAsync();
+
+    // Caller decides to stop while refresh() is mid-await.
+    setTimeoutSpy.mockClear();
+    backgroundRefresh.stop();
+
+    // Complete the in-flight refresh(). Without the stopped-flag guard, refresh()
+    // would reinstall a setTimeout here and pin the event loop indefinitely.
+    resolveListStacks({ StackSummaries: [] });
+    // Drain microtasks thoroughly so the in-flight refresh() fully unwinds.
+    for (let i = 0; i < 50; i++) {
+      await Promise.resolve();
+    }
+    await jest.advanceTimersByTimeAsync(0);
+
+    expect(setTimeoutSpy).not.toHaveBeenCalled();
+    expect(jest.getTimerCount()).toBe(0);
+  });
 });
 
 describe('ProgressPrinter', () => {


rix0rrr (approved):
Thanks!

PR #1387, opened 2026-04-17
fix(cli): deploy with no changes skips outputs and stack ARN

### LINKED ISSUE #1386: (core): Unchanged Stacks does not run exports anymore
### Describe the bug

With `2.250.0` Exports does not run on unchanged stacks anymore:
```
MyStack/GeneralVpc (no changes)
```

With `2.240.0` we see the exports running
```
MyStack/GeneralVpc (no changes)

 
✨  Deployment time: 0.26s
 
Outputs:
MyStackGeneralVpc4AD906FB.ExportsOutputFnEndpoint248E82BEDnsEntries6213B1EB = .......
```

### Regression Issue

_No response_

### Expected Behavior

Exports should be exported on unchanged stacks to be used downstream.

### Current Behavior

exports are missing downstream

### Reproduction Steps

Export on stack and deploy unchanged.

### Possible Solution

_No response_

_No response_

2.250.0`

### AWS CDK CLI version

2.1115.1

### Node.js Version

v24.14.1

### OS

Ubuntu

### Language

Python

### Language Version

_No response_

### Other information

_No response_

### LINKED ISSUE #1389: (cdk deploy): --outputs-file does not produce a new file anymore
### Describe the bug

There seems to be a regression in v2.1118.1 where `--outputs-file` does not write the file anymore.

### Regression Issue

v2.1118.0 (build 2a848bc)

### Expected Behavior

`cdk deploy --outputs-file outputs.json` will produce a JSON file containing the CloudFormation outputs. This is working up to `2.1118.0 (build 2a848bc)`.

### Current Behavior

`cdk deploy --outputs-file outputs.json` deploys successfully but `outputs.json` is not written anywhere.

### Reproduction Steps

1. Install `aws-cdk@2.1118.0` globally.
2. Run `cdk deploy --outputs-file outputs.json` on a project with Cloudformation outputs. 
3. The deploy should succeed and the `outputs.json` is written.

Repeat steps 1 and 2 with `2.1118.1`. The deploy will still succeed but the `outputs.json` is missing.

Repeat steps 1 and 2 with latest aws-cdk (`2.1118.2` as of this writing). The deploy will also succeed but the `outputs.json` is missing.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.1118.1 (build 3a90862)

### Framework Version

2.248.0

### Node.js Version

25.9.0

### OS

Linux (WSL)

### Language

Python

### Language Version

3.14.4

### Other information

_No response_

mrgrain:
Fixes #1386
Fixes #1389

Since the two-phase deploy was introduced in #1273, an empty change set returned early from the deploy function before printing outputs, stack ARN, and deployment timings. For stacks without changes, downstream consumers that depend on CloudFormation exports saw nothing after the "(no changes)" line, breaking workflows that assumed outputs are always visible on successful deploys.

The fix lets the no-op result flow through the normal output formatting path by including it in the `prepareIsFinal` check. The existing output, timing, and stack ARN notifications fire as they did before the regression. No additional change set is created because the prepared result is already final — the deploy loop is skipped entirely when `prepareResult.noOp` is true.

The same bug existed in both the CLI and the toolkit-lib deploy implementations, and both are fixed here. It also caused `--outputs-file` to be skipped on no-change deploys, since the file is written in the deploy function's `finally` block that was never reached. Unit tests in each package assert that outputs, deployment time, stack ARN, and the `--outputs-file` are emitted on a no-change deploy. Integration tests redeploy a stack with an output and assert both the stdout outputs and `--outputs-file` are present in both runs.

### Checklist

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/deploy/cdk-deploy-outputs-file-written-on-no-change-deploy.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/deploy/cdk-deploy-outputs-file-written-on-no-change-deploy.integtest.ts
@@ -1,0 +1,28 @@
+import { promises as fs } from 'fs';
+import * as path from 'path';
+import { integTest, withDefaultFixture } from '../../../lib';
+
+integTest(
+  'outputs-file is written on initial and no-change deploys',
+  withDefaultFixture(async (fixture) => {
+    const outputsFile = path.join(fixture.integTestDir, 'outputs', 'outputs.json');
+    await fs.mkdir(path.dirname(outputsFile), { recursive: true });
+
+    // First deploy — creates the stack and writes the outputs file
+    await fixture.cdkDeploy('outputs-test-1', { options: ['--outputs-file', outputsFile] });
+    const firstOutputs = JSON.parse(await fs.readFile(outputsFile, 'utf-8'));
+    expect(firstOutputs).toEqual({
+      [`${fixture.stackNamePrefix}-outputs-test-1`]: {
+        TopicName: `${fixture.stackNamePrefix}-outputs-test-1MyTopic`,
+      },
+    });
+
+    // Delete the file so we can assert it gets recreated on the no-change deploy
+    await fs.rm(outputsFile);
+
+    // Second deploy — no changes, outputs file must still be written
+    await fixture.cdkDeploy('outputs-test-1', { options: ['--outputs-file', outputsFile] });
+    const secondOutputs = JSON.parse(await fs.readFile(outputsFile, 'utf-8'));
+    expect(secondOutputs).toEqual(firstOutputs);
+  }),
+);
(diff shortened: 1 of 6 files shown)


zined:
Thanks for the quick fix!

One small suggestion for the test coverage: the integration test and the unit
tests both assert that `Outputs:` and the output key appear on stdout, but not 
that `--outputs-file` contains them. The concrete failure we hit in CI was the 
file, not the stdout: a no-op stack deploy produced an `outputs.json` missing
the entries for that stack, and a later step reading the file via `jq` got 
`null` and passed it as an IAM `Resource`, which CloudFormation rejected.

A minimal addition that would have caught this directly:

```ts
const outputsFile = path.join(workDir, 'outputs.json');

await cdkToolkit.deploy({
  // ... existing setup ... 
  outputsFile,
});

expect(await fs.readJson(outputsFile)).toEqual({
  'Test-Stack-A': { BucketName: 'my-bucket' },
});
```

Happy to send it as a follow-up PR if that is useful, or you can fold it into
this one.

Also, for cross-referencing: issue #1389 looks like a duplicate of #1386 and 
could be linked or closed once this merges.

mrgrain:
@zined Yep, added the additional tests!

otaviomacedo (approved):

PR #1399, opened 2026-04-20
feat(cli): `cdk orphan` command detaches resources from a stack, allowing resource type upgrades (behind `--unstable`)

LeeroyHannigan:
ref: https://github.com/aws/aws-cdk-cli/pull/1324

Adds a new `cdk orphan` command (behind `--unstable=orphan`) that safely detaches resources from a CloudFormation stack without deleting them. This enables resource type migrations (e.g., DynamoDB `Table` → `TableV2`) with zero downtime or data loss.

## Usage

```console
# Orphan a single resource
$ cdk orphan --unstable=orphan MyStack/MyTable

# Orphan multiple resources
$ cdk orphan --unstable=orphan MyStack/MyTable MyStack/MyBucket
```

## How it works

The command performs three sequential CloudFormation deployments:

1. **Resolve references** — Injects temporary stack Outputs to resolve `{Ref}`, `{Fn::GetAtt}`, and `{Fn::Sub}` values for the targeted resources via `DescribeStacks`
2. **Isolate resources** — Sets `DeletionPolicy: Retain`, replaces all cross-resource references with resolved literal values, and removes `DependsOn` entries
3. **Remove resources** — Deletes the resources from the CloudFormation template (they continue to exist in the AWS account)

After orphaning, the command outputs a `cdk import --resource-mapping-inline` command with the resource mapping, so users can immediately re-import the resource under a new construct type.

### New: `cdk orphan` command
- `packages/@aws-cdk/toolkit-lib/lib/api/orphan/` — Core `ResourceOrphaner` class and reference resolution helpers
- `packages/aws-cdk/lib/cli/` — CLI wiring: command definition, argument parsing, user input types
- `packages/aws-cdk/README.md` — Documentation with usage examples and a DynamoDB migration walkthrough

### Enhanced: `cdk import` inline resource mapping
- `packages/@aws-cdk/toolkit-lib/lib/api/resource-import/importer.ts` — Added `loadResourceIdentifiers()` accepting a JSON string or object (in addition to the existing file-based `loadResourceIdentifiersFromFile()`)

### Tests
- `packages/@aws-cdk/toolkit-lib/test/api/orphan/orphan.test.ts` — Unit tests for the orphaner (394 lines)
- `packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/orphan/` — Integration test: deploys a DynamoDB table + Lambda consumer, orphans the table, verifies template changes and data integrity

Depends-On: https://github.com/aws/aws-cdk-cli/pull/1403

---


packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
   IAM permissions to the `deploy-role`.
 
 
+### `cdk orphan`
+
+`cdk orphan` is currently experimental, meaning we reserve the right to change 
+options and flag names in the future. Pass the `--unstable=orphan` flag when using
+this command and be aware of this when using it in scripts.
```
rix0rrr:
```suggestion
> `cdk orphan` is currently experimental, meaning we reserve the right to change 
> options and flag names in the future. Pass the `--unstable=orphan` flag when using
> this command and be aware of this when using it in scripts.
```

rix0rrr:
The build is failing, and needs AppSec involvement because of the bootstrap stack change. Other than that I am good with this!

packages/@aws-cdk-testing/cli-integ/package.json
mrgrain, line 1: This needs to be reverted to keep the updated client versions. (As it is, t's downgrading client versions.)

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1153,6 +1155,72 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+    await ioHelper.defaults.info(`✅ Resources orphaned from ${plan.stackName}`);
+    await ioHelper.defaults.info('');
+    await ioHelper.defaults.info('Next steps:');
+    await ioHelper.defaults.info('  1. Update your CDK code to use the new resource type');
+    if (Object.keys(result.resourceMapping).length > 0) {
+      const mappingJson = JSON.stringify(result.resourceMapping);
+      await ioHelper.defaults.info(`  2. cdk import --resource-mapping-inline '${mappingJson}'`);
+    } else {
+      await ioHelper.defaults.info('  2. cdk import');
+    }
```
mrgrain: Make this a single message please. We do not use a line emitting system (like `console.log`) but an event bus-like system with complete messages.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1153,6 +1155,72 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+    });
+
+    const plan = await orphaner.makePlan(stack, parsed.constructPaths);
+
+    // Show the plan
+    await ioHelper.defaults.info(`Stack: ${plan.stackName}`);
+    await ioHelper.defaults.info(`Resources to orphan (${plan.orphanedResources.length}):`);
+    for (const r of plan.orphanedResources) {
+      await ioHelper.defaults.info(`  ${r.logicalId} (${r.resourceType}) — ${r.cdkPath}`);
+    }
```
mrgrain: Make this a single message please. We do not use a line emitting system (like console.log) but an event bus-like system with complete messages.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1584,3 +1652,38 @@ async function synthAndMeasure(
...
   return { asMs: 0, asSec: 0 };
 }
+
+/**
+ * Parse construct paths like `/MyStack/MyTable` or `MyStack/MyTable` into
+ * a stack construct ID and construct-level paths.
+ *
+ * All paths must reference the same stack.
+ */
+function parseConstructPaths(paths: string[]): { stackId: string; constructPaths: string[] } {
```
mrgrain: suggestion (non-blocking): can we move this to a utils or helper file instead? Seems generic enough and will clutter the main class less.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1584,3 +1652,38 @@ async function synthAndMeasure(
...
+  }
+
+  const constructPaths: string[] = [];
+  let stackId: string | undefined;
+
+  for (const raw of paths) {
+    const p = raw.replace(/^\//, ''); // strip leading slash
+    const slashIdx = p.indexOf('/');
+    if (slashIdx < 0) {
+      throw new ToolkitError('InvalidConstructPath', `Construct path '${raw}' must be in the format StackId/ConstructPath (e.g. MyStack/MyTable)`);
```
mrgrain: nit: I think this error happens if you give it just a stack name `MyStack` or just an id `MyTable`. maybe we can improve the error message here and make it more friendly.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: nit: this seems to mix multiple concerns other than just parsing a construct path, e.g. it also enforces all paths to be in the same stack

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -1026,9 +1035,14 @@ export class CdkToolkit {
     }
 
     // Prepare a mapping of physical resources to CDK constructs
-    const actualImport = !options.resourceMappingFile
-      ? await resourceImporter.askForResourceIdentifiers(additions)
-      : await resourceImporter.loadResourceIdentifiers(additions, options.resourceMappingFile);
+    let actualImport: Awaited<ReturnType<typeof resourceImporter.askForResourceIdentifiers>>;
+    if (options.resourceMappingInline) {
+      actualImport = await resourceImporter.loadResourceIdentifiers(additions, JSON.parse(options.resourceMappingInline));
```
mrgrain: suggestion: what about JSON.parse errors? I'd be inclined to have `loadResourceIdentifiers` receive a raw JSON string and handle JSON errors there gracefully.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -292,6 +292,11 @@ export async function makeConfig(): Promise<CliConfig> {
             desc: 'If specified, CDK will use the given file to map physical resources to CDK resources for import, instead of interactively ' +
               'asking the user. Can be run from scripts',
           },
+          'resource-mapping-inline': {
+            type: 'string',
+            requiresArg: true,
+            desc: 'Inline JSON resource mapping, e.g. \'{"MyResource":{"TableName":"my-table"}}\'',
+          },
```
mrgrain, line 299: thought: Annoyingly we have taken `-m` already. do we need this to be a separate flag?

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -494,6 +499,21 @@ export async function makeConfig(): Promise<CliConfig> {
...
+          path: {
+            type: 'array',
+            requiresArg: true,
+            desc: 'Construct path(s) to orphan, e.g. MyStack/MyTable',
+          },
+          force: {
+            type: 'boolean',
+            default: false,
+            desc: 'Do not ask for confirmation before orphaning',
+          },
```
mrgrain: question: Is this not the same as `--yes`? Do we need a separate flag?

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
...
+3. Removing the resources from the CloudFormation template (they continue to exist in your AWS account)
+
+After orphaning, you can update your CDK code and use `cdk import` to bring the resource back under management with the new construct type.
+
+```console
+$ # Orphan a single resource
+$ cdk orphan --unstable=orphan --path MyStack/MyTable
+
+$ # Orphan multiple resources
+$ cdk orphan --unstable=orphan --path MyStack/MyTable --path MyStack/MyBucket
```
mrgrain: Why `--path`? This looks nicer to me:

```suggestion
$ cdk orphan --unstable=orphan MyStack/MyTable MyStack/MyBucket
```

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
...
+
+1. Resolving any cross-resource references (`Ref`, `Fn::GetAtt`) to the orphaned resources and replacing them with their current physical values
+2. Setting the `DeletionPolicy` to `Retain` and decoupling the resources from the stack
+3. Removing the resources from the CloudFormation template (they continue to exist in your AWS account)
+
+After orphaning, you can update your CDK code and use `cdk import` to bring the resource back under management with the new construct type.
+
+```console
+$ # Orphan a single resource
+$ cdk orphan --unstable=orphan --path MyStack/MyTable
```
mrgrain: Does this support wildcards?

E.g.

```
cdk orphan --unstable=orphan MyStack*/MyTable
```

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
...
 
+### `cdk orphan`
+
+> `cdk orphan` is currently experimental, meaning we reserve the right to change 
+> options and flag names in the future. Pass the `--unstable=orphan` flag when using
+> this command and be aware of this when using it in scripts.
+
+Detaches one or more resources from a CloudFormation stack without deleting them.
+This is useful when you need to migrate a resource from one construct type to another
+(for example, migrating a DynamoDB `Table` to `TableV2`) without any downtime or data loss.
```
mrgrain: Not quite sure, but want to make this less scary sounding. It's safe. Trust us.

```suggestion
Safely detaches one or more resources from a CloudFormation stack without deleting them.
This is useful when you need to migrate a resource from one construct type to another
(for example, migrating a DynamoDB `Table` to `TableV2`) without any downtime or data loss.
```

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
...
+
+The orphan command works by:
+
+1. Resolving any cross-resource references (`Ref`, `Fn::GetAtt`) to the orphaned resources and replacing them with their current physical values
+2. Setting the `DeletionPolicy` to `Retain` and decoupling the resources from the stack
```
mrgrain: nit: Feels a bit short to just say "decoupling". what is it actually doing?

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
...
+> options and flag names in the future. Pass the `--unstable=orphan` flag when using
+> this command and be aware of this when using it in scripts.
+
+Detaches one or more resources from a CloudFormation stack without deleting them.
+This is useful when you need to migrate a resource from one construct type to another
+(for example, migrating a DynamoDB `Table` to `TableV2`) without any downtime or data loss.
+
+The orphan command works by:
+
+1. Resolving any cross-resource references (`Ref`, `Fn::GetAtt`) to the orphaned resources and replacing them with their current physical values
```
mrgrain: nit: Okay but why?

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
...
+2. Setting the `DeletionPolicy` to `Retain` and decoupling the resources from the stack
+3. Removing the resources from the CloudFormation template (they continue to exist in your AWS account)
```
mrgrain, line 810: Nice!

packages/aws-cdk/README.md
```diff
@@ -792,6 +793,71 @@ This feature currently has the following limitations:
   IAM permissions to the `deploy-role`.
 
 
+### `cdk orphan`
```
mrgrain, line 796: nit: Doesn't mention the single stack limitation.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1584,3 +1652,38 @@ async function synthAndMeasure(
...
+    }
+
+    const thisStack = p.substring(0, slashIdx);
+    const constructPath = p.substring(slashIdx + 1);
+
+    if (stackId && thisStack !== stackId) {
+      throw new ToolkitError('MultipleStacks', `All construct paths must reference the same stack, but got '${stackId}' and '${thisStack}'`);
+    }
```
mrgrain: q: But why?

packages/@aws-cdk/toolkit-lib/lib/api/orphan/orphaner.ts
mrgrain, line 1: task: Implementation classes currently live in `lib/api`, not `lib/actions`. There's no good reason for it except consistency and maybe slightly less risk of exporting something that shouldn't be.

packages/@aws-cdk/toolkit-lib/lib/actions/orphan/orphaner.ts
```diff
@@ -0,0 +1,335 @@
...
+    const currentTemplate = await this.deployments.readCurrentTemplate(stack);
+    const resources = currentTemplate.Resources ?? {};
+
+    const logicalIds: string[] = [];
+    for (const p of constructPaths) {
+      logicalIds.push(...findResourcesByPath(resources, stack.stackName, p));
+    }
+
+    if (logicalIds.length === 0) {
+      const hint = !hasAnyCdkPathMetadata(resources)
```
mrgrain: Q: Why do we need this? We have already synth'd the template earlier. Can we not (also) use the local app to get a logical id for a construct path? We can still fail if we cannot find a physical id for it.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -292,6 +292,11 @@ export async function makeConfig(): Promise<CliConfig> {
             desc: 'If specified, CDK will use the given file to map physical resources to CDK resources for import, instead of interactively ' +
               'asking the user. Can be run from scripts',
           },
+          'resource-mapping-inline': {
```
mrgrain, line 295: nit: This could technically be its own PR

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
LeeroyHannigan: Ack, the function is small enough that I think the single-responsibility tradeoff is acceptable here, but happy to split if you feel strongly.

packages/aws-cdk/lib/cli/cli-config.ts
LeeroyHannigan: Interesting, what we could do is always output to a file and keep it usage only with `-m`?

packages/aws-cdk/lib/cli/cli-config.ts
LeeroyHannigan: Good catch, this was left over from the initial implementation.

packages/aws-cdk/README.md
LeeroyHannigan: No, wildcards aren't supported. Paths are matched as exact prefixes. Do you want me to glob for wildcards?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
LeeroyHannigan: No strong technical reason. We could group by stack and run the flow for each stack independently. Kept it to a single stack for simplicity in the initial release.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: I think it's okay to have it do both - but then it should be named accordingly.

packages/aws-cdk/README.md
mrgrain: Not necessarily if there's good reason not to. From my perspective, it's just a little unexpected since most other commands accept patters. I think the minimum would be to call this limitation out in the readme.

packages/aws-cdk/lib/cli/cli-config.ts
mrgrain: I was wondering if we could overload it? But I am okay with the current solution.

packages/aws-cdk/README.md
mrgrain: I still feel that the single stack limitation is unnecessary. But if it makes it easier to land this, I am happy to. Let's just call it out in the readme and open an issue to note that we should revisit that decision.

mrgrain:
approved in principle. we need the additional PR #1403 merged first.

mrgrain (approved):

PR #1400, opened 2026-04-20
fix(toolkit-lib): hotswap deployments swap all changes since first template of session

### LINKED ISSUE #885: (hotswap): Hotswap always deploys all resources changed since last full deployment
Running `cdk deploy --hotswap` will update all hotswapped resources since the last CloudFormation deployment.

Suppose a stack contains two lambda functions, `lambda1` and `lambda2`. Then a change to `lambda1` is deployed with `--hotswap`. Now a change to `lambda2` is made. When these changes to `lambda2` are deployed with `--hotswap`, the changes to `lambda1` are also deployed, even though they were deployed by a previous hotswap.

This happens because all potentially hotswappable changes are found by diffing the current app template with the template in CloudFormation, which is not updated by hotswaps. Thus subsequent `--hotswap` deployments do not know about previous `--hotswap` deployments; that is, hotswaps are not aware of drift introduced by previous hotswaps.

This could incur a performance penalty over time. The time it takes for a `--hotswap` deployment to complete is proportional to the number of changed resources, not the number of changes. If many resources are changed over the course of many `--hotswap` deployments, then the performance will degrade. If many changes are made to only a few resources, the performance change is negligible.

To avoid this performance loss, perform a regular deployment. This will remove all drift, meaning that the next hotswap will not suffer the performance penalty from the previously accumulated drift. 

## Edit by @mrgrain 

https://github.com/aws/aws-cdk-cli/issues/888 poses a related problem. When reverting back to the original template after a change that was hotswapped, hotswap does not recognise this as a change anymore and therefore will not run.

ShadowCat567:
Fixes #885 

Previously when performing a hotswap deployment, in order to figure out what changed, the CLI would create a diff between the Cloudformation template from the current hotswap deployment and the deployed Cloudformation state. 
This strategy had some issues which are outlined in issue #885 when multiple hotswap deployment happen in a row without a Cloudformation deployment between them. 

In this PR, a hotswap cache is introduced which saves the Cloudformation template created by a hotswap deployment in a folder called `.hotswap-cache` in `cdk.out`. If another hotswap deployment is performed afterwards, it will perform a diff against the template that is in this folder rather than getting the currently deployed Cloudformation template. When a full Cloudformation deployment happens through `cdk deploy` or if non hotswappable changes are detected when running `cdk hotswap --hotswap-fallback`, then the template for that stack will be cleared from the cache.
Whenever a hotswap deployment happens and there is no template in the hotswap cache for the current stack, the template from Cloudformation will be diff-ed against to detect hotswappable changes. 

Added some integ tests for these changes and enforced that the CCAPI tests which involve Bedrock Agents only run in regions that support Bedrock Agents. 

---


rix0rrr (commented):
Just based on reading the PR description, why not keep this cache in memory? 

Is that to catch `cdk deploy --hotswap` cases (as opposed to `cdk watch`) ?

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-template-cache.ts
```diff
@@ -0,0 +1,146 @@
...
+  const cachedPath = cachePath(assemblyDir, stackName);
+  await fs.ensureDir(path.dirname(cachedPath));
+  await fs.writeJson(cachedPath, state, { spaces: 2 });
+}
+
+/**
+ * Invalidate the hotswap cache for a stack (e.g. after a full CloudFormation deploy).
+ */
+export async function invalidateHotswapTemplateCache(assemblyDir: string, stackName: string): Promise<void> {
+  await fs.remove(cachePath(assemblyDir, stackName));
```
rix0rrr: `{ force: true }` ?

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -378,6 +379,11 @@ class FullCloudFormationDeployment {
   public async performDeployment(): Promise<DeployStackResult> {
     const deploymentMethod = this.deploymentMethod ?? { method: 'change-set' };
 
+    // if there is a hotswap cache, clear it when a full Cloudformation of any kind happens
+    if (existingHotswapCache(this.stackArtifact.assembly.directory, this.stackArtifact.stackName)) {
+      await invalidateHotswapTemplateCache(this.stackArtifact.assembly.directory, this.stackArtifact.stackName);
```
rix0rrr: Technically speaking this is a TOCTOU. It might be rare, but it's a pattern to be cognizant of.

Any type of code that looks like:

```ts
if (condition) {
  // 1️⃣
  performActionBasedOnCondition();
}
```

Is vulnerable to a race condition that causes `condition` to be false at 1️⃣.

* For fully in-memory conditions:
  * In a single-threaded application (the common case) that can't happen, and because that is so common that lulls you into forgetting about this problem.
  * For multi-threaded applications, you now need to start paying attention to whether some other thread can concurrently invalidate 1️⃣. In JavaScript, which is single-threaded, that either also can't happen or it will be obvious if it can or cannot due to `await` statements.
* If `condition` is based on an resource that's external to your application, like the filesystem or a web service or what have you -- this is quite relevant! 

If you do:

```ts
if (fileExists) {
  readFile();
}
```

Then you are vulnerable to `readFile` exploding with a `FileNotFound` error, if a concurrent process happens to delete it between the existence check and the readFile! 

A traditional vulnerability in UNIX systems is a vulnerability to a file being replaced with a symlink in between some access check and actually operating on the file, and presto, you now have performed an action on a file you didn't intend to!

----

Common fixes to this are: 

* In the condition check, return/operate on an unfakeable handle that cannot disappear or change meaning. For example, a `file handle` instead of a `file path`.
* Don't do a `predictSuccess + act`, instead do an `act + recover`. Try to perform the action, and if the result is a failure try to determine after the fact whether that error was actually acceptable/expeted and in that case ignore it.

In this case, you can try the removal and catch + ignore any errors. Or you can use a built-in mechanism for that: `rm(..., { force: true })`.

rix0rrr (commented):
Love it! 

Is there a way in which we can write an integ test against this? And have you tested it manually?

ShadowCat567:
> Love it!
> 
> Is there a way in which we can write an integ test against this? And have you tested it manually?

Yes I have tested manually (and caught several nested stack issues while testing), I'll see about adding an integ test.

ShadowCat567:
> Just based on reading the PR description, why not keep this cache in memory?
> 
> Is that to catch `cdk deploy --hotswap` cases (as opposed to `cdk watch`) ?

Yes, this solution is more focused on `cdk deploy --hotswap` cases ~~(have not tested with `watch` yet)~~. This was the solution I had in mind since the RFC, so I did not consider caching in memory does it make sense for me to look into whether that would work for this?

Have tested with both `cdk deploy --hotswap` and `cdk watch` - this works with both which is what I wanted out of this change.

rix0rrr (approved):

PR #1402, opened 2026-04-21
fix: skip require-approval prompt when change set is empty

### LINKED ISSUE #1401: toolkit-lib: deploy() fires require-approval prompt even when change set is empty
### Package

`@aws-cdk/toolkit-lib` (tested on 1.22.1)

### Context

[aws-cdk-cli#1273](https://github.com/aws/aws-cdk-cli/pull/1273) changed the `change-set` deployment flow so that the approval diff is computed from a real CloudFormation change set instead of a local template comparison. That's great for accuracy.

However, when `requireApproval` is enabled and the change set contains no changes, the toolkit still prompts the user to approve the deployment.

### Observed behaviour

Deploying a stack that has no changes, with `--require-approval` enabled (or via `toolkit-lib`'s equivalent), produces a prompt like:

```
Stack my-stack
There were no differences

"--require-approval" is enabled and stack includes updates.
Do you wish to deploy these changes (y/n)?
```

The diff explicitly reports *no differences*, yet the user is still asked to approve "updates".

### Expected behaviour

When the prepared change set has no changes (`prepareResult.noOp === true`), the approval prompt should be skipped and the toolkit should emit a "no changes, skipping deployment" notification instead.

### Root cause

In [`lib/toolkit/toolkit.ts:709-742`](https://github.com/aws/aws-cdk-cli/blob/@aws-cdk/toolkit-lib@v1.22.1/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L709-L742), the deploy stack loop does:

```ts
const prepareResult = isChangeSetDeployment(options.deploymentMethod)
  ? await deployments.prepareStack({
    ...sharedDeployOptions,
    deploymentMethod: options.deploymentMethod,
    cleanupOnNoOp: isExecutingChangeSetDeployment(options.deploymentMethod),
  })
  : undefined;

const formatter = new DiffFormatter({ /* ... */ });
const securityDiff = formatter.formatSecurityDiff();
const stackDiff = formatter.formatStackDiff();

const hasSecurityChanges = securityDiff.permissionChangeType !== PermissionChangeType.NONE;
const deployMotivation = hasSecurityChanges
  ? '"--require-approval" is enabled and stack includes security-sensitive updates.'
  : '"--require-approval" is enabled and stack includes updates.';
const diffOutput = hasSecurityChanges ? securityDiff.formattedDiff : stackDiff.formattedDiff;
const deployQuestion = `${diffOutput}\n\n${deployMotivation}\nDo you wish to deploy these changes`;
const deployConfirmed = await ioHelper.requestResponse(IO.CDK_TOOLKIT_I5060.req(deployQuestion, { /* ... */ }));
```

`prepareResult.noOp` is set to `true` in the change-set-no-changes branch of `FullCloudFormationDeployment.performDeployment` ([`lib/api/deployments/deploy-stack.ts:427-432`](https://github.com/aws/aws-cdk-cli/blob/@aws-cdk/toolkit-lib@v1.22.1/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts#L427-L432)) and is consulted later at [`toolkit.ts:761`](https://github.com/aws/aws-cdk-cli/blob/@aws-cdk/toolkit-lib@v1.22.1/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L761) for `prepareIsFinal`, but it is not consulted before the `CDK_TOOLKIT_I5060` request. The prompt therefore fires regardless of whether the change set is empty.

### Suggested fix

Before the `CDK_TOOLKIT_I5060` request, short-circuit when the change set is a no-op and emit a warn-level notification analogous to the existing [`CDK_TOOLKIT_W5021`](https://github.com/aws/aws-cdk-cli/blob/@aws-cdk/toolkit-lib@v1.22.1/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L588) (used for stacks with no resources):

```ts
if (prepareResult?.noOp) {
  await ioHelper.notify(IO.CDK_TOOLKIT_W5023.msg(
    `${chalk.bold(stack.displayName)}: stack has no changes, skipping deployment.`,
  ));
  // record result and continue to next stack
  continue;
}
```

`CDK_TOOLKIT_W5023` would be a new code in the same `W502x` family as `W5021`/`W5022`.

### Reproduction

1. Deploy a stack with `@aws-cdk/toolkit-lib` using the default `change-set` deployment method.
2. Immediately deploy the same stack again without changing anything.
3. Observe the approval prompt appearing even though the diff reports "There were no differences".

### Additional context

This regression is visible after #1273 landed. The earlier public proposal for change-set review (#892, closed unmerged) included an explicit early-exit when the change set had no changes, avoiding this prompt — I humbly suggest we preserve that behaviour.

orien:
Fixes #1401

## Context

When `--require-approval` is enabled and the prepared change set has no changes, the toolkit still asks the user to approve the deploy. The output reads:

```
There were no differences

"--require-approval" is enabled and stack includes updates.
Do you wish to deploy these changes (y/n)?
```

The diff explicitly reports there are no differences, yet the user is prompted to confirm "updates" that don't exist.

The prompt lives in both `@aws-cdk/toolkit-lib`'s `toolkit.ts` deploy loop and the CLI's `cdk-toolkit.ts`. `prepareResult.noOp` is already consulted later (for `prepareIsFinal`) but is not checked before the `CDK_TOOLKIT_I5060` confirmation request.

## Changes

- Guard the `CDK_TOOLKIT_I5060` request in both `@aws-cdk/toolkit-lib` and the CLI's `cdk-toolkit` with `!prepareResult?.noOp`.
- Add a unit test in each package asserting that `CDK_TOOLKIT_I5060` is not issued when `prepareStack` returns `noOp: true`, even with `requireApproval=any-change`.
- Add an integration test that deploys a stack and then redeploys it with `--require-approval=any-change --method=change-set`, asserting the second run completes and the prompt text never appears in the output.

The indent was increased due to the new conditional in the toolkit-lib. Review with whitespace ignored: https://github.com/aws/aws-cdk-cli/pull/1402/changes?w=1

## Consequences

- No-op deploys with `--require-approval` enabled no longer prompt. The user is not asked to approve changes when there are no changes to apply.
- The no-op result continues to flow through the normal output/ARN/timing path introduced in #1387, so the "(no changes)" success message, stack outputs, `--outputs-file`, and stack ARN are still emitted.
- No message registry codes added; the fix is a guard on an existing path.

## Considerations

The original issue also suggests emitting a new `CDK_TOOLKIT_W5023` warn-level notification analogous to `W5021`/`W5022` — for example `"<stack>: stack has no changes."`. I've left this out because the existing `(no changes)` suffix on the `CDK_TOOLKIT_I5900` success message already conveys the same information, and a separate event felt like duplication.

Happy to add a dedicated `W5023` code if you'd like a distinct structured event for tooling to hook into — it's a purely additive change and wouldn't affect the outputs/ARN/timing path. Let me know and I'll extend this PR.

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/deploy/cdk-deploy-skips-approval-prompt-on-no-change-deploy.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/deploy/cdk-deploy-skips-approval-prompt-on-no-change-deploy.integtest.ts
@@ -1,0 +1,26 @@
+import { integTest, withDefaultFixture } from '../../../lib';
+
+integTest(
+  'deploy with --require-approval skips the approval prompt on a no-change deploy',
+  withDefaultFixture(async (fixture) => {
+    // First deploy — creates the stack.
+    await fixture.cdkDeploy('test-2');
+
+    // Second deploy — no changes. With --require-approval=any-change and no
+    // --yes, the CLI must *not* prompt for approval: there is nothing for the
+    // user to approve. If the bug regresses, this call will hang waiting for
+    // stdin and the test will time out.
+    const output = await fixture.cdkDeploy('test-2', {
+      options: ['--require-approval=any-change', '--method=change-set'],
+      neverRequireApproval: false,
+      modEnv: {
+        FORCE_COLOR: '0',
+      },
+    });
+
+    // The deploy completed and reported no changes — and, crucially, never
+    // asked the user to confirm "updates" that don't exist.
+    expect(output).toContain('(no changes)');
+    expect(output).not.toContain('Do you wish to deploy these changes');
+  }),
+);
--- a/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
@@ -714,37 +714,42 @@ export class Toolkit extends CloudAssemblySourceBuilder {
         })
         : undefined;
 
-      const formatter = new DiffFormatter({
-        templateInfo: {
-          oldTemplate: currentTemplate,
-          newTemplate: stack,
-          changeSet: prepareResult?.changeSet,
-        },
-      });
+      // Skip the approval prompt entirely when the prepared change set has no
+      // changes — there is nothing for the user to approve. Outputs, stack ARN,
+      // and timings are still emitted via the normal no-op deploy path below.
+      if (!prepareResult?.noOp) {
+        const formatter = new DiffFormatter({
+          templateInfo: {
+            oldTemplate: currentTemplate,
+            newTemplate: stack,
+            changeSet: prepareResult?.changeSet,
+          },
+        });
 
-      const securityDiff = formatter.formatSecurityDiff();
-      const stackDiff = formatter.formatStackDiff();
+        const securityDiff = formatter.formatSecurityDiff();
+        const stackDiff = formatter.formatStackDiff();
 
-      // Send a request response with the diff as part of the message,
-      // and the template diff as data
-      // (IoHost decides whether to print depending on permissionChangeType)
-      const hasSecurityChanges = securityDiff.permissionChangeType !== PermissionChangeType.NONE;
-      const deployMotivation = hasSecurityChanges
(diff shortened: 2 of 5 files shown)


mrgrain (approved):
Great PR, thank you! :shipit:

PR #1419, opened 2026-04-27
fix: empty diff with only hidden CDK metadata changes requests approval from user without explanation

orien:
## Context

When deploying with the changeset method, it's possible for the changeset to be non-empty (not a CloudFormation no-op) while the diff presented to the user shows nothing. This happens when the only template differences are `AWS::CDK::Metadata` resources or `CheckBootstrapVersion` rules — both of which are filtered by `obscureDiff()` unless `cdk diff --strict` is passed.

The user sees a blank "There were no differences" message, then is prompted to approve the deployment with no explanation of why; very confusing.

## Changes

In `DiffFormatter.formatStackDiffHelper()`, capture whether the diff was non-empty before `obscureDiff()` runs. If `obscureDiff()` filtered everything away, append a grey hint to the "no differences" message:

> There were no differences *(CDK metadata changes were hidden, run cdk diff --strict to show)*

This applies everywhere `formatStackDiff()` is called: `cdk diff`, the deploy approval prompt, and `cdk import`.

## Consequences

Users will have a clear signal explaining why the diff appears empty and how to see the full picture. The `--strict` flag is specific to `cdk diff` and already exists to show these changes; this change makes users aware of it when it's relevant.

---
_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license._

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/diff/cdk-cdk-diff-doesnt-show-resource-metadata-changes.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/diff/cdk-cdk-diff-doesnt-show-resource-metadata-changes.integtest.ts
@@ -14,8 +14,9 @@ integTest(
       },
     });
 
-    // Assert there are no changes
+    // Assert no visible changes, but hint indicates hidden metadata changes
     expect(diff).toContain('There were no differences');
+    expect(diff).toContain('CDK metadata changes were hidden, run cdk diff --strict to show');
   }),
 );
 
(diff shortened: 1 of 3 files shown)


mrgrain (approved):

mrgrain:
@orien thanks for the fix! there appears to be another integ test failing.

orien:
Thanks, I'll fix it up.

orien:
I've reverted the change to the metadata integration test, since it tests a different scenario.

An integration test will require adding a dedicated fixture stack to the app and deploying it to AWS, I'm not convinced this change is worth it.

PR #1467, opened 2026-05-01
chore(toolkit-lib): recommend to use drift aware change sets when deploying to Cloudformation after hotswapping

ShadowCat567:
Drift aware change sets take into account current resource state when making a changeset and will not make changes if the new template and the resource state align, otherwise will prefer changes from the new template. 
When recovering from hotswap deployments, drift aware changesets will do a better job of getting your Cloudformation state and your current resource state back into sync than a regular Cloudformation deployment (using changesets or update stack) would. Adds a message in the CLI that recommends to use drift aware changesets when deploying to Cloudformation after performing hotswap deployments. 

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
@@ -343,7 +343,7 @@ export async function deployStack(options: DeployStackOptions, ioHelper: IoHelpe
       deploymentMethod = deploymentMethod.fallback;
     } else {
       await ioHelper.defaults.info(format(
-        `Your next regular deployment with ${stackEnv.name} should use 'cdk deploy --revert-drift' to resolve the drift that was introduced while hotswapping.`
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -320,6 +320,11 @@ export async function deployStack(options: DeployStackOptions, ioHelper: IoHelpe
       );
 
       if (hotswapDeploymentResult) {
+        if (deploymentMethod.fallback) {
```
rix0rrr: This is true even if we never do the fallback ourselves.

As soon as we hotswap we have introduced drift, and we should print "next non-hotsweap deployment should include --revert-drift".

rix0rrr: If we do a fallback deployment WITHOUT `--revert-drift`, then that's an entirely different problem :D

rix0rrr (approved):

PR #1478, opened 2026-05-06
feat: CLI send in performance profile if emitted by app

rix0rrr:
If the CDK app emits a performance profile (a set of counters explaining where time was spent in the CDK app), the CLI will attach those to the SYNTH event.

The data itself depends on what the CDK app is going to emit (not fully written yet), but is going to look something like this (feature added [here](https://github.com/aws/aws-cdk/pull/37843)):

```js
{
  'phase:Construction': 12154,
  'phase:Construction(cnt)': 1,
  'bundle:NodejsFunction': 12133,
  'bundle:NodejsFunction(cnt)': 2,
  'DockerImage.fromBuild': 10957,
  'DockerImage.fromBuild(cnt)': 1,
  'phase:Load': 3940,
  'phase:Load(cnt)': 1,
  'AssetBundlingBindMount.run': 1176,
  'AssetBundlingBindMount.run(cnt)': 1,
  'phase:Synthesis': 26,
  'phase:Synthesis(cnt)': 1,
  'Stack.resolve': 7,
  'Stack.resolve(cnt)': 47,
  'FileSystem.fingerprint': 1,
  'FileSystem.fingerprint(cnt)': 1,
  'FileSystem.isEmpty': 0,
  'FileSystem.isEmpty(cnt)': 1
}
```

---


packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -669,6 +679,81 @@ function isFeatureEnabled(configuration: Configuration, featureFlag: string) {
...
+      return;
+    }
+
+    const choices = ['yes', 'no', 'always', 'never', 'view', 'y', 'n', 'a', 'r', 'v'] as const;
+    type Choice = (typeof choices)[number];
+
+    // Ask the user
+    let decision: boolean | undefined = undefined;
+    while (decision === undefined) {
+      const answer: Choice = await promptly.prompt(`🔍 ${chalk.cyan('Your CDK app produced a performance profile to help AWS diagnose slow synthesis. Send in for analysis?')} (Y)es (N)o (A)lways Neve(R) (V)iew`, {
```
otaviomacedo: Why not use `IoHelper` here?

packages/aws-cdk/lib/cli/telemetry/telemetry-prefs.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface TelemetryPrefsFile {
+  /**
+   * If the user has given an Always/Never response to sending performance counters
+   */
+  defaultSendPerfCounters?: boolean;
+
+  /**
+   * If present, will be sent in every event
+   */
+  permanentCounters?: Record<string, number>;
```
otaviomacedo: How is this supposed to be used?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -224,7 +230,11 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
   }
 
   try {
-    return await main(cmd, argv);
+    const ret = await main(cmd, argv);
+
+    await maybeCommitPerfCounters(ioHost);
```
otaviomacedo: Do we care about performance metrics only in case of success?

mrgrain (changes requested):
Main blocker for me is that when asking for "always" approve, adding to the data might be an issue. We either need to make this abundantly clear on the opt-in, or version the data schema and store the opt-in only per schema version.

Similarly, if we phrase this exclusively as performance data, we cannot reuse the same opt-in for e.g. error logs in future. This might be by design, but wanted to call this out.

packages/aws-cdk/lib/cli/util/fs-util.ts
```diff
@@ -0,0 +1,12 @@
+import * as fs from 'fs-extra';
```
mrgrain: nit: can we implement this without a dep?

packages/aws-cdk/lib/cli/telemetry/telemetry-prefs.ts
```diff
@@ -0,0 +1,33 @@
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import { cdkHomeDir } from '../../util';
+import { tryReadJson } from '../util/fs-util';
+
+const TELEMETRY_PREFS_PATH = path.join(cdkHomeDir(), 'telemetry-prefs.json');
```
mrgrain: why is this a new file and not just `cdk.json`?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -3,6 +3,7 @@ import * as cxapi from '@aws-cdk/cx-api';
 import type { ChangeSetDeployment, DeploymentMethod, DirectDeployment, StackSelector as LibStackSelector } from '@aws-cdk/toolkit-lib';
 import { ExpandStackSelection, StackSelectionStrategy, ToolkitError, Toolkit } from '@aws-cdk/toolkit-lib';
 import * as chalk from 'chalk';
+import * as promptly from 'promptly';
```
mrgrain: pls not. I spend some time to limit promptly to the IoHost...

packages/aws-cdk/lib/cli/telemetry/telemetry-prefs.ts
rix0rrr: Because it was per-machine and not per-project.

Removed now.

packages/aws-cdk/lib/cli/cli.ts
rix0rrr: Aaalllright :)

mrgrain (approved):
Looks good. Could make a note somewhere that `performance-counters.json` is an inherently temporary file.

PR #1506, opened 2026-05-13
feat(cli): add --region as a global CLI option

### LINKED ISSUE #928: (cli): (If default region not specified in CLI profile, accounts not in the Commercial partition (e.g. GovCloud) will not bootstrap --region does not work)
### Describe the bug

If you do not specify a region in your AWS CLI profile, and the account is not in the main AWS partition, (like an account in GovCloud), the AWS CDK CLI will fail, indicating it cannot access the AWS account.  Specifying the region as a CLI parameter --region <region-name> doesn't work; it always defaults to us-east-1.

### Regression Issue

None

### Expected Behavior

I expect the CLI to use the region specified in the command, if the CLI profile does not contain a region setting.  (us-east-1 _is_ a sensible default if no region is specified at all.)

### Current Behavior

`cdk bootstrap --profile abc_123 --region us-gov-west-1` [note that the region is being specified]
...
> [14:40:01] **_Setting "CDK_DEFAULT_REGION" environment variable to us-east-1_**
> [14:40:02] Looking up default account ID from STS
> [14:40:02] Unable to determine the default AWS account (InvalidClientTokenId): The security token included in the request is invalid
> [14:40:02] outdir: cdk.out
> [14:40:02] env: {
>   CDK_DEFAULT_REGION: 'us-east-1',
>   CDK_OUTDIR: 'cdk.out',
>   CDK_CLI_ASM_VERSION: '48.0.0',
>   CDK_CLI_VERSION: '2.1031.1'
> }
> Unable to resolve AWS account to use. It must be either configured when you define your CDK Stack, or through the environment

As you can see, it's ignoring the region specified in the cdk command.

Complete verbose output included.

[cdk.log](https://github.com/user-attachments/files/23344584/cdk.log)

### Reproduction Steps

Attempt a cdk bootstrap against a GovCloud account without specifying a region in the AWS CLI profile.  (cdk synth works fine (because that doesn't require talking to aws), I suspect a cdk deploy would also fail, though I have not tried it.)

cdk bootstrap --profile abc_123 --region us-gov-west-1

If you (person on the development team) do not have access to GovCloud, contact me at @pmescher, and I'd be happy to assist in further re-creates or fix testing.

### Possible Solution

At whatever step in the code that involves authenticating with AWS, use the profile specified in the --region CLI option, if present.  Also, if the command fails, specify the (attempted) credentials in the credentials error message.

_No response_

### CDK CLI Version

2.1031.1

### Framework Version

_No response_

### Node.js Version

22.19.0

### OS

MacOS Sequoia 15.7.1

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

sanjanaravikumar-az:
## Summary

- Adds `--region` as a global CLI option (like `--profile`), giving it highest priority in region resolution — overriding environment variables, profile config, and the `us-east-1` default fallback.

Closes #928

## Motivation

GovCloud/China partition users whose AWS profile doesn't specify a region get `InvalidClientTokenId` because the CLI defaults to `us-east-1` — credentials for non-commercial partitions can't authenticate against commercial endpoints. The `--region` flag was silently ignored because it wasn't a registered option.

## Test plan

- [x] `awscli-compatible.test.ts`: 4 new tests verifying region override takes priority over env vars, profile, and default
- [x] `check-unknown-options.test.ts`: 8 tests covering known options, aliases, camelCase variants, negation keys, and unknown detection
- [x] `sdk-provider.test.ts`: All 31 existing tests pass (signature change is backward-compatible)
- [x] `configuration.test.ts`: All 11 existing tests pass (settings flow works)
- [ ] Integration test with GovCloud profile (manual — need GovCloud credentials)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk-provider.ts
```diff
@@ -109,9 +109,9 @@ export class SdkProvider {
    * @param profile - Profile to read from ~/.aws
```
mrgrain, line 109: task: document region here

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -99,6 +100,12 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
   await ioHost.defaults.debug('CDK Toolkit CLI version:', versionWithBuild());
   await ioHost.defaults.debug('Command line arguments:', argv);
 
+  const unknownOptions = findUnknownOptions(argv);
+  if (unknownOptions.length > 0) {
+    const formatted = unknownOptions.map((o) => `--${o}`).join(', ');
+    await ioHost.defaults.warn(`Unknown option(s): ${formatted}. These will be ignored. Run 'cdk --help' to see available options.`);
+  }
```
mrgrain: suggestion (non blocking): this would absolutely be worth splitting into two separate PRs.

packages/aws-cdk/README.md
mrgrain, line 1: nit: this probably needs more explanation on how `profile` and `region` interact with each other. Not in `JSON Configuration files` section, but somewhere else.

mrgrain (changes requested):
small feedback. I also suggest you split this into two separate PRs but I won't block on it.

packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/sdk-provider.ts
sanjanaravikumar-az: done!

packages/aws-cdk/lib/cli/cli.ts
sanjanaravikumar-az: yes, i agree, have split them into separate prs

packages/aws-cdk/README.md
sanjanaravikumar-az: added a block

mrgrain (approved):
Looks good now. Please update the PR body to remove the warning feature before merging. Also, the list of changes is quite redundant. A git log provides the same information.

PR #1508, opened 2026-05-13
feat(toolkit-lib): validate command

kaizencc:
## Summary

Adds a `validate()` method to the `Toolkit` class in `@aws-cdk/toolkit-lib`. This command:
- Synthesizes the CDK app
- Reads the `policy-validation-report.json` from the cloud assembly output directory (written by aws-cdk-lib during synthesis)
- Returns a structured `ValidateResult` with overall pass/fail status and per-plugin reports

This is the toolkit-lib layer only — CLI wiring (`cdk validate`) will follow separately.

## Changes

- **New action types** (`lib/actions/validate/index.ts`): `ValidateOptions`, `ValidateResult` — uses schema types from `@aws-cdk/cloud-assembly-schema` (PR #1515) rather than redefining them
- **New `validate()` method** on the `Toolkit` class — reads the report JSON and casts to typed schema, no manual field mapping
- **IO messages**: `CDK_TOOLKIT_I9600` (passed), `CDK_TOOLKIT_E9600` (failed), `CDK_TOOLKIT_I9601` (no report found)
- **`ToolkitAction`**: added `validate` to the union type

## Dependencies

- [x] #1515 — schema types in `@aws-cdk/cloud-assembly-schema` (must merge first)
- [ ] Corresponding change in `aws-cdk-lib` to always write `policy-validation-report.json` to the assembly directory (currently gated behind a context flag)

## Test plan

- [x] Unit tests covering: success, failure, missing report, IO message emission, violation parsing, plugin version handling
- [ ] Integration test with a CDK app that has validation plugins configured (blocked on aws-cdk-lib change)

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
@@ -670,7 +670,7 @@ export class Toolkit extends CloudAssemblySourceBuilder {
         conclusion: 'success',
         pluginReports: [],
       };
-      await ioHelper.notify(IO.CDK_TOOLKIT_I9601.msg('No validation plugins configured. Add a plugin to your CDK app to enable policy validation.'));
+      await ioHelper.notify(IO.CDK_TOOLKIT_I9601.msg('No validation plugins configured. Add a plugin to your CDK app to enable validation.'));
       return result;
     }
 
@@ -687,9 +687,9 @@ export class Toolkit extends CloudAssemblySourceBuilder {
     };
 
     if (conclusion === 'failure') {
-      await ioHelper.notify(IO.CDK_TOOLKIT_E9600.msg('❌ Validation found policy violations', result));
+      await ioHelper.notify(IO.CDK_TOOLKIT_E9600.msg('❌ cdk validate found problems', result));
     } else {
-      await ioHelper.notify(IO.CDK_TOOLKIT_I9600.msg('✅ All policy checks passed', result));
+      await ioHelper.notify(IO.CDK_TOOLKIT_I9600.msg('✅ No problems found', result));
     }
 
     return result;
--- a/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
@@ -57,7 +57,7 @@ import type { PublishAssetsOptions, PublishAssetsResult } from '../actions/publi
 import type { RefactorOptions } from '../actions/refactor';
 import { type RollbackOptions } from '../actions/rollback';
 import { type SynthOptions } from '../actions/synth';
-import type { ValidateOptions, ValidateResult, PolicyValidationReportJson } from '../actions/validate';
+import type { ValidateOptions, ValidateResult, PolicyValidationReportJson, PolicyValidationReportStatus } from '../actions/validate';
 import type { IWatcher, WatchOptions } from '../actions/watch';
 import { countAssemblyResults } from './private/count-assembly-results';
 import { WATCH_EXCLUDE_DEFAULTS } from '../actions/watch/private';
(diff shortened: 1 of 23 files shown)


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -649,6 +652,55 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      status,
+      title: report.title,
+      pluginReports: report.pluginReports,
+    };
+
+    if (status === 'failure') {
+      await ioHelper.notify(IO.CDK_TOOLKIT_E9600.msg('Policy validation failed', result));
+    } else {
+      await ioHelper.notify(IO.CDK_TOOLKIT_I9600.msg('Policy validation passed', result));
+    }
```
rix0rrr: Can we pick slightly less mechanical language here?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -649,6 +652,55 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+   *
+   * Synthesizes the CDK app and reads the policy validation report
+   * from the cloud assembly output directory.
+   */
+  public async validate(cx: ICloudAssemblySource, options: ValidateOptions = {}): Promise<ValidateResult> {
+    const ioHelper = asIoHelper(this.ioHost, 'validate');
+    const selectStacks = stacksOpt(options);
+    await using assembly = await synthAndMeasure(ioHelper, cx, selectStacks);
+
+    const reportPath = path.join(assembly.directory, POLICY_VALIDATION_REPORT_FILE);
```
rix0rrr, line 666: Should be an artifact in the manifest, no?

Although I can live with this for a bit.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -649,6 +652,55 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+
+    if (!await fs.pathExists(reportPath)) {
+      const result: ValidateResult = {
+        status: 'success',
+        pluginReports: [],
+      };
+      await ioHelper.notify(IO.CDK_TOOLKIT_I9601.msg('No policy validation report found'));
```
rix0rrr: This is very mechanical. What does this mean to me as a user?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -649,6 +652,55 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+      return result;
+    }
+
+    const reportJson = await fs.readJson(reportPath);
+
+    if (!Array.isArray(reportJson.pluginReports)) {
```
rix0rrr: Can we do a JSON schema validation so we don't have to bother with this?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
rix0rrr: "policy policy policy"

Let's speak like a human.

"cdk validate found problems" (or something to that effect), "no problems found", let's not pretend to be IBM and we all wear ties in triplicate.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
kaizencc: https://github.com/orgs/aws/projects/178/views/44?pane=issue&itemId=189647151

i tracked it and will do it separately

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
kaizencc: Right now it means no plugins / no construct annotations / default rule set is turned off. 

I could be convinced for this to not be possible and we always write a report no matter what

kaizencc: there's no default rule set right now so the message wont mention that for now

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
kaizencc: done

rix0rrr (approved):

PR #1512, opened 2026-05-14
refactor: extract zip utilities into shared `@aws-cdk/tools` package

mrgrain:
Both `cdk-assets-lib` and `toolkit-lib` carried their own copy of the zip utilities (`zipDirectory`, `zipString`) wrapping `archiver` and `fast-glob`. This duplication made it easy for the implementations to drift apart — the `toolkit-lib` version had already lost the deterministic date reset, for example.

This PR introduces a new private (unpublished) `@aws-cdk/tools` package that hosts small, self-contained utilities. Each tool lives in its own subdirectory under `lib/`, is type-checked by `tsc`, and bundled by `esbuild` with `--packages=external` — so the bundled output contains **only the code we wrote**, while third-party packages stay as plain `require(...)` calls.

Consumer packages cherry-pick tools at pre-compile time via a `useTools` option on the new `CdkTypeScriptWorkspace` projen construct, using typed tool references (e.g. `useTools: [tools.zip]`). For each tool it:

- copies the bundled `.js`/`.d.ts` into `lib/private/tools/<name>/`. These files are committed, so `tsc --build` of dependent projects (e.g. the CLI) can resolve them without first running this package's pre-compile step.
- adds the tool's declared runtime dependencies (e.g. `archiver`, `fast-glob`) to the consuming package.

Keeping those dependencies **external** to the bundle (rather than inlining them) is deliberate. The `aws-cdk` CLI bundles `cdk-assets-lib` and `toolkit-lib` together; if each shipped its own inlined copy of `archiver` (and its `lodash`/`fast-glob` deps), the CLI bundle would carry the implementation multiple times. Declaring the deps and keeping them external lets the bundler resolve them to a single shared copy, matching the pre-refactor footprint.

Adding a new tool is just: create `lib/<tool>/index.ts` in the tools package, declare its third-party deps in `toolDeps`, and `useTools: [tools.<tool>]` in each consumer.

### Checklist

---


.projenrc.ts
```diff
@@ -695,23 +696,52 @@ const cliPluginContract = configureProject(
...
+});
+configureProject(tools);
+
 // #endregion
 //////////////////////////////////////////////////////////////////////
 // #region @aws-cdk/cdk-assets-lib
 
 const cdkAssetsLib = configureProject(
-  new yarn.TypeScriptWorkspace({
+  new CdkTypeScriptWorkspace({
```
rix0rrr: See this is the thing that is bothersome about inheritance.

Can this not be written as 

```ts
new ToolUse(cdkAssetsLib, {
  tools: [tools.zip],
}));
```

Thereby foregoing the new projen class?

mrgrain: probably can be a mixin to be honest!

rix0rrr (approved):

PR #1514, opened 2026-05-14
feat(cli): warn on unrecognized CLI options

sanjanaravikumar-az:
## Summary

- Adds a warning when unrecognized CLI options are passed, since yargs currently silently ignores them
- Handles edge cases: env-injected keys (CDK_* via `.env('CDK')`), undefined defaults, camelCase variants, boolean negations, negativeAliases

This implements a non-breaking warning approach. Unrecognized options are still processed by yargs as before, but users now get feedback that their flag was not recognized.

## Changes

| File | Change |
|------|--------|
| `check-unknown-options.ts` | New utility that detects unrecognized options |
| `cli.ts` | Calls `findUnknownOptions` and emits a warning |
| `check-unknown-options.test.ts` | 10 tests covering all edge cases |

## Test plan

- [x] 10 unit tests passing (known options, aliases, camelCase, negation, env-injected, unknown detection)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/aws-cdk/test/cli/util/check-unknown-options.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    const argv = {
+      _: ['deploy'],
+      $0: 'cdk',
+      R: true,
+      rollback: false,
+    };
+    expect(findUnknownOptions(argv)).toEqual([]);
+  });
+
+  test('does not report keys injected by yargs .env("CDK") from environment variables', () => {
```
mrgrain, line 79: but why?

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Detect unrecognized CLI options and emit warnings.
+ *
+ * Yargs does not enable strict option checking by default, so unknown flags
+ * like `--region` (before it was added) are silently swallowed. This function
+ * compares the parsed argv keys against the known global and command options
+ * from the CLI type registry and warns for any that don't match.
+ */
+export function findUnknownOptions(argv: any): string[] {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const config = require('../cli-type-registry.json');
```
mrgrain: nit: this doesn't change, so can be outside the function as a "global" variable. In practice this doesn't matter much because `findUnknownOptions` is only called once and `require` is cached, but you should start to thinking about these kind of things.

The same applies to much of the "setup" code later on: sorting global and command options etc.

A good question to ask is: Which parts of the code are actually dependent on the `argv` input? Which parts only depend on the static `cli-type-registry.json`? This will give you an idea how to split it.

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (commandArg) {
+    yargsInternals.add(commandArg);
+  }
+
+  const unknown: string[] = [];
+  for (const key of Object.keys(argv)) {
+    if (argv[key] === undefined) continue;
+    if (yargsInternals.has(key)) continue;
+    if (globalOptions.has(key)) continue;
+    if (commandOptions.has(key)) continue;
```
mrgrain: can be one `if`

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
```diff
@@ -0,0 +1,92 @@
...
+    // yargs creates camelCase versions of kebab-case options — skip those
+    const kebab = camelToKebab(key);
+    if (kebab !== key && (globalOptions.has(kebab) || commandOptions.has(kebab))) continue;
+
+    // yargs creates "noFoo" keys for --no-foo boolean negations — skip those
+    if (key.startsWith('no') && key.length > 2 && key[2] === key[2].toUpperCase()) {
+      const positiveKey = key[2].toLowerCase() + key.slice(3);
+      const positiveKebab = camelToKebab(positiveKey);
+      if (globalOptions.has(positiveKey) || commandOptions.has(positiveKey) ||
+          globalOptions.has(positiveKebab) || commandOptions.has(positiveKebab)) continue;
```
mrgrain: we should enforce this as a linter rule, but imho these lines are hard to parse and read. Prefer always using `{}`.

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    unknown.push(key);
+  }
+
+  return unknown;
+}
+
+function camelToKebab(str: string): string {
+  return str.replace(/[A-Z]/g, (m) => `-${m.toLowerCase()}`);
+}
```
mrgrain: This probably happens to work in this case, but is likely not working in the general case. For your code, it would be much easier to go the other way: From kebab to camel.

mrgrain (dismissed):
small feedback.

packages/aws-cdk/test/cli/util/check-unknown-options.test.ts
sanjanaravikumar-az: I dug into it and the test was actually wrong. I ran the actual CDK parser and `yargs` doesn't create `noFoo `keys from `--no-foo` flags, it just sets `foo: false`. The only way a `noFoo` key ends up in argv is if someone sets `CDK_NO_FOO` as an env var (via .`env('CDK')`), which the env prefix check already covers. Removed.

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
sanjanaravikumar-az: That's a good point, thank you! Moved the registry load and global options set to module scope since they never change. Only the command-specific options are computed inside the function now since those depend on which command was invoked.

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
sanjanaravikumar-az: I actually removed this entirely since it's the same issue as the first comment. The negation check was solving a problem that doesn't exist with how yargs works.

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
sanjanaravikumar-az: got it, changed it

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
sanjanaravikumar-az: ah okay, switched it

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
```diff
@@ -0,0 +1,86 @@
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const config = require('../cli-type-registry.json');
+
+/**
+ * Build a set of all known option names for a given option definitions object.
+ * Includes the kebab-case name, its camelCase equivalent, all aliases, and negativeAliases.
+ */
+function collectKnownOptions(optionDefs: Record<string, any>): Set<string> {
```
iankhou, line 12: `config.globalOptions` that is passed in contains strings with kebab-case. However, the function signature and JSDoc comments don't indicate that the input must be kebab-case to work correctly. There should be some indication of how to use the function correctly, even though it's not public.

packages/aws-cdk/test/cli/util/check-unknown-options.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  test('detects unknown options', () => {
+    const argv = {
+      _: ['bootstrap'],
+      $0: 'cdk',
+      profile: 'my-profile',
+      fakeOption: 'value',
+    };
+    const unknown = findUnknownOptions(argv);
+    expect(unknown).toContain('fakeOption');
```
iankhou: We can be stricter with this check. It shouldn't just contain `'fakeOption'`, it should ONLY contain `'fakeOption'`.

packages/aws-cdk/test/cli/util/check-unknown-options.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    try {
+      const argv = {
+        _: ['deploy'],
+        $0: 'cdk',
+        integAtmospherePool: 'test-pool',
+        totallyFakeOption: 'value',
+      };
+      const unknown = findUnknownOptions(argv);
+      expect(unknown).not.toContain('integAtmospherePool');
+      expect(unknown).toContain('totallyFakeOption');
```
iankhou: For this test, I would recommend comparing unknown to an exact object.

iankhou (changes requested):

packages/aws-cdk/lib/cli/util/check-unknown-options.ts
sanjanaravikumar-az: added!

packages/aws-cdk/test/cli/util/check-unknown-options.test.ts
sanjanaravikumar-az: got it, changed to `expect(unknown).toEqual(['fakeOption']);`

packages/aws-cdk/test/cli/util/check-unknown-options.test.ts
sanjanaravikumar-az: same change, updated to `expect(unknown).toEqual(['totallyFakeOption']);`

iankhou (approved):

PR #1515, opened 2026-05-14
feat(cloud-assembly-schema): add policy validation report schema types

kaizencc:
Define the JSON schema types for `policy-validation-report.json` in the shared cloud-assembly-schema package. This file is written by aws-cdk-lib during synthesis and consumed by toolkit-lib's validate command. Having the types in the shared contract prevents drift between producer and consumer.

## Corresponding types in aws-cdk-lib

These schema types mirror the structures defined in aws-cdk-lib:
- [`packages/aws-cdk-lib/core/lib/validation/validation.ts`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/validation/validation.ts) — `IPolicyValidationPluginBeta1`, `PolicyViolationBeta1`, `PolicyViolatingResourceBeta1`
- [`packages/aws-cdk-lib/core/lib/validation/private/report.ts`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/validation/private/report.ts) — report formatting and JSON serialization

## Schema sufficiency for RFC 899

We evaluated whether schema changes are needed to support [CDK Comprehensive Validation (RFC 899)](https://github.com/aws/aws-cdk-rfcs/pull/899) and its proposed output format. The existing schema is sufficient — all data required by the RFC is either already present or derivable at display time:

| RFC requirement | How it's satisfied |
|---|---|
| **Severity grouping** (Fatal/Error/Warning) | `PolicyViolationJson.severity` is already present. CLI groups violations by severity at display time. Defaults to Warning if absent. |
| **Source name** (e.g. "ValidationEngine (Default)", "CdkNagValidator") | Derivable from `PluginReportJson.summary.pluginName` |
| **Suppression/acknowledge ID** | Derivable as `${pluginName}::${ruleName}` (both fields present). Spaces replaced with dashes per [aws-cdk#37808](https://github.com/aws/aws-cdk/pull/37808). |
| **Suppress instruction** | Computed by CLI: `Acknowledge '${pluginName}::${ruleName}'`. Omitted for Fatal violations. |
| **Summary counts** | Computed from violation arrays at render time |
| **Source file location** | `ViolatingConstructJson.constructStack.location` |
| **Construct path** | `ViolatingConstructJson.constructPath` |

For Construct Annotations specifically, aws-cdk-lib (via [#37808](https://github.com/aws/aws-cdk/pull/37808)) converts annotations into `PolicyViolation` objects before writing the report — extracting the `[ack: id]` tag as `ruleName` and mapping the level to severity. By the time the CLI reads the report, it's already structured.

No additional schema fields are needed. The RFC output is purely a CLI display concern.

---


packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,217 @@
...
+ */
+
+/**
+ * The top-level structure of the policy validation report file.
+ */
+export interface PolicyValidationReportJson {
+  /**
+   * Report title.
+   */
+  readonly title: string;
```
rix0rrr: Required?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,217 @@
...
+   * @default - no fix provided
+   */
+  readonly fix?: string;
+
+  /**
+   * The severity of the violation.
+   *
+   * @default - no severity
+   */
+  readonly severity?: string;
```
rix0rrr: No enumeration here?

That means there will not be an interpretation to any of the values here? No colorization? No sorting?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * Resources that violated the rule.
+   */
+  readonly violatingResources: ViolatingResourceJson[];
+
+  /**
+   * Constructs that violated the rule.
+   */
+  readonly violatingConstructs: ViolatingConstructJson[];
```
rix0rrr: Does that mean violating resources will never have a construct path?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
kaizencc: this will all have to be done in a bespoke way, as in we'll have to match `error` with a color things we don't match will have a default `warning` value (or `info`)

otherwise we can't fit existing plugins into the system

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
kaizencc: https://github.com/aws/aws-cdk/blob/0d088ca0a6e720c25132d177e675ecfe08caff2f/packages/aws-cdk-lib/core/lib/validation/private/report.ts#L213-L224

```ts
            violatingResources: violation.violatingResources,
            violatingConstructs: violation.violatingResources.map(resource => {
              // Use constructPath from the input if provided (e.g. annotations),
              // otherwise derive it from the logical ID via the construct tree.
              const constructPath = resource.constructPath ?? (
                resource.templatePath && resource.resourceLogicalId
                  ? this.tree.getConstructByLogicalId(
                    path.basename(resource.templatePath),
                    resource.resourceLogicalId,
                  )?.node.path
                  : undefined
              );
```

right now the report maps resources to constructs so they will have construct path if avail

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
kaizencc: It was previously required, I don't think it's that problematic though

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
rix0rrr: You could have `warning | error | info | custom` and use `custom` for the enum levels of plugins if they're weird. The naked `string` is making me uncomfortable.

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
rix0rrr: What's the use though? Why not combine these?

kaizencc:
```ts
export interface ValidationViolatingConstruct extends report.PolicyViolatingResource {
  /**
   * The construct path as defined in the application.
   *
   * @default - construct path will be empty if the cli is not run with `--debug`
   */
  readonly constructPath?: string;

  /**
   * A stack of constructs that lead to the violation.
   *
   * @default - stack will be empty if the cli is not run with `--debug`
   */
  readonly constructStack?: ConstructTrace;
}
```

they _should_ be combined in the schema, but changing this breaks the existing schema and I'd rather not unless we have to. my thoughts are that we already have some sort of report, and if it can be repurposed into what we need, even if there's some redundancy, that's better than creating a new-but-slightly-different report that we then have to explain migration path or hide behind a feature flag.

kaizencc: eh, i'll remove it

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+/**
+ * The severity of a policy violation.
+ *
+ * If you need to use a severity level that doesn't exist as a static member,
+ * use `PolicyViolationSeverity.custom('critical')`.
+ */
+export class PolicyViolationSeverity {
+  static readonly ERROR = new PolicyViolationSeverity('error');
```
rix0rrr: This is a schema library. We have no need of values, and I have a suspicion they will hinder more than they will help.

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * The construct creation stack trace.
+   *
+   * @default - no stack trace
+   */
+  readonly constructStack?: ConstructTraceJson;
+
+  /**
+   * Locations within the construct where the violation was detected.
```
rix0rrr: What is the formalism here? Property names? JSON paths? What?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+   *
+   * @default - no locations
+   */
+  readonly locations?: string[];
+
+  /**
+   * The logical ID of the resource in the CloudFormation template.
+   */
+  readonly resourceLogicalId: string;
```
rix0rrr: Needs to be optional if this is about constructs.

(In fact: how do we have optional construct PATHS, but definitely logical IDs?)

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * The path to the CloudFormation template containing this resource.
+   */
+  readonly templatePath: string;
```
rix0rrr: Same here, this cannot be required.

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * A construct that violated a policy rule.
+ */
+export interface ViolatingConstructJson {
+  /**
+   * The construct path as defined in the application.
+   *
+   * @default - no construct path
+   */
+  readonly constructPath?: string;
```
rix0rrr: Why would we ever not have a construct path?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+/**
+ * A node in the construct creation stack trace.
+ */
+export interface ConstructTraceJson {
+  /**
+   * The construct ID.
+   */
+  readonly id: string;
```
rix0rrr: ID is unnecessary, it's the last part of teh path.

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+   * @default - this is the leaf node
+   */
+  readonly child?: ConstructTraceJson;
+
+  /**
+   * The fully qualified name of the construct class.
+   *
+   * @default - no construct info
+   */
+  readonly construct?: string;
```
rix0rrr:
```suggestion
  readonly classFqn?: string;
```

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * The version of the library that contains this construct.
```
rix0rrr: Library name and version? Or just version?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+   *
+   * @default - no version info
+   */
+  readonly libraryVersion?: string;
+
+  /**
+   * The source code location where this construct was created.
```
rix0rrr: Do you mean stack trace?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
rix0rrr: But what is the "single title" of the entire report? Isn't it always going to be a constant?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+  /**
+   * Reports from all validation plugins that ran during synthesis.
+   */
+  readonly pluginReports: PluginReportJson[];
+}
+
+/**
+ * A report from a single validation plugin.
+ */
+export interface PluginReportJson {
```
rix0rrr, line 31: Should we not report on suppressions in here as well? `cdk-nag` needs that, doesn't it?

packages/@aws-cdk/cloud-assembly-schema/lib/cloud-assembly/validation-report-schema.ts
```diff
@@ -0,0 +1,212 @@
...
+export interface ViolatingConstructJson {
+  /**
+   * The construct path as defined in the application.
+   *
+   * @default - no construct path
+   */
+  readonly constructPath?: string;
+
+  /**
+   * The construct creation stack trace.
```
rix0rrr: There is a lot of data in that struct that is not a stack trace though.

rix0rrr (approved):

packages/@aws-cdk/cloud-assembly-schema/lib/manifest.ts
```diff
@@ -188,7 +200,7 @@ export abstract class Manifest {
     manifest: any,
     schema: jsonschema.Schema,
     options?: LoadManifestOptions,
-  ): asserts manifest is assembly.AssemblyManifest {
+  ): void {
```
kaizencc, line 203: `validate()` is called in `saveManifest` and `loadManifest` for its side effect (throwing on invalid input). It happens to work for the other loaders because `AssetManifest` and `IntegManifest` both have `version: string`, making them structurally compatible with `AssemblyManifest`. We can't get it to work for `PolicyValidationReportJson` because it has no version field. The asserts narrowing isn't meant to be there — `validate()` validates against arbitrary schemas, not specifically `AssemblyManifest`

kaizencc: Subsequent research shows me that we should actually have a `version: string` property if we want the policy report as part of the manifest. So I've added it in, and now this line would work as is. However, it's still wrong for many of the functions in `Manifest`, and just happens to work because they all share `version: string`. So I'm keeping this change

PR #1531, opened 2026-05-19
fix(toolkit-lib): deploy does not report nested stack errors

### LINKED ISSUE #635: (deploy) Deploy error reporting is not showing the reason of failure when using nested stacks.
main/root stack is not properly reporting errors from nested stacks, there is no way to know why it actually failed because the error is generalized that some resources could not be created, but why, there is no reason. The only chance to see the error is to constantly monitor the deployment on AWS CloudFormation Console, and once it starts to roll back and delete nested stacks you can quickly see why the nested stack failed in events tab.

E.g.: 
```
creating CloudFormation changeset...
 0/3 | 2:07:56 PM | CREATE_IN_PROGRESS   | AWS::CDK::Metadata         | CDKMetadata
 0/3 | 2:07:56 PM | CREATE_IN_PROGRESS   | AWS::CloudFormation::Stack | appsync-bizon.NestedStack/appsync-bizon.NestedStackResource (appsyncbizonNestedStackappsyncbizonNestedStackResource814A53CF)
 0/3 | 2:07:57 PM | CREATE_IN_PROGRESS   | AWS::CloudFormation::Stack | appsync-bizon.NestedStack/appsync-bizon.NestedStackResource (appsyncbizonNestedStackappsyncbizonNestedStackResource814A53CF) Resource creation Initiated
 0/3 | 2:07:58 PM | CREATE_IN_PROGRESS   | AWS::CDK::Metadata         | CDKMetadata Resource creation Initiated
 1/3 | 2:07:58 PM | CREATE_COMPLETE      | AWS::CDK::Metadata         | CDKMetadata
 2/3 | 2:08:19 PM | CREATE_FAILED        | AWS::CloudFormation::Stack | appsync-bizon.NestedStack/appsync-bizon.NestedStackResource (appsyncbizonNestedStackappsyncbizonNestedStackResource814A53CF) Embedded stack arn:aws:cloudformation:eu-west-2:452364846299:stack/ProductName-bizon-appsyncbizonNestedStackappsyncbizonNestedStackResource814A53CF-4NAE12ZLOQGT/a92a4b40-3c46-11ea-9fec-06452e71541c was not successfully created: The following resource(s) failed to create: [dynamodbstackbizonNestedStackdynamodbstackbizonNestedStackResource0AACEAF0, productnameresolversNestedStackproductnameresolversNestedStackResource42DF9B6A].
```
as you can see there is no reason why it failed, just saying that some resources failed to create. While actually the error on the nested stack events tab is:

E.g.:
```
Service role arn not specified (Service: AWSAppSync; Status Code: 400; Error Code: BadRequestException; Request ID: 5a9b00be-f21c-4cea-97ad-3d0e8552c5e5)
```
or 
```
Only one resolver is allowed per field. (Service: AWSAppSync; Status Code: 400; Error Code: BadRequestException; Request ID: 39813c06-7d28-4d72-be32-f05785f4338e)
  QueryListEvents, WikiArticleLastRevision
```
or
```
Property validation failure: [Value for property {/Name} does not match pattern {[_A-Za-z][_0-9A-Za-z]*}]
```

So there is no way to know about those errors because on roll back those stacks are removed, and developer will never know why it failed.

### Reproduction Steps

Just add a nested stack and try to simulate the above error examples:
1. Create AppSync nested stack but don't create service roll for it, and try to create another resource which should rely on that service role.
2. Name a resource in a nested stack improperly with `-` or `.`
3. Create 2 same resolver in one nested stack.
etc ... basically any error that will be thrown inside nested stack

### Error Log

already specified above

### Environment

  - **CLI Version      : 1.21.0**
  - **Framework Version:**
  - **OS               : Windows**
  - **Language         : Typescript**

### Other

I found this issue opened as a feature aws/aws-cdk#4489 but I don't think it is a feature, rather it is a bug, since the feature of having Nested Stacks exists, this means that error reporting for nested stack to be chained in the parent wasn't.

--- 

This is :bug: Bug Report

mrgrain:
Fixes #635

Re-enables `IncludeNestedStacks: true` on the deploy changeset creation path. This was originally added in #1292 but reverted in #1307 due to a CloudFormation server-side bug where intrinsic functions (like `Fn::Join`) in export names caused false "duplicate Export names" errors. That server-side issue has now been resolved.

With `IncludeNestedStacks` enabled, CloudFormation validates nested stack templates together during changeset creation. This means deployment errors in nested stacks are properly reported to the user, instead of the generic "The following resource(s) failed to create" message. It also allows CloudFormation to accurately detect "no changes" for nested stacks, avoiding unnecessary changeset executions.

The diff changeset path is simplified to always pass `includeNestedStacks: true` unconditionally (previously it was conditional on `templateContainsNestedStacks`). The `createChangeSetAndCleanup` function is un-exported since it was only exported for testing — the test is moved up a layer to test through the deploy flow instead.

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/test/api/deployments/cloudformation-deployments.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/deployments/cloudformation-deployments.test.ts
@@ -1,4 +1,3 @@
-import * as cxapi from '@aws-cdk/cloud-assembly-api';
 import {
   ContinueUpdateRollbackCommand,
   DescribeStackEventsCommand,
@@ -7,16 +6,11 @@ import {
   RollbackStackCommand,
   type StackResourceSummary,
   StackStatus,
-  DescribeChangeSetCommand,
-  ChangeSetStatus,
-  CreateChangeSetCommand,
 } from '@aws-sdk/client-cloudformation';
 import { GetParameterCommand } from '@aws-sdk/client-ssm';
 import { CloudFormationStack } from '../../../lib/api/cloudformation';
 import { Deployments } from '../../../lib/api/deployments';
-import * as cfnApi from '../../../lib/api/deployments/cfn-api';
 import { deployStack } from '../../../lib/api/deployments/deploy-stack';
-import { CloudFormationStackDiagnoser } from '../../../lib/api/diagnosing/stack-diagnoser';
 import { ToolkitInfo } from '../../../lib/api/toolkit-info';
 import { testStack } from '../../_helpers/assembly';
 import {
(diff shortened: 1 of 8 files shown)


packages/@aws-cdk/toolkit-lib/test/api/deployments/cloudformation-deployments.test.ts
```diff
@@ -1194,49 +1188,6 @@ describe('stackExists', () => {
   });
 });
 
-test('tags are passed along to create change set', async () => {
```
mrgrain, line 1197: The function under test (`createChangeSetAndCleanup`) was only exported to test it here. This makes the code less clear and harder to refactor. 

Removed the test in favor of adding a new assertion in `deploy-stack.test.ts`.

otaviomacedo (approved):

PR #1557, opened 2026-05-27
feat(cli): pass `CDK_CLI_USERAGENT` env var through to telemetry config

sarayev:
## What

Adds support for the `CDK_CLI_USERAGENT` environment variable, recording it in `command.config.cdkCliUserAgent`.

## Why

Allows wrapper tools (like AWS Blocks) to identify themselves in telemetry data for traffic attribution.

## Format

Only values matching `aws-blocks/{version}/{mode}` (where mode is `sandbox` or `production`) are recorded. Invalid values are silently ignored.

## Encoding

Uses the existing ConfigEntry pattern (`{ "aws-blocks/1.2.3/sandbox": true }`) for the `cdkCliUserAgent` config key.

---


packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -76,6 +76,7 @@ export class TelemetrySession {
           parameters,
           config: {
             context: sanitizeContext(this.props.context),
+            ...(process.env.CDK_CLI_WRAPPER ? { wrapper: { [process.env.CDK_CLI_WRAPPER]: true } } : {}),
```
mrgrain: I think we would want this as `CDK_CLI_USERAGENT` under `identifiers`

sarayev: After internal discussions, we agreed with @kaizencc that, `event.command.config` is the best place for this.

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -240,3 +243,17 @@ function isAbortedError(error?: ErrorDetails) {
...
 }
+
+/**
+ * Validates that the CDK_CLI_USERAGENT env var value matches
+ * the expected format: `aws-blocks/<version>/<mode>` where mode is
+ * either `sandbox` or `production`.
+ */
+export function isValidWrapperUserAgent(value: string | undefined): value is string {
+  if (!value) return false;
+  if (!value.startsWith('aws-blocks/')) return false;
```
kaizencc: can we extract this out to a `VALID_USER_AGENT` constant with the idea that we are creating a mechanism to report user agents we control (right now `aws-blocks` is the only one).

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -240,3 +249,17 @@ function isAbortedError(error?: ErrorDetails) {
...
+ * Validates that the CDK_CLI_USERAGENT env var value matches
+ * the expected format: `<name>/<version>/<mode>` where name is one of
+ * VALID_USER_AGENTS and mode is either `sandbox` or `production`.
+ */
+export function isValidWrapperUserAgent(value: string | undefined): value is string {
+  if (!value) return false;
+  const parts = value.split('/');
+  if (parts.length !== 3) return false;
+  const [name, _version, mode] = parts;
+  if (!VALID_USER_AGENTS.includes(name as any)) return false;
```
kaizencc:
```suggestion
  if (!VALID_USER_AGENTS.includes(name)) return false;
```

packages/aws-cdk/lib/cli/telemetry/session.ts
```diff
@@ -17,6 +17,12 @@ import { USER_INTERRUPTED_CODE } from './error';
 
 const ABORTED_ERROR_MESSAGE = '__CDK-Toolkit__Aborted';
 
+/**
+ * Valid user agent prefixes that are allowed to report through CDK_CLI_USERAGENT.
+ * This creates a mechanism to report user agents we control.
+ */
+const VALID_USER_AGENTS = ['aws-blocks'] as const;
```
kaizencc:
```suggestion
const VALID_USER_AGENTS = ['aws-blocks'];
```

kaizencc (approved):

PR #1561, opened 2026-05-27
feat: synth daemon

megha-narayanan:
Long-running daemon which coordinates synth runs for LSP servers 

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
@@ -25,7 +25,10 @@ export async function connectToDaemon(projectDir: string): Promise<DaemonConnect
 
 async function connectSocket(socketPath: string): Promise<net.Socket> {
   return new Promise((resolve, reject) => {
-    const socket = net.connect(socketPath, () => resolve(socket));
+    const socket = net.connect(socketPath, () => {
+      socket.removeListener('error', reject);
+      resolve(socket);
+    });
     socket.on('error', reject);
   });
 }
--- a/packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
@@ -1,0 +1,160 @@
+import * as net from 'net';
+import { ClientMessage, DaemonMessage, PROTOCOL_VERSION } from '../protocol';
+import { LineParser } from './line-parser';
+import { socketPathForProject } from './socket-path';
+
+export interface DaemonConnection {
+  readonly messages: AsyncIterable<DaemonMessage>;
+  send(msg: ClientMessage): void;
+  close(): void;
+}
+
+export async function connectToDaemon(projectDir: string): Promise<DaemonConnection> {
+  const socketPath = socketPathForProject(projectDir);
+  const socket = await connectSocket(socketPath);
+
+  const { accepted, leftover } = await performHandshake(socket);
+  if (!accepted) {
+    socket.destroy();
+    throw new Error('Daemon rejected handshake: protocol version mismatch');
+  }
+
+  return createConnection(socket, leftover);
+}
+
+async function connectSocket(socketPath: string): Promise<net.Socket> {
+  return new Promise((resolve, reject) => {
+    const socket = net.connect(socketPath, () => resolve(socket));
+    socket.on('error', reject);
+  });
+}
+
+interface HandshakeResult {
+  accepted: boolean;
+  leftover: string;
+}
+
+const HANDSHAKE_TIMEOUT_MS = 5000;
+
+async function performHandshake(socket: net.Socket): Promise<HandshakeResult> {
+  socket.write(JSON.stringify({ type: 'handshake', version: PROTOCOL_VERSION }) + '\n');
+
+  return new Promise((resolve, reject) => {
+    let buffer = '';
+    const timer = setTimeout(() => {
+      cleanup();
+      reject(new Error('Handshake timed out'));
+    }, HANDSHAKE_TIMEOUT_MS);
+
+    const cleanup = () => {
+      clearTimeout(timer);
+      socket.removeListener('data', onData);
+      socket.removeListener('error', onError);
+    };
+
+    const onData = (data: Buffer) => {
+      buffer += data.toString();
+      const newlineIndex = buffer.indexOf('\n');
+      if (newlineIndex !== -1) {
+        cleanup();
+        const line = buffer.slice(0, newlineIndex);
+        const leftover = buffer.slice(newlineIndex + 1);
+        let msg: DaemonMessage;
+        try {
+          msg = JSON.parse(line) as DaemonMessage;
+        } catch {
+          reject(new Error('Invalid JSON in handshake response'));
+          return;
+        }
+        if (msg.type === 'handshakeAck') {
+          resolve({ accepted: msg.accepted, leftover });
+        } else {
+          reject(new Error(`Expected handshakeAck, got ${msg.type}`));
+        }
+      }
+    };
+    const onError = (err: Error) => {
+      cleanup();
+      reject(err);
+    };
+    socket.on('data', onData);
+    socket.on('error', onError);
+  });
+}
+
+function createConnection(socket: net.Socket, leftover: string): DaemonConnection {
+  return {
+    messages: createMessageIterable(socket, leftover),
+    send(msg: ClientMessage): void {
+      if (!socket.destroyed) {
+        socket.write(JSON.stringify(msg) + '\n');
+      }
+    },
+    close(): void {
+      socket.destroy();
+    },
+  };
+}
+
+function createMessageIterable(socket: net.Socket, leftover: string): AsyncIterable<DaemonMessage> {
+  return {
+    [Symbol.asyncIterator](): AsyncIterableIterator<DaemonMessage> {
+      const parser = new LineParser<DaemonMessage>();
+      const queue: DaemonMessage[] = [];
+      let resolve: ((value: IteratorResult<DaemonMessage>) => void) | undefined;
+      let done = false;
+
+      const enqueue = (messages: DaemonMessage[]) => {
+        for (const msg of messages) {
+          if (resolve) {
+            const r = resolve;
+            resolve = undefined;
+            r({ value: msg, done: false });
+          } else {
+            queue.push(msg);
+          }
+        }
+      };
+
+      // Process any data that arrived during handshake
+      if (leftover.length > 0) {
+        enqueue(parser.feed(leftover));
+      }
+
+      socket.on('data', (data) => {
+        enqueue(parser.feed(data.toString()));
+      });
+
+      socket.on('close', () => {
+        done = true;
+        if (resolve) {
+          const r = resolve;
+          resolve = undefined;
+          r({ value: undefined as never, done: true });
+        }
+      });
+
+      socket.on('error', () => {
+        done = true;
+        if (resolve) {
+          const r = resolve;
+          resolve = undefined;
+          r({ value: undefined as never, done: true });
+        }
+      });
+
+      return {
+        next(): Promise<IteratorResult<DaemonMessage>> {
+          if (queue.length > 0) {
+            return Promise.resolve({ value: queue.shift()!, done: false });
+          }
+          if (done) {
+            return Promise.resolve({ value: undefined as never, done: true });
+          }
+          return new Promise((r) => { resolve = r; });
+        },
+        [Symbol.asyncIterator]() { return this; },
+      };
+    },
+  };
+}
--- a/packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
@@ -97,23 +97,34 @@ function createConnection(socket: net.Socket, leftover: string): DaemonConnectio
   };
 }
 
+const DONE: IteratorReturnResult<undefined> = { value: undefined, done: true };
+
 function createMessageIterable(socket: net.Socket, leftover: string): AsyncIterable<DaemonMessage> {
   return {
     [Symbol.asyncIterator](): AsyncIterableIterator<DaemonMessage> {
       const parser = new LineParser<DaemonMessage>();
       const queue: DaemonMessage[] = [];
-      let resolve: ((value: IteratorResult<DaemonMessage>) => void) | undefined;
+      let pending: ((value: IteratorResult<DaemonMessage>) => void) | undefined;
       let done = false;
 
       const enqueue = (messages: DaemonMessage[]) => {
         for (const msg of messages) {
-          if (resolve) {
-            const r = resolve;
-            resolve = undefined;
-            r({ value: msg, done: false });
+          if (pending) {
+            const settle = pending;
+            pending = undefined;
+            settle({ value: msg, done: false });
           } else {
             queue.push(msg);
           }
+        }
+      };
+
+      const terminate = () => {
+        done = true;
+        if (pending) {
+          const settle = pending;
+          pending = undefined;
+          settle(DONE);
         }
       };
 
@@ -126,23 +137,8 @@ function createMessageIterable(socket: net.Socket, leftover: string): AsyncItera
         enqueue(parser.feed(data.toString()));
       });
 
-      socket.on('close', () => {
-        done = true;
-        if (resolve) {
-          const r = resolve;
-          resolve = undefined;
-          r({ value: undefined as never, done: true });
-        }
-      });
-
-      socket.on('error', () => {
-        done = true;
-        if (resolve) {
-          const r = resolve;
-          resolve = undefined;
-          r({ value: undefined as never, done: true });
-        }
-      });
+      socket.on('close', terminate);
+      socket.on('error', terminate);
 
       return {
         next(): Promise<IteratorResult<DaemonMessage>> {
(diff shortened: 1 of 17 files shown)


packages/@aws-cdk/cdk-explorer/lib/protocol.ts
```diff
@@ -0,0 +1,86 @@
...
+  readonly socketPath: string;
+  readonly version: string;
+  readonly startedAt: number;
+  readonly logFile: string;
+}
+
+// ---------------------------------------------------------------------------
+// Protocol version — manually bumped when the wire protocol changes.
+// Used for daemon version handshake: if client and daemon versions differ,
+// the daemon shuts down so a new one matching the client can be spawned.
```
ShadowCat567: I don't think I am entirely understanding what this is describing. Why would our client and daemon version differ? What causes the protocol version to change? What does wire protocol mean?

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/**
+ * Ensures a daemon is running for the given project directory
+ * and returns a connected client.
+ *
+ * Uses an exclusive lock file to prevent concurrent spawns.
+ * If a daemon is already running, connects to it directly.
+ */
+export async function ensureDaemon(options: EnsureDaemonOptions): Promise<DaemonConnection> {
```
ShadowCat567: nit: I don't really like `ensure` as a part of this name, I would prefer `checkDaemon` a bit better, but I am happy to hear other alternatives

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  // daemon already running
+  try {
+    return await connectToDaemon(projectDir);
+  } catch {
+    // Connection failed, need to spawn or clean up
+  }
+
+  const lockPath = lockPathForProject(projectDir);
+  const lockFd = acquireLock(lockPath);
```
ShadowCat567, line 35: Is this locking step so we don't attempt to spawn a new daemon when something else has already spawned a new daemon?

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  try {
```
ShadowCat567, line 37: Why is this a try block (attempting to connect to the daemon after acquiring the lock)? 
I feel a bit weird about these nested try blocks and I am wondering if there is a reason why you chose this method or if there is a better way

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
```diff
@@ -0,0 +1,180 @@
...
+  } finally {
+    releaseLock(lockFd, lockPath);
+  }
+}
+
+function acquireLock(lockPath: string): number {
+  try {
+    const fd = fs.openSync(lockPath, 'wx');
+    fs.writeSync(fd, String(process.pid));
+    return fd;
```
ShadowCat567: this section of code is repeated in a few places it might be a good candidate for a small function

packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
```diff
@@ -0,0 +1,165 @@
...
+      socket.on('data', (data) => {
+        enqueue(parser.feed(data.toString()));
+      });
+
+      socket.on('close', () => {
+        done = true;
+        if (resolve) {
+          const r = resolve;
+          resolve = undefined;
+          r({ value: undefined as never, done: true });
```
ShadowCat567: If the value can be undefined shouldn't it be optional? Why are we casting to never? (Also what is `r`)

packages/@aws-cdk/cdk-explorer/lib/daemon/entry.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  const shutdown = () => void server.stop();
+  process.on('SIGTERM', shutdown);
+  process.on('SIGINT', shutdown);
+
+  server.start().then(
+    () => {
+      // eslint-disable-next-line no-console
+      console.log(`Daemon started: pid=${process.pid} socket=${socketPath}`);
```
ShadowCat567, line 23: CDK might have some kind of console writing function you can make use of instead of `console.log`, can you look into that?

packages/@aws-cdk/cdk-explorer/test/daemon/connect.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      socketPath,
+      projectDir: TEST_PROJECT,
+      logFile: socketPath + '.log',
+      onSynth: options?.onSynth ?? (() => Promise.resolve()),
+      idleTimeoutMs: 60_000,
+    });
+    return server.start();
+  }
+
+  test('connects and completes handshake', async () => {
```
ShadowCat567: I feel like this test isn't really doing that, we are not checking to see whether the handshake went through. 
I am not entirely understanding how `connection.send` and `connection.close` being `functions` relates to whether the handshake was completed

packages/@aws-cdk/cdk-explorer/test/daemon/connect.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    connection.send({ type: 'subscribe' });
+    connection.send({ type: 'requestSynth', triggerFile: 'fail.ts' });
+
+    const iterator = connection.messages[Symbol.asyncIterator]();
+    const result = await iterator.next();
+    expect(result.done).toBe(false);
+    expect(result.value.type).toBe('synthFailed');
+  });
+
+  test('multiple messages arrive in order', async () => {
```
ShadowCat567, line 98: I like what is going on in this test :D

packages/@aws-cdk/cdk-explorer/test/daemon/server.test.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  beforeEach(() => {
+    socketPath = uniqueSocketPath();
+  });
+
+  afterEach(async () => {
+    if (server) {
+      await server.stop();
+    }
+    try {
```
ShadowCat567, line 72: why is there a try/catch block here? What would cause `unlinkSync` to fail in a test? Is there anything we can do to make sure it doesn't fail?

ShadowCat567 (commented):
For future PRs can you give more of an overview of what I am looking at/reviewing in the PR description since even though I know the architecture of what you are going for, figuring out which file made most sense to start in and how the code in the different files worked together was a bit challenging. It would also be helpful to have some more comments describing what is going on with your functions since many of them took me some time to parse and figure out what they were trying to accomplish (especially because this kind of architecture is not commonly used in CDK).
Otherwise it looks pretty good overall, I left some comments for you to take a look at.

packages/@aws-cdk/cdk-explorer/lib/protocol.ts
megha-narayanan: The wire protocol was teh format of the JSON message that we sent over this socket. This was for if you update your cli, so we have a new version of the client (LSP server/web) which may have different signals, but the daemon from the OLD version could still be running. It doesn't necessarily shut down when your CLI updates. So I put this in as a catch for this case, so a new daemon from the new version of the client can be spun up.  Otherwise we could silently miss messages.  Updated the documentation here to reflect that.

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
megha-narayanan: Renamed to `acquireDaemon`, wanted to make it clear that we don't only check and see if a daemon is there, but also spin one up if there isn't

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
megha-narayanan: Yep, its so that if a daemon dies, two clients (like an LSP and the Web Server) can't both spin up a new daemon at the same time. I put this in docstring of the function, lmk if that can be clearer

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
megha-narayanan: Because after we acquire the lock, if we were, say, the loser of a race, the try connect would succeed. else, if the daemon still doesn't exist, it would error, so that's what we would catch (and when we would make a new daemon). I can take this into a tryConnect helper which will return undefined on failure, to avoid the nested try-catch blocks here. teh outer try-finally is to make sure that no matter what happens, we HAVE TO release the lock. even if we have a timeout or a hang or unexpected failure, like if spawn daemon times out.

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
megha-narayanan: done, also did unlinkIfExists

packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
megha-narayanan: The casting as never was a workaround to make TypeScript happy.....when iteration is done, there's no value to return, but the type still demands one. I've cleaned this up there's now a shared `DONE` constant with the right type `(IteratorReturnResult<undefined>)`.

`r` is the stored promise resolver, I named it r to avoid shadowing the outer `resolve`, but you're right it's cryptic. Renamed to `pending` (someone is waiting for the next message) and `settle` (we're handing them their answer). The grab-then-null pattern `(const settle = pending; pending = undefined; settle(...))` just prevents accidentally resolving the same promise twice.

packages/@aws-cdk/cdk-explorer/lib/daemon/entry.ts
megha-narayanan: yeaaaaa.... there's one in `toolkit-lib`. It does bring in some stuff we don't need, but I do think we should use this eventually. One side effect is that that function requires an `action` field, which would have to be `synth` here even though the daemon also logs non-synth lifecycle stuff (startup, idle timeout, shutdown) ... this doesn't feel totally correct to me but I don't think altering the `ToolkitAction` type is the correct thing to do either, since everything else in that type is a user-facing CLI command (packages/@aws-cdk/toolkit-lib/lib/api/io/toolkit-action.ts). Happy to hear thoughts and alternatives.

packages/@aws-cdk/cdk-explorer/test/daemon/connect.test.ts
megha-narayanan: `connectToDaemon` performs the handshake internally and throws if it fails, so this is basically testing that (interally) when the daemon is running, it doesn't fail.  I do think this test is redundant though, the happy-path is already covered by the other tests in this file that successfully connect, subscribe, and exchange messages (like the next test after this one, receives messages via async iterable) which is more robust.   

I removed it and added a proper test that proves the handshake is a real gate: 'rejects connection on protocol version mismatch'. It mocks `PROTOCOL_VERSION` to a wrong value and asserts that connectToDaemon throws. That proves the handshake actually happens and rejects mismatched clients.

packages/@aws-cdk/cdk-explorer/test/daemon/server.test.ts
megha-narayanan: the try catch block was because the socket file SHOULD have been deleted already because stop() usually removes the file first. this was a safety net to prevent leaking socket files between tests, even if stop fails/hangs, but I realize that actually isn't doing what it needs to as is. fixed to put stop instdie the try block, so we actually cover the failure case with clean up.

packages/@aws-cdk/cdk-explorer/lib/daemon/spawn.ts
```diff
@@ -0,0 +1,187 @@
...
+  const lockFd = acquireLock(lockPath);
+
+  try {
+    // Re-check after acquiring lock (another process may have won the race)
+    const raceWinner = await tryConnect(projectDir);
+    if (raceWinner) return raceWinner;
+
+    await cleanupStaleState(projectDir);
+    await spawnDaemon(projectDir);
+    return await connectToDaemon(projectDir);
```
ShadowCat567, line 44: Why do we still go to this section if we are not the `raceWinner`?

packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
```diff
@@ -0,0 +1,164 @@
...
+        socket.write(JSON.stringify(msg) + '\n');
+      }
+    },
+    close(): void {
+      socket.destroy();
+    },
+  };
+}
+
+const DONE: IteratorReturnResult<undefined> = { value: undefined, done: true };
```
ShadowCat567, line 103: I would prefer if this were at the top of the file to make it easier to find

ShadowCat567 (commented):
Please add some comments to the functions in this PR! Otherwise this looks pretty good, have a few minor suggestions

packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
```diff
@@ -0,0 +1,164 @@
...
+      if (leftover.length > 0) {
+        enqueue(parser.feed(leftover));
+      }
+
+      socket.on('data', (data) => {
+        enqueue(parser.feed(data.toString()));
+      });
+
+      socket.on('close', terminate);
+      socket.on('error', terminate);
```
9pace, line 144: here if daemon crashes, consumers can’t distinguish graceful shutdown from a crash. can we preserve the error and propagate it upwards?

packages/@aws-cdk/cdk-explorer/lib/daemon/connect.ts
```diff
@@ -0,0 +1,164 @@
...
+    socket.on('data', onData);
+    socket.on('error', onError);
+  });
+}
+
+function createConnection(socket: net.Socket, leftover: string): DaemonConnection {
+  return {
+    messages: createMessageIterable(socket, leftover),
+    send(msg: ClientMessage): void {
+      if (!socket.destroyed) {
```
9pace, line 93: If the daemon has already idle-timed-out, socket.destroyed makes send() silently no-op. so requests can be dropped while the consumer waits forever for a response.

packages/@aws-cdk/cdk-explorer/lib/daemon/server.ts
```diff
@@ -0,0 +1,248 @@
...
+      void this.stop();
+    }
+  }
+
+  private handleSubscribe(socket: net.Socket): void {
+    this.subscribers.add(socket);
+    this.clearIdleTimer();
+  }
+
+  private handleRequestSynth(): void {
```
9pace, line 165: A non-subscribing requestSynth can be killed by an  idle timer that is already going... if the timer was set before synth starts, stop() can destroy the socket mid-synth. Is this scenario possible?
we could call clearIdleTimer().

packages/@aws-cdk/cdk-explorer/lib/daemon/server.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  private handleHandshake(socket: net.Socket, clientVersion: string): void {
+    const accepted = clientVersion === PROTOCOL_VERSION;
+    this.send(socket, {
+      type: 'handshakeAck',
+      version: PROTOCOL_VERSION,
+      accepted,
+    });
+
+    if (!accepted) {
```
9pace, line 155: imagine: user has 1 editor windows running the LSP, all subscribed to a v1 daemon. They npm install a new CLI version. A nother process 2 connects with v2. This branch fires and the entire daemon shuts down, causing editor 1's connection to drop, maybe mid synth. Maybe instead of shutting down the whole daemon, we destroy the socket.

packages/@aws-cdk/cdk-explorer/lib/daemon/server.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+export class DaemonServer {
+  private readonly socketPath: string;
+  private readonly projectDir: string;
+  private readonly logFile: string;
+  private readonly onSynth: () => Promise<void>;
+  private readonly idleTimeoutMs: number;
+  private readonly synthTimeoutMs: number;
+  private readonly latch = new SynthLatch();
```
9pace, line 34: why not manage synthstate on the server `SynthState = 'idle' | 'synthesizing' | 'queued';` instead of having it as a separate class.

rix0rrr (changes requested):
A bit of architectural choices in the PR body (or potentially in an ADR doc) would be extremely helpful. It would cover things like:

- What is the daemon for?
- How/when is it started
- How/when is it shut down
- How do we deal with multiple daemons running at the same time
- How/when is it discovered by clients
- What happens to the CDK app's `stdout/stderr`, that customers are probably relying on?
- What if the app's behavior is different whether attached to a terminal or not?
- What are all the failure modes we thought of, and how are we preventing them?

This is a big architectural change/choice, and I'm a little concerned that it is being taken rather quickly.

I would like your work to land on `main`, rather than be stuck in a branch forever. This particular change is making me uneasy, and I'm concerned we're going to have to disentangle it from the rest of your work in the future if we decide against it after all. To make that easier/less risky, I'd rather have this as an add-on at the very end. Or not have it at all.

For example, a simpler solution that you haven't proposed any counterarguments for (those would go into said ADR): all processes try to lock the output directory, one of them wins and that one synthesizes, the others just wait, and when the write lock is lifted assume that the assembly is up-to-date and just read it.

PR #1565, opened 2026-05-28
refactor: use WorkspaceJsiiBuild from cdklabs-projen-project-types

mrgrain:
Replaces the local `JsiiBuild` component with `WorkspaceJsiiBuild` from `cdklabs-projen-project-types` for the `cloud-assembly-schema` package. This moves us to the upstream implementation, reducing local maintenance burden and keeping us aligned with the shared build infrastructure used across cdklabs projects.

The remaining local `JsiiBuild` class has its NPM publishing job check commented out since `WorkspaceJsiiBuild` handles this differently.

### Checklist

---


projenrc/jsii.ts
mrgrain, line 1: removing the now redundant file

.projenrc.ts
```diff
@@ -462,7 +461,7 @@ const cloudAssemblySchema = configureProject(
 );
 fixupTestTask(cloudAssemblySchema);
 
-new JsiiBuild(cloudAssemblySchema, {
+cloudAssemblySchema.with(new yarn.WorkspaceJsiiBuild({
```
mrgrain, line 464: this is the actual change. everything else is just version updates.

packages/@aws-cdk/cloud-assembly-schema/.projen/tasks.json
```diff
@@ -150,8 +150,7 @@
       "description": "Creates the distribution package",
       "steps": [
         {
-          "spawn": "package:js",
-          "condition": "node -e \"if (!process.env.CI) process.exit(1)\""
+          "spawn": "package:js"
```
mrgrain, line 153: This is the only functional difference to before.

Previously we wouldn't package `@aws-cdk/cloud-assembly-schema` locally at all during build. I believe this was not intentional, since all other packages are packaged during build.

What was intentional however was the to remove local packaging for all jsii languages (which is super slow). I have kept this removal and just re-established the `package:js` step.

In CI, it is important we only run `package:js` during build since the other packaging tasks are run in a separate job.

rix0rrr (approved):

PR #1578, opened 2026-06-01
feat: add web server scaffold and `cdk explore` command

megha-narayanan:
- Adds a minimal Express web server in `@aws-cdk/cdk-explorer` with a `/api/health` endpoint
- Wires up `cdk explore` as a CLI subcommand that starts the server and blocks until SIGINT/SIGTERM
- Port behavior: auto-increments from 4200 if default is taken; throws if an explicit -`-port` is unavailable

Fixes #

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/web/server.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/web/server.ts
@@ -2,7 +2,7 @@ import * as http from 'http';
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import express = require('express');
 
-const DEFAULT_PORT = 4200;
+export const DEFAULT_PORT = 4200;
 const MAX_PORT_ATTEMPTS = 100;
 
 export interface WebServerOptions {
--- a/packages/@aws-cdk/cdk-explorer/lib/web/server.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/web/server.ts
@@ -1,0 +1,90 @@
+import * as http from 'http';
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import express = require('express');
+
+const DEFAULT_PORT = 4200;
+const MAX_PORT_ATTEMPTS = 100;
+
+export interface WebServerOptions {
+  readonly port?: number;
+  readonly host?: string;
+}
+
+export interface WebServer {
+  readonly url: string;
+  stop(): Promise<void>;
+}
+
+/**
+ * Starts the CDK Explorer web server.
+ *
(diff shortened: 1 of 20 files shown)


packages/@aws-cdk/cdk-explorer/test/web/server.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  test('binds to 127.0.0.1 by default', async () => {
+    server = await startWebServer();
+    expect(server.url).toMatch(/^http:\/\/127\.0\.0\.1:\d+$/);
+  });
+
+  test('auto-increments port when default is taken', async () => {
+    const first = await startWebServer();
+    server = await startWebServer();
+
+    expect(server.url).not.toBe(first.url);
```
ShadowCat567: Can we check directly that it is using `default port number + 1`? (EX: `expect(server.url).toEqual(default+1)`)

packages/@aws-cdk/cdk-explorer/lib/web/server.ts
```diff
@@ -0,0 +1,90 @@
+import * as http from 'http';
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import express = require('express');
```
ShadowCat567, line 3: I think express can be imported using `import express from 'express'` (or something similar)

ShadowCat567 (commented):
Can you add a small test for the explore command to ensure it starts up the server and prints out the message?
Otherwise this looks pretty good!

packages/@aws-cdk/cdk-explorer/lib/web/server.ts
megha-narayanan: from what I encountered, it's required because this monorepo has `esModuleInterop: false `in `tsconfig`.

With `esModuleInterop: false`, `import express from 'express'` fails at runtime because TypeScript expects a default export property on the module, but express exports a function directly. Let me know if there's a better way for getting around this. 

The same approach is used elsewhere in the monorepo (aws-cdk imports various CJS modules this way).

packages/@aws-cdk/cdk-explorer/test/web/server.test.ts
megha-narayanan: yep, added this

packages/@aws-cdk/cdk-explorer/lib/web/server.ts
ShadowCat567: D: too bad, curse you express!

ShadowCat567 (approved):

PR #1579, opened 2026-06-01
chore: Synth Daemon ADR

megha-narayanan:
Architecture Decision Record for the per-project synth daemon

Covers:
- Why a daemon vs. no coordination vs. directory locking (flock)
- Daemon lifecycle (start, discovery, shutdown, singleton enforcement)
- CDK app stdout/stderr handling (via IIoHost)
- TTY behavior 
- Failure modes and recovery
- Assumptions that would invalidate this decision

### Checklist

---


packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+Each client triggers its own synth independently. The last one to
+finish wins. This is the simplest approach.
+
+Problems:
+
+- N clients × M saves = N×M synth operations. A user with an LSP
+  and a web server open gets 2 synths per save. With rapid saves
+  during a refactor, both processes spend 100% of their time
+  synthesizing.
+- No client knows whether its `cdk.out/` is fresh. Each must either
```
rix0rrr, line 41: Could we solve this with additional bookkeeping though?

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+  poll the filesystem or re-synth speculatively.
+- Concurrent writes to `cdk.out/` from two simultaneous synths could
```
rix0rrr, line 43: Won't happen. There is already locking on the cdk.out directory.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+  produce inconsistent output.
+
+#### Option B: Directory locking (flock on `cdk.out/`)
```
rix0rrr, line 46: We don't necessarily need to use `flock()`, there is already a locking system. It may not be great, but it's a small delta from what we currently use. Could potentially be improved by using `flock()`, who knows.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+
+Processes compete for a write lock on `cdk.out/`. Winner synthesizes,
+losers wait and read. This prevents concurrent writes but creates a
+dilemma:
+
+- *If losers synth unconditionally after acquiring the lock
+  ("optimistic"):* This is wasted compute — each client re-synths
```
rix0rrr, line 53: I'm not sure what this says exactly. "If losers retry until they acquire the lock, nothing changes in between, but the loser still triggers a resynth, then that would be wasted effort".

True.

But:

1. That's not what your diagram shows. Your diagram shows an actual change in between clients A and B (*), so I think the sequence of what happens in `optimistic-flock.png` is correct. First we have version A, and then we have version C. We can display both, as soon as we have it.
2. It's also not necessary to unconditionally resynth. With some additional bookkeeping we can avoid it.

> (*) Diagramming tip: if you call your clients A and B, name your versions something else (1, 2, 3, or X, Y, Z). I would have liked to just refer to "A" but that is now ambiguous whether I mean a client or a version.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+  the same source state another client just finished synthesizing.
+  With 3 clients and frequent saves, the system may spend near 100%
+  of wall-clock time synthesizing. Since there is no method to
```
rix0rrr, line 56: Yes, but watch out. Ask yourself what the essence is of what is happening. 

If there are frequent saves (as in, saves come in faster than the time it takes to synthesize), how does a daemon prevent 100% wall clock time spent in synth?

I think you might be getting at a different fundamental trade-off here, CPU usage vs staleness in the face of frequent changes:

image

## But note that this trade-off is exactly the same with or without a daemon!

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+  broadcast, only the last client to finish has fresh data; the
+  others are already stale by the time they complete.
```
rix0rrr, line 58: But how does a daemon solve that? 

Probably either one of:

- Clients poll the daemon for the current synth status; or
- The daemon *notifies* clients when a synth succeeds

Well, if we didn't have a daemon we have another component that can fulfill exactly those functions: that is, the *filesystem*! You can poll the file system, or listen for file changes. So clients can re-read the files on disk if they change.

rix0rrr: Speaking of the filesystem, you need to model that as part of your system as well. I don't know if you get into that in the rest of your document, but another interesting problem that is the opposite of what we are talking about here, is the *opposite of staleness*: how do we avoid unnecessary resynths in the face of multiple clients:

image

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+
+![Optimistic Flock](optimistic-flock.png)
+
+- *If losers skip synth and just read ("pessimistic"):* This is
```
rix0rrr, line 62: As discussed previously, this section has the finger on the right pulse for a fundamental trade-off, but misattributes it to something that a daemon will magically solve.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+  version 1 data. The losers read it and assume it's fresh, but it
+  doesn't reflect version 2. Synth time is managed, but all clients
+  show stale data with no mechanism to detect it, and no re-synth
+  is triggered.
+
+![Pessimistic Flock](pessimistic-flock.png)
+
+Additionally, flock on `cdk.out/` interferes with normal
+`cdk synth` — the user's manual synth would either block on the
+lock (unacceptable) or bypass it entirely (reintroduces all the
```
rix0rrr, line 74: Well an interaction with `cdk synth` hasn't been discussed at all. 

It's kind of hard to compare solutions here since we haven't discussed expectations at all 🤔.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+state without unbounded cost.
+
+All connected clients receive a `synthComplete`/`synthFailed`
+broadcast when synthesis finishes.
+
+**What it does NOT do:**
+
+- Does not interfere with `cdk synth` or `cdk deploy`. Interactive
+  CLI commands never go through the daemon — they are too critical
+  to depend on another process.
```
rix0rrr, line 107: But what does that mean? As you have brought up above, it is dangerous to have 2 processes write into the same directory.

Yet if `cdk synth` and the daemon are unconnected, and someone runs `cdk synth` while the daemon is running in response to a `Cmd-S`, **exactly that will happen, the processes will step on each other**.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+- Does not watch files. Clients decide when to request synth.
+- Does not lock `cdk.out/`. A user running `cdk synth` in their
```
rix0rrr, line 109: Well that is what you said 2 bullets above.

And because the CLI already does locking, this is also not what happens.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+  terminal is never blocked or interfered with by the daemon.
+
+**Why now, not later:**
+
+- Multi-client is a day-1 scenario: the editor LSP and
+  `cdk explore` web server both trigger synth on the same project
+  from the start.
```
rix0rrr, line 116: [Assumption] Depends on the user audience. Only if they run both at the same time, constantly.

[Alternate solution] Could also be solved by having `cdk explore` not synth automatically but based on a button. I would totally accept that for an MVP.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+- An LSP serving stale data is useless. Freshness must be solved
+  before building diagnostic features on top.
```
rix0rrr, line 118: "Useless", really? I don't think so. Slightly annoying to work with perhaps. Definitely not useless.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+- Unbounded synth cost from uncoordinated clients is unacceptable
+  for a background tool.
```
rix0rrr, line 120: False premise: no one said uncoordinated.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+
+**Lifecycle:**
+
+- Started by: first client that calls `acquireDaemon(projectDir)`
+  (attempts connection, spawns if absent)
+- Discovered by: deterministic socket path derived from project
+  directory (`/tmp/cdk-synth-<sha256>.sock`)
```
rix0rrr, line 127: Windows?

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+  from the start.
+- An LSP serving stale data is useless. Freshness must be solved
+  before building diagnostic features on top.
+- Unbounded synth cost from uncoordinated clients is unacceptable
+  for a background tool.
+
+**Lifecycle:**
+
+- Started by: first client that calls `acquireDaemon(projectDir)`
+  (attempts connection, spawns if absent)
```
rix0rrr, line 125: What client? Only LSP and `cdk explore`, is that right?

Any plans of getting `cdk watch` in there? `cdk synth` perhaps?

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+- Shut down by: idle timeout (5 min with zero subscribers),
+  protocol version mismatch, SIGTERM
+- Singleton enforced by: exclusive lock file during spawn,
+  deterministic socket path
+
+**CDK app stdout/stderr:**
+
+The daemon implements `IIoHost` from `@aws-cdk/toolkit-lib`. All
+CDK app output during synthesis arrives as structured `IoMessage`
+objects via `notify()`. Nothing is silently lost.
```
rix0rrr, line 137: I'm pretty sure today just attach the subprocess to the parent process' terminal, so this is not true.

The question is to engage with what will happen if we change that assumption. It may be hard for you to gauge the effect of this, as it's quite esoteric, I understand. Search the bug tracker for what happened last time we disconnected a user app from the terminal 😉

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+- **App `console.log` output:** routed to the editor's Output
+  channel via LSP `window/logMessage`. Users who need interactive
+  stdout run `cdk synth` directly, which bypasses the daemon.
+- **Daemon operational messages:** written to a log file at
+  `<socketPath>.log`.
+
+**Terminal attachment (TTY):**
+
+Not a concern. The CDK app is always spawned with piped stdio by
+`execInChildProcess` in toolkit-lib — `process.stdout.isTTY` is
```
rix0rrr, line 151: Perhaps in `toolkit-lib`, but not in `aws-cdk`.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+`false` regardless of whether the caller is the CLI, CI, or the
+daemon. The daemon does not change the app's TTY status.
+
+**Failure modes:**
+
+| Failure                        | Recovery                                    |
+| ------------------------------ | ------------------------------------------- |
+| Daemon crashes mid-synth       | Client detects socket close, respawns       |
+| Concurrent spawn race          | Exclusive lock file; loser re-checks socket |
+| Synth hangs                    | 5-min timeout → `synthFailed` broadcast     |
```
rix0rrr, line 161: Synth sometimes takes 40 minutes for some people.

packages/@aws-cdk/cdk-explorer/doc/adr/0001-per-project-synth-daemon.md
```diff
@@ -0,0 +1,202 @@
...
+problem because nothing reads `cdk.out/` while it's being written.
+You run synth, it finishes, then you look at the output.
+
+The daemon does not change this risk. When the daemon triggers synth,
+clients wait for `synthComplete` before re-reading, which naturally
+avoids the race. The torn-read window only exists when an external
+process (manual `cdk synth`, `cdk watch`) writes to `cdk.out/` while
+clients happen to be reading — the same situation that exists today.
+The file-write phase is fast (milliseconds), so the window is tiny
+in practice.
```
rix0rrr, line 179: Not true, it can takes dozens of seconds. Where do the sources for this statement come from?

rix0rrr (changes requested):

PR #1585, opened 2026-06-01
feat(toolkit-lib): simplify deploy approval message and surface auto-confirm

### LINKED ISSUE #828: toolkit-lib: suppress deployment notice about --require-approval
### Describe the bug

I'm deploying stacks with toolkit-lib and the stack has iam role. On deployment it gives the changes (which is fine) but then puts a message like 

```
"--require-approval" is enabled and stack includes security-sensitive updates.
Do you wish to deploy these changes
```

Then proceeds to deploy. I added `    "@aws-cdk/toolkit:requireApproval": "never",` in cdk.json of my root so it passes but I would like to suppress that message. 

Total logs
```
Stack8
Stack Stack8
IAM Statement Changes
┌───┬───────────────────────┬────────┬────────────────┬───────────────────────────┬───────────┐
│   │ Resource              │ Effect │ Action         │ Principal                 │ Condition │
├───┼───────────────────────┼────────┼────────────────┼───────────────────────────┼───────────┤
│ + │ ${S3ReadOnlyRole.Arn} │ Allow  │ sts:AssumeRole │ Service:ec2.amazonaws.com │           │
└───┴───────────────────────┴────────┴────────────────┴───────────────────────────┴───────────┘
IAM Policy Changes
┌───┬───────────────────┬──────────────────────────────────────────────────────────────┐
│   │ Resource          │ Managed Policy ARN                                           │
├───┼───────────────────┼──────────────────────────────────────────────────────────────┤
│ + │ ${S3ReadOnlyRole} │ arn:${AWS::Partition}:iam::aws:policy/AmazonS3ReadOnlyAccess │
└───┴───────────────────┴──────────────────────────────────────────────────────────────┘
(NOTE: There may be security-related changes not in this list. See https://github.com/aws/aws-cdk/issues/1299)



"--require-approval" is enabled and stack includes security-sensitive updates.
Do you wish to deploy these changes
Stack8: deploying... [9/10]
Stack8: creating stack...
Stack8 |   0 | 4:00:55 PM | CREATE_IN_PROGRESS   | AWS::CloudFormation::Stack | Stack8 User Initiated
Stack8 |   0 | 4:00:57 PM | CREATE_IN_PROGRESS   | AWS::CDK::Metadata         | CDKMetadata/Default (CDKMetadata) 
Stack8 |   0 | 4:00:57 PM | CREATE_IN_PROGRESS   | AWS::IAM::Role             | S3ReadOnlyRole (S3ReadOnlyRoleFA2E4025) 
Stack8 |   0 | 4:00:58 PM | CREATE_IN_PROGRESS   | AWS::IAM::Role             | S3ReadOnlyRole (S3ReadOnlyRoleFA2E4025) Resource creation Initiated
Stack8 |   0 | 4:00:58 PM | CREATE_IN_PROGRESS   | AWS::CDK::Metadata         | CDKMetadata/Default (CDKMetadata) Resource creation Initiated
Stack8 |   1 | 4:00:59 PM | CREATE_COMPLETE      | AWS::CDK::Metadata         | CDKMetadata/Default (CDKMetadata) 
Stack8 |   2 | 4:01:14 PM | CREATE_COMPLETE      | AWS::IAM::Role             | S3ReadOnlyRole (S3ReadOnlyRoleFA2E4025) 
Stack8 |   3 | 4:01:15 PM | CREATE_COMPLETE      | AWS::CloudFormation::Stack | Stack8 

 ✅  Stack8
✨  Deployment time: 21.81s
```

Stack used
```typescript
import * as iam from 'aws-cdk-lib/aws-iam';
// import * as sqs from 'aws-cdk-lib/aws-sqs';

export class Cdk20250826Stack extends cdk.Stack {
    super(scope, id, props);

    // IAM role with S3 readonly access
    new iam.Role(this, 'S3ReadOnlyRole', {
      assumedBy: new iam.ServicePrincipal('ec2.amazonaws.com'),
      managedPolicies: [
        iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonS3ReadOnlyAccess')
      ]
    });
  }
}

```

### Regression Issue

_No response_

### Expected Behavior

an option to suppress the message 
```
"--require-approval" is enabled and stack includes security-sensitive updates.
Do you wish to deploy these changes
```

### Current Behavior

even though require-approval is not on it still shows up in the logs

### Reproduction Steps

[cdk_20250826.ts](https://github.com/user-attachments/files/22060670/cdk_20250826.ts)
Reduce the loop size to 1 and deploy with command `node bin/cdk_20250826.js deploy` 

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.1026.0

### Framework Version

_No response_

### Node.js Version

v20.19.3

### OS

macOS 15.6

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

sai-ray:
Closes #828.

  `@aws-cdk/toolkit-lib`'s `deploy()` previously printed `"--require-approval" is enabled and stack includes security-sensitive updates. Do you wish to deploy these changes` and then silently auto-confirmed via the default `NonInteractiveIoHost`. The question-shaped text without a visible answer-shaped action confused programmatic users.

  This PR:
- keeps the I5060 confirmation request in `_deploy` (it's the IoHost's job to decide what to do with it).
- drops the `--require-approval` reference from the library's vocabulary.
- makes `NonInteractiveIoHost`'s auto-response visible.
- moves the CLI-side `--require-approval` framing into a listener registered by the deploy command.

#### Changes

- **`@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts`**: motivation drops the `--require-approval` reference but keeps the existing security-sensitive vs non-security-sensitive branching:
    - Security-sensitive change: `Stack includes security-sensitive updates`
    - Non-security change: `Stack includes updates`
    
- **`@aws-cdk/toolkit-lib/lib/toolkit/non-interactive-io-host.ts`**: `requestResponse` annotates the message based on the actual default response value:
    - `defaultResponse: true`: `(auto-confirmed)`
    - `defaultResponse: false`: `(auto-denied)`
    - non-boolean default: `(auto-responded with default: <value>)`
    
 - **`aws-cdk/lib/cli/cdk-toolkit.ts`**: `deploy()` registers a `rewrite(IO.CDK_TOOLKIT_I5060, …)` listener on `CliIoHost` that reframes the library's bare-fact question with the `--require-approval` value, keyed off `msg.data.permissionChangeType`. Both the CLI's own deploy path and toolkit-lib's deploy action resolve through the same host, so one listener covers both. Registered up front and removed in a `finally`, matching the `cdk destroy` reroute. The complete sentences it produces:
    - `BROADENING` + security change: `Stack includes security-sensitive updates and "--require-approval" is set to 'broadening'. Do you wish to deploy these changes?`
    - `ANYCHANGE` + security change: `Stack includes security-sensitive updates and "--require-approval" is set to 'any-change'. Do you wish to deploy these changes?`
    - `ANYCHANGE` + non-security change: `Stack includes updates and "--require-approval" is set to 'any-change'. Do you wish to deploy these changes?`
    
- **`@aws-cdk/toolkit-lib/test/actions/deploy.test.ts`**: motivation assertions updated for the new bare-fact text.

  #### Example

  Same stack, before vs after this PR:

  **Before**
  ```
  "--require-approval" is enabled and stack includes security-sensitive updates.
  Do you wish to deploy these changes
  Cdk828Probe: deploying... [1/1]
  ```

  **After (`cdk deploy`, default `--require-approval=broadening`, broadening change)**
  ```
  Stack includes security-sensitive updates and "--require-approval" is set to 'broadening'.
  Do you wish to deploy these changes? (y/n)
  ```

  **After (`cdk deploy --require-approval=any-change`, broadening change)**
  ```
  Stack includes security-sensitive updates and "--require-approval" is set to 'any-change'.
  Do you wish to deploy these changes? (y/n)
  ```

  **After (`cdk deploy --require-approval=any-change`, non-security change)**
  ```
  Stack includes updates and "--require-approval" is set to 'any-change'.
  Do you wish to deploy these changes? (y/n)
  ```

  **After (`await toolkit.deploy()` programmatic, default `NonInteractiveIoHost`, security-sensitive change)**
  ```
  Stack includes security-sensitive updates. 
  Do you wish to deploy these changes? (auto-confirmed)
  Cdk828Probe: deploying... [1/1]
  ```

  **After (`await toolkit.deploy()` programmatic, default `NonInteractiveIoHost`, non-security change)**
  ```
  Stack includes updates. 
  Do you wish to deploy these changes? (auto-confirmed)
  Cdk828Probe: deploying... [1/1]
  ```

  #### Test plan

  - [x] Unit tests in `test/actions/deploy.test.ts` (motivation assertions updated for the bare-fact text).
  - [x] Unit tests in `test/cli/io-host/cli-io-host.test.ts`: the four wording branches over the `rewrite(I5060)` listener (broadening/any-change × security/non-security, plus non-I5060 passthrough).
  - [x] End-to-end test in `test/commands/deploy.test.ts`: runs `deploy()` and asserts the recorded, listener-reframed prompt, plus an IO snapshot.
  - [x] Unit tests in `test/toolkit/non-interactive-io-host.test.ts`: the auto-response annotation for `true`, `false`, and non-boolean defaults.
  - [x] Verified live against a sandbox account: a real IAM (broadening) diff renders `Stack includes security-sensitive updates and "--require-approval" is set to 'broadening'. Do you wish to deploy these changes?`

  ### Checklist

  ---


packages/@aws-cdk/toolkit-lib/test/actions/deploy.test.ts
```diff
@@ -105,9 +105,9 @@ IAM Statement Changes
   });
 
   test('request response contains stack diff when there are no security changes', async () => {
-    // WHEN
+    // WHEN — 'any-change' opts in to a prompt even without security changes
     const cx = await cdkOutFixture(toolkit, 'stack-with-bucket');
-    await toolkit.deploy(cx);
+    await toolkit.deploy(cx, { requireApproval: 'any-change' });
```
sai-ray: Updated `request response contains stack diff when there are no security changes` to pass `requireApproval: 'any-change'`, since the new default no longer prompts on non-security changes.

mrgrain (changes requested):
Need to look at this in more detail. It was by design that toolkit lib doesn't get interactions.

rix0rrr (changes requested):
While wer'e doing this, can we rename it to `require-confirmation` ?

"Approval" is actually something else than approval.

mrgrain (changes requested):
Right, I don't think I want to encode this into `toolkit-lib`. We are already ask for confirmation, it is up to the implementor to decide how to proceed here.

For now, we should improve the message/motivation to simply be something like

```
Changes detected. Do you wish to deploy these changes?
```

Additionally, the `NonInteractiveIoHost` should be clearer about what's happening, e.g. print the auto-acceptance to the output like the `--yes` flag does.

---

In future, we should have an easy way to drop a specific message:

```ts
const ioHost = new NonInteractiveIoHost();
ioHost.on('I5060', (msg) => {
  // do nothing
});
```

rix0rrr:
I don't disagree @mrgrain, but then shouldn't we pull this all the way through and require the confirmation always, regardless of the change type? And give enough information for the client to correlate the change impact with the command line flag?

sai-ray:
Thanks @mrgrain and @rix0rrr. Dropped the `requireApproval` option, so `_deploy` always emits I5060 on any change and the IoHost decides what to do with it. Simplified the motivation to `Changes detected.` so the `--require-approval` reference doesn't leak out of the library. And `NonInteractiveIoHost` now annotates auto-replies with `(auto-confirmed)` or `(auto-responded with default: …)`, similar to `--yes` mode.

  @rix0rrr on your second point : if I understand correctly,  you're saying the I5060 payload should give the IoHost enough to make its own decision. `permissionChangeType` and `templateDiffs` are already on the payload, so the IoHost can correlate the change to whatever flag it exposes (that's what `CliIoHost.skipApprovalStep` uses today). The message itself still branches on security vs stack diff inside `_deploy` . should I move that out too, or is it fine where it is? Happy to do it here or as a follow-up.

packages/@aws-cdk/toolkit-lib/lib/toolkit/non-interactive-io-host.ts
```diff
@@ -120,12 +121,20 @@ export class NonInteractiveIoHost implements IIoHost {
...
+   * The non-interactive IoHost cannot prompt the user, so it always returns
+   * the request's `defaultResponse`. To make this visible, the message is
+   * annotated to indicate the auto-response — same convention as the CLI's
+   * `--yes` mode.
    */
   public async requestResponse<DataType, ResponseType>(msg: IoRequest<DataType, ResponseType>): Promise<ResponseType> {
-    // in the non-interactive IoHost, no requests are promptable
-    await this.notify(msg);
+    const annotation = typeof msg.defaultResponse === 'boolean'
+      ? '(auto-confirmed)'
```
rix0rrr: You are assuming that all `boolean`s are always `true`, apparently.

The `defaultResponse` boolean could have been `false`, in which case `auto-confirmed` is not correct. It could have been auto-denied as well, you don't have enough information to tell.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -858,14 +858,12 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
         // Send a request response with the diff as part of the message,
-        // and the template diff as data
-        // (IoHost decides whether to print depending on permissionChangeType)
+        // and the template diff as data. The IoHost decides how to handle the
+        // request — interactively prompt the user, auto-confirm, or suppress.
         const hasSecurityChanges = securityDiff.permissionChangeType !== PermissionChangeType.NONE;
-        const deployMotivation = hasSecurityChanges
-          ? '"--require-approval" is enabled and stack includes security-sensitive updates.'
-          : '"--require-approval" is enabled and stack includes updates.';
+        const deployMotivation = 'Changes detected.';
```
rix0rrr: I think I prefer the specificy of the original message. If the problem is that you cannot refer to `--require-approval`, then remove that part. But now you've removed *all* specificity.

Here is the question I want you to ask yourself, at all times: "What is better/clearer for the user?"

So, now answer that question: why is it better for the user to get a message like:

```
Changes detected. Do you wish to deploy these changes?
```

Put yourself in the shoes of the user. You've just made changes. You're now running `cdk deploy` to deploy your changes. Then the tool triumphantly tells you "there are changes!". What would you think at that point?

rix0rrr: I think the point was: The original message was good and to the point. But the *authority* that comes up with that message is in the wrong place, because the `toolkit-lib` has no business talking about command-line arguments.

So:

* You need to produce an *accurate* message here that does not refer to command line arguments.
* You then need to produce an *even more accurate* message in the CLI, that *does* refer to command line arguments.

rix0rrr (dismissed):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -858,14 +858,14 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
-        // (IoHost decides whether to print depending on permissionChangeType)
+        // and the template diff as data. The IoHost decides how to handle the
+        // request — interactively prompt the user, auto-confirm, or suppress.
         const hasSecurityChanges = securityDiff.permissionChangeType !== PermissionChangeType.NONE;
         const deployMotivation = hasSecurityChanges
-          ? '"--require-approval" is enabled and stack includes security-sensitive updates.'
-          : '"--require-approval" is enabled and stack includes updates.';
+          ? 'Stack includes security-sensitive updates.'
+          : 'Stack includes updates.';
         const diffOutput = hasSecurityChanges ? securityDiff.formattedDiff : stackDiff.formattedDiff;
```
rix0rrr: Please put the messaging about the flags back in the `CliIoHost`.

packages/@aws-cdk/toolkit-lib/lib/toolkit/non-interactive-io-host.ts
sai-ray: Fixed. `(auto-confirmed)` for `true`, `(auto-denied)` for `false`, `(auto-responded with default: <value>)` for non-boolean.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
sai-ray: Restored the security-vs-non-security branching, just dropped the `--require-approval` reference. Library now says `Stack includes security-sensitive updates.` / `Stack includes updates.`.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -398,6 +398,44 @@ export class CliIoHost implements IIoHost {
...
+   * Augment toolkit-lib's flag-free I5060 motivation with the
+   * `--require-approval` framing so CLI users see flag context.
+   */
+  private augmentDeployApprovalMessage<DataType, ResponseType>(msg: IoRequest<DataType, ResponseType>): IoRequest<DataType, ResponseType> {
+    const approvalToolkitCodes = ['CDK_TOOLKIT_I5060'];
+    if (!(msg.code && approvalToolkitCodes.includes(msg.code))) {
+      return msg;
+    }
+
+    const data = (msg.data ?? {}) as { motivation?: string; permissionChangeType?: string };
```
rix0rrr: Is there not an existing type definition that models the `data` field here? I feel like we require this.

In fact, there should be a `CDK_TOOLKIT_I5060.is(msg)` that will enforce a type guard that will automatically make `msg.data` have the right type.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -398,6 +398,44 @@ export class CliIoHost implements IIoHost {
...
+    const baseMotivation = data.motivation;
+    if (!baseMotivation) {
+      return msg;
+    }
+
+    let prefix: string;
+    switch (this.requireDeployApproval) {
+      case RequireApproval.ANYCHANGE:
+        prefix = `"--require-approval" is set to '${RequireApproval.ANYCHANGE}'. `;
```
rix0rrr: Because you want to reuse the original message, and both need to make sense in their own context, the final sentence here will be very staccato:

```
require-approval is set to anychange. Stack contains updates. Do you want to continue?
```

Instead, you can include `hasSecurityChanges: true|false` into the `data` object and format a complete sentence here, rather than scrapping about with fragments.

You might have noticed already, but I care quite a lot how CDK "feels". Not just: what the CLI says is technically accurate, but also that it speaks in a normal voice to a normal user. No formal IBM-speak, no vague Microsoft-speak.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -464,13 +502,18 @@ export class CliIoHost implements IIoHost {
         return true;
       }
 
+      // The library emits I5060 with a flag-free motivation. The CLI is the
+      // layer that knows about `--require-approval` and adds it to the
+      // user-facing message before any downstream rendering.
+      const promptMessage = this.augmentDeployApprovalMessage(msg).message;
```
rix0rrr: No need for this function to return a full `IoRequest` object if all we're going to do is pull out a single field from that object and throw the rest away. You can just have this return a `string`.

rix0rrr (commented):
Almost there!

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
sai-ray: Done. Added `hasSecurityChanges: boolean` to the `data` object, and the augmenter now writes one of three full sentences instead of stitching fragments onto the library's text:

  - BROADENING + has-security: 
  ```
Stack includes security-sensitive updates and "--require-approval" is set to 'broadening'.
  Do you wish to deploy these changes?
```
  - ANYCHANGE + has-security: 
   ```
  Stack includes security-sensitive updates and "--require-approval" is set to 'any-change'.
  Do you wish to deploy these changes?
```
  - ANYCHANGE + non-security: 
```  
Stack includes updates and "--require-approval" is set to 'any-change'.  
Do you wish to deploy these changes?
```
  On the voice, this was what I thought as being both natural and not too informal. Any thoughts/ suggestions?

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
sai-ray: Yeah, the type already exists `DeployConfirmationRequest` in `payloads/deploy.ts`. It's been wired to I5060 via `make.confirm<DeployConfirmationRequest>(...)` in `messages.ts` since before this PR. The untyped cast you commented on is gone and is replaced by the suggested type guard.
  
Added `is()` on `IoRequestMaker` in `message-maker.ts` (it was missing, only `IoMessageMaker` had it) and `hasSecurityChanges: boolean` on `DeployConfirmationRequest`. The augmenter now uses `IO.CDK_TOOLKIT_I5060.is(msg)` to narrow `msg.data` and reads `hasSecurityChanges` to pick the right phrasing. So the prompt branch is driven by a typed payload field instead of re-deriving from `permissionChangeType`.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -398,6 +398,30 @@ export class CliIoHost implements IIoHost {
     }
   }
 
+  /**
+   * Add the `--require-approval` flag to the deploy approval prompt.
+   */
+  private augmentDeployApprovalMessage(msg: IoRequest<any, any>): string {
```
rix0rrr: I'm entirely ready to ship this now. HOWEVER, it would be even better if this used the message formatting feature that @mrgrain just formalized here: https://github.com/aws/aws-cdk-cli/pull/1603

So oooooooooone final thing 😉

packages/@aws-cdk/toolkit-lib/lib/payloads/deploy.ts
```diff
@@ -34,6 +34,12 @@ export interface DeployConfirmationRequest extends ConfirmationRequest {
    */
   readonly permissionChangeType: PermissionChangeType;
 
+  /**
+   * True when the change includes security-sensitive updates (IAM or security
+   * group changes). IoHosts use this to phrase the prompt accordingly.
+   */
+  readonly hasSecurityChanges: boolean;
```
mrgrain: this is redundant, `permissionChangeType` already includes this.
```suggestion
```

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -857,18 +857,19 @@ export class Toolkit extends CloudAssemblySourceBuilder {
...
+          ? 'Stack includes security-sensitive updates'
+          : 'Stack includes updates';
         const diffOutput = hasSecurityChanges ? securityDiff.formattedDiff : stackDiff.formattedDiff;
-        const deployQuestion = `${diffOutput}\n\n${deployMotivation}\nDo you wish to deploy these changes`;
+        const deployQuestion = `${diffOutput}\n\n${deployMotivation}. Do you wish to deploy these changes?`;
         const deployConfirmed = await ioHelper.requestResponse(IO.CDK_TOOLKIT_I5060.req(deployQuestion, {
           motivation: deployMotivation,
           concurrency,
           permissionChangeType: securityDiff.permissionChangeType,
+          hasSecurityChanges,
```
mrgrain: redundant `permissionChangeType: securityDiff.permissionChangeType` has the same info

```suggestion
```

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -398,6 +398,30 @@ export class CliIoHost implements IIoHost {
...
+    // Library-emitted I5060 messages always carry data, but defend against
+    // partially-shaped inputs (e.g. tests, stubs) so we never crash a prompt.
+    if (!IO.CDK_TOOLKIT_I5060.is(msg) || !msg.data) {
+      return msg.message;
+    }
+
+    const hasSecurityChanges = msg.data.hasSecurityChanges;
```
mrgrain: check `msg.data.permissionChangeType` instead

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -2347,6 +2349,7 @@ class WorkGraphDeploymentActions implements WorkGraphActions {
               motivation,
               concurrency: this.options.concurrency,
               permissionChangeType: securityDiff.permissionChangeType,
+              hasSecurityChanges,
```
mrgrain: see above

packages/aws-cdk/lib/cli/cdk-toolkit.ts
mrgrain: nit: For consistency with `toolkit-lib` change `:` to `.`?

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -398,6 +398,30 @@ export class CliIoHost implements IIoHost {
...
+
+    if (this.requireDeployApproval === RequireApproval.BROADENING && hasSecurityChanges) {
+      return 'Stack includes security-sensitive updates and "--require-approval" is set to \'broadening\'.\nDo you wish to deploy these changes?';
+    }
+    if (this.requireDeployApproval === RequireApproval.ANYCHANGE && hasSecurityChanges) {
+      return 'Stack includes security-sensitive updates and "--require-approval" is set to \'any-change\'.\nDo you wish to deploy these changes?';
+    }
+    if (this.requireDeployApproval === RequireApproval.ANYCHANGE && !hasSecurityChanges) {
+      return 'Stack includes updates and "--require-approval" is set to \'any-change\'.\nDo you wish to deploy these changes?';
+    }
```
mrgrain: Hmm. I feel this much harder to read than the old system. I had to read three times to see the differences here. Also this if is going to be annoying if we are adding more states.

I suggest you keep this as a templating string:

```ts
const updateTypeText = hasSecurityChanges ? "security-sensitive updates" : "updates";
const requireApproval = this.requireDeployApproval; // maybe? not sure.

return `Stack includes ${updateTypeText} and "--require-approval" is set to '${requireApproval}'.\nDo you wish to deploy these changes?`;
```

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/payloads/deploy.ts
sai-ray: updated and removed `hasSecurityChanges`

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
sai-ray: switched to the suggested template string form. it's now in the rewrite listener in `cdk-toolkit.ts`

sai-ray:
Reworked to use the `CliIoHost` listener mechanism.

`--require-approval` framing now lives in a `rewrite(IO.CDK_TOOLKIT_I5060, …)` listener registered in `deploy().` toolkit-lib emits the bare motivation, the CLI adds the flag framing. One listener covers both deploy paths.

Updated this PR now that the request-listener support from #1666 and #1679 is available. This also relies on the cork-replay fix in #1701, which has now merged, to avoid re-applying the rewrite during replay.

mrgrain (approved):

PR #1587, opened 2026-06-02
feat: lsp skeleton

megha-narayanan:
LSP server skeleton: handles initialize/shutdown lifecycle, filters didSave notifications, triggers synth via injectable callback
Reuses `toolkit-lib`'s `createIgnoreMatcher` + `WATCH_EXCLUDE_DEFAULTS` for file filtering (same exclusion logic as `cdk watch`)
Adds `@aws-cdk/toolkit-lib` as a dependency in `.projenrc.ts` (enables `toolkit-internals.ts` re-export pattern matching packages/aws-cdk/lib/api-private.ts)

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/toolkit-internals.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/toolkit-internals.ts
@@ -1,0 +1,5 @@
+// Re-exports from @aws-cdk/toolkit-lib internals (same pattern as packages/aws-cdk/lib/api-private.ts)
+/* eslint-disable import/no-relative-packages */
+export { createIgnoreMatcher } from '../../toolkit-lib/lib/util/glob-matcher';
+export type { GlobMatcherOptions } from '../../toolkit-lib/lib/util/glob-matcher';
+export { WATCH_EXCLUDE_DEFAULTS } from '../../toolkit-lib/lib/actions/watch/private/helpers';
--- a/packages/@aws-cdk/cdk-explorer/lib/toolkit-internals.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/toolkit-internals.ts
@@ -1,5 +1,0 @@
-// Re-exports from @aws-cdk/toolkit-lib internals (same pattern as packages/aws-cdk/lib/api-private.ts)
-/* eslint-disable import/no-relative-packages */
-export { createIgnoreMatcher } from '../../toolkit-lib/lib/util/glob-matcher';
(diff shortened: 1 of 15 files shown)


packages/@aws-cdk/cdk-explorer/lib/toolkit-internals.ts
```diff
@@ -0,0 +1,5 @@
+// Re-exports from @aws-cdk/toolkit-lib internals (same pattern as packages/aws-cdk/lib/api-private.ts)
+/* eslint-disable import/no-relative-packages */
+export { createIgnoreMatcher } from '../../toolkit-lib/lib/util/glob-matcher';
+export type { GlobMatcherOptions } from '../../toolkit-lib/lib/util/glob-matcher';
+export { WATCH_EXCLUDE_DEFAULTS } from '../../toolkit-lib/lib/actions/watch/private/helpers';
```
ShadowCat567: Why can't these be imported like `@aws-cdk/toolkit-lib`? (like how `toolkit-lib` imports other `@aws-cdk` packages)

ShadowCat567 (commented):
Other than the import comment, I think this looks pretty reasonable, looking forward to seeing the LSP be able to actually do something :D

packages/@aws-cdk/cdk-explorer/lib/toolkit-internals.ts
megha-narayanan: Short answer: they aren't public, and probably shouldn't be if they aren't yet.

megha-narayanan: this follows the same pattern used by the `aws-cdk` CLI package see [packages/aws-cdk/lib/api-private.ts](https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/api-private.ts#L1-L10), which imports `createIgnoreMatcher` and other internal symbols from `toolkit-lib` via relative paths with the same `eslint-disable`.

ShadowCat567: based on what other packages in `@aws-cdk` do we might be able to do the deep import in the file where we need to get the info from and not have the intermediate file

packages/@aws-cdk/cdk-explorer/test/index.test.ts
ShadowCat567, line 1: you shouldn't need `index.test.ts` file, I think you can remove this

ShadowCat567 (approved):
Overall looks good, take a look at the `index.test.ts` file before merging!

PR #1591, opened 2026-06-03
fix(cdk-assets): re-tag docker image for multi-destination publishes

calabrla:
## Summary

- Fixes multi-region container image deployments failing with "An image does not exist locally with the tag" for the second+ destination
- When `WorkGraphBuilder` deduplicates the `asset-build` node for same-source assets, only one destination gets `docker tag`'d during `build()` — subsequent `publish()` calls for other destinations fail because the local tag doesn't exist
- Adds a lightweight `docker inspect` + `docker images --filter` + `docker tag` fallback in `publish()` to ensure the image is tagged locally before pushing

## Root Cause

`WorkGraphBuilder.addAsset()` creates one `asset-build` node per unique source (`assetId` + `genericSource` hash). When the same image is deployed to multiple regions, only the first destination's handler gets `build()` called by the graph executor. The other destinations only receive `publishEntry()`, which calls `docker push` on a URI that was never tagged locally.

## Test plan

- [ ] New test: `publish re-tags image when local tag is missing for a destination` — verifies full build+publish flow with multi-dest manifest
- [ ] New test: `publish re-tags from existing image when build was not called for destination` — simulates the graph executor scenario (`buildAssets: false`)
- [ ] Manual verification: deployed same container image to `ca-central-1` and `eu-west-2` simultaneously — both regions succeeded

---


mrgrain (changes requested):
Looks good, but we will need an integration test for this

mrgrain (approved):

PR #1539, opened 2026-05-21
feat(cli): cdk validate --online

kaizencc:
## Summary

Screenshot 2026-05-21 at 7 56 05 PM

Adds `cdk validate --online` which submits synthesized CloudFormation templates for early validation via non-executing change sets. This catches errors that offline policy plugins can't — invalid resource types, property validation failures, name conflicts, etc.

- Adds `online` option to `ValidateOptions` (defaults to `true`)
- Refactors `waitForChangeSet` into `waitForChangeSetReport` (non-throwing) + wrapper, extracting shared `prepareChangeSetEnv` helper to deduplicate with the diff change set path
- Adds `createValidationChangeSet()` — creates a non-executing change set per stack, collects early validation errors via the diagnoser, reports them as a synthetic "CloudFormation" plugin, always cleans up
- Extends `validate()` to merge offline (policy report) + online (CFN) results
- Wires `--online` / `--no-online` flag through CLI (`cli-config.ts`, `cli.ts`)
- Unit tests for the online path (mocked `createValidationChangeSet`)
- Integration tests with a fixture app using `AWS::Fake::DoesNotExist` to trigger CFN rejection

### Depends on
- #1527 (`cdk validate` CLI command behind `--unstable` flag)

## Test plan
- [x] Unit tests: offline path (12 tests), online path (4 tests), validate-formatting (9 tests)
- [ ] Integration tests: `cdk validate --online` catches invalid resource type, passes valid template, `--no-online` skips CFN validation
- [ ] Manual: `cd packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/validate-online-app && STACK_NAME_PREFIX=test node ../../../../../aws-cdk/bin/cdk --app "node app.js" --unstable=validate validate --online`

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license

PR #1592, opened 2026-06-03
feat: surface validation violations and CFN resources in LSP

megha-narayanan:
-  Adds an assembly reader that joins `cdk.out/`'s `tree.json` with each stack's manifest metadata into a `ConstructNode` tree carrying `logicalId`, CFN type, and source location.
- LSP publishes CDK validation violations as `Diagnostic`s anchored to the construct's TypeScript source line, with rule-level severity. 
- LSP serves `CodeLens` entries above each construct creation site summarising the CFN resources it produces.
- Source resolution covers `.ts` and `.js` (with sibling `.js.map`); non-TS apps degrade gracefully (no crash, no source-linked features). 

Fixes #

### Checklist

---


packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,177 @@
...
+    const assembly = new CloudAssembly(assemblyDir);
+    const rawTree = loadTree(assemblyDir);
+    const stacks = assembly.stacksRecursively;
+
+    const stackIndex = buildStackIndex(stacks);
+    const tree = rawTree
+      ? buildTree(rawTree, stackIndex, onWarn)
+      : [];
+
+    let violations: PolicyValidationReportJson | undefined;
```
ShadowCat567, line 56: Can you check whether online validations appear in this report? 
Since I think it might be possible that a customer runs `cdk validate` themselves and online validations appear in the report with offline validations.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,177 @@
...
+    return {
+      status: 'success',
+      data: { tree, violations, violationsError },
+    };
+  } catch (err) {
+    return { status: 'error', message: err instanceof Error ? err.message : String(err) };
+  }
+}
+
+// Raw tree.json node. The root tree.json holds the FULL hierarchy including
```
ShadowCat567: This comment is a little awkward to read, is the comment on the following line supposed to be a part of it?

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,130 @@
...
+    // don't re-attempt every lookup, and surface it once so a broken map
+    // doesn't masquerade as "no source map at all".
+    traceMapCache.set(jsFile, null);
+    onWarn?.(`Source map ${mapPath} failed to load: ${err instanceof Error ? err.message : String(err)}`);
+    return null;
+  }
+}
+
+/** Test-only: clear the source map cache between test cases. */
+export function _clearTraceMapCache(): void {
```
ShadowCat567: If this is test only logic, I don't think it should be here. Is there a way to have this logic live in the test file?

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+  if (loc.line < 1 || loc.column < 1) {
+    return { error: `invalid line/column: ${loc.line}:${loc.column}` };
+  }
+
+  // LSP positions are 0-based; sourceLocation is 1-based.
+  // Span to end-of-line so the squiggle is visible. The LSP spec defines
+  // Number.MAX_VALUE as the canonical end-of-line sentinel (vscode-
+  // languageserver-types' Position.create normalizes it to uinteger.MAX_VALUE),
+  // and conformant clients clamp character values past line length.
```
ShadowCat567: I don't think we need this much information about `Number.MAX_VALUE`
I would consider this to be sufficient information: 
```
  // LSP positions are 0-based; sourceLocation is 1-based.
  // Span to end-of-line with Number.MAX_VALUE so the squiggle is visible.
```

packages/@aws-cdk/cdk-explorer/test/core/assembly-reader.test.ts
```diff
@@ -0,0 +1,249 @@
...
+  test('parses tree from a flat assembly', () => {
+    dir = buildFlatAssembly({
+      stacks: [
+        { id: 'Stack1', resources: [{ id: 'MyBucket', logicalId: 'MyBucketF68F3FF0', cfnType: 'AWS::S3::Bucket' }] },
+        { id: 'Stack2', resources: [{ id: 'MyQueue', logicalId: 'MyQueueE6CA6235', cfnType: 'AWS::SQS::Queue' }] },
+      ],
+    });
+
+    const result = readAssembly(dir);
+    if (result.status !== 'success') throw new Error('expected success');
```
ShadowCat567: Why would this test not be successful? (same for the other tests that throw an error if `result.status !== 'success'`)

ShadowCat567 (commented):
Initial comments, will come back and do more of a review of the logic for the codelens + diagnostics. 
In general for LSP features, it would be helpful to have some screenshots of what you are implementing looks like in the IDE

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: yes, reformatted to make this clear.

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
megha-narayanan: fixed.

packages/@aws-cdk/cdk-explorer/test/core/assembly-reader.test.ts
megha-narayanan: fixed so we dont throw in tests

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: yep, moved.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,177 @@
...
+  | { readonly status: 'success'; readonly data: AssemblyData }
+  | { readonly status: 'not-found' }
+  | { readonly status: 'error'; readonly message: string };
+
+/**
+ * Reads a cdk.out/ directory and joins tree.json with each stack's manifest
+ * metadata to produce a ConstructNode tree where every CFN resource carries
+ * its logicalId, CFN type, and source location.
+ *
+ * Supports:
```
ShadowCat567: is there anything you know of that we don't support with this?

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,177 @@
...
+    const tree = rawTree
+      ? buildTree(rawTree, stackIndex, onWarn)
+      : [];
+
+    let violations: PolicyValidationReportJson | undefined;
+    let violationsError: string | undefined;
+    try {
+      violations = loadViolations(assemblyDir);
+    } catch (err) {
+      violationsError = err instanceof Error ? err.message : String(err);
```
ShadowCat567: errors are usually Objects, `JSON.stringify()` would probably give you nicer looking output than `String()` - similar comment with the other places you are using `String()` instead of `JSON.stringify()`

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,177 @@
...
+    type: cfnType,
+    logicalId,
+    sourceLocation,
+    children: Object.values(raw.children ?? {})
+      .filter((child) => !isCdkInternal(child.id))
+      .map((child) => buildNode(child, stackIndex, metadata, onWarn)),
+  };
+}
+
+const CDK_INTERNAL_IDS = new Set(['Tree', 'CDKMetadata', 'BootstrapVersion', 'CheckBootstrapVersion']);
```
ShadowCat567: this is mostly a style nit, but I would prefer of all of your const variables that are not in the scope of some function are grouped together at the top of the file beneath the imports

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: Yep, I think (in the actual version of validate) they do. the schema doesn't distinguish between online and offline violations, so we can't really filter them out. My thought would be that this is a good thing though, right? If we have violations detected, we should surface them the same way. ofc, if tehres no source location thats a separate issue and they get filtered out anyway.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Resolve a construct's creation site from its metadata entries. Returns
+ * undefined when no trace is present (non-TS apps) or when every frame is
+ * a skip-placeholder (framework-only call sites).
+ */
+export function resolveSourceLocation(
+  metadataEntries: readonly MetadataEntry[] | undefined,
+  cache: SourceMapCache,
+  onWarn?: WarnFn,
+): SourceLocation | undefined {
+  const frames = pickCreationFrames(metadataEntries);
```
ShadowCat567: what are `frames` in this context?

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
```diff
@@ -0,0 +1,126 @@
...
+    range: { start: { line, character }, end: { line, character: Number.MAX_VALUE } },
+  };
+}
+
+function buildDiagnostic(violation: PolicyViolationJson, range: Range): Diagnostic {
+  return {
+    range,
+    severity: severityFor(violation.severity),
+    code: violation.ruleName,
+    source: 'CDK Synth',
```
ShadowCat567: Why do all of our violations have a source of `CDK synth` could this instead be used to surface which plugin it came from?

megha-narayanan:
Ugly but functional surfacing of resources with source resolution
Screenshot 2026-06-04 at 4 17 04 PM

packages/@aws-cdk/cdk-explorer/test/_fixtures/README.md
```diff
@@ -0,0 +1,25 @@
+# Test Fixtures
+
+Most tests build their `cdk.out/` programmatically via `builders.ts` —
+keeps test intent in TypeScript and avoids drift when aws-cdk-lib or the
+cloud-assembly schema upgrades. `builders.test.ts` sanity-checks each
```
ShadowCat567: Do we stay on a certain version of aws-cdk-lib/cloud-assembly schema or does this update with them?

packages/@aws-cdk/cdk-explorer/test/_fixtures/README.md
ShadowCat567, line 1: I think I need more context about what these fixtures are and why we need them before I can properly review them

megha-narayanan:
And violations: Screenshot 2026-06-04 at 4 24 48 PM

ShadowCat567 (commented):
The biggest question I have is related to the fixtures, I don't think it's a pattern that is used elsewhere in this repo and I would like a bit more explanation in the readme about what their deal is

ShadowCat567:
> And violations: Screenshot 2026-06-04 at 4 24 48 PM

:O Do the squiggles change color based on severity?

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,185 @@
...
+import {
+  ArtifactMetadataEntryType,
+  Manifest,
+  type MetadataEntry,
+  type PolicyValidationReportJson,
+} from '@aws-cdk/cloud-assembly-schema';
+import { createSourceMapCache, resolveSourceLocation, type SourceLocation, type SourceMapCache, type WarnFn } from './source-resolver';
+
+/** A construct from tree.json plus the CFN metadata the LSP surfaces. */
+export interface ConstructNode {
```
otaviomacedo: Eventually we'll also need to store information about properties and their assignment locations. So, just think about how you're going to model it then, and if the current model is extensible for that. For example, you could add a new attribute to this interface called something like `properties`. Or you could decide to have a tree in which the nodes may be constructs or properties (and properties are children of constructs).

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
rix0rrr, line 1: Shouldn't this code live in `@aws-cdk/cloud-assembly-api` ?

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
rix0rrr: Pretty sure that online validations are not part of this report.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
rix0rrr: Just assume that `err` is always an `Error`, and reject any code that throws anything else during code review.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * Supports:
+ *  - Flat assemblies (stacks directly under App).
+ *  - Stage-based apps: each Stage produces a nested cloud-assembly artifact
+ *    under assembly-STAGE/. stacksRecursively descends into them.
+ *  - NestedStack contents: aws-cdk-lib emits the per-resource metadata for
+ *    nested-stack constructs into the PARENT stack's artifact metadata,
+ *    keyed by full construct path.
+ */
+export function readAssembly(assemblyDir: string, onWarn?: WarnFn): AssemblyReadResult {
```
rix0rrr: Why not return warnings on the successful result?

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// renderCallStackJustMyCode emits frames as "    at <name> (<file>:<line>:<col>)".
+// Anchoring on "(" avoids capturing the leading "at " into the file group.
+const FRAME_RE = /\(([^()\s][^()]*?):(\d+):(\d+)\)\s*$/;
+
+/**
+ * Mirrors toolkit-lib's findCreationStackTrace preference: prefer the
+ * aws:cdk:logicalId.trace, fall back to aws:cdk:creationStack.data.
+ */
```
rix0rrr: This shouldn't just *mirror*, it should literally be the same code path.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,140 @@
...
+  const line = Number(m[2]);
+  const column = Number(m[3]);
+  if (!Number.isFinite(line) || !Number.isFinite(column)) return undefined;
+  return { file: m[1], line, column };
+}
+
+/**
+ * Map a .js location to its original .ts via a sibling .js.map. Returns
+ * undefined when there's no map; the caller falls back to the .js location.
+ */
```
rix0rrr: Don't do that. You are now putting work on the caller. The caller is forced to be like:

```ts
// ❌ Why make me do extra work?
const realLoc = mapJsToOriginalSource(loc) ?? loc;

// ✅ Just always return the best value of this object
const realLoc = mapJsToOriginalSource(loc);
```

Also, the very first line of this function belies the "I will return `undefined` if I have nothing to do" contract.

In actual fact, the contract seems to be "I will sometimes return the original object and sometimes a new one and sometimes nothing, in which case you should probably fall back to the original object".

That doesn't seem like a great contract.

Here is a golden programming rule: reduce the number of `if`s as much s you can. Fewer ifs = fewer code paths = more executions hit the common paths = better testing coverage and easier to reason about

(And ??` is an `if` in disguise!)

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,140 @@
...
+    : path.resolve(path.dirname(loc.file), orig.source);
+
+  return { file: resolvedFile, line: orig.line, column: orig.column + 1 };
+}
+
+function loadTraceMap(jsFile: string, cache: SourceMapCache, onWarn?: WarnFn): TraceMap | null {
+  const cached = cache.get(jsFile);
+  if (cached !== undefined) return cached;
+
+  const mapPath = jsFile + '.map';
```
rix0rrr: Source maps can also be inlined. You can't rely on just this file existing. 

In fact the filename can be different, and the real filename is in the source file.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,140 @@
...
+  return { file: m[1], line, column };
+}
+
+/**
+ * Map a .js location to its original .ts via a sibling .js.map. Returns
+ * undefined when there's no map; the caller falls back to the .js location.
+ */
+function mapJsToOriginalSource(
+  loc: SourceLocation,
+  cache: SourceMapCache,
```
rix0rrr: Since `cache` needs to be passed around a bunch, I wonder if it makes sense to put all these functions on a class that has the cache as state.

The ideal number of parameters to a function is as few as possible. If you can reduce that count by sharing state, go for it.

packages/@aws-cdk/cdk-explorer/lib/core/tree-utils.ts
```diff
@@ -0,0 +1,15 @@
+import type { ConstructNode } from './assembly-reader';
+
+/** Flatten a construct tree into a Map keyed by node.path for O(1) lookups. */
+export function indexNodesByPath(tree: readonly ConstructNode[]): Map<string, ConstructNode> {
```
rix0rrr: I feel there is a data type trying to come out here. `class ConstructIndex {}` ?

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
```diff
@@ -0,0 +1,69 @@
...
+function titleFor(resources: readonly ResourceLensInfo[]): string {
+  if (resources.length === 1) {
+    const r = resources[0];
+    return `Creates: ${r.cfnType} [logical: ${r.logicalId}]`;
+  }
+  const ids = resources.map((r) => r.logicalId).join(', ');
+  return `${resources.length} resources: ${ids}`;
+}
+
+function walk(nodes: readonly ConstructNode[], visit: (node: ConstructNode) => void): void {
```
rix0rrr: Duplication of `walk` function.

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
```diff
@@ -0,0 +1,69 @@
+import { pathToFileURL } from 'url';
+import { type CodeLens, type Range } from 'vscode-languageserver/node';
+import type { ConstructNode } from '../core/assembly-reader';
+
+/**
+ * Build CodeLens entries for a single source file. For every construct whose
+ * sourceLocation matches fileUri, group by line and emit one lens per line
+ * summarising the CFN resources produced there.
+ */
+export function codeLensesForFile(tree: readonly ConstructNode[], fileUri: string): CodeLens[] {
```
rix0rrr: Shouldn't this function take a `ConstructIndex` class if we define one?

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
```diff
@@ -0,0 +1,69 @@
...
+  const matches: Array<{ line: number; resource: ResourceLensInfo }> = [];
+  walk(tree, (node) => {
```
rix0rrr: Feels nicer if you make this an `array.filter().map()`.

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  // Multiple resources can map to one line when an L2 construct fans out
+  // (e.g. an L2 producing a primary resource + auxiliary resources).
+  const byLine = new Map<number, ResourceLensInfo[]>();
+  for (const m of matches) {
+    const list = byLine.get(m.line);
+    if (list) list.push(m.resource);
+    else byLine.set(m.line, [m.resource]);
+  }
```
rix0rrr: Make a `groupBy()` combinator, then use that.

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const lenses: CodeLens[] = [];
+  for (const [line, resources] of byLine) {
+    // Empty command name = title-only lens; click does nothing for now.
+    lenses.push({
+      range: lineRange(line),
+      command: { title: titleFor(resources), command: '' },
+    });
+  }
+  return lenses;
```
rix0rrr: Much nicer to use a `.map()`.

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
```diff
@@ -0,0 +1,126 @@
...
+    readonly constructPath: string;
+    readonly reason: string;
+  }>;
+}
+
+/**
+ * Convert a validation report into LSP diagnostics keyed by file URI.
+ * Violations whose construct path is unknown, has no source location, or
+ * points outside TypeScript are dropped (with a reason) rather than thrown.
+ */
```
rix0rrr, line 32: I thought we were going to represent warnings without a source location in some default location. Is this an initial iteration, or is this the final plan?

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
```diff
@@ -0,0 +1,126 @@
...
+): ResolvedAnchor | { readonly error: string } {
+  if (!constructPath) return { error: 'empty constructPath' };
+
+  const node = nodesByPath.get(constructPath);
+  if (!node) return { error: 'no tree node for constructPath' };
+
+  const loc = node.sourceLocation;
+  if (!loc) return { error: 'node has no sourceLocation (non-TS or framework-only trace)' };
+
+  if (!loc.file.endsWith('.ts') && !loc.file.endsWith('.tsx')) {
```
rix0rrr: I'd rather have "limiting to TypeScript for now" be a type of processing we do consciously do at the start, rather than something in the middle of different kinds of error recovery code.

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+function anchor(
+  constructPath: string | undefined,
+  nodesByPath: Map<string, ConstructNode>,
+): ResolvedAnchor | { readonly error: string } {
+  if (!constructPath) return { error: 'empty constructPath' };
+
+  const node = nodesByPath.get(constructPath);
+  if (!node) return { error: 'no tree node for constructPath' };
```
rix0rrr: What are these errors for?

If we show them to users, I would like them to look like messages that are appropriate to show to a user `(could not find this node in tree.json)`.

If we're not going to show this to a user, then they could be error codes or we don't have this field at all.

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const loc = node.sourceLocation;
+  if (!loc) return { error: 'node has no sourceLocation (non-TS or framework-only trace)' };
+
+  if (!loc.file.endsWith('.ts') && !loc.file.endsWith('.tsx')) {
+    return { error: `non-TypeScript source: ${loc.file}` };
+  }
+  if (loc.line < 1 || loc.column < 1) {
+    return { error: `invalid line/column: ${loc.line}:${loc.column}` };
```
rix0rrr: But we could still relate it to a file? Can't we stick it at line 1 of this file then?

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -96,13 +176,19 @@ export function startServer(options: LspServerOptions): void {
     new StreamMessageWriter(options.writable),
   );
 
-  const onSynthRequest = options.onSynthRequest ?? (() => {
+  const handlers = createLspHandlers({
+    onSynthRequest: options.onSynthRequest,
+    readAssembly: options.readAssembly,
+    logger: connection.console,
+    onPublishDiagnostics: (uri, diagnostics) => {
+      void connection.sendDiagnostics({ uri, diagnostics });
```
rix0rrr, line 186: What's the `void` here, and why is that okay?

Is that a silenced "you shall `await` this" linter error?

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
kaizencc: depends. if running `cdk synth`, online validation is not a part of the report. if running `cdk validate --online`, then it will be a part of the report.

the plugin name will be `CloudFormation` -- see https://github.com/aws/aws-cdk-cli/pull/1539/changes#diff-efa522e30d6c1c6a99aa8a5da647de2173348e9dd3c571bc2f37ec0916adbc33R758-R762 -- so we can filter those out if needed. or just run `validate --offline`

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: Not that I know of, mostly added this comment for clarity of what I had tested and designed for.

packages/@aws-cdk/cdk-explorer/lib/core/tree-utils.ts
megha-narayanan: yep, made this a new data type, and moved all of this logic to cloud-assembly-api (see prev comments)

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
megha-narayanan: see top comment but I put ConstructIndex in cloud-assembly-api so it's reusable, which is why it's generic over the node type, the explorer uses ConstructIndex<ConstructNode> to keep sourceLocation. If you'd rather it stay explorer local, easy to change, lmk.

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
megha-narayanan: yep. rewrote this whole file in a more functional-programming style

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
megha-narayanan: done. see aboev

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
megha-narayanan: fixed

packages/@aws-cdk/cdk-explorer/lib/lsp/codelens.ts
megha-narayanan: fixed

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: Done. Moved the tree builder into `@aws-cdk/cloud-assembly-api` as `buildConstructTree(assembly, decorate)` it owns the tree.json parse + stack-metadata join and produces a generic `ConstructTreeNode`. The explorer passes a decorate callback that adds `sourceLocation`; that stays here since it needs trace-mapping (dependency). I also moved `ConstructIndex` there so it's reusable, which is why it's generic over the node type (ConstructIndex<T>); the explorer uses ConstructIndex<ConstructNode> to keep its sourceLocation. If you'd rather the index stay specific to teh explorer, easy to change. lmk if this looks good, it adds a small public API surface.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: Yep, i think the model is extensible for this. I'd add an optional `properties?: PropertyNode[]` to the node (PropertyNode { name; sourceLocation }) rather than making properties tree children, so the construct index/iteration and CodeLens stay construct-only. The data's already available `PROPERTY_ASSIGNMENT` metadata ({ propertyName, stackTrace }) flows through the same `buildConstructTree` decorate callback that resolves creation-site locations, so resolving an assignment location reuses the exact source-map machinery. Planning to land the field + populate it as a follow-up to keep this PR scoped.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: lines of the creation stack trace. made this better documented

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: done!

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: you're totally right. I was trying to evoid messing with APIs, but I need to do it at some point. I exported `findCreationStackTrace` from toolkit-lib's source-tracing so this can be the literal same code path; it takes (stack, constructPath). While there I also exported findMutationStackTraces for the property-assignment locations otavio raised. Source-map decode stays in the explorer.  ok to add these two to toolkit-lib's public API?

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: fixed.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: Got it, made this a class so we no longer pass in cache and onWarn

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: done, warnings are part of the result now: `readAssembly` returns `warnings: string[] `on the success data instead of taking an onWarn callback; SourceResolver accumulates them and the server logs from the returned array.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: fixed

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
megha-narayanan: fixed this. pass the plugin through.

packages/@aws-cdk/cdk-explorer/test/_fixtures/README.md
megha-narayanan: there's no version pin. The package depends on `@aws-cdk/cloud-assembly-schema` with an `any-future `version range, so when the schema bumps we pick it up automatically. 
 
The only place a version literal appears is `ASSEMBLY_SCHEMA_VERSION = '53.0.0'` in `builders.ts`. That's the version we write into test fixtures, not a version we read against.

packages/@aws-cdk/cdk-explorer/test/_fixtures/README.md
megha-narayanan: The reader consumes a real cdk.out/, so tests need realistic cdk.out/ directories.  I do that programmatically rather than checking in pre-synthesized cdk.out/s for two reasons: (1) tests read as a typed spec of the shape under test (buildNestedAssembly, buildNonTypeScriptAssembly, etc.) rather than an opaque directory, and (2) checked-in cdk.out/s drift every time aws-cdk-lib or the cloud-assembly schema bumps, and the diffs are over generated JSON. 

The one exception is source-maps/: real tsc output, because a builder can fabricate stack traces but can't fabricate a .js.map tsc would accept.

I updated the readme... does this help? happy to chat if it needs more explanation/justfiication

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
megha-narayanan: This was an initial iteration, but the more I think about it I should just update this now. Fixed. File-known-but-no-line violations now anchor at the top of the file rather than dropping (see the comment below). Truly location-less violations (no file at all) can't become LSP diagnostics, so they're dropped and surfaced in the warning log. That's intentional for now; lmk if you think there's some good project-level default instead you'd prefer over dropping.

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
megha-narayanan: They're logged to the user's Output panel (in VSCode), so I reworded them as nicer human readable messages as you suggested

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
megha-narayanan: refactored to do this

packages/@aws-cdk/cdk-explorer/lib/lsp/diagnostics.ts
megha-narayanan: Done, when the construct's file is known but the line/column aren't, it now anchors at the top of the file ({0,0}) instead of being dropped.

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
megha-narayanan: yep, thats exactly what it is. connection.sendDiagnostics is an LSP notification, which is fire-and-forget, so theres no response to await, so the void is to silence the linter.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: Very good point. Adressed this, `loadTraceMap` now honors the` //# sourceMappingURL= `directive via `convert-source-map,` so it handles maps inlined as a data: URI as well as external maps under any filename. Sources are resolved through the map's URL, so sourceRoot and maps that live in a different directory than the .js resolve correctly too. Added tests for the inline, renamed-external, and sourceRoot cases.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+    return undefined;
+  }
+
+  /**
+   * Map a .js location to its original .ts via a sibling .js.map. Returns the
+   * input location unchanged when it isn't a .js file or has no usable map.
+   */
+  private mapJsToOriginalSource(loc: SourceLocation): SourceLocation {
+    if (loc.file.endsWith('.ts') || loc.file.endsWith('.tsx') || !loc.file.endsWith('.js')) return loc;
```
otaviomacedo: This function assumes that the file extension is either ".ts", ".tsx" or ".js". But there is nothing guaranteeing this. If we add Python stack traces along with the JS ones, for example, this will start receiving file names ending in ".py". We need to filter out everything that is not in this allow-list before calling the function.

otaviomacedo (changes requested):

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: good catch for the incoming jsii changes, just pushed an update, let me know what you tjink

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
rix0rrr: Does `cdk validate` *rewrite* the files in `cdk.out` ?

.projenrc.ts
```diff
@@ -1681,10 +1681,13 @@ const cdkExplorer = configureProject(
       'vscode-languageserver-textdocument@^1',
       'vscode-jsonrpc@^8',
       'express@^4',
+      '@jridgewell/trace-mapping@^0.3',
+      'convert-source-map@^2',
```
rix0rrr, line 1685: Every new dependency is a liability.

What's the research thas has gone into picking these specific ones?

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -0,0 +1,81 @@
+import * as fs from 'fs';
```
rix0rrr, line 1: Didn't you say you were moving this code?

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,139 @@
...
+  readonly column: number;
+}
+
+/**
+ * Resolves construct creation stack traces to user source locations, caching
+ * parsed source maps across calls. Source-map parsing is not free and one
+ * cdk.out tree asks for the same .js.map many times, so create one resolver
+ * per assembly read and reuse it for every construct.
+ */
+export class SourceResolver {
```
rix0rrr: Let's call this `SourceMapResolver`, because that's what it is.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    // aws-cdk-lib's renderCallStackJustMyCode (in node_modules/aws-cdk-lib/core/
+    // lib/stack-trace.js) pre-filters node_modules/node:internal frames into
+    // skip-placeholder lines. Those don't match FRAME_RE, so the first frame
+    // that parses IS the user call site.
+    for (const frame of frames) {
+      const parsed = parseFrame(frame);
+      if (!parsed) continue;
+      if (!isSupportedSourceFile(parsed.file)) return undefined;
+      return this.mapJsToOriginalSource(parsed);
```
ShadowCat567, line 51: for source tracing with non-TS apps would we need to make more methods like this or is there a way we can extend this logic to work for any language? (we might not know the answer to this until source tracing with non-TS gets implemented)

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+    return undefined;
+  }
+
+  /**
+   * Map a .js location to its original .ts via a sibling .js.map. Returns the
+   * input location unchanged when it isn't a .js file or has no usable map.
+   */
+  private mapJsToOriginalSource(loc: SourceLocation): SourceLocation {
```
ShadowCat567, line 60: Javascript is supported as a language in CDK, what does this do if someone has written their app entirely in JS (no TS files)?

packages/@aws-cdk/cdk-explorer/test/_fixtures/builders.ts
```diff
@@ -0,0 +1,392 @@
...
+ * Programmatic fixture builders. Each builder writes a minimal cdk.out/ to a
+ * temp dir and returns its path; cleanupFixture removes it. CloudAssembly
+ * construction enforces the schema, which is what catches drift in builders.
+ *
+ *   const dir = buildFlatAssembly({ stacks: [...] });
+ *   try { ... } finally { cleanupFixture(dir); }
+ */
+
+/** Tracks @aws-cdk/cloud-assembly-schema's revision (currently 53). */
+const ASSEMBLY_SCHEMA_VERSION = '53.0.0';
```
ShadowCat567: does this need to get manually updated whenever `@aws-cdk/cloud-assembly-schema` version changes?

packages/@aws-cdk/cdk-explorer/test/_fixtures/builders.ts
```diff
@@ -0,0 +1,392 @@
...
+
+  const parentChildren: Record<string, unknown> = {};
+  for (const r of parent.resources) {
+    parentChildren[r.id] = resourceTreeNode(parent.id, r);
+  }
+  for (const ns of parent.nestedStacks) {
+    parentChildren[ns.id] = {
+      id: ns.id,
+      path: `${parent.id}/${ns.id}`,
+      constructInfo: { fqn: 'aws-cdk-lib.NestedStack', version: '2.245.0' },
```
ShadowCat567: version is hard-coded here, do you expect this to get updated with cdk version or does the value not really matter so it does not need to change?
Whenever you have kind of magic numbers like this (since several other things have a version) can you put it in a constant so if it does need to manually change the change only happens in one place?

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: it'll return the original loc in the .js, which is teh correct source mapping

.projenrc.ts
megha-narayanan: Here's where I am in that research. The two libs do different jobs, `convert-source-map` extracts the map from the .js, `trace-mapping` decodes it to the original .ts position.  Right now, both of these deps are already in our tree, but only as dev-indirect deps which both resolve in main's yarn.lock today via jest/vitest/babel.

For what it's worth on the safety signals both are MIT, have 0 advisories in OSV across all versions, `convert-source-map` has zero dependencies, and `trace-mapping` only pulls its author's own `resolve-uri/sourcemap-codec` and is maintained by a Babel core member (it's what Babel/Jest/Vite use).  Neither have a dependabot config, and 'trace-mapping' doesn't have a security policy (but 'convert-source-map' does). Also, `convert-source-map` isn't actively maintained, but it looks to be feature complete.  

The extractor is basically two regexes + base64 + JSON.parse, so I can inline the slice I use and drop that dep. The one thing I actually need from it is their sourceMappingURL regexes, and I definitely don't think I can lift those verbatim without attribution. What's our convention there? Is there a method for attribution, or would you rather I just keep the dep?

For the decoder, looking for alternatives, I checked native Node, `module.findSourceMap() `only works for files the process has actually loaded with `--enable-source-maps`, and we read emitted .js out of cdk.out that we never execute, so it can't help with extraction. The SourceMap class could do the decode though.  The catch is it returns the raw source path without applying `sourceRoot` or resolving relative sources, so I'd have to do that myself. Most of the edge cases trace-mapping handles don't (shouldn't :P) really show up in cdk.out output, those are web-bundler artifacts. So native decode is viable also; just depends on if its worth owning the edge cases. 

I'm not sure what our bar is for a runtime dep, so would really appreciate your input, given this info.  Lmk what you think.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: Yep, the generic tree machinery moved, so `buildConstructTree`, `ConstructIndex`, and `ConstructTreeNode` are in `@aws-cdk/cloud-assembly-api `now, and this file imports them. What's left here is the explorer-specific layer, so `readAssembly`, the `ConstructNode` type that adds `sourceLocation`, and `loadViolations`. Those stay because they pull in t`oolkit-lib` (`findCreationStackTrace`) and the source-map resolver. `cloud-assembly-api` can't depend on toolkit-lib, since that would be circular. So cloud-assembly-api builds the generic tree and the explorer decorates it with source locations via a callback

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: it does call `cdk synth`, which does rewrite them.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: done

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: we shouldn't need a separate resolver per language.... source maps are a TSthing, so a Python/Java creation frame already points at the original source, nothing to decode. 

So extending is mostly (1) add the extension to the allow-list and (2) parse that language's frame format, depending on how that stabilizes upstream.

packages/@aws-cdk/cdk-explorer/test/_fixtures/builders.ts
megha-narayanan: No, and I've changed it so it can't drift. The package depends on cloud-assembly-schema and the schema is forward-readable, the loader only throws if a manifest's major is higher than the installed schema, so a fixture written at an older revision always loads. Hardcoding '53.0.0' was just the current value; I've replaced it with `Manifest.version()`, so fixtures always carry exactly the installed schema version

packages/@aws-cdk/cdk-explorer/test/_fixtures/builders.ts
megha-narayanan: these dont actually NEED to be bumped ever, but I pulled them into constants regardless

.projenrc.ts
rix0rrr: If it's maintained by someone whose full-time job is JavaScript things, and they don't have too many transitive dependencies (and are somewhat popular?) I'm fine with adding these. Thanks for the digging.

packages/@aws-cdk/cdk-explorer/test/_fixtures/builders.ts
rix0rrr: I think keeping a literal `53.0.0` in there is the correct approach. As you said, we should always be able to read any old assembly, so if nothing else this is another implicit test that we uphold that contract.

In fact, tracking the "most recent" version number while keeping the contents of the file static might be a bug in disguise: what if (purely made up scenario), in version `60.0.0`, we have decided that all keys should be `UPPERCASE`? We could write the CLI to check the version number and read the file appropriately: `>=60` = keys are uppercase, `<60` keys are lowercase  -- but now the manifest you would be generating here would be lying, because it would **pretend** to be version `60` while actually having keys in lowercase.

packages/@aws-cdk/cdk-explorer/test/_fixtures/builders.ts
rix0rrr: I appreciate the "insist on high standards" here, but I think this is a bit of overkill :).

I would recommend pulling out into constants if:

- The value is likely to need to be changed
- You can attach a NAME to that value that will make it easier to understand.

In this case, I think neither of those rules apply, so you don't strictly need to.

But that's just my opinion, and it could be the case (in fact, as demonstrated -- **is** the case) that people will find the code harder to read if the values *aren't* extracted. So this one is debatable.

My general rule would be: if you can, do things for objective reasons 😉

rix0rrr (approved):
Provisional approval! That means I expect you to the address the last few comments, but I trust you to fix those and I don't need to see the PR again before you merge.

packages/@aws-cdk/cdk-explorer/test/_fixtures/builders.ts
megha-narayanan: Ah, true. Good call, i reverted this.

ShadowCat567 (approved):

Issue #1217, opened 2026-03-06
(cli) CDK CLI hangs indefinitely after command completion in CI environments (GitLab/Kubernetes) - v2.1034.0+


jshah04:
### Describe the bug

After upgrading AWS CDK CLI from v2.1033.0 to v2.1034.0+, the CLI process hangs indefinitely after cdk deploy or cdk diff commands complete successfully in GitLab CI (Kubernetes executor). The command output shows completion (e.g., "✨ Number of stacks with differences: 0") but the process never exits, requiring manual cancellation.
This issue does not occur when running the same commands locally on macOS - only in CI environments.

### Investigation findings:
Telemetry connects despite being disabled: Setting CDK_DISABLE_CLI_TELEMETRY=1 shows disableTelemetry: 1 in verbose output, but logs still show Endpoint Telemetry connected and Telemetry Sent Successfully.

--no-telemetry flag partially helps: Using --no-telemetry correctly shows telemetry: false and Endpoint Telemetry NOT connected, but the process still hangs.

Another likely root cause: AWS SDK v3 HTTP connections (to STS, CloudFormation, S3, SSM) are not being properly closed, keeping the Node.js event loop alive. In CI environments without a TTY, these connections don't trigger cleanup.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Last Known Working CDK Library Version

2.1033.0

### Expected Behavior

The CDK CLI process should exit cleanly (exit code 0) after the command completes, just as it does in local terminal environments.

### Current Behavior

1. Command runs and completes successfully
2. Output shows completion message (e.g., ✨ Command time: 69.51s)
3. With CDK_DISABLE_CLI_TELEMETRY=1: Telemetry still connects and sends data, then process hangs
4. With --no-telemetry: Telemetry doesn't connect, but process still hangs
5. Process must be manually killed/canceled after several minutes

### Reproduction Steps

1. Create a GitLab CI pipeline using Kubernetes executor
2. Run any CDK command (cdk diff, cdk deploy, cdk synth)
3. Observe the command completes successfully with output
4. Process hangs indefinitely instead of exiting

```
# .gitlab-ci.yml
cdk-diff:
  image: node:22
  script:
    - npm install -g aws-cdk
    - export CDK_DISABLE_CLI_TELEMETRY=1
    - cdk diff --ci -c env=development -v -v
```

### Additional Information/Context

Relevant log snippets:

# With env var only (telemetry still connects despite disable flag)
[...]

PR #1601, opened 2026-06-05
feat(cli): [WIP] extend `--debug` to report leaked async handles at exit

sai-ray:
WIP
 
Extends the existing `--debug` flag to also dump Node async resources still keeping the CLI alive at exit, with the source location where each was created. Helps diagnose CLI hangs in CI environments. Refs #1217.

When `--debug` is not set, no tracking happens (zero overhead).
 
### Checklist

---


sai-ray:
Closing in favor of #1651, which rebuilds this cleanly on top of the `--debug-cli` flag (#1604).

PR #1603, opened 2026-06-05
refactor(cli): use the list action from toolkit-lib in the CLI

mrgrain:
Relates to #959

The `cdk list` command kept its own stack-selection and synthesis logic (the `listStacks` helper) in parallel to the toolkit-lib `list` action. That duplicates behavior that already lives in the toolkit and means the two can drift. This routes the CLI through the toolkit-lib `list` action so there is a single source of truth for how stacks are selected and synthesized.

The wrinkle is output. The `list` action emits its result as a single `CDK_TOOLKIT_I2901` message, but the CLI needs to render that differently depending on `--long`, `--json` and `--show-dependencies`, and those flags are not something the action should know about. Rather than special-casing the message inside the io host, this introduces a small event-listener API on `CliIoHost`: `on`/`once` to observe a message by its code, and `transform`/`transformOnce` to replace a message or consume it entirely by returning `undefined`. The `list` command registers a one-shot transformer that formats the payload (extracted into a pure, unit-tested `formatStackList` helper), so the presentation logic stays with the command that owns it.

The same mechanism replaces the previously hardcoded `isStackActivity` branch in `notify`: stack-activity messages (`CDK_TOOLKIT_I5501`/`I5502`/`I5503`) are now routed to the activity printer and consumed via transformers registered at construction time, which keeps `notify` free of message-specific knowledge.

### Checklist

---

### DIFF
--- a/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
+++ b/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
@@ -3,11 +3,11 @@ import * as util from 'node:util';
 import { RequireApproval } from '@aws-cdk/cloud-assembly-schema';
 import { ToolkitError } from '@aws-cdk/toolkit-lib';
 import type { HotswapResult, IIoHost, IoMessage, IoMessageCode, IoMessageLevel, IoRequest, ToolkitAction } from '@aws-cdk/toolkit-lib';
-import type { Context } from '@aws-cdk/toolkit-lib/lib/api';
 import * as chalk from 'chalk';
 import * as promptly from 'promptly';
-import type { IoHelper, ActivityPrinterProps, IActivityPrinter } from '../../../lib/api-private';
+import type { IoHelper, ActivityPrinterProps, IActivityPrinter, IoMessageMaker, IoDefaultMessages } from '../../../lib/api-private';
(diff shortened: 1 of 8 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -324,32 +344,148 @@ export class CliIoHost implements IIoHost {
...
+          listeners.splice(index, 1);
+        }
+      }
+
+      if (listener.transform) {
+        if (result === undefined) {
+          // The transformer consumed the message; stop and signal a drop.
+          return undefined;
+        }
+        current = result;
```
rix0rrr: The transformer should not be allowed to change the message code, otherwise the `listeners` list is now incorrect.

More broadly: isn't this capability too powerful for what we need? We want to apply a formatter to replace `message`, and nothing else. Why then add powerful capabilities that we don't need, with the result that they can interact in unexpected ways with other features and become hard to reason about?

mrgrain: Yeah that's fair. I had started with just the on/once listeners but that felt hard work for each listener to also be responsible to do the actual logging. Good idea to lock it down to allow only updating the message. 👍

rix0rrr (approved):

PR #1604, opened 2026-06-05
feat(cli): add --debug-app and --debug-cli flags

mrgrain:
## Granular debug flags

The `--debug` flag now splits into two independent targets:

- **`--debug-app`** — Sets `CDK_DEBUG=true` for the synthesized app. Logs additional information during synthesis (e.g. token creation stack traces). Slows down synthesis.
- **`--debug-cli`** — Enables verbose AWS SDK tracing in the CLI itself.

`--debug` continues to work as before and implies both `--debug-app` and `--debug-cli`.

### Configuration

The flags are also available in `cdk.json` / `~/.cdk.json`:

```json
{
  "debugApp": true,
  "debugCli": true
}
```

An existing `"debug": true` in config files implies both targets for backwards compatibility.

### `cdk doctor` improvements

`cdk doctor` now reports some CLI configuration (verbosity level, active debug targets).

```
$ cdk doctor --debug
```

image

---


packages/aws-cdk/lib/cli/util/debug.ts
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Debug the CLI itself (enables AWS SDK tracing).
+   */
+  readonly cli: boolean;
+}
+
+const DEBUG_ALL: DebugMode = { app: true, cli: true };
+const DEBUG_NONE: DebugMode = { app: false, cli: false };
+const DEBUG_APP: DebugMode = { app: true, cli: false };
+const DEBUG_CLI: DebugMode = { app: false, cli: true };
```
mrgrain: instead, can we just return a list of enums that are selected for debug? than the check is `debug.includes(DebugMode.APP)`. though I like the simplicity of `debugMode.cli`. But maintaining these constants is yuck.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -84,12 +85,13 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
-  // If users request debug information (which switches on CDK_DEBUG=1 to get the
-  // CDK app to emit more information), they probably are troubleshooting and we've
-  // historically decided they then want more logging from the CLI as well.
-  // There is currently no way to get the CDK to emit tracing information that
-  // does not cause the CLI to barf logs to the console.
-  setSdkTracing(argv.debug || argv.verbose > 2);
+  // `--debug` is a multi-state flag (see `normalizeDebugArgs`/`interpretDebugMode`):
+  //   --debug / --debug=all  -> debug everything (app + CLI)
+  //   --debug=app            -> debug only the CDK app (sets CDK_DEBUG, see user-configuration.ts)
+  //   --debug=cli            -> debug only the CLI (verbose AWS SDK tracing)
```
mrgrain: can be shorter

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -100,6 +102,14 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
   await ioHost.defaults.debug('CDK Toolkit CLI version:', versionWithBuild());
   await ioHost.defaults.debug('Command line arguments:', argv);
 
+  if (!isRecognizedDebugValue(argv.debug)) {
+    await ioHost.defaults.warn(
+      `Unrecognized value '${argv.debug}' for --debug; enabling full debug output. ` +
+      `Expected no value, or one of: ${DEBUG_VALUE_CHOICES.map((c) => `--debug=${c}`).join(', ')}. ` +
+      'Note that --debug takes its value with "=" (e.g. --debug=app), otherwise the following argument may be consumed.',
+    );
+  }
```
mrgrain: can we integrate this with the unknown options check? it feels different than yargs as well.

packages/aws-cdk/lib/cli/user-configuration.ts
```diff
@@ -303,7 +304,10 @@ export async function commandLineArgumentsToSettings(ioHelper: IoHelper, argv: A
     build: argv.build,
     caBundlePath: argv.caBundlePath,
     context,
-    debug: argv.debug,
+    // `--debug` is a multi-state flag; only "app" debugging (--debug / --debug=all /
+    // --debug=app) should set CDK_DEBUG for the app. CLI-only debugging (--debug=cli)
+    // must not. See cloud-assembly/environment.ts which turns this setting into CDK_DEBUG.
+    debug: interpretDebugMode(argv.debug).app,
```
mrgrain: well this is awakard. Now settings.get('debug') isn't clear on what this is

packages/aws-cdk/test/cli/cli.test.ts
```diff
@@ -181,6 +195,37 @@ describe('exec verbose flag tests', () => {
   });
 });
 
+describe('exec --debug flag wiring', () => {
```
mrgrain: we can skip this

packages/aws-cdk/lib/cli/util/debug.ts
mrgrain: Done — `interpretDebug()` now returns `DebugMode[]` (enum `APP`/`CLI`) and the constant objects are gone. Callers read as `debug.includes(DebugMode.APP)` / `.includes(DebugMode.CLI)`.

packages/aws-cdk/lib/cli/cli.ts
mrgrain: Trimmed to a single line.

packages/aws-cdk/lib/cli/cli.ts
mrgrain: I dropped the bespoke warning entirely rather than integrating it. After `normalizeDebugArgs` rewrites a bare `--debug` to `--debug=all`, the only way to hit an unrecognized value is an explicit `--debug=typo`, and `interpretDebug` already fails open to full debug — so a silent fail-open felt cleaner than a yargs-flavoured message. Happy to wire it into the unknown-options warning instead if you'd prefer an explicit heads-up.

packages/aws-cdk/lib/cli/user-configuration.ts
mrgrain: Reworked to `debug: interpretDebug(argv.debug).includes(DebugMode.APP)` with a one-line comment. I kept the `debug` settings key as-is because it's the cdk.json-facing key that `environment.ts` turns into `CDK_DEBUG`, so renaming it would be a breaking change to user config. Open to renaming/splitting the setting if you think it's worth it.

packages/aws-cdk/test/cli/cli.test.ts
mrgrain: Removed — also dropped the `setSdkTracing` mock/import and the debug handling I'd added to the parse mock.

rix0rrr:
> This is additive rather than breaking: a bare --debug is unchanged, and --debug=app/--debug=cli previously coerced to false under the boolean option, so they did nothing before.

At the expense of losing the way to write `--debug app` I suppose? It *must* be written as `--debug=app` ?

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/environment.ts
```diff
@@ -89,7 +89,9 @@ export function synthParametersFromSettings(settings: Settings): {
     env: {
       // An environment variable instead of a context variable, so it can also
       // be accessed in framework code where we don't have access to a construct tree.
-      ...settings.get(['debug']) ? { CDK_DEBUG: 'true' } : {},
+      // `debugApp` is the dedicated setting; `debug` is honored for backwards
+      // compatibility with older configurations that set it directly.
+      ...(settings.get(['debugApp']) ?? settings.get(['debug'])) ? { CDK_DEBUG: 'true' } : {},
```
mrgrain: this should only check for `debugApp`. `--debug` or a `cdk.json` `debug:true` should however imply both `debugApp: true` and `debugCli: true`. add tests as well.

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -29,7 +29,9 @@ export async function makeConfig(): Promise<CliConfig> {
...
       'json': { type: 'boolean', alias: 'j', desc: 'Use JSON output instead of YAML when templates are printed to STDOUT', default: false },
       'verbose': { type: 'boolean', alias: 'v', desc: 'Show debug logs (specify multiple times to increase verbosity)', default: false, count: true },
-      'debug': { type: 'boolean', desc: 'Debug the CDK app. Log additional information during synthesis, such as creation stack traces of tokens (sets CDK_DEBUG, will slow down synthesis)', default: false },
+      'debug': { type: 'boolean', default: undefined, desc: 'Produce more detailed output to help diagnose unexpected behavior. Implies --debug-app and --debug-cli. Note that this will slow everything down' },
+      'debug-app': { type: 'boolean', default: undefined, desc: 'Debug the CDK app. Logs additional information during synthesis, such as creation stack traces of tokens (sets the CDK_DEBUG environment variable, will slow down synthesis)' },
+      'debug-cli': { type: 'boolean', default: undefined, desc: 'Debug the CDK CLI itself (enables verbose AWS SDK tracing)' },
```
mrgrain:
```suggestion
      'debug': { type: 'boolean', default: false, desc: 'Produce more detailed output to help diagnose unexpected behavior. Implies --debug-app and --debug-cli. Note that this will slow everything down' },
      'debug-app': { type: 'boolean', default: false, desc: 'Debug the CDK app. Logs additional information during synthesis, such as creation stack traces of tokens (sets the CDK_DEBUG environment variable, will slow down synthesis)' },
      'debug-cli': { type: 'boolean', default: false, desc: 'Debug the CDK CLI itself (enables verbose AWS SDK tracing)' },
```

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -88,12 +88,10 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
 
-  // If users request debug information (which switches on CDK_DEBUG=1 to get the
-  // CDK app to emit more information), they probably are troubleshooting and we've
-  // historically decided they then want more logging from the CLI as well.
-  // There is currently no way to get the CDK to emit tracing information that
-  // does not cause the CLI to barf logs to the console.
-  setSdkTracing(argv.debug || argv.verbose > 2);
+  // `--debug` implies both `--debug-app` and `--debug-cli`. CLI debugging (and the
+  // highest verbosity, `-vvv`) turns on verbose AWS SDK tracing.
+  const debugCli = Boolean(argv.debug || argv.debugCli);
```
mrgrain: i do not want see `debug` used anywhere anymore.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -323,6 +321,9 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
         ioHost.currentAction = 'doctor';
         return doctor({
           ioHelper,
+          debugApp: Boolean(argv.debug || argv.debugApp),
+          debugCli: Boolean(argv.debug || argv.debugCli),
+          verbose: argv.verbose,
```
mrgrain: can this just get settings?

packages/aws-cdk/lib/cli/user-input.ts
```diff
@@ -237,12 +237,26 @@ export interface GlobalOptions {
   readonly verbose?: number;
 
   /**
-   * Debug the CDK app. Log additional information during synthesis, such as creation stack traces of tokens (sets CDK_DEBUG, will slow down synthesis)
+   * Produce more detailed output to help diagnose unexpected behavior. Implies --debug-app and --debug-cli. Note that this will slow everything down
    *
-   * @default - false
+   * @default - undefined
```
mrgrain: why not keep false?

packages/aws-cdk/lib/commands/doctor.ts
```diff
@@ -5,13 +5,43 @@ import type { IoHelper } from '../api-private';
...
+   */
+  readonly debugCli?: boolean;
+
+  /**
+   * The configured verbosity level (number of `-v` flags).
+   *
+   * @default 0
+   */
+  readonly verbose?: number;
+}
```
mrgrain, line 19: just use settings

packages/aws-cdk/lib/commands/doctor.ts
```diff
@@ -24,6 +54,17 @@ const verifications: Array<(ioHelper: IoHelper) => boolean | Promise<boolean>> =
...
+async function displayCliConfiguration(
+  ioHelper: IoHelper,
+  flags: { debugApp: boolean; debugCli: boolean; verbose: number },
+) {
+  const onOff = (enabled: boolean) => (enabled ? chalk.green('enabled') : chalk.gray('disabled'));
+  await ioHelper.defaults.info('ℹ️ CLI configuration:');
+  await ioHelper.defaults.info(`  - App debugging (--debug-app): ${onOff(flags.debugApp)}`);
+  await ioHelper.defaults.info(`  - CLI debugging (--debug-cli): ${onOff(flags.debugCli)}`);
+  await ioHelper.defaults.info(`  - Verbosity level: ${chalk.green(String(flags.verbose))}`);
+}
```
mrgrain: let's move this below `CDK Version` and make it a single line string in gray

mrgrain (commented):
Almost there, but the whole`--debug` implies both targets things needs better modelling. we can change the configuration generator package.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/environment.ts
mrgrain: Done — `synthParametersFromSettings` checks only `debugApp` now. The implication lives upstream: `--debug` switches on both targets via the generator middleware, and a config-file `debug: true` is normalized to both when settings are assembled. Tests added in environment.test.ts (CDK_DEBUG mapping), user-config.test.ts (config-file implication) and configuration.test.ts.

packages/aws-cdk/lib/cli/cli-config.ts
mrgrain: Done, kept `default: false` for all three.

packages/aws-cdk/lib/cli/cli.ts
mrgrain: Done — `argv.debug` is no longer referenced. `--debug` is resolved to `debugApp`/`debugCli` during parsing, so SDK tracing now keys off `argv.debugCli`.

packages/aws-cdk/lib/cli/cli.ts
mrgrain: Done — doctor now receives `configuration.settings`.

packages/aws-cdk/lib/cli/user-input.ts
mrgrain: Kept `false` (regenerated from the config).

packages/aws-cdk/lib/commands/doctor.ts
mrgrain: Done — `doctor` now takes `settings` and reads the debug/verbosity state from it.

packages/aws-cdk/lib/commands/doctor.ts
mrgrain: Done — collapsed into a single gray line printed right below the CDK version.

packages/aws-cdk/lib/cli/cli.ts
mrgrain: Obsolete after the redesign: the flags are plain booleans now, so there is no value-parsing and no warning.

packages/aws-cdk/lib/cli/user-configuration.ts
mrgrain: Obsolete after the redesign: settings now carry explicit `debugApp`/`debugCli` instead of a derived `debug`.

mrgrain:
Reworked the `--debug` implication into the config generator as you suggested. `@aws-cdk/user-input-gen` now supports an option-level `implies` field, and `debug` declares `implies: ['debug-app', 'debug-cli']`. The generator emits a yargs middleware (`yargsImplies`) that switches the implied flags on during parsing, so by the time anything reads `argv`, `debugApp`/`debugCli` are already `true` — no hand-written `argv.debug || ...` anywhere. A config-file `debug: true` is normalized to both targets when settings are assembled, and `environment.ts` keys `CDK_DEBUG` off `debugApp` only. All comment threads addressed and resolved.

packages/aws-cdk/lib/commands/context.ts
mrgrain, line 1: this is always shown, removing the double up

rix0rrr (approved):

PR #1606, opened 2026-06-05
feat: cloudscape frontend

megha-narayanan:
Builds the `cdk explore` web explorer UI, with a React + Cloudscape single-page app is served by the existing
Express server, alongside file-serving API endpoints.

- **API** (`lib/web/`): `GET /api/health`, `GET /api/files?dir=`,`GET /api/file?path=`, 
- **SPA** (`frontend/`): Cloudscape `ContentLayout` with four regions. The two center/right panes are functional server-backed file viewers (browse + view). Header shows a "last updated" placeholder.
- **Build**: `frontend/` is compiled by esbuild (separate `tsconfig.frontend.json`, kept off the Node `tsc` path) in `post-compile`, then embedded into `web-assets.generated.json` and served from memory so it travels in the
  bundled CLI. 

### Checklist

---


megha-narayanan:
Frontend: 
Screenshot 2026-06-05 at 12 50 28 PM

packages/@aws-cdk/cdk-explorer/build-tools/bundle-frontend.ts
ShadowCat567, line 1: Can you explain a bit why you are doing this unique path to bundling the code for the web explorer? Is it because of the frontend code?

packages/@aws-cdk/cdk-explorer/frontend/App.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Container from '@cloudscape-design/components/container';
+import ContentLayout from '@cloudscape-design/components/content-layout';
+import Grid from '@cloudscape-design/components/grid';
+import Header from '@cloudscape-design/components/header';
+import SpaceBetween from '@cloudscape-design/components/space-between';
+import * as React from 'react';
+import { FilePane } from './components/FilePane';
+
+/** Web explorer shell: Resource Tree (left), two file panes, Violations (bottom). Tree/violations are placeholders until the cloud-assembly reader is wired in. */
+export function App(): JSX.Element {
```
ShadowCat567, line 11: Is there a way to make this more flexible in the future so customers can resize/rearrange their panels?

packages/@aws-cdk/cdk-explorer/lib/web/safe-path.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Resolve a client-supplied, root-relative path to an absolute path guaranteed
+ * to stay inside `root`. Returns `undefined` when the request escapes the root
+ * (via `..` or a symlink pointing outside), which callers must treat as a 403.
+ *
+ * `root` is expected to be absolute. A leading `/` on `requested` is stripped so
+ * absolute-looking inputs cannot jump to the filesystem root.
+ */
+export function resolveWithinRoot(root: string, requested: string): string | undefined {
```
ShadowCat567, line 12: will this work for Windows file paths?

packages/@aws-cdk/cdk-explorer/test/web/safe-path.test.ts
```diff
@@ -0,0 +1,32 @@
+import * as path from 'path';
+import { resolveWithinRoot } from '../../lib/web/safe-path';
+
+const root = path.resolve('/srv/app');
+
+describe('resolveWithinRoot', () => {
```
ShadowCat567, line 6: Can you add a test for Windows file paths?

packages/@aws-cdk/cdk-explorer/build-tools/bundle-frontend.ts
megha-narayanan: yep, it's driven entirely by the frontend, and there are two separate problems this handles:

First, the frontend is browser code, not server code, so it can't go through the normal per-file `tsc` compile. esbuild bundles `frontend/index.tsx` into a single `bundle.js/bundle.css`. esbuild only transpiles, so we typecheck the frontend separately with `tsc --noEmit -p tsconfig.frontend.json`

Second, getting those assets into the published CLI, which is the more complicated part. The aws-cdk CLI ships as one bundle that only includes what's reachable through the require() graph. Assets served off disk via `express.static/sendFile` are invisible to the bundler and get dropped from the npm tarball, so the explorer would 404 in a real install even though it works from source. To avoid that, the script writes the built assets into l`ib/web/web-assets.generated.json`, and `web-assets.ts` does a static `require('./web-assets.generated.json').` That require carries the bytes into the CLI bundle, and the server serves them from memory rather than disk.

packages/@aws-cdk/cdk-explorer/frontend/App.tsx
megha-narayanan: TL; DR: Kind of, but the UI looks way worse. I played around with this for a few hours yesterday.  In the next PR, I have code that makes the construct tree resizable horizontally. In my RFC, i had re-arrangable panes as a stretch goal. With cloudscape I've actually been finding that pretty hard, so my plan was to leave it as a stretch goal for now, lmk if you think its important enough to devote time to now.

packages/@aws-cdk/cdk-explorer/lib/web/safe-path.ts
megha-narayanan: Yep, all path manipulation goes through Node's path module, which is platform aware. Traversal via ..\\ and drive-absolute inputs like C:\\Windows get rejected by the same `isInside` check.

packages/@aws-cdk/cdk-explorer/test/web/safe-path.test.ts
megha-narayanan: Yep, added

packages/@aws-cdk/cdk-explorer/frontend/App.tsx
ShadowCat567: I think having more static panels is fine for now, I was just wondering if you looked into it and how easy/hard it would be to add later

ShadowCat567 (approved):

PR #1614, opened 2026-06-08
fix: not enough information to count sequential failures

rix0rrr:
We would like to know how many sequential failures a user encounters. Count them up in a state file and send that information along when we send deployment statistics.

---

### DIFF
--- a/packages/aws-cdk/lib/cli/telemetry/installation-id.ts
+++ b/packages/aws-cdk/lib/cli/telemetry/installation-id.ts
@@ -2,8 +2,7 @@ import { randomUUID } from 'node:crypto';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import type { IoHelper } from '../../api-private';
-import { cdkCacheDir, ensureCacheDir } from '../../util';
-
+import { cdkCacheDir } from '../../util';
 
 /**
  * Get or create installation id
@@ -11,7 +10,6 @@ import { cdkCacheDir, ensureCacheDir } from '../../util';
 export async function getOrCreateInstallationId(ioHelper: IoHelper) {
   // Do this quite lazily, so we can mock `cdkCacheDir` during tests
   const installationIdPath = path.join(cdkCacheDir(), 'installation-id.json');
-  console.log('path', installationIdPath);
 
   try {
     // Check if the installation ID file exists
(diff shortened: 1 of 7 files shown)


packages/aws-cdk/lib/cli/telemetry/installation-id.ts
```diff
@@ -31,17 +27,31 @@ export async function getOrCreateInstallationId(ioHelper: IoHelper) {
...
     } catch (e: any) {
       // If we can't write the file, still return the generated ID
       // but log a trace message about the failure
-      await ioHelper.defaults.trace(`Failed to write installation ID to ${INSTALLATION_ID_PATH}: ${e}`);
+      await ioHelper.defaults.trace(`Failed to write installation ID to ${installationIdPath}: ${e}`);
     }
     return newId;
   } catch (e: any) {
+    // eslint-disable-next-line no-console
+    console.log(e);
```
kaizencc: is this a leftover debug?

packages/aws-cdk/lib/cli/telemetry/telemetry-state.ts
```diff
@@ -0,0 +1,53 @@
+import * as fs from 'fs/promises';
+import * as path from 'path';
+import { ensureCacheDir } from './installation-id';
+import { cdkCacheDir } from '../../util';
+
+/**
+ * Telemetry state across CLI invocations
+ */
+export interface TelemetryState {
```
kaizencc, line 9: can we enforce that `TelemetryState` only ever includes properties of type `boolean` or `number` (or `enum` i guess). my worry is that a user can modify their telemetry file however they want and `TelemetryState` is the only gate we have to enforce that the data is what we expect. If we in the future add `previousInstallationIds: string[]` (we definitely _shouldn't_ do this but I digress) then we open the doors to collecting customer content in these strings

packages/aws-cdk/test/cli/telemetry/installation-id.test.ts
kaizencc, line 1: can we add a test that confirms malformed telemetry state is not passed on?

kaizencc (approved):
approved modulo minor thoughts

packages/aws-cdk/test/cli/telemetry/installation-id.test.ts
rix0rrr: Are you under the impression that the whole of `TelemetryState` gets passed on to the server side automatically?

kaizencc: not today. i am concerned in the future someone adds to TelemetryState a string property that is directly passed on to the server

PR #1625, opened 2026-06-14
fix: bucket URL is incorrect for local S3 emulators

peter-smith-phd:
The AWS SDK addresses S3 buckets using virtual-hosted-style URLs by default (`https://<bucket>.s3.amazonaws.com`). When targeting custom or local S3-compatible endpoints (for example LocalStack or MinIO), path-style addressing (`https://<endpoint>/<bucket>`) is often required instead.

There are two ways to enable path-style addressing on the S3 clients, used for asset publishing (`cdk-assets-lib`) and by the toolkit (`toolkit-lib`).

1. Setting `AWS_ENDPOINT_URL` (or `AWS_ENDPOINT_URL_S3`) to use a local loopback host, such as `localhost`, `127.0.0.1` or `::1`. 
2. Setting the `CDK_S3_FORCE_PATH_STYLE` environment variable to explicitly enable the feature. This name follows the existing convention of CDK-specific environment variables such as `CDK_DOCKER` and `CDK_HOME`.

### Checklist
- [ ] Integration tests added/updated (if deploying new resource types or cross-service interactions)

---

### DIFF
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -708,11 +708,11 @@ const tools = defineTools({
   ...genericCdkProps({ private: true }),
   parent: repo,
   tools: {
-    subprocess: {
+    'subprocess': {
       deps: ['cross-spawn@^7.0.6'],
       devDeps: ['@types/cross-spawn'],
     },
-    zip: {
+    'zip': {
       deps: ['yazl@^3.3.1', 'fast-glob@^3.3.3'],
       devDeps: ['@types/yazl', 'jszip', 'timezone-mock'],
     },
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -716,6 +716,9 @@ const tools = defineTools({
       deps: ['yazl@^3.3.1', 'fast-glob@^3.3.3'],
       devDeps: ['@types/yazl', 'jszip', 'timezone-mock'],
     },
+    's3-path-style': {
+      deps: [],
+    },
   },
 });
 configureProject(tools);
@@ -773,6 +776,7 @@ const cdkAssetsLib = configureProject(
   }),
 );
 cdkAssetsLib.with(tools.zip);
+cdkAssetsLib.with(tools['s3-path-style']);
 fixupTestTask(cdkAssetsLib);
 
 // Prevent imports of private API surface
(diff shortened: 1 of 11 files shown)


rix0rrr:
Is there a way to autoderive the correct behavior here? Zero-conf is better than an option fleet...

peter-smith-phd:
> Is there a way to autoderive the correct behavior here? Zero-conf is better than an option fleet...

Great question. This option is solely necessary when the user sets a custom `AWS_ENDPOINT_URL_S3` . If the URL they provide doesn't support subdomains properly (which is a problem for `127.0.0.1` or `localhost`), then S3 path-style addressing is required.

So, we could automatically set `forcePathStyle` if there's a change to `AWS_ENDPOINT_URL_S3` (or `AWS_ENDPOINT_URL` that implicitly sets `AWS_ENDPOINT_URL_S3`). This is fine if the new URL refers to a non-AWS location (like an emulator), but could be problematic if the new URL points so something like a FIPS endpoint (prefers virtual host for performance reasons).

The most accurate solution would be to detect `127.0.0.1` and `localhost` explicitly, and automatically set the flag only in those cases. Would that be more acceptable?

rix0rrr:
I think I would like that better, yes. I'm debating whether it would be beneficial to check the URL and see if it ends in one of the AWS domains (`*.amazonaws.com` etc), and disable the feature if not. Should work for FIPS endpoints, VPC endpoints, and if our domain list is exhaustive enough also for ADC regions.

But to be frank, I don't know enough about the various network configurations people could run to know if that's exhaustive enough or not, or people would want virtual hosts with custom domains as well... so probably doing the following will do the most good for the most people:

- Keep the environment variable for explicit control
- Autodetect to path-based routing for special hosts like `localhost` and `127.0.0.1`

peter-smith-phd:
Thanks, that sounds like a plan. I suspect that `localhost` and `127.0.0.1` are the far most common hostnames that would be used, whereas trying to enumerate all those ADC regions would end up with yearly maintenance 😄 

I've updated the PR, and have created a shared package (based off how `zip` is shared). Hopefully this is the correct approach.

rix0rrr (approved):

peter-smith-phd:
@rix0rrr Thanks for approving this PR. Is there anything else I need to do at this point in time? (I don't have a "Merge" button).

mrgrain (approved):

PR #1631, opened 2026-06-15
fix: constrain assembly-derived file reads to the project dir

megha-narayanan:
The LSP followed source/template paths from the cloud assembly (stack-trace
frames, source maps, template paths) with only an extension allow-list. If
cdk.out is tampered with, a crafted path like `../../../etc/passwd.js` could be
read or surfaced as a nav target.

Adds a symlink-aware `isWithinRoot` and gates every read: source paths against
the project dir, `templateFile` against the assembly dir. Out-of-root paths are
dropped as not-navigable (same as non-TS apps). No change for valid assemblies.

Fixes #

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
@@ -1,6 +1,6 @@
 import * as fs from 'fs';
 import * as path from 'path';
-import { buildConstructTree, CloudAssembly, type ConstructTreeNode } from '@aws-cdk/cloud-assembly-api';
+import { buildConstructTreeAsync, CloudAssembly, type ConstructTreeNode } from '@aws-cdk/cloud-assembly-api';
 import { VALIDATION_REPORT_FILE, type PolicyValidationReportJson } from '@aws-cdk/cloud-assembly-schema';
 import { findCreationStackTrace } from '@aws-cdk/toolkit-lib';
 import { SourceMapResolver, isWithinRoot, type SourceLocation } from './source-resolver';
@@ -33,9 +33,9 @@ export type AssemblyReadResult =
 /**
  * Decorates the cloud assembly's construct tree with the source location of
  * each node and attaches any policy-validation violations. Tree construction
- * (tree.json + stack-metadata join) is delegated to buildConstructTree.
+ * (tree.json + stack-metadata join) is delegated to buildConstructTreeAsync.
  */
-export function readAssembly(assemblyDir: string): AssemblyReadResult {
+export async function readAssembly(assemblyDir: string): Promise<AssemblyReadResult> {
   const manifestPath = path.join(assemblyDir, 'manifest.json');
   if (!fs.existsSync(manifestPath)) {
     return { status: 'not-found' };
@@ -52,17 +52,17 @@ export function readAssembly(assemblyDir: string): AssemblyReadResult {
     // One resolver per readAssembly call: caches parsed source maps across
     // constructs, scoped so a fresh synth observes any moved/edited maps.
     const sourceResolver = new SourceMapResolver(projectRoot);
-    const tree = buildConstructTree<ConstructNode>(assembly, (fields, stack, constructPath) => ({
+    const tree = await buildConstructTreeAsync<ConstructNode>(assembly, async (fields, stack, constructPath) => ({
       ...fields,
       // templateFile comes from the manifest / a nested stack's aws:asset:path,
       // both attacker-influenceable if cdk.out is tampered with. Drop any that
       // escape the assembly dir so the template read in resourceTarget stays
       // contained
-      templateFile: fields.templateFile && isWithinRoot(assemblyDir, fields.templateFile)
+      templateFile: fields.templateFile && (await isWithinRoot(assemblyDir, fields.templateFile))
         ? fields.templateFile
         : undefined,
       sourceLocation: stack
-        ? sourceResolver.resolveFrames(findCreationStackTrace(stack, constructPath))
+        ? await sourceResolver.resolveFrames(findCreationStackTrace(stack, constructPath))
         : undefined,
     }));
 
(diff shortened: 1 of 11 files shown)


packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -48,6 +58,9 @@ export class SourceMapResolver {
       const parsed = parseFrame(frame);
       if (!parsed) continue;
       if (!isSupportedSourceFile(parsed.file)) return undefined;
+      // The frame's file path is assembly-derived and attacker-influenceable.
+      // Never read or surface a location outside the project (T-003/M-003/M-004).
```
ShadowCat567: You probably want to update this comment. This `(T-003/M-003/M-004)` means nothing to future contributors unless they know to look at the threat model

packages/@aws-cdk/cdk-explorer/test/core/source-resolver.test.ts
```diff
@@ -117,8 +129,91 @@ describe('source-map resolution', () => {
...
     fs.writeFileSync(jsPath, `${sampleBody()}\n//# sourceMappingURL=renamed.map\n`);
 
+    const resolver = new SourceMapResolver(dir);
     const result = resolver.resolveFrames([`    at greet (${jsPath}:5:10)`]);
     expect(result!.file).toBe(path.join(dir, 'nested', 'sample.ts'));
     expect(result!.line).toBe(2);
   });
 });
+
+// Test-004 / M-003 / M-004: paths from the (attacker-influenceable) cloud
```
ShadowCat567: This comment also has info you need the threat model to understand

ShadowCat567 (commented):
Overall this looks good, take another look through your comments to make sure you don't have info that only makes sense if you have access to internal documentation

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: ah, got it. fixed!

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -43,11 +43,24 @@ export function readAssembly(assemblyDir: string): AssemblyReadResult {
...
+    // The project root is the parent of the assembly dir: server.ts always
+    // builds it as <projectDir>/cdk.out, so dirname() recovers the project dir.
+    // (If --output support is added to relocate cdk.out, thread the IDE-provided
+    // application dir here instead.) Used to keep every file the resolver reads
+    // within the project
+    const projectRoot = path.dirname(assemblyDir);
     // One resolver per readAssembly call: caches parsed source maps across
     // constructs, scoped so a fresh synth observes any moved/edited maps.
-    const sourceResolver = new SourceMapResolver();
+    const sourceResolver = new SourceMapResolver(projectRoot);
```
rix0rrr, line 52: Is(n't?) this code also going into the LSP?

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -137,3 +158,25 @@ function parseFrame(frame: string): SourceLocation | undefined {
...
+export function isWithinRoot(root: string, candidate: string): boolean {
+  const realRoot = realOrSelf(path.resolve(root));
+  const realCandidate = realOrSelf(path.resolve(candidate));
+  return realCandidate === realRoot || realCandidate.startsWith(realRoot + path.sep);
+}
+
+/** Real path with symlinks resolved, or the resolved input if it does not exist. */
+function realOrSelf(p: string): string {
+  try {
+    return fs.realpathSync(p);
```
rix0rrr: Try to avoid sync filesystem functions as much as possible. Use the `async` variants instead.

Reason being this will block whatever thread this code is running on, and if that happens to be thread that does other useful work (like, oh, serving HTTP requests or handling UI updates) it now no longer can. Whereas if this was `async` it could all happen in the background while other work can continue.

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: Yep, this core gets imported by the LSP.  The LSP doesn't have its own source-map resolver. It calls `readAssembly`, which runs this one and stores the resolved locations on the tree.

packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
megha-narayanan: Agreed on the principle.

The wrinkle is that I can't find a natural place where the sync/async seam can sit. This `realpathSync` backs`isWithinRoot`, the containment predicate. That predicate runs per node inside the visitor we pass to `buildConstructTree` in `assembly-reader.ts`: once for the `templateFile` gate, and again inside `resolveFrames` for the source-map paths (which also read from disk). Since a Promise can't be read synchronously, the visitor has to become async, and `buildConstructTree` itself doesn't do this I/O, so
the only way to let the visitor await is to make `buildConstructTree` accept an async visitor and return a Promise. That is a signature change on the `cloud-assembly-api` export, and it propagates up through `readAssembly` to the server's refresh path.

So the minimal form of your suggestion (that i can see) is  to make `buildConstructTree` async, including that shared-package signature change. Is that the direction you want? Just wanted to check before making a change which would touch so many files

rix0rrr: If it's a change on the `cloud-assembly-api` export, then we can't change it! A public contract is a public contract, and we can't just go around changing functions in backwards breaking ways.

I think what we would do then is add a `buildConstructTreeAsync` function.

megha-narayanan: got it, just implemented that. looks like I touched a ton of files but its really just adding the async keyword everywhere :) let me know if this seam doesn't align with what you were thinking/

rix0rrr: >  I touched a ton of files but its really just adding the async keyword everywhere

[Yes, that is unfortunately just what it is.](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/)

rix0rrr (approved):

PR #1634, opened 2026-06-16
feat: synth triggering

megha-narayanan:
Adds manual synth and auto-synth-on-save to the CDK LSP. The cdk.out watcher (landed in #1630) handles all diagnostics/lens refreshes -- these commands just add new ways to produce a fresh assembly.

Features:
↻ Synth now CodeLens at the top of any CDK source file (only when auto-synth is off)
▶ Enable auto-synth / ⏹ Disable auto-synth toggle
Auto-synth starts disabled
When auto-synth is on, saving any non-ignored file in the project triggers a synth
All feedback/output goes to the Output panel, no popups

Design decisions (the important ones):

Concurrent synths are supressed: if a save fires during a slow synth, that save is skipped. The next save picks it up.
cdk.json is read once at startup. Changing it requires an LSP restart.
The toggle/synth lenses only appear on files that already have L1 resource lenses. Files with no CDK resources see nothing.
Toggle state resets to disabled on LSP restart (not persisted).

NOTE: One thing not in this PR (but must land before prod) is a workspace trust gate. We should verify with the user that they trust this workspace before running any synth.  I may need to set up some way to preserve this between sessions. 

Open questions / things I need some feedback on:
- Is "auto-synth" the right name? Alternatives: "synth on save", "live synth"?
- Where should the toggle live? Line 0 of a CDK file works for LSP-only, but it's only visible when you're in a file with constructs. Status bar item would be better UX but needs a client extension.  Is there a better alternative?
- Should synth failures (app compile errors) be more visible than the Output panel? A diagnostic on the first line of the failing file, for example? 
- If the app has context lookups and no cached cdk.context.json, synth fails with an auth/context error. Should the error message detect this and suggest running cdk synth in terminal first?

Screenshot 2026-06-16 at 3 41 01 PM
Screenshot 2026-06-16 at 3 41 10 PM

### Checklist

---


rix0rrr:
I appreciate you calling out the design decisions, thanks!

> Concurrent synths coalesce: if a save fires during a slow synth, that save is skipped. The next save picks it up.

So we are stale on the second save? I don't think I would call this "coalesce" -- to my mind that would be "multiple saves combine into one synth". This is more like "suppress" -- multiple saves, everything after the first is ignored.

It's probably fine for the first iteration, regardless of what we call it 😄. We'll see how it feels in practice.

> cdk.json is read once at startup. Changing it requires an LSP restart.

This feels like behavior I would hate as a user, because it's not clear what/why or how things are inconsistent. Especially when you go to change feature flags, that should trigger a reload. Is it too much effort to watch this file and do a reload whenever necessary? (Not a full restart of the entire LSP, keep the state of the autosynth toggle)

> We should verify with the user that they trust this workspace before running any synth

Yes good call -- though I wonder if the implicit VSCode plugin loading dialog would not qualify for this already? I mean this one:
image

> Is "auto-synth" the right name? Alternatives: "synth on save", "live synth"?

Auto-synth is probably fine. `▶ Synth on save`/`⏹ Stop synthing on save` might do as well for explicitness.

> Should synth failures (app compile errors) be more visible than the Output panel? A diagnostic on the first line of the failing file, for example?

Yes, definitely a diagnostic!  A diagnostic on `cdk.json` would work, if we didn't have anything better. If you can get to the failing file, so much the better (but you might not be able to...). 

> If the app has context lookups and no cached cdk.context.json, synth fails with an auth/context error. Should the error message detect this and suggest running cdk synth in terminal first?

Lovely, yes! You can also run synth with `--no-lookups` (I think) to force this behavior with a nicer error message.

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
```diff
@@ -0,0 +1,79 @@
...
+    cached = await options.toolkit.synth(cx);
+  } catch (err) {
+    return classify(err);
+  }
+
+  try {
+    await cached.dispose();
+  } catch (err) {
+    // Synth succeeded and `cdk.out` is on disk, but we could not release the
+    // read lock. Surface as `error` so the caller can warn the user; without
```
rix0rrr: Did this happen in practice, or is this just overly defensive programming?

Also, is it even safe to release the lock here already? Shouldn't we keep the lock for as long as we're reading from the assembly? 

Why does this function not return an assembly object?

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
```diff
@@ -0,0 +1,79 @@
...
+  return { status: 'success' };
+}
+
+function classify(err: unknown): SynthRunResult {
+  if (ToolkitError.isToolkitError(err)) {
+    // ToolkitError stores its discriminating code in `name` (the constructor's
+    // first arg overrides the default Error name), not in a `code` property.
+    // These literals come from RWLock (toolkit-lib/lib/api/rwlock.ts). No
+    // named constants are exported; if upstream renames them, classification
+    // falls through to 'error' silently -- update here and in tests if so.
```
rix0rrr: Does `cdk-explorer` not depend on `toolkit-lib`? If so, it should be possible to expose these error names as constants and consume them as constants here?

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (err.name === 'ConcurrentWriteLock' || err.name === 'ConcurrentReadLock') {
+      return { status: 'lock-conflict' };
+    }
+    if (ToolkitError.isAssemblyError(err)) {
+      return { status: 'app-failure', message: err.message };
+    }
+    return { status: 'error', message: err.message };
+  }
+  return { status: 'error', message: (err as Error).message };
```
rix0rrr: These lines are the same, so could be combined.

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
rix0rrr: Or you could add `ToolkitError.isLockError()` or something.

packages/@aws-cdk/cdk-explorer/lib/lsp/io-host.ts
```diff
@@ -0,0 +1,45 @@
...
+        // Suppress noisy levels; keeps the Output panel readable.
+        break;
+      default:
+        this.console.info(msg.message);
+    }
+  }
+
+  public async requestResponse<T>(msg: IoRequest<unknown, T>): Promise<T> {
+    await this.notify(msg);
+    return msg.defaultResponse;
```
rix0rrr, line 44: Might be worth logging to the console what the question and the automatic response were.

packages/@aws-cdk/cdk-explorer/lib/lsp/io-host.ts
```diff
@@ -0,0 +1,45 @@
...
+    switch (msg.level) {
+      case 'error':
+        this.console.error(msg.message);
+        break;
+      case 'warn':
+        this.console.warn(msg.message);
+        break;
+      case 'debug':
+      case 'trace':
+        // Suppress noisy levels; keeps the Output panel readable.
```
rix0rrr, line 34: Doesn't that panel have its own verbosity level filtering?

packages/@aws-cdk/cdk-explorer/lib/lsp/main.ts
```diff
@@ -1,9 +1,24 @@
...
+  const config = readCdkConfig(projectDir);
+
   startServer({
     readable: process.stdin,
     writable: process.stdout,
+    synthAvailable: config.app !== undefined,
+    // buildSynthRunner is called once after the LSP connection is established,
+    // so LspIoHost can receive a real connection.console to route Toolkit
+    // output to the editor's Output panel.
+    buildSynthRunner: config.app !== undefined ? (console) => {
```
rix0rrr: We might call this a "factory", that's a somewhat standardized name for "an object that produces the object you're *actually* interested in".

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -61,11 +63,24 @@ export interface LspHandlerOptions {
    * overridden in tests to drive refreshes deterministically.
    */
   readonly startAssemblyWatcher?: (options: AssemblyWatcherOptions) => AssemblyWatcher;
+  /** Runs a synth and returns its typed outcome. Injected by main.ts; omitted in tests that don't exercise synth. */
+  readonly synthRunner?: () => Promise<SynthRunResult>;
+  /** Whether cdk.json exists and has a valid `app` key. Controls whether synthNow is available. */
+  readonly synthAvailable?: boolean;
```
rix0rrr: Are we ever in a situation where it makes sense to have one of these and not the other?

If not, we could say that `synthAvailable ≡ (synthRunner !== undefined)`

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -61,11 +63,24 @@ export interface LspHandlerOptions {
...
 }
 
 export interface LspServerOptions extends LspHandlerOptions {
   readonly readable: NodeJS.ReadableStream;
   readonly writable: NodeJS.WritableStream;
+  /**
+   * Optional factory called once after the connection is established.
+   * Receives `connection.console` so the returned runner can route Toolkit
+   * IO to the editor's Output panel. Takes precedence over `synthRunner`
+   * when both are provided.
```
rix0rrr: Not necessary to have both of these then. `buildSynthRunner` is strictly more expressive than `synthRunner`, so you can just have the one?

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -163,6 +203,21 @@ export function createLspHandlers(options: LspHandlerOptions = {}): LspHandlers
...
+  // Shared synth invocation used by both the manual CodeLens command and
+  // auto-synth-on-save. The in-flight latch prevents concurrent synths from
+  // the same LSP instance. A second call while the first is running returns
+  // lock-conflict immediately (the Toolkit's RWLock would do the same, but
+  // this short-circuits before any setup work).
+  async function guardedSynth(): Promise<SynthRunResult> {
+    if (synthInFlight) return { status: 'lock-conflict' };
+    synthInFlight = true;
+    try {
+      return await (synthRunner ? synthRunner() : Promise.resolve({ status: 'error', message: 'No synth runner configured' } as const));
```
rix0rrr: Wait, where is `synthRunner` coming from here? Is that a global variable? Feels like it should be an instance variable of some long-lived object? And same for `synthInFlight`.

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
megha-narayanan: Yep, these are derived from the same place in main.ts so I've collapsed them to one.

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
megha-narayanan: yep, the type was extending LspHandlerOptions but I separated these out now and resolved to a runner before passing

packages/@aws-cdk/cdk-explorer/lib/lsp/main.ts
megha-narayanan: yep, renamed

megha-narayanan:
> So we are stale on the second save? I don't think I would call this "coalesce" -- to my mind that would be "multiple saves combine into one synth". This is more like "suppress" -- multiple saves, everything after the first is ignored.

Yep, redid the docs to reflect the suppress behavior. 


> Yes, definitely a diagnostic! A diagnostic on cdk.json would work, if we didn't have anything better. If you can get to the failing file, so much the better (but you might not be able to...).

Added a diagnostic for the correct file when we can get there


> Is it too much effort to watch this file and do a reload whenever necessary? (Not a full restart of the entire LSP, keep the state of the autosynth toggle)

Sorry, I think I miscommunicated this in my original description, I meant that we cache the APP field, so that is only read once. Changing feat flags would be re-read on the next synth. Reading app just once at construction means "runner exists" means "synth is available." Because main.ts only builds the synthRunnerFactory when config.app !== undefined, the presence of a runner doubles as the availability signal.  Would you prefer a design that doesn't cache the app? it is definitely SUPER rare that it would change, mostly just if you rename your entry file or running cdk init. Let me know what you think is better design, I do think it could go both ways.  

Also, added code to detect context lookups and ask user to run synth in the terminal. For the trust gate, yes, for VSCode the existing trust gate means no extension will run in an untrusted workspace. So its fine for VSCode. The question is how we deal with other IDEs

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
megha-narayanan: runSynth is a pure producer. `toolkit.synth()` takes the write lock on cdk.out, writes the assembly, then converts to a read lock that the returned assembly holds. We dispose() right away to release that read lock, and we have to: the next synth's acquireWrite() would otherwise see our own lingering reader and throw ConcurrentReadLock, so we'd self-conflict.

We never read the assembly here. The cdk.out watcher is the single consumer, and it reads lock-free (plain JSON.parse), so keeping the lock wouldn't protect its read anyway, and returning the assembly would create a second refresh path racing the watcher. Keeping runSynth write-only means one refresh path for every producer, whether that's our synth, or an external cdk synth.

On the defensive programming bit, you're right. dispose() only fails on a rare fs error deleting the lock file. I trimmed it to a line. I kept the try/catch itself so runSynth always resolves to a typed SynthRunResult rather than rejecting, since the command layer depends on that.

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
megha-narayanan: Added isLockError into the upstream.

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
megha-narayanan: done

packages/@aws-cdk/cdk-explorer/lib/lsp/io-host.ts
megha-narayanan: added a line logging what we answered automatically.

packages/@aws-cdk/cdk-explorer/lib/lsp/io-host.ts
megha-narayanan: Some clients do, some don't, so I didn't want to rely on it. connection.console emits window/logMessage with a MessageType. New VS Code versions route that into a channel with user-selectable levels, but older versions and several non-VS Code clients write everything to a plain channel with no level filter. trace is also a separate channel, not a logMessage level. So I kept the suppression to keep the Output panel readable regardless of client.

If you'd rather defer to the client, lmk. Happy to go either way.

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
megha-narayanan: Neither are globals, synthRunner and synthInFlight are both closure state of `createLspHandlers`. The handlers object that createLspHandlers returns is the long-lived instance, and guardedSynth (along with the other handlers) closes over both. The diff hunk just didn't show the declarations further up in the function.

megha-narayanan:
Read cdk.json's app on demand per synth instead of adding a watcher. Nothing is cached, so editing the app (or cdk init in an open folder) applies on the next synth with no restart, and the autosynth toggle is preserved. Feature flags were never cached by the LSP anyway: each synth re-reads cdk.json in a fresh subprocess.

Also unified the project root behind one currentProjectDir() (synth was using process.cwd(), the watcher applicationDir) so they can't disagree.

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
rix0rrr: > runSynth is a pure producer. toolkit.synth() takes the write lock on cdk.out, writes the assembly, then converts to a read lock that the returned assembly holds. We dispose() right away to release that read lock, [...]  The cdk.out watcher is the single consumer, and it reads lock-free (plain JSON.parse), so keeping the lock wouldn't protect its read anyway,

Hmm. I see. Your first question might be: why do we bother with read locking in the first place? Because *every* reader could choose to ignore the lock, and if we did that we wouldn't have to bother with locks at all, and isn't that much easier? 

Hopefully the answer to that should reveal why we do locking in the first place, and so an unlocked read is probably a bad idea altogether. What we would probably want is for the reader to hold a reader lock for the duration that it's reading the assembly directory. 

So given that we want the reader to hold a lock, the question becomes: how does it obtain that lock? 

1. It can acquire one independently
2. It so happens that the synther has a read lock available; perhaps it could hand that over.

The problem with (1) would be that the reader would race against a pending new synth, and the problem with (2) would be that the synther and the reader are currently fully disconnected, and we don't really have a good mechanism to hand ownership of the lock over.

We can pick one of those 2 solution paths, and come up with a solution to overcome the downside of each of these solutions.

## 1. Reader acquires its own lock

*Challenge:* There is a race between a pending re-synth (acquiring a fresh writer lock) and the reader (acquiring a reader lock).

*Potential solution:* Reading should be swift, and synthing is slow already. We can add a bit of artificial delay (of a couple 100ms) between resynths. This should be enough time for the reader to detect the change, acquire a lock, read the files its interested in, and release the lock. And ~100ms on the time scale of a synth and human interaction in the IDE disappears in the noise, so should be barely perceptible.

## 2. Synther communicates the lock to the reader

*Benefit*: We don't have any artificial delay and no lock contention, we go as fast as possible.

*Challenge:* We don't have a communication channel, currently. Plus, the reader may also trigger on changes external to the IDE (where it will still need to be able to acquire its own lock), plus there may be timing issues between it detecting on-disk changes and it receiving the lock from the synther. So we would need some sort of protocol with a small state machine:

image

## What do we do?

Given the relative complexity between these 2 solutions... I vote we add a little delay to the synther 😉 . And I think we put the delay *before* releasing the read lock. That way, we are guaranteed that nobody else comes in to overwrite the directory when we're not looking.

And it would be good to add a comment to the synther why we're doing it that way: give the IDE reader an opportunity to acquire a read lock without racing another writer.

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
rix0rrr: If your closures get huge, it might be [worthwhile to rewrite them into objects with member fields](https://wiki.c2.com/?ClosuresAndObjectsAreEquivalent). Not strictly necessary, but it's just more clear to other people what's going on there.

Of course, if you're doing that and you're handing out callbacks all over the place, you need to be aware of the `this` binding problem of JavaScript. I *hope* and *think* we have an ESLint rule to remind you about that.

packages/@aws-cdk/cdk-explorer/lib/core/synth-runner.ts
megha-narayanan: Adding a delay to the synther seems like a great solution. Thanks for all the detail here, it really does help. In order to actually make this do what we want we still need to make the reader actually hold a read lock while reading, instead of reading lock-free. That needs RWLock exported from toolkit-lib, the delay alone only narrows that window. 

Are you ok exporting the lock from toolkit-lib? I'm pushing a commit with the whole fix, lmk what you think.

rix0rrr: If exporting is necessary then that's what we need to do.

But I feel like the lock should already be part of the `CloudAssembly` class you can obtain from the toolkit.

megha-narayanan: Ah, you're right and I understand now. toolkit.fromAssemblyDirectory(dir).produce() already returns an IReadableCloudAssembly that holds a read lock for its lifetime, so the refresh takes its lock through that and RWLock stays internal to toolkit-lib. I also fixed one more behavior, the refresh now retries the acquire, instead of giving up after one try and leaving diagnostics stale.

rix0rrr (approved):

PR #1636, opened 2026-06-16
chore(cli): validate via report instead of stack metadata

kaizencc:
in the cli we validate via stack metadata. this metadata gets written when construct annotations are added. this PR moves validation to the right place -- the `validation-report.json` that we get when we synthesize. 

in conjunction with https://github.com/aws/aws-cdk/pull/38155, annotations will always be written to the report. 

---


rix0rrr:
Superseded by https://github.com/aws/aws-cdk-cli/pull/1650

PR #1637, opened 2026-06-16
fix(cli): cdk list pollutes stdout with synthesis time

### LINKED ISSUE #1632: (cli/list): including informational output in stdout stream in CI environments
### Describe the bug

`cdk list` command is including `✨  Synthesis time: X.XXs` in the output in CI environments (when `--ci` flag or `CI=true`) as of the most recent release `2.1127.0`. This is impacting our use case of piping the resulting output to `jq` for further processing in CI environments.

### Regression Issue

2.1126.0

### Expected Behavior

Informational output is not piped to stdout in CI environment alongside JSON

### Current Behavior

Informational output is being included alongside JSON output in CI environment.

## Reproduce
`npx cdk init sample-app --language=typescript`
`npx cdk synth`

### `cdk@2.1127.0`

`npx cdk@2.1127.0 list --json --no-notices`

```
✨  Synthesis time: 4.21s
ExampleStack
```

### `cdk@2.1126.0`

`npx cdk@2.1126.0 list --json --no-notices`

```
ExampleStack
```

### Possible Solution

We've worked around the issue by cutting out the `Synthesis time` line manually, but this info should not be included alongside the JSON in the CI environment.

Possibly related to [this commit](https://github.com/aws/aws-cdk-cli/commit/a0588770764d3ec3d83d1596ed26739610e16739) that changed the `list` command to use `toolkit-lib`?

### CDK CLI Version

2.1127.0

### Framework Version

_No response_

### Node.js Version

24

### OS

Widnows

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

sai-ray:
Fixes #1632

`cdk list` prints `✨  Synthesis time: X.XXs` to stdout in CI mode (regressed in 2.1127.0). That extra line shows up in machine-readable output, so piping `cdk list` into a parser (e.g. `cdk list --json | jq`) breaks.

For example, `$ cdk list --json` in CI mode gives you:

```
✨  Synthesis time: 3.72s
MyStack
```

The parser only expects the stack list, so the `Synthesis time` line makes it fail.

#1603 routed `cdk list` through the toolkit-lib `list` action, which synthesizes via `synthAndMeasure`. That helper opens a timing span, and ending the span emits `CDK_TOOLKIT_I1000` ("Synthesis time"). In CI mode the CLI sends non-error messages to stdout, so the line lands next to the listing.

The CLI now drops the synth-time message (CDK_TOOLKIT_I1000) on the list path before it's written, so stdout carries only the listing.

`$ cdk list --json` # after fix, CI mode

```
MyStack
```

## Testing

- Unit test: `cdk list` registers a one-shot suppressor for `CDK_TOOLKIT_I1000`, so the synth-time line is not printed.
- Integ test: `cdk ls` in CI produces stdout with only the stack listing, no synthesis-time line.
- Verified manually against a sample app (`cdk list --json` in CI → clean stdout).
 
### Checklist

---


packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -596,7 +596,9 @@ export class Toolkit extends CloudAssemblySourceBuilder {
   public async list(cx: ICloudAssemblySource, options: ListOptions = {}): Promise<StackDetails[]> {
     const ioHelper = asIoHelper(this.ioHost, 'list');
     const selectStacks = stacksOpt(options);
-    await using assembly = await synthAndMeasure(ioHelper, cx, selectStacks);
+    // Not `synthAndMeasure`: `list` is a read-only query and its synth-time message
+    // would pollute stdout when the result is piped (e.g. `list --json | jq`).
+    await using assembly = await assemblyFromSource(ioHelper, cx);
```
rix0rrr: I would say this is the wrong fix.

Wouldn't it make more sense to suppress the "synthesis time" message in `--json` mode, or something to that effect?

Either that, or say "this is query and therefore we're not printing anything other than the output" (which you said on Slack that I also like), but I wouldn't implement that by bypassing the entire method `synthAndMeasure` method; all synths get routed through that function and I would like to keep it that way. The measuring is NOT JUST printing to the screen, it's also measuring internally. 

Maybe we suppress printing by the the `IoHost` instead or something. 

This fix should be about what gets **printed**, not about what gets **done**.

mrgrain: +100

mrgrain (dismissed):

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
sai-ray: Updated to follow the same pattern as the `rewriteOnce(I2901)` line right above, a host-side listener on a library message. That one reshapes the listing, this one suppresses the synth-time line via `once(I1000, () => ({ preventDefault: true }))`. `synthAndMeasure` stays untouched so synth is still measured, only what's printed changes.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -960,6 +960,11 @@ export class CdkToolkit {
   ): Promise<number> {
     this.ioHost.rewriteOnce(IO.CDK_TOOLKIT_I2901, (msg) => formatStackList(msg.data.stacks, options));
 
+    // `list` is a read-only query whose stdout is commonly piped (e.g. `list --json | jq`).
+    // Synthesis still runs and is measured, but the "Synthesis time" line (I1000) must not
+    // be printed, since in CI mode non-error output goes to stdout and would corrupt the listing.
+    this.ioHost.once(IO.CDK_TOOLKIT_I1000, () => ({ preventDefault: true }));
```
iliapolo: Lets make this even more targeted with:

`if (options.json) { ... }`

This establishes a good pattern where machine parsable output is only expected if `--json` is used. Otherwise, we are free to add additional prints later on.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/list/cdk-cdk-ls-ci-no-synth-time-on-stdout.integtest.ts
```diff
@@ -0,0 +1,22 @@
...
+
+integTest(
+  'cdk ls in CI does not print synthesis time to stdout',
+  withDefaultFixture(async (fixture) => {
+    // In CI mode, `cdk ls` output goes to stdout and is often piped, so it must
+    // be only the stack listing, not status lines like "✨ Synthesis time: ...".
+    const listing = await fixture.cdk(['ls'], {
+      verbose: false, // no -v, so stdout is just the listing
+      captureStderr: false, // stdout only
+      modEnv: { CI: 'true' },
```
iliapolo: Why is this relevant? does the code do something different if this is not set?

iliapolo (changes requested):

packages/aws-cdk/lib/cli/cdk-toolkit.ts
sai-ray: Updated, gated behind `if (options.json)`.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/list/cdk-cdk-ls-ci-no-synth-time-on-stdout.integtest.ts
sai-ray: Yeah, the behavior changes with these. With the default fixture options we don't get a clean listing, we get the full verbose dump with stderr. So the flags strip that back to just the stack listing on stdout:

- `CI: 'true'` puts non-error output on stdout (default routes it to stderr, where the bug wouldn't surface).
- `verbose: false` drops the `-v` debug dump. An earlier run without it failed on the very first line: https://github.com/aws/aws-cdk-cli/actions/runs/27655942876/job/81790649855
- `captureStderr: false` keeps stderr noise out so only stdout is asserted on.

iliapolo (approved):

PR #1638, opened 2026-06-17
chore: ignore workspace packages in Dependabot config

9pace:
Yarn berry records workspace deps declared as `"<pkg>": "^0.0.0"` with an explicit `npm:^0.0.0` descriptor. Dependabot reads the descriptor and treats it as a registry dep, producing spurious bumps like #1613 (reverted in #1629)

Adds an `ignore` list derived from `TypeScriptWorkspace` subprojects, so new workspaces are picked up automatically on `npx projen`.

Fixes #

### Checklist

---


rix0rrr (approved):

.projenrc.ts
```diff
@@ -1722,6 +1686,10 @@ new pj.YamlFile(repo, '.github/dependabot.yml', {
         'allow': [{
           'dependency-type': 'production',
         }],
+        'ignore': repoProject.subprojects
+          .filter((p): p is yarn.TypeScriptWorkspace => p instanceof yarn.TypeScriptWorkspace)
```
mrgrain: subprojects already is guaranteed to be a `TypeScriptWorkspace`. the filter we have in other places is outdated and unnecessary.

.projenrc.ts
```diff
@@ -1722,6 +1686,10 @@ new pj.YamlFile(repo, '.github/dependabot.yml', {
...
+          .map(p => ({ 'dependency-name': p.name }))
+          .sort((a, b) => a['dependency-name'].localeCompare(b['dependency-name'])),
```
mrgrain, line 1727: nit: seems unnecessary, but fine 🤷🏻

mrgrain (commented):
some nits

mrgrain (approved):

PR #1643, opened 2026-06-18
fix: remove broken cdk alias entrypoint metadata

Pugsin:
## Summary
- set the `cdk` alias package `entrypoint` to an empty string in projen config
- remove generated `main` and `types` fields from `packages/cdk/package.json`
- keep the existing CLI shim `bin/cdk` and `aws-cdk` runtime dependency

## Why
The published `cdk@2.1128.0` package is a CLI alias package. Its tarball includes `bin/cdk`, but does not include `lib/index.js` or `lib/index.d.ts`, while package metadata advertises those files through `main` and `types`.

## Validation
- Checked `npm pack cdk@2.1128.0`: tarball includes `bin/cdk`, package metadata points to missing `lib/index.js` / `lib/index.d.ts`
- Checked projen `NodePackageOptions`: `entrypoint: ''` omits `main` from package metadata
- Verified PR diff only removes the broken generated metadata and adds the projen source setting


mrgrain:
@Pugsin looks good but you'll need to include the contribution statement. Best to always read the pull request template of a repo you are contributing to.

mrgrain:
@Pugsin still interested in fixing this?

PR #1644, opened 2026-06-18
fix(aws-cdk): correct project-name argument mapping in cdk init

### LINKED ISSUE #31992: cdk: add option to specify templateName to `cdk init`
### Describe the feature

When running `cdk init`, I want to pass templateName as a command option.  
Note: The templateName context is taken from the following code. 
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/init.ts#L234

### Use Case

I'm always frustrated when I have to manually change the stack name and file names generated by `cdk init`...  
This could be resolved by making a `templateName` option available.

### Proposed Solution

I think the steps for possible implementation are as follows

1. Add `name` option

Add the `name` option to the following code.  
(I think `n` is a good alias)

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/cli.ts#L304-L308

2. Modify the implementation of `packages/aws-cdk/lib/init.ts`

see: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/init.ts

Add the `name` property to `CliInitOptions` and modify the `install` method as follows

```diff
-  public async install(language: string, targetDirectory: string, stackName?: string) {
+  public async install(language: string, targetDirectory: string, name?:string, stackName?: string) {
  // ... some implementations
    const projectInfo: ProjectInfo = {
-     name: decamelize(path.basename(path.resolve(targetDirectory))),
+     name: decamelize(name ?? path.basename(path.resolve(targetDirectory))),
      stackName,
    };
  // ... some implementations
  }

```

### Other Information

_No response_

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.146.0

macOS Sonoma v14.0

TORIFUKUKaiou:
The `--project-name` (`-n`) option for `cdk init` was silently ignored because the legacy code path in `cli.ts` referenced `args.name` (a non-existent property) instead of `args.projectName`.

Yargs automatically converts the kebab-case option `--project-name` to `args.projectName` via camelCase conversion, but the mapping incorrectly used `args.name`, causing the value to always be undefined and falling back to the directory name.

Also fixes the test description that incorrectly referred to the option as `--name` instead of `--project-name`.

Fixes aws/aws-cdk#31992

### Validation

I verified the fix locally in the devcontainer environment by running the modified CLI and comparing the outputs.

1. **Verify `yargs` argument parsing**:
   Added a new unit test in `packages/aws-cdk/test/cli/cli-arguments.test.ts` to verify that the `--project-name` option correctly maps to the `projectName` property in the config.

2. **Verify generated project structure and files**:
   I generated two projects with the same target project name (`awesome-app`)—one by relying on the default fallback (directory name) and the other by explicitly specifying `--project-name`:

   ```bash
   # 1. Project without --project-name (inheriting directory name)
   mkdir /tmp/awesome-app && cd /tmp/awesome-app
   node /workspace/aws-cdk-cli/packages/aws-cdk/bin/cdk init app --language typescript --generate-only

   # 2. Project with --project-name in a different directory
   mkdir /tmp/cdk-verify-test && cd /tmp/cdk-verify-test
   node /workspace/aws-cdk-cli/packages/aws-cdk/bin/cdk init app --language typescript --project-name awesome-app --generate-only

   # 3. Compare the two generated projects
   diff -r /tmp/awesome-app /tmp/cdk-verify-test
   ```

   **Result**: 
   The `diff` command yielded **no differences**, confirming that `--project-name awesome-app` was correctly respected and generated identical files (including `package.json`, stack file names, and class names inside `lib/` and `bin/`) regardless of the directory name.

### Checklist

---


packages/aws-cdk/test/cli/cli-arguments.test.ts
```diff
@@ -122,6 +122,21 @@ describe('yargs', () => {
       globalOptions: expect.anything(),
     });
   });
+
+  test('init --project-name is correctly passed through', async () => {
```
sai-ray, line 126: This test checks that the `--project-name` flag gets parsed into the right field, which is fine. That part already worked before the fix though, so it passes either way and won't catch the bug coming back. The bug was actually a layer down, the parsed value wasn't making it into `cliInit` and came through as `undefined`. Could we add a test that runs the `init` command and confirms the name reaches `cliInit`?

sai-ray (changes requested):
Hey @TORIFUKUKaiou, nice catch and thanks for fixing this! It's been silently doing nothing since it was added, the handler was reading the wrong property, so it always fell back to the folder name.

One thing: we'll need a test that actually runs the command with option and confirms the name comes through, so this can't quietly break again.

packages/aws-cdk/test/cli/cli-arguments.test.ts
TORIFUKUKaiou: Thank you for the catch and the helpful feedback, @sai-ray!

I agree that we need a test that runs the actual command to prevent future regressions. I added a CLI command test in `packages/aws-cdk/test/commands/init.test.ts` that runs `cdk init` with `--project-name` and verifies both the generated `package.json` name and the stack file/class name.

I kept the parser test in `cli-arguments.test.ts` as well, so the yargs mapping remains covered separately.

The new commit has been pushed. Thanks again for the guidance!

sai-ray:
Could you also update your branch with the latest main before we run CI? It's a bit behind, and I'd like CI to run against the current base. Thanks!

TORIFUKUKaiou:
Done! I've rebased my branch with the latest main and pushed the updates. The branch is now up to date with the current base. Thanks!

sai-ray (approved):

PR #1651, opened 2026-06-19
feat(cli): `--debug-cli` reports active handles that block process exit

sai-ray:
In some CI environments a `cdk` command finishes its work but the process never exits, and the job hangs until the runner kills it (issue #1217). It's intermittent and environment-specific, so there's no way today to see what is keeping the process alive.

This PR adds that visibility behind the existing `--debug-cli` flag. If the process is still alive shortly after the work is done, the CLI prints the handles holding it open and where each was created. It's a diagnostic only and on a normal run nothing is printed.

**How it works**

One module, `lib/cli/debug-handles.ts`, built on Node's `async_hooks`. `LeakedHandleTracker` watches every async resource as it's created (recording its type and creation stack) and forgets it when it's cleaned up. At exit, `report()` lists the ones still keeping the process alive, skipping handles that were garbage collected or `.unref()`'d since those don't block exit.

Only libuv handle types are tracked, the 14 in `TRACKED_TYPES`. Handles hold the event loop open and expose `hasRef()`, while requests are transient and cannot cause a hang. This is also what keeps the cost down, a `cdk synth` creates 1,046 async resources but only 6 of them are a tracked type, because `PROMISE` (899 of them) and `TickObject` (43) are excluded by the allowlist.

Wiring in `cli.ts` is two lines. `exec()` starts tracking, because only it has the parsed arguments and tracking has to begin before the CLI opens anything:

```ts
handleTracker = argv.debugCli ? trackLeakedHandles() : undefined;
```

and `cli()` schedules the report from its `finally`, so a command that fails before reaching `exec`'s own cleanup still gets one, and so the grace period starts after telemetry has finished its network calls rather than during them:

```ts
handleTracker?.scheduleReport(ioHelper);
```

The grace timer lives inside the tracker rather than in `cli.ts`, so `report()` can never be scheduled without it. That timer is the key bit, a handle open the instant work finishes might just be closing on a clean shutdown. The timer lets the normal exit path finish first, and because it's `.unref()`'d it never fires on a clean run. The report only shows up when the process is genuinely still alive after the window, which is likely the hang.

**Behavior change to declare**

`--debug-cli` now raises the CLI log level from `info` to `debug`. Without this the report it exists to print is filtered out and the flag does nothing visible. It stops at `debug` and deliberately does not go to `trace`, since `trace` also unmasks AWS SDK request logging whose payloads include whole CloudFormation templates. `verbose` is only ever raised above this level, so taking the more verbose of the two can never walk back what `-v` asked for.

Consequence, `--debug-cli` and `--debug` (which implies it) are more verbose than before.

**What the user sees**

The frames a user sees point into the bundle they installed from npm, not into any code of their own. Real output from a CLI installed from a tarball with `npm install`:

```cli
2 handles still keeping the CLI process alive:

# TCPWRAP (open network connection)
  call stack:
    <anonymous> (/private/tmp/customer-test/node_modules/aws-cdk/lib/index.js:319869:29)
      return socket.addRequest(req, connectOpts);

# TCPWRAP (open network connection)
  (opened entirely inside Node internals, no CLI frames to show)
```

The report gives a position in the bundle and the line of code at it. The bundle has no package names left in it, so identifying the library is a second step, grepping node_modules for that source line. Here return socket.addRequest(req, connectOpts) is unique to agent-base, the proxy layer proxy-agent sits on. That's the whole diagnostic path, and today a user gets none of it, just a command that finishes and a process that sits there.

Each entry shows the type, a plain-language description, the function it was created in when that name is meaningful, and every application frame with the line of code beneath it (greyed). Handles opened entirely inside Node are still listed so nothing is hidden. When no tracked handle explains the hang the report says so explicitly and hands over what Node itself reports as active, rather than printing an empty list.

Reproducing this needs the environment to hold the socket, so the run above wraps `net.Socket.prototype.connect` from a `--require` script to open one duplicate socket alongside the CLI's own and leave the original untouched. The wrapper's own two frames sat above the bundle frame in the raw output and are elided here. The bundle frame is the one a real hang produces, and it is unmodified.

The other handle types were checked the same way, by compiling deliberate leaks of six types into `cli.ts` itself, bundling, packing a tarball and installing it with `npm install`, so nothing is injected at runtime and every frame is a position in the shipped `index.js`. That scaffolding is not part of this PR. Output:

```cli
7 handles still keeping the CLI process alive:

# Timeout (timer from setTimeout or setInterval)
  created in leakRepeatingTimer()
  call stack:
    leakRepeatingTimer (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312739:3)
      setInterval(() => {
    installLeakProbe (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312771:3)
      leakRepeatingTimer();
    exec (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:368650:3)
      installLeakProbe();

# FSEVENTWRAP (file-system watcher)
  created in leakDirectoryWatcher()
  call stack:
    leakDirectoryWatcher (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312759:8)
      fs51.watch(os14.tmpdir(), () => {
    ...

# STATWATCHER (file-system stat watcher)
  created in leakStatWatcher()
  call stack:
    leakStatWatcher (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312763:8)
      fs51.watchFile(path48.join(os14.tmpdir(), "cdk-leak-probe-target"), () => {
    ...

# TCPSERVERWRAP (listening TCP server)
  (opened entirely inside Node internals, no CLI frames to show)

# TCPWRAP (open network connection)
  created in leakOutboundSocket()
  call stack:
    leakOutboundSocket (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312756:7)
      net.connect(port, "127.0.0.1");
    <anonymous> (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312776:33)
      leakListeningServer((port) => leakOutboundSocket(port));
    <anonymous> (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312752:5)
      onPort(server.address().port);

# TCPWRAP (open network connection)
  (opened entirely inside Node internals, no CLI frames to show)

# Immediate (pending setImmediate callback)
  created in again()
  call stack:
    again (/private/tmp/leaktest/node_modules/aws-cdk/lib/index.js:312744:5)
      setImmediate(again);
```

`fs51.watch` and `os14.tmpdir()` are esbuild's renamed imports, which is what confirms the snippets are read out of the bundle rather than out of the original source.

Locations are captured with `Error.captureStackTrace` and a structured `prepareStackTrace`, so line and column come from V8's own parse metadata. The source line printed underneath is a separate step, read from disk at report time, and it degrades independently, if the file cannot be read the location still prints.

The tracker also only sees handles created in the CLI process. The CDK app runs as a separate child process, so a user's own app code can never appear in a report.

Signal handlers are deliberately absent from reports. `SIGNALWRAP` is in `TRACKED_TYPES`, but a handler registered with `process.on('SIGUSR2', ...)` reports `hasRef()` as `false`, since Node unrefs signal handles. Confirmed directly rather than inferred, a process whose only handle is a signal handler exits immediately. So the `hasRef()` filter drops it, correctly.

**Tests**

`test/cli/debug-handles.test.ts`, 21 tests using real resources with only the IoHost faked. Covers a leaked timer and a real leaked TCP connection being reported, `.unref()`'d and garbage-collected and already-fired handles being excluded, promise and tick noise being filtered, the report being emitted at `debug` and not `info`, per-column location, the minified-line window, the unreadable-file path, the capacity cap, and `scheduleReport` staying silent until the grace period passes and swallowing a failure rather than crashing the process it is watching.

`cdk-debug-cli-handle-report.integtest.ts` asserts the log-level raise survives bundling, which is the part unit tests cannot reach. It passes `verbose: false` explicitly, because the fixture injects `-v` by default and that would make the test pass whether or not the flag did anything.

Manually verified end to end from a tarball built at HEAD and installed with `npm install`, the way a user would get it. First against a healthy CLI, to check the flag stays out of the way:

| Command | Result |
|---|---|
| `cdk --help` | `--debug-cli` documented |
| `cdk synth --debug-cli`, healthy environment | exit 0, 2.046s, no report |
| `cdk synth --debug-cli`, socket held open | report names the bundle location that opened it |
| `cdk synth`, same socket held open | hangs, 2 lines of stderr, no explanation |
| `cdk synth --debug`, same socket held open | report appears |

Row four is the before case, an unexplained hang, and row three is the same hang with this change.

Then against a tarball with deliberate leaks of six handle types compiled into `cli.ts`, which is where the bundled output above comes from. This is the flag matrix, since the leak is unconditional and every run either reports or does not:

| Command | Report | Note |
|---|---|---|
| `cdk doctor --debug-cli` | 7 handles, 6 types | full output above |
| `cdk synth --debug-cli` | 7 handles, 6 types | synth output itself still correct |
| `cdk doctor` | none | 9 stderr lines, no explanation, the status quo |
| `cdk doctor --debug` | 7 handles | the implied flag works |
| `cdk doctor -vvv` | none | verbosity alone does not enable tracking |
| `cdk doctor --debug-cli -vvv` | 7 handles | `-vvv` is not downgraded by the level floor |

### Checklist
- [x] Unit tests added
- [x] Integration test added

---


rix0rrr:
> What the user sees

Can you confirm this with a packaged CLI? Because we bundle all the source, and I'm pretty sure source locations  are going to end up as `cli.js:1:35891`.

I don't think we have to go to source maps (though we can), but we should at least have the trace of all functions.

sai-ray:
> > What the user sees
> 
> Can you confirm this with a packaged CLI? Because we bundle all the source, and I'm pretty sure source locations are going to end up as `cli.js:1:35891`.
> 
> I don't think we have to go to source maps (though we can), but we should at least have the trace of all functions.

In the bundle the locations just came out as `index.js`. Function names survive bundling fine, so updated the report to lead with the function and shows the line of code that opened the handle:
```cli
# Timeout (timer from setTimeout or setInterval)
  created in scheduleRefresh()
  call stack:
    setInterval(() => this.refresh(), 60_000)
```
I'd like to get the actual file and line too and was curious how you'd approach it. From what I found, shipping a source map would do it since `node-backpack` already passes a sourcemap option through to `esbuild`, it just isn't exposed on the `pack` command in use, and the tracker would switch to reading `error.stack` instead of CallSites since that path is sourcemap aware. Any ideas on the cleanest way, or something simpler?

iankhou (changes requested):
Why gate it behind `--debug-cli` and not `--verbose`, which we already have? Does the feature introduce additional response time/latency? If so, what is that cost?

If we have a separate flag, I think it should be more specific. From reading the name, I wouldn't know the intended difference between `--debug-cli` and providing `--verbose`, which sets our log level to DEBUG.

mrgrain:
> Why gate it behind `--debug-cli` and not `--verbose`, which we already have? Does the feature introduce additional response time/latency? If so, what is that cost?
> 
> If we have a separate flag, I think it should be more specific. From reading the name, I wouldn't know the intended difference between `--debug-cli` and providing `--verbose`, which sets our log level to DEBUG.

There was a whole discussion. tl;dr
- `--verbose` should not change behavior, just increase output. The log level `DEBUG` is a somewhat standard naming that we only really use internally and don't expose to customers.
- `--debug`  is a shortcut for `--debug-app` and `--debug-cli`. Because we have two different entities to debug here that one might want to distinguish on. Either of these debugs will increase the respective log level, but also implies functional changes to the execution. With the most relevant effect that it will be slower.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -258,6 +270,15 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
       await trapErrors(ioHelper, 'Could not display notices', () => notices.display());
     }
+
+    if (argv.debugCli) {
+      // If the process is still alive after the grace period, something is
+      // keeping the event loop busy. Dump the leaked handles so the user can
+      // see why. .unref() so this timer itself doesn't keep us alive.
+      setTimeout(() => {
+        void reportLeakedHandles(ioHelper);
+      }, HANDLE_DUMP_GRACE_MS).unref();
```
mrgrain, line 280: nit: It'd probably have this inside a helper function `detectLeakedHandles(ioHelper)` so that the `cli.ts` file is less busy. It sounds like `reportLeakedHandles` really should never be called without this `setTimeout`.

packages/aws-cdk/lib/cli/debug-handles.ts
```diff
@@ -0,0 +1,280 @@
...
+    // doesn't flood the report.
+    return line && line.length > 200 ? `${line.slice(0, 200)}…` : line;
+  } catch {
+    // The source file may not be readable (e.g. a bundled or eval'd frame).
+    // The location is still useful on its own, so reporting continues without it.
+    return undefined;
+  }
+}
+
+const tracker = new LeakedHandleTracker();
```
mrgrain, line 263: Importing this file now has a side effect that might not be clear or intentional. instead we can do this inside a `tracLeakedHandles`  method that creates the class and calls start.

packages/aws-cdk/lib/cli/debug-handles.ts
```diff
@@ -0,0 +1,280 @@
...
+      const handle = r.handleRef.deref();
+      // Already garbage collected, so no longer keeping the loop alive.
+      if (handle === undefined) {
+        return false;
+      }
+      return handle.hasRef?.() ?? true;
+    });
+    this.watched.clear();
+
+    await ioHelper.defaults.info(`${leaks.length} ${leaks.length === 1 ? 'handle' : 'handles'} still keeping the CLI process alive:`);
```
mrgrain, line 169: Are these really at `info` level?

mrgrain (changes requested):

packages/aws-cdk/lib/cli/cli-config.ts
```diff
@@ -31,7 +31,7 @@ export async function makeConfig(): Promise<CliConfig> {
       'verbose': { type: 'boolean', alias: 'v', desc: 'Show debug logs (specify multiple times to increase verbosity)', default: false, count: true },
       'debug': { type: 'boolean', default: false, implies: ['debug-app', 'debug-cli'], desc: 'Produce more detailed output to help diagnose unexpected behavior for the CDK app and CDK CLI. Note that this will significantly slow down synthesis time.' },
       'debug-app': { type: 'boolean', default: false, desc: 'Debug the CDK app. Logs additional information during synthesis, such as creation stack traces and sets the CDK_DEBUG environment variable. Will slow down synthesis.' },
-      'debug-cli': { type: 'boolean', default: false, desc: 'Debug the CDK CLI itself.' },
+      'debug-cli': { type: 'boolean', default: false, desc: 'Debug the CDK CLI itself. Records where the CLI opens every network connection, timer and other async resource, so that a CLI which fails to exit can report what is keeping it alive, and raises the CLI log level to show that report. Captures a stack trace per resource, so it will slow down execution.' },
```
mrgrain, line 34: we don't need all that detail here

packages/aws-cdk/lib/cli/cli.ts
sai-ray: Done. The timer lives in the tracker now as `scheduleReport()`, along with the grace constant, so `report()` can't be scheduled without it. `cli.ts` is down to two lines, one to start tracking and one in `cli()`'s `finally` so a command that throws still gets a report.

packages/aws-cdk/lib/cli/debug-handles.ts
sai-ray: Done, `trackLeakedHandles()` creates it and calls `start()` now

packages/aws-cdk/lib/cli/debug-handles.ts
sai-ray: It's at `debug` now. It needed the level raise to go with it, since `--debug-cli` wasn't raising anything and the report was getting filtered out.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -80,6 +99,15 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
...
   }
 
+  // `--debug-cli` raises the CLI-side log level, which is what makes its own
+  // handle report visible without a second flag. Only as far as DEBUG, never
+  // TRACE: TRACE also unmasks AWS SDK request logging, whose payloads include
+  // whole CloudFormation templates. `verbose` is only ever raised above this, so
+  // taking the more verbose of the two can never walk back what `-v` asked for.
+  if (argv.debugCli && ioMessageLevel === 'info') {
+    ioMessageLevel = 'debug';
+  }
```
mrgrain, line 109: A/ Combine this with the block above, probably a helper function to determine the `ioMessageLevel`. 

B/ Just `== 'info'` might not be what you want. What if someone specified error? what is YOUR intention here? Change it when no specific level is provided? Or force debug level no matter what?

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -971,6 +999,15 @@ export function cli(args: string[] = process.argv.slice(2)) {
       } catch (e: any) {
         await CliIoHost.get()?.asIoHelper().defaults.trace(`Ending Telemetry failed: ${e.message}`);
       }
+
+      // Last thing we do, on both the success and the failure path: if the
+      // process is still alive after the grace period, something is keeping the
+      // event loop busy, so report the leaked handles. All of the CLI's own work
+      // is finished by now, so anything left is genuinely unaccounted for.
```
mrgrain, line 1006: So this is scheduling something to check later? comment doesn't make it clear.

mrgrain (changes requested):
Some small comments left. Otherwise looks good!

packages/aws-cdk/lib/cli/cli-config.ts
sai-ray: shortened the description

packages/aws-cdk/lib/cli/cli.ts
sai-ray: Yes, and the comment is reworded now. It arms an unref'd timer instead of printing anything there. On a clean exit the process is gone before the timer fires, so nothing prints. If it does fire, something is still holding the event loop open and the CLI's own work is already done.

packages/aws-cdk/lib/cli/cli.ts
sai-ray: Both blocks are one `determineIoMessageLevel(argv)` now. The intention is a minimum. `--debug-cli` needs `debug` or its report gets filtered out, and anything louder is left alone, so `-v -v --debug-cli` still gives `trace`. Without `--debug-cli` nothing changes, no flags, `-v`, and `-v -v `behave as before.

mrgrain (approved):

PR #1653, opened 2026-06-19
feat: construct tree and violations

megha-narayanan:
Adds two read-only views to the `cdk explore` web UI.

- **Construct tree** (`GET /api/tree`): the synthesized construct hierarchy. Each
  node is flagged with the highest severity of any policy violation on it, so
  problem areas are visible at a glance.
- **Policy-validation panel** (`GET /api/policy-validation`): violations grouped
  by rule, sorted by severity, labeled with the originating plugin.

The server reads the cloud assembly in `cdk.out` and returns a wire-stable,
app-relative view. The backend owns the full transform from core construct node
to displayed node, including default-child collapse and the per-node
highest-severity join, so the frontend renders without re-deriving any of it.

## Out of scope 

- Navigation between a construct, its synthesized template, and its source. The
  wire model already carries `templateFile` and `sourceLocation`, but no 
  navigation UI yet.
- Frontend tests (adds deps, so future PR)

frontend:
Screenshot 2026-07-01 at 11 32 09 AM

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/web/severity.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/web/severity.ts
@@ -1,0 +1,67 @@
+/**
+ * Severity vocabulary for CDK policy violations, shared by the server (ranking
+ * violations onto construct-tree nodes) and the SPA (sorting and coloring the
+ * violations panel) so the rules live in exactly one place.
+ */
+
+import type { WebViolation, WebViolationSeverity } from './protocol';
+
+/** Severities CDK policy validation reports natively. */
+const KNOWN_SEVERITIES = new Set<string>(['fatal', 'error', 'warning', 'info', 'custom']);
+
+/**
+ * Classify a severity from a raw, unvalidated policy-validation report into the
+ * wire model. Known values and a missing severity pass through unchanged; a
+ * value outside the set is mapped to `custom`, keeping the raw value as the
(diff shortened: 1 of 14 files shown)


packages/@aws-cdk/cdk-explorer/lib/web/severity.ts
```diff
@@ -0,0 +1,67 @@
...
+const KNOWN_SEVERITIES = new Set<string>(['fatal', 'error', 'warning', 'info', 'custom']);
+
+/**
+ * Classify a severity from a raw, unvalidated policy-validation report into the
+ * wire model. Known values and a missing severity pass through unchanged; a
+ * value outside the set is mapped to `custom`, keeping the raw value as the
+ * label, so a newer or non-conforming report can never produce a wire value
+ * that violates {@link WebViolationSeverity}.
+ */
+export function classifyReportSeverity(
```
ShadowCat567, line 19: I don't think I entirely understand what this function is doing. Is there a possibility that a construct can get assigned a severity and it is not one we recognize? (Error, Fatal, Warning, etc)

megha-narayanan: Severity comes from the validator plugin, where it's just a free-form string (so it can be anything), and construct annotations, where addWarning/addError/addInfo map to warning/error/info.

ShadowCat567 (approved):

PR #1655, opened 2026-06-21
chore: restrict dependabot to security updates only

iliapolo:
In https://github.com/aws/aws-cdk-cli/pull/1035, we added a dependabot configuration that adds an `auto-approve` label to NPM update PRs created by dependabot. This was done so that such PRs can be auto merged. However, we inadvertently also increased the scope of which PRs should depedabot create. 

**Prior to https://github.com/aws/aws-cdk-cli/pull/1035**

The `npm` package registry was missing - which means dependabot would **only** create PRs that resolve a security alert, as specified in our repository settings:

Screenshot 2026-06-21 at 7 13 00 PM

**After https://github.com/aws/aws-cdk-cli/pull/1035**

Dependabot creates PRs for regular npm dependency upgrades, duplicating our custom [projen based workflow](https://github.com/aws/aws-cdk-cli/blob/main/.github/workflows/upgrade.yml). For example:

- https://github.com/aws/aws-cdk-cli/pull/1622
- https://github.com/aws/aws-cdk-cli/pull/1621
- https://github.com/aws/aws-cdk-cli/pull/1620
- https://github.com/aws/aws-cdk-cli/pull/1619
- https://github.com/aws/aws-cdk-cli/pull/1618

----

According to [GitHub docs](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/configure-security-updates#overriding-the-default-behavior-with-a-configuration-file):

> If you only require security updates and want to exclude version updates, you can set open-pull-requests-limit to 0 in order to prevent version updates for a given package-ecosystem.

We now use this mechanism in order to limit dependabot to security alerts but preserve the addition of the `auto-approve` lablel. 

### Checklist

---


.projenrc.ts
```diff
@@ -1714,14 +1718,11 @@ new pj.YamlFile(repo, '.github/dependabot.yml', {
...
-        'schedule': { interval: 'weekly' },
-        'cooldown': {
-          'default-days': dependabotCooldown,
-        },
+        'schedule': { interval: 'daily' },
+        // disable version updates, leaving only security updates.
+        // see https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/configure-security-updates#overriding-the-default-behavior-with-a-configuration-file
+        'open-pull-requests-limit': 0,
         'labels': ['auto-approve'],
-        'allow': [{
```
iliapolo, line 1722: Removed because we want to resolve security alerts in dev deps as well.

.projenrc.ts
```diff
@@ -1714,14 +1718,11 @@ new pj.YamlFile(repo, '.github/dependabot.yml', {
     updates: [
       {
         'package-ecosystem': 'npm',
-        'schedule': { interval: 'weekly' },
-        'cooldown': {
```
iliapolo, line 1718: Removed because we want to grab security fixes as fast as possible, assuming they themselves don't need a cooldown period.

mrgrain: I'd like to keep some cooldown if we are auto merging packages. It can be shorter then before. I'm not sure I trust Dependabot/our understanding of the config enough here.

mrgrain (changes requested):
Can we keep a cooldown please?

mrgrain (approved):

PR #1657, opened 2026-06-22
fix(integ-runner): stack traces are in golden snapshot manifest

rix0rrr:
The golden snapshots contain stack trace information captured by the AWS CDK Library. This PR sets a context flag to ask the CDK Library to not do that.

Upon diving into this code, I got an itch I couldn't resist, so I refactored it *a lot* (and still in the middle of doing more):

- Both snapshot validator and test runner inherited from some base class (`IntegRunner`) that they only inherited from to get access to some helper functions. Renamed this class to `CdkIntegHelper` and used composition instead of inheritance to make these routines available in the other "runners".
- The tests do an awful lot of `synth`ing just to get at the test definition information that is *inside* of the CDK app. In this change, we at least remove one instance of `synths`: when a snapshot is being checked for invalidation, we used to do 2 `synths` but are now doing just 1 (mostly). The reason was `enableLookups`, see below.
- Snapshot generation logic is now more centralized; other synths (actual deployments do *another* synth just to get the test definition, even though we just got it from snapshot invalidation) could potentially be refactored and deduplicated in the future, but haven't done so yet. Plus there is still a lot of code duplication that we could probably move into the helper.

enableLookups
-------------

This setting is slightly weird: it is supposed to control whether the app can/is going to do context lookups.

If this is enabled, the integ runner passes fake context to the golden snapshot generator, and if it's not enabled, it doesn't.

The problem is that the value of this flag is determined by the CDK app, and we must already have decided on its value before we invoke the CDK app. This is a dependency that is hard to shake, and the reason the original code did 2 synths: one to get the value of `enableLookups`, then another one to do the "actual" synth.

Since the golden snapshot lookups are always faked, we could have always passed the context and have been fine. The problem is that the fake context affects the generated templates (for example: the number of subnets in a VPC is affected by these settings), so we can't wholesale change this setting now for all future runs, because it would cause snapshot invalidations.

So what we will do instead is read the `enableLookups` value from the golden snapshot settings, and use the same value to do the new snapshot synthesis. If after synthesizing we find out that we guessed wrong, we synth again with the right setting. This should almost never be the case.

We can ratchet to a future in which this setting is always turned on and the code that deals with backwards compatibility is removed, but for that we need to be able to change `integ.json` from the integ runner, and we don't have code to do that right now. This will be an exercise for the future.

---

### DIFF
--- a/packages/@aws-cdk/integ-runner/lib/runner/engine.ts
+++ b/packages/@aws-cdk/integ-runner/lib/runner/engine.ts
@@ -6,7 +6,7 @@ import { ToolkitLibRunnerEngine } from '../engines/toolkit-lib';
  *
  * Only the toolkit-lib engine is supported.
  */
-export function makeEngine(options: CdkIntegHelperOptions): ToolkitLibRunnerEngine {
+export function makeEngine(options: EngineOptions): ToolkitLibRunnerEngine {
   return new ToolkitLibRunnerEngine({
     workingDirectory: options.test.directory,
     showOutput: options.showOutput,
@@ -17,3 +17,5 @@ export function makeEngine(options: CdkIntegHelperOptions): ToolkitLibRunnerEngi
     caBundlePath: options.caBundlePath,
   });
 }
+
+export type EngineOptions = Pick<CdkIntegHelperOptions, 'test' | 'showOutput' | 'env' | 'region' | 'profile' | 'proxy' | 'caBundlePath'>;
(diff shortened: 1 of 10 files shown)


packages/@aws-cdk/integ-runner/lib/runner/engine.ts
```diff
@@ -1,12 +1,12 @@
...
+import type { CdkIntegHelperOptions } from './cdk-integ-helper';
 import { ToolkitLibRunnerEngine } from '../engines/toolkit-lib';
 
 /**
  * Creates the engine for running integration tests.
  *
  * Only the toolkit-lib engine is supported.
  */
-export function makeEngine(options: IntegRunnerOptions): ToolkitLibRunnerEngine {
+export function makeEngine(options: CdkIntegHelperOptions): ToolkitLibRunnerEngine {
```
mrgrain: nit: I get how this falls out of your change, but it now reads very strange. These options aren't really to configure the helper, but actually to configure "the runner" (whatever that is).

mrgrain (approved):

mrgrain:
nit: This introduces the concept of a "golden snapshot" but doesn't really explain anywhere what this is compared to other snapshots. I get this is code internal, but would be nice to have that written down somewhere.

PR #1660, opened 2026-06-22
chore(integ-runner): remove CLI as run-time dependency

rix0rrr:
This is probably not necessary anymore.

---


.projenrc.ts
```diff
@@ -1495,7 +1495,7 @@ const integRunner = configureProject(
     deps: [
       cloudAssemblySchema.customizeReference({ versionType: 'any-future' }),
       cloudAssemblyApi.customizeReference({ versionType: 'exact' }),
-      cli.customizeReference({ versionType: 'exact' }),
+      // cli.customizeReference({ versionType: 'exact' }),
```
mrgrain:
```suggestion
```

mrgrain (approved):
Nice, can we just remove it instead of commenting out?

PR #1663, opened 2026-06-22
fix(cli): `cdk list` pollutes stdout with synth and dependency status lines

sai-ray:
`cdk ls` prints status lines like `✨  Synthesis time: X.XXs` and `Including dependency stacks: ...` alongside the stack listing. In CI these go to stdout, next to the listing, which breaks scripts that parse the output (issue aws/aws-cdk#38165).

#1637 tried to fix this by suppressing the synth-time message under `--json`, but plain `cdk ls` in CI was still affected, and the dependency line was never handled.

Both lines now get suppressed on the list path so stdout is only the listing. The dependency line got a message code in #1677 so it can be targeted by code, same as the synth-time line:

```ts
this.ioHost.once(IO.CDK_TOOLKIT_I1000, () => ({ preventDefault: true }));
this.ioHost.once(IO.CDK_TOOLKIT_I1002, () => ({ preventDefault: true }));
```
This applies in all cases, not just CI or `--json`.

#1603 added the synth-time line and the dependency line predates it. Both are info-level status that cdk ls should not print, so both are suppressed in all cases. CI is just where they corrupted stdout.

Pre #1603:

Screenshot 2026-06-30 at 9 43 07 AM

Current:

Screenshot 2026-06-29 at 10 41 11 PM

After:

Screenshot 2026-06-29 at 10 54 56 PM

**Tests**

- Unit: both suppressors are registered on the list path.
- Integ: `cdk ls` in CI produces only the stack listing, no status lines.

### Checklist
---


packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -966,8 +966,10 @@ export class CdkToolkit {
   ): Promise<number> {
     this.ioHost.rewriteOnce(IO.CDK_TOOLKIT_I2901, (msg) => formatStackList(msg.data.stacks, options));
 
-    // With `--json`, stdout must stay machine-parsable, so suppress the synth-time line (I1000).
-    if (options.json) {
+    // In CI, info-level output (incl. the synth-time line, I1000) routes to stdout next to the
+    // listing and breaks parsers. With --json, stdout is an explicit machine-readable contract.
+    // In both cases, suppress I1000 so stdout carries only the listing.
+    if (options.json || this.ioHost.isCI) {
```
mrgrain: Just disable them for any case please. Let us not get confused by reports saying this happens in CI. It happens everywhere.

mrgrain (changes requested):

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -966,8 +966,10 @@ export class CdkToolkit {
...
       this.ioHost.once(IO.CDK_TOOLKIT_I1000, () => ({ preventDefault: true }));
```
mrgrain: I am CERTAIN there is a second message we need to silence as well.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
sai-ray: Converted to draft while I rework this.

The current approach raises the log level to warn during listing. It works, but it's pretty blunt, it mutates shared ioHost state and ends up swallowing other info output during the call too.

So I was thinking of giving the "Including dependency stacks" line its own message code. Then I'd suppress that and the synth-time line by code on the list path. Feels self contained and matches what we already do. Does that sound right to you, would you go about it differently?

mrgrain (changes requested):
I struggle understanding what we are trying to achieve here. Can you please structure the PR to show relevant screenshots of the pre #1603 output (have one for every mode you are interested in) and than include screenshots of today's output.

This way it will be abundantly clear which messages got introduced and how this PR is fixing it.

packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -966,17 +966,19 @@ export class CdkToolkit {
   ): Promise<number> {
     this.ioHost.rewriteOnce(IO.CDK_TOOLKIT_I2901, (msg) => formatStackList(msg.data.stacks, options));
 
-    // With `--json`, stdout must stay machine-parsable, so suppress the synth-time line (I1000).
-    if (options.json) {
-      this.ioHost.once(IO.CDK_TOOLKIT_I1000, () => ({ preventDefault: true }));
+    // Only the listing should reach stdout; drop info status lines (synth time, dependency notes).
+    const previousLevel = this.ioHost.logLevel;
+    this.ioHost.logLevel = 'warn';
```
mrgrain: this doesn't seem like the right approach

sai-ray: reworked to suppress both lines by message code (no gate). Second message was the dependency line, registered it in #1677 and suppressed here. Updated the PR description with pre and post screenshots.

mrgrain (approved):

PR #1666, opened 2026-06-23
chore(cli): `CliIoHost` listeners can set level handle and requests

mrgrain:
Two small `CliIoHost` listener capabilities that the upcoming `cdk destroy` reroute needs. Both extend the existing `on`/`once`/`rewrite` family and change no command behavior on their own.

1. Rewrite listeners can override a message's level. A listener result may now carry a `level`, and `rewrite`/`rewriteOnce` take an optional third argument to set it. The level drives verbosity filtering and stream selection, so this lets a rerouted command re-categorize a message the action emits (for example, present a `result` as a plain `info` status line) without the action knowing. Level-only changes use `on`/`once` returning `{ level }`.

2. Listeners handle requests too. Requests now flow through the same listener pipeline as notifications, so `on`/`once`/`rewrite`/`rewriteOnce` also accept request codes. A listener can reword a question (or change its level), or answer it via a `respond` value so the host neither prompts nor prints. `respond(code, value)` and `respondOnce(code, value)` are sugar for the common "just answer it" case. This is how `cdk destroy --force` will confirm silently once it is routed through toolkit-lib.

Builds on the IoHost recording groundwork from #1661.

Fixes #

### Checklist

---

### DIFF
--- a/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
+++ b/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
@@ -5,7 +5,7 @@ import { ToolkitError } from '@aws-cdk/toolkit-lib';
 import type { HotswapResult, IIoHost, IoMessage, IoMessageCode, IoMessageLevel, IoRequest, ToolkitAction } from '@aws-cdk/toolkit-lib';
 import * as chalk from 'chalk';
 import * as promptly from 'promptly';
-import type { IoHelper, ActivityPrinterProps, IActivityPrinter, IoMessageMaker, IoDefaultMessages } from '../../../lib/api-private';
+import type { IoHelper, ActivityPrinterProps, IActivityPrinter, IoMessageMaker, IoRequestMaker, IoDefaultMessages } from '../../../lib/api-private';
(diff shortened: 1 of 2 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
sai-ray, line 793: Question on the `autoRespond` path. Listeners run once at the top of `requestResponse`, and the non-promptable branch uses `writeMessage(applied.message)` so they don't run again. But the two `autoRespond` branches still call `this.notify(...)`, and `notify` runs the listeners a second time.                                                                                                         

So under `--yes`, a listener on a request code fires twice. The visible case is a `rewrite` overwriting the `(auto-confirmed)` suffix on the second pass, but anything the listener does runs twice. I don't think this reaches `destroy --force` though, it looks latent rather than on the real path.

Is the re-notify on the `autoRespond` branches deliberate, or should they skip the second listener pass like the non-promptable one?

mrgrain: Good catch, fixed!

sai-ray (approved):

PR #1667, opened 2026-06-23
fix(cli): a declined `cdk destroy` now returns exit code 1

mrgrain:
Declining a confirmation prompt was inconsistent: `cdk destroy` exited 0 while other commands exited non-zero, some "no" answers showed a generic error messages, others a more specific one. Now declining any prompt behaves the same way: the command stops with a short, command-specific message and a non-zero exit code, shown as a notice rather than a red error.

### Before

image

### After

image

## User-facing changes

When you answer "no" to a confirmation:

- `cdk deploy`: `Deployment cancelled`
- `cdk destroy`: `Deletion cancelled` (now exits non-zero; previously exited 0)
- `cdk import`: `Import cancelled`
- `cdk rollback`: `Rollback cancelled`
- `cdk orphan`: `Orphaning cancelled`
- `cdk gc`: `Garbage collection cancelled`
- `cdk flags`: `Flag update cancelled` (now also cleans up its temporary files)

## Technical details

This changes the responsibility of who owns throwing exceptions. Previously we had this centralized inside the CLI's `CliIoHost`. However this was inconsistent with how `toolkit-lib` and _some_ other code paths worked. Now the `CliIoHost` no longer aborts on a declined confirmation; it returns the answer and it is the command's responsibility to throw (or do something else).

This is more consistent for a commands, since `requestResponse` always _seemed_ like it would return a response. Commands can now check the response and act accordingly. This also allows for more specific errors and messages to be thrown.

- `cdk destroy` no longer emits `CDK_TOOLKIT_E7010`.
- The reported error name (and telemetry) is now command-specific (`DeployAborted`, `DestroyAborted`, `ImportAborted`, `RollbackAborted`, `ReplacementRollbackAborted`, `OrphanAborted`, `DeletionAborted`, `FlagsAborted`) instead of the generic `AbortedByUser`.
- toolkit-lib: new exported `AbortError` (a `ToolkitError` subclass, `AbortError.isAbortError(x)`, default message `Operation cancelled`). `deploy`, `rollback`, `orphan`, and `gc` throw it on decline; the message text changes to `<Operation> cancelled` while the error code is unchanged.
- The convention is documented in `packages/aws-cdk/docs/confirmation-prompts.md`.

Fixes #

### Checklist

---

### DIFF
--- a/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
+++ b/packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
@@ -768,7 +768,7 @@ export class CliIoHost implements IIoHost, ObservableIoHost {
       return msg.defaultResponse;
     }
 
-    const response = await this.withCorkedLogging(async (): Promise<string | number | true> => {
+    const response = await this.withCorkedLogging(async (): Promise<string | number | boolean> => {
       // prepare prompt data
       // @todo this format is not defined anywhere, probably should be
       const data: {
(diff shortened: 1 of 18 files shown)


packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -733,12 +733,13 @@ export class CliIoHost implements IIoHost, ObservableIoHost {
 
       // Basic confirmation prompt
       // We treat all requests with a boolean response as confirmation prompts
+      // Basic confirmation prompt
+      // We treat all requests with a boolean response as confirmation prompts.
```
sai-ray:
```suggestion
```

sai-ray (approved):

PR #1669, opened 2026-06-24
feat: LSP hover with resolved CFN properties and template locations

megha-narayanan:
- Adds LSP hover on construct creation lines showing the resolved CloudFormation logical ID, resource type, construct path, and top-level CFN properties with values
- Each property value links to its exact line in the synthesized template; the logical ID header links to the resource block
- Multi-resource constructs show auxiliaries inline (≤5 listed individually, >5 collapsed to a type histogram like "6× Subnet, 3× RouteTable")
- Extends `template-ranges.ts` with `resolveResourceRanges` / `indexTemplateRanges` to resolve per-property character ranges in a single parse
- Exports `cfnProperties` from the construct tree builder so the hover has access to resolved property values

example hover:
Screenshot 2026-06-24 at 12 12 52 PM

Screenshot 2026-06-24 at 12 13 44 PM

Fixes #

### Checklist

---


packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
```diff
@@ -0,0 +1,305 @@
...
+ * Template line targets for a hover, resolved from the synthesized template(s).
+ * `blocks` is keyed by construct path (the hovered resource and its auxiliaries);
+ * paths are globally unique, where stack-relative logical ids can collide across
+ * templates. `properties` is keyed by the lower-cased property name of the
+ * primary resource, so the L1 camelCase name and the template PascalCase name
+ * match regardless of acronym casing. Absent when the template can't be read,
+ * in which case values render without links.
+ */
+export interface HoverLinks {
+  readonly blocks: Record<string, LinkTarget>;
```
ShadowCat567: What does `blocks` refer to in this context? Can it get a more descriptive name?

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
```diff
@@ -0,0 +1,305 @@
...
+    return `{Fn::GetAtt ${Array.isArray(target) ? target.join('.') : String(target)}}`;
+  }
+  if (key.startsWith('Fn::')) {
+    return `{${key} …}`;
+  }
+  return undefined;
+}
+
+/** Group auxiliary resources by short CFN type, most common first, e.g. "8× Subnet". */
+function histogram(nodes: readonly ResourceConstruct[]): string {
```
ShadowCat567, line 302: What does this look like in the hover UI?

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
```diff
@@ -0,0 +1,305 @@
...
+export function buildHover(
+  nodes: readonly ResourceConstruct[],
+  selection: PrimarySelection | undefined,
+  links: HoverLinks | undefined,
+  range: Range,
+): Hover | undefined {
+  if (nodes.length === 0) {
+    return undefined;
+  }
+  const value = selection === undefined
```
ShadowCat567: Can `selection` instead be called `defaultChild`? Since it seems like the logic here revolves around whether the construct we are hovering over has a default child or not

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
```diff
@@ -0,0 +1,305 @@
...
+    lines.push('', alsoCreates(others, links));
+  }
+  return lines.join('\n');
+}
+
+function propertyLines(primary: ResourceConstruct, links: HoverLinks | undefined): string[] {
+  const properties = primary.cfnProperties ?? {};
+  const keys = Object.keys(properties);
+  const lines = keys.slice(0, MAX_PROPERTIES).map((key) => {
+    const value = `\`${renderValue(properties[key])}\``;
```
ShadowCat567, line 228: For nested properties like `S3Bucket` and `S3Key` in `Code` for Lambda functions, any particular reason why you render it like:
```
code: { s3bucket, s3key }
```
instead of: 
```
code: {
   s3bucket: <bucket name>
   s3key: <key zip file>
}
```

ShadowCat567 (commented):
Looks good overall, have some comments about variable names and UI details

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
megha-narayanan: `blocks` is the name that was used for the chunk of the template associated with a resource from the template range resolver. Renamed to `resourceLocations` which better describes what these are, a map from construct path to the location of that resource's definition in the synthesized template. Pushed.

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
megha-narayanan: realized I needed to include the case where the default is same depth as everything else, so just pushed that bit of code, but it looks like this: 

Image

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
megha-narayanan: It's not TOTALLY the default child, `selectPrimary` picks the uniquely shallowest resource on the line, which is usually the default child but could be any uniquely-shallow resource. Renamed to primaryResource, which feels more descriptive than selection without implying the default child relationship specifically. Looks ok?

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
megha-narayanan: mostly just to save hover space, nested CFN values are usually intrinsics or deeply nested objects that make the tooltip very long. Showing key names felt like a good middle ground, and the property value is clickable, so it links to the exact line in the template where the full expansion is visible. happy to expand shallow objects if feels too terse, lmk.

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/** 1-based template line of a range's start (the file: `#L` fragment is 1-based). */
+function lineOf(text: string, range: OffsetRange): number {
+  return offsetsToRange(text, range).start.line + 1;
+}
+
+/**
+ * Builds the hover for the resource(s) created on a line, given the primary
+ * primaryResource. When one resource is the construct's primary (default) child, its
```
ShadowCat567: nit: this comment is confusing if the primary resource is not actually the default child (default child of construct), can you change the wording of this comment and comments like this to describe what you mean by primary child and probably remove the reference to default child if primary child != default child

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
ShadowCat567: If we have the link to get the information, then this should be fine

packages/@aws-cdk/cdk-explorer/lib/lsp/hover.ts
ShadowCat567: Do you have a test for that? (default same depth as the other constructs)

megha-narayanan: added

ShadowCat567 (approved):

PR #1670, opened 2026-06-24
feat(cli): add 'cdk lsp' command to start the CDK Language Server over stdio

megha-narayanan:
Editors and AI agents start a language server by running a command over stdio,
but the CDK Language Server (in `@aws-cdk/cdk-explorer`) has no such entry
point. This adds one, so any LSP-capable editor or agent can get CDK
diagnostics and CodeLens by pointing at `cdk lsp`.

Wires up `cdk lsp` as a CLI subcommand that starts the server on stdin/stdout
and runs until the client closes the channel.

- Adds `@aws-cdk/cdk-explorer` as a runtime dependency so the server ships in the CLI bundle, version-aligned with the user's CDK toolchain.
- Writes nothing to stdout (CLI logs already go to stderr), keeping the JSON-RPC channel clean. `applicationDir` comes from the LSP `initialize` options, so the command takes no arguments.
- Mirrors the `cdk explore` wiring from #1598 

Fixes #

### Checklist

---


packages/aws-cdk/.projen/deps.json
```diff
@@ -197,6 +197,10 @@
       "name": "@aws-cdk/cdk-assets-lib",
       "type": "runtime"
     },
+    {
+      "name": "@aws-cdk/cdk-explorer",
+      "type": "runtime"
```
ShadowCat567, line 202: Why add the explorer as a runtime dependency in this PR?

ShadowCat567 (commented):
Looks good overall, just have one question

packages/aws-cdk/.projen/deps.json
megha-narayanan: cdk-explorer is the name of my whole package, that has both the LSP and the web side. so its a runtime dep now because `cdk lsp` needs startServer from the LSP code to run.

ShadowCat567 (approved):

PR #1672, opened 2026-06-25
feat(cli): allow customizing role session name via CDK_ROLE_SESSION_NAME

### LINKED ISSUE #1120: allow customizing role session name
### Describe the feature

When CDK automatically assumes a role (such as cdk-hnb659fds-deploy-role), currently it hard-codes the role session name to be "aws-cdk-\<username\>". This is not particularly useful for auditing when deployments are made via a CICD pipeline.

https://github.com/aws/aws-cdk/blob/400ad91cb926fb0a6d71039f8eba3bb63e7c8ca8/packages/aws-cdk/lib/api/aws-auth/sdk-provider.ts#L348

Instead, it should allow specifying the role session name, either through a command line flag or through an environment variable.

### Use Case

See above.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.28.0 (build ba233f0)

Alpine 3.16, Python 3.10.5

UTKARSH698:
Closes #1120

When the CDK CLI assumes a role (e.g. `cdk-hnb659fds-deploy-role`), it currently hard-codes the STS role session name to `aws-cdk-<username>`. When deployments run from a shared CI/CD pipeline this makes CloudTrail attribution unhelpful, since every deployment shows the same machine user.

This adds an opt-in `CDK_ROLE_SESSION_NAME` environment variable that overrides the session name when set, falling back to the existing `aws-cdk-<username>` default when unset. The variable name follows the `CDK_` env var naming convention.

### Notes
- This revives the approach from aws/aws-cdk#22631, which passed its build but was auto-closed for author inactivity rather than rejected on merit.
- The change lives in `toolkit-lib`'''s `SdkProvider.withAssumedRole`; reading an env var here is consistent with the surrounding `awscli-compatible.ts`, which already reads `AWS_PROFILE`, `AWS_REGION`, etc.

### Testing
- Added a unit test asserting that `CDK_ROLE_SESSION_NAME` overrides the default session name, alongside the existing default/sanitization tests.

---


mrgrain (changes requested):
CDK env vars must start with `CDK_`. 

What is the reason to use an env var over a cli flag? Or both? (You might be right, but I would like to see weighing up the options here.)

UTKARSH698:
Thanks for the review @mrgrain.

**Naming:** Renamed `AWS_CDK_ROLE_SESSION_NAME` -> `CDK_ROLE_SESSION_NAME` to follow the `CDK_` convention. Updated the code, the unit test, and the README env-var section (pushed in 0025193).

**Env var vs CLI flag:** A few reasons I leaned to an env var here, though I'm happy to add a flag too if you'd prefer:

- The role session name is *ambient, per-environment* context ("which pipeline/runner is assuming this role"), set once for the environment rather than per command. That maps naturally onto an env var, the same way `CDK_DEFAULT_ACCOUNT` / `CDK_DEFAULT_REGION` already work.
- The main motivation is CloudTrail attribution in CI/CD, where it should apply to *every* `cdk` invocation (`deploy`, `diff`, `bootstrap`, ...). With an env var you set it once in the job; a flag would have to be threaded onto every command.
- It also keeps parity with `toolkit-lib`'s existing env-driven auth config (`AWS_PROFILE`, `AWS_REGION`, etc. via `awscli-compatible.ts`).

A CLI flag would be the better fit for *per-invocation overrides*. If you'd like both, I'm glad to add `--role-session-name` with the flag taking precedence over the env var. Let me know which direction you prefer and I'll update.

mrgrain (approved):

PR #1673, opened 2026-06-25
feat(integ-runner): context is loaded from integ.context.json or cdk.json

rix0rrr:
It used to be that the context if integ tests could not be controlled very well, and the feature flags would float with the version of `aws-cdk-lib` that the runner was built against.

We change this to make the flags in `integ.context.json` or `cdk.json` take precedence to the floating behavior: when found, those will be loaded and configured instead.

This makes integ tests stable during `integ-runner` upgrades, and gives a single place where context flags for 100s of tests can be controlled easily.

---

### DIFF
--- a/packages/@aws-cdk/integ-runner/README.md
+++ b/packages/@aws-cdk/integ-runner/README.md
@@ -172,7 +172,7 @@ Tests:    1 failed, 9 total
 Error: Some snapshot tests failed!
 To re-run failed tests run: integ-runner --update-on-failed
     at main (packages/@aws-cdk/integ-runner/lib/cli.js:90:15)
-error Command failed with exit code 1. 
+error Command failed with exit code 1.
 ```
 
 To re-run the integration test for the failed tests you would then run:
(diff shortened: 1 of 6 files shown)


packages/@aws-cdk/integ-runner/README.md
```diff
@@ -364,3 +364,22 @@ integ-runner --help
...
+upstream from the test, context values from those files will be used and
+completely replace the built-in flags. This makes sure that your context stays
+the same, even if you upgrade `integ-runner`. It also gives you the ability
+to change a context value for all tests at once. If there are multiple files
+they will not be merged: the closest file to the test will be used.
+
+If `integ.context.json` is found, it is expected to be a one-level dictionary of
+context keys to values. If `cdk.json` is found, its `"context"` key will be used
+as a dictionary of context keys and values (same as the file that typically
+controls a CDK app's context and feature flags).
```
mrgrain, line 386: nit: What happens if both are found?

mrgrain (approved):
Suggestion (non-blocking): People tend to not like hardcoded magic file names. Would be nice to give them a config flag to point to any custom context file.

rix0rrr:
> Would be nice to give them a config flag to point to any custom context file.

I see how this would be a wise thing to do. Not going to do it just yet though, we'll see how this picks up first.

PR #1674, opened 2026-06-25
feat: support source tracing for python

megha-narayanan:
Add utilities to source resolver to parse jsii host-language creation-stack frames for Python, and drop the TypeScript-only diagnostics gate. Go-to-definition (source → template, and reverse) and policy-violation diagnostics now work for Python apps, not just TypeScript.

Requires synthing with `JSII_HOST_STACK_TRACES=1` (opt-in; jsii 1.137+ / aws-cdk-lib 2.260+). Auto-enabling it for LSP-triggered synth is a follow-up.  Java support separated into new PR due to FQN path parsing complexities. 

Example: 
Screenshot 2026-06-25 at 1 30 10 PM

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
@@ -44,24 +44,27 @@ export class SourceMapResolver {
   /**
    * Resolve the user source location from a creation stack trace (the frames
    * produced by toolkit-lib's findCreationStackTrace). Returns undefined when
-   * there's no trace (non-TS apps) or every frame is a skip-placeholder
-   * (framework-only call sites).
+   * there's no trace or no frame resolves to an in-root user source file.
    */
   public async resolveFrames(frames: readonly string[] | undefined): Promise<SourceLocation | undefined> {
     if (!frames) return undefined;
 
-    // aws-cdk-lib's renderCallStackJustMyCode (in node_modules/aws-cdk-lib/core/
-    // lib/stack-trace.js) pre-filters node_modules/node:internal frames into
-    // skip-placeholder lines. Those don't match FRAME_RE, so the first frame
-    // that parses IS the user call site.
+    // Scan for the first frame that is a supported source file inside the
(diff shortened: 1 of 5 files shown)


packages/@aws-cdk/cdk-explorer/lib/core/source-resolver.ts
```diff
@@ -140,21 +139,32 @@ export class SourceMapResolver {
...
+  if (TS_JS_EXTENSIONS.some((ext) => file.endsWith(ext))) return 'tsjs';
+  if (HOST_LANGUAGE_EXTENSIONS.some((ext) => file.endsWith(ext))) return 'host';
+  return undefined;
+}
 
-function isSupportedSourceFile(file: string): boolean {
-  return SUPPORTED_SOURCE_EXTENSIONS.some((ext) => file.endsWith(ext));
+// Host frame columns are 0-indexed (0 = unavailable); SourceLocation is 1-based.
+function normalizeHostFrame(loc: SourceLocation): SourceLocation {
+  return { file: loc.file, line: loc.line, column: loc.column + 1 };
```
otaviomacedo: Shouldn't it be:

```suggestion
  return { file: loc.file, line: loc.line, column: Math.max(1, loc.column) };
```

If the column is greater than zero, we want to preserve it.

megha-narayanan: As of right now at least, jsii's Python runtime always hardcodes the column to 0 as an "unavailable" sentinel and the kernel omits column 0 from the frame string, so every real Python/Java frame arrives with no column. So this is all code to support in the future if columns are sent in the frames. My logic here was to treat host columns as 0-based (which is Python's col_offset convention), so + 1 converts a 0-based column for the SourceLocation which is 1-based (at least the way I have done it now). Let me know if you still think this needs a change.

otaviomacedo (approved):

packages/@aws-cdk/cdk-explorer/test/core/source-resolver.test.ts
ShadowCat567, line 1: Should we have at least 1 test for Java as well?

PR #1677, opened 2026-06-25
feat(toolkit-lib): register message codes for dependency-stack selection lines

sai-ray:
This PR adds message codes to two stack-selection lines:

- `CDK_TOOLKIT_I1002` — "Including dependency stacks" (upstream)
- `CDK_TOOLKIT_I1002` — "Including depending stacks" (downstream)

They were emitted without a code, so integrators couldn't target them. Text is unchanged.

#1663  will use these codes to suppress the lines for `cdk list`.

### Checklist

---


mrgrain (changes requested):
This is great! Minor note: the `CDK_ASSEMBLY_` messages are reserved for messaged originating from the user's CDK app. This message is from the toolkit, despite being from a class called `StackAssembly` 🙈

sai-ray:
> This is great! Minor note: the `CDK_ASSEMBLY_` messages are reserved for messaged originating from the user's CDK app. This message is from the toolkit, despite being from a class called `StackAssembly` 🙈

Wasn't totally sure where these belonged, so I followed `CDK_TOOLKIT_I1001` and put them in the synth band next to it since it's also a stack-selection message that fires from lots of commands and not just synth.

mrgrain (approved):

PR #1617, opened 2026-06-11
feat: explorer codelens nav + template resource locations

megha-narayanan:
Builds on #1592, which surfaced display-only CodeLens entries for the CFN
resources each construct produces. This PR makes them **clickable**. Selecting a
lens jumps to the resource's definition in the synthesized CloudFormation template.

Features:
- **Clickable navigation** — each lens carries an `openResource` command that opens
  the resource's template file at its logical-ID line.
- **Multi-resource picker** — constructs producing several resources show a QuickPick;
  single-resource constructs open directly.
- **Positional `templateFile` resolution (cloud-assembly-api)** — `buildConstructTree`
  threads the owning template through the tree, switching at NestedStack boundaries.
- **Clearer titles** — `Creates AWS::S3::Bucket`, or
  `Creates 3 resources: AWS::S3::Bucket, AWS::S3::BucketPolicy, AWS::KMS::Key`.

### Checklist
- [ ] This change contains a major version upgrade for a dependency and I confirm all breaking changes are addressed
  - Release notes for the new version:

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license

PR #1593, opened 2026-06-03
chore: lsp branch refactor

megha-narayanan:
Refactor: put LSP skeleton in a unique branch from web explorer

- [ ] This change contains a major version upgrade for a dependency and I confirm all breaking changes are addressed
  - Release notes for the new version:

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license

---------

Fixes #

### Checklist
- [ ] This change contains a major version upgrade for a dependency and I confirm all breaking changes are addressed
  - Release notes for the new version:

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license

PR #1681, opened 2026-06-29
feat: integrate lsp feature branch

megha-narayanan:
Merges the`feat/cdk-lsp' branch into `main`. The change is additive and introduces no behavior change to existing CLI commands.

- New `@aws-cdk/cdk-explorer` package containing the Language Server under `lib/lsp` (server, diagnostics, CodeLens, template locator, position mapping).
- Extends `@aws-cdk/cloud-assembly-api` with two parsing modules consumed by the server: `construct-tree.ts` (builds the construct tree from a cloud assembly) and `template-ranges.ts` (resolves a logical ID or property to its byte range in the template).

Capabilities (folds in #1559, #1593, #1592, #1617, #1624, #1630, #1631, #1662, #1634, #1674):

- Diagnostics: surfaces synth errors and policy-validation violations in the editor, mapped back to the source.
- Surfaces CFN resources and adds CodeLens navigation from a construct to its template resource.
- Navigation between construct source and the synthesized template in both directions.
- Live refresh: diagnostics and CodeLens update when `cdk.out` changes.
- Reads are constrained to the project directory, and template reads run off the LSP event loop.

This PR is the server and parsing foundation. It does not add a shipped CLI command or the web explorer.

### Checklist

---


packages/@aws-cdk/toolkit-lib/lib/api/source-tracing/index.ts
```diff
@@ -1 +1,2 @@
 export * from './types';
+export { findCreationStackTrace, findMutationStackTraces } from './private/stack-source-tracing';
```
ShadowCat567: I am a little bit concerned about this. Are we exporting any new API surface area that we shouldn't be?

megha-narayanan: Going to switch this to a import directly from the private surface and remove from API. Opening a PR for this now

ShadowCat567 (approved):

PR #1689, opened 2026-06-30
feat(cli): Integrate with express mode using `--express` flag for deploy, destroy, and bootstrap

ShadowCat567:
Adds `--express` flag to `cdk bootstrap`, `cdk destroy`, and `cdk deploy`. This flag enables stacks to be bootstrapped, deployed, and destroyed using Cloudformation express mode. 
Hotswap fallback deployments are also now completed with express mode (affects `cdk deploy --hotswap-fallback`).

I have also made some updates to the console output when deploying with express mode. When resources have been created or updated their `CREATE_COMPLETE`/`UPDATE_COMPLETE` message will linger on the screen for a few seconds with the status reason: 
```
Resource operation completed using Express Mode. It may continue becoming available in the background.
```
`DELETE_COMPLETE` messages will not get the same treatment because this same message appears as the status reason for `DELETE_COMPLETE` and it is confusing since resources are being torn down not becoming available. 

`cdk deploy`, `cdk bootstrap`, and `cdk destroy` also show a message at the end of the deployment/teardown that resources are still stabilizing.

deploy:
image

if there are more than 5 resources affected by the express mode deployment, they get truncated like this:
```
⚠️  Stack deployed using Express Mode. Resources still stabilizing: SomeResourceSDL00192, ApiF70053CD, ApiNoneDSB4E6495F, ApiSchema510EECD7, ApiDefaultApiKeyF991C37B, ...and 1 more...
```

bootstrap (gets the same disappearing messages as deploy does): 
image

destroy:
image

#### Rollback
By default express mode deployments do not use automatic rollback. If you want to enable automatic rollback you have to use:
```
$ cdk deploy --express --rollback
```
When performing an express mode where rollback is required because of resource replacement, we take the same strategy as `cdk deploy --no-rollback` where we ask the customer if they want to perform a deployment without rollback. This was previously called a 'regular deployment' but I find that terminology is confusing when it is applied to express mode (does regular deployment mean deployment without express mode?), so I have changed the messaging to say we will be 'Performing a deployment with rollback enabled'.

#### ✅ intentation
The intent that appears before ✅ has been removed so everything in the output is aligned on the same line

---


packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/template-body-parameter.ts
```diff
@@ -145,7 +145,7 @@ export async function restUrlFromManifest(url: string, environment: Environment)
     // SDK v3 no longer allows for getting endpoints from only region.
     // A command and client config must now be provided.
     const s3 = new S3Client({ region });
-    const endpoint = await getEndpointFromInstructions({}, HeadObjectCommand, {
+    const endpoint = await getEndpointFromInstructions({}, HeadObjectCommand as any, {
```
ShadowCat567, line 148: This and the other `as any` cast are a result of changes in the SDK - related to this: https://github.com/aws/aws-sdk-js-v3/issues/8138

rix0rrr (approved):

Temikus:
FYI, this is causing hot-swap stacks to get wedged via this death loop:

RollbackStack → "not supported for stacks that were last updated using EXPRESS deployment mode and are in a failed state". DisableRollback: false → "non-terminal state". Mode: STANDARD → "must use EXPRESS until *_COMPLETE".

mrgrain:
> FYI, this is causing hot-swap stacks to get wedged via this death loop:
> 
> RollbackStack → "not supported for stacks that were last updated using EXPRESS deployment mode and are in a failed state". DisableRollback: false → "non-terminal state". Mode: STANDARD → "must use EXPRESS until *_COMPLETE".

@Temikus please create an issue for this so we can triage and review. Thanks!

Temikus:
@mrgrain case 178591708800284

Or do you mean a github issue? Happy to do that as well.

mrgrain:
> @mrgrain case 178591708800284
> 
> Or do you mean a github issue? Happy to do that as well.

Thanks! Yes, a GitHub issue is what I meant. Support case is always good as well though!

PR #1693, opened 2026-07-01
chore: dedupe packages and upgrade sdk version

ShadowCat567:
Upgrade `@aws-sdk` packages to consume a bug fix. Update code that was using a workaround that is no longer needed with most recent sdk version. 
Dedupe `@smithy` and `@aws-sdk` package versions to improve binary size.

---


projenrc/cdk-cli-integ-tests.ts
```diff
@@ -247,9 +247,6 @@ export class CdkCliIntegTestsWorkflow extends Component {
       storage: './storage',
       auth: { htpasswd: { file: './htpasswd' } },
       uplinks: { npmjs: { url: 'https://registry.npmjs.org/' } },
-      // The bundled `aws-cdk` tarball is larger than Verdaccio's default 10mb
-      // body limit
-      max_body_size: '100mb',
```
iliapolo, line 252: How close are we though? we should have at least a 10% buffer I feel or just leave it as is if we aren't really tracking tarball size anyway.

rix0rrr: Yeah, please just leave this in. We are quite close, 100s of kBs close to it.

rix0rrr:
```suggestion
      max_body_size: '100mb',
```

ShadowCat567: yep, we don't seem to track tarball size, revived `max_body_size` in the verdaccio config

rix0rrr (approved):

PR #1698, opened 2026-07-01
feat: explorer web live refresh

megha-narayanan:
Adds live refresh to the cdk explore web UI. When the cloud assembly on disk changes, the server pushes an event and the SPA re-fetches, so the construct tree and violations stay current without a manual reload.

- SSE stream (GET /api/events): a long-lived server-sent events channel. SseBroadcaster tracks connected clients and pushes to all of them, evicting any that disconnect.
- cdk.out watcher: same watcher as LSP
- SPA: subscribes on load and re-fetches the tree and violations on each event.
- The event carries no payload. The server holds no assembly state, so the client just re-reads /api/tree and /api/policy-validation on receipt. Also narrows the web routes' injectable readAssembly seam to Promise<AssemblyReadResult> to match the real reader.

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/web/events.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/web/events.ts
@@ -1,14 +1,6 @@
 import type { Request, Response } from 'express';
 
 /**
- * SSE event name signalling that the cloud assembly was rewritten. The event
- * carries no payload: the web server holds no assembly state, so a client
- * re-fetches the construct tree and violations through the regular GET
- * endpoints whenever it sees this.
- */
-export const ASSEMBLY_CHANGED = 'assembly-changed';
-
-/**
  * Tracks connected Server-Sent Events clients and pushes events to all of them.
  * One instance lives per web server. `close()` ends every open stream on
  * shutdown so the HTTP server can stop cleanly.
@@ -25,7 +17,7 @@ export class SseBroadcaster {
     res.set({
       'Content-Type': 'text/event-stream',
       'Cache-Control': 'no-store',
-      Connection: 'keep-alive',
+      'Connection': 'keep-alive',
     });
     res.flushHeaders();
     this.clients.add(res);
--- a/packages/@aws-cdk/cdk-explorer/lib/web/events.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/web/events.ts
@@ -1,0 +1,58 @@
+import type { Request, Response } from 'express';
+
+/**
+ * SSE event name signalling that the cloud assembly was rewritten. The event
+ * carries no payload: the web server holds no assembly state, so a client
+ * re-fetches the construct tree and violations through the regular GET
+ * endpoints whenever it sees this.
+ */
+export const ASSEMBLY_CHANGED = 'assembly-changed';
+
+/**
+ * Tracks connected Server-Sent Events clients and pushes events to all of them.
+ * One instance lives per web server. `close()` ends every open stream on
+ * shutdown so the HTTP server can stop cleanly.
+ */
+export class SseBroadcaster {
+  private readonly clients = new Set<Response>();
+
+  /**
+   * Express handler for `GET /api/events`. Opens a long-lived SSE stream and
+   * registers the client, removing it when the request closes or the socket
+   * errors so a vanished client is never written to.
+   */
+  public readonly handle = (req: Request, res: Response): void => {
+    res.set({
+      'Content-Type': 'text/event-stream',
+      'Cache-Control': 'no-store',
+      Connection: 'keep-alive',
+    });
+    res.flushHeaders();
+    this.clients.add(res);
+
+    const remove = (): void => {
+      this.clients.delete(res);
+    };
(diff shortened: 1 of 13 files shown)


packages/@aws-cdk/cdk-explorer/lib/web/server.ts
```diff
@@ -36,12 +53,19 @@ export interface WebServer {
...
 
   const app = express();
 
-  registerApi(app, { appDir, assemblyDir: options.assemblyDir });
+  registerApi(app, { appDir, assemblyDir });
+
+  // Live-refresh stream: browsers subscribe here and re-fetch when the assembly
+  // changes. Registered before the /api catch-all so it is not treated as unknown.
+  const events = new SseBroadcaster();
+  app.get('/api/events', events.handle);
```
rix0rrr: This only works if `events.handle` doesn't use any `this` references.

It's a bit hard to explain and different from all other programming languages, but `this` is dynamically bound in JavaScript, and it only gets a value at the exact moment when you *call* a function.

If you "take" a function from an object and pass it around, `this` is unbound and when you call a function without an object it gets bound to the global object.

Now, it might be that that's not a problem because `events.handle` is a closure that never uses `this`, but I can't tell that from this call site, so this looks supicious. As a rule:

**I would like to be able to reason about things locally.** If they look locally correct, they should be globally correct. This might still be globally correct, but it looks locally incorrect, so breaks my heuristics. 

Use the following instead:

```ts
// This bakes the value of "this" into the function
app.get('/api/events', events.handle.bind(events));
```

I thought we had a linter rule for this... 🤔

packages/@aws-cdk/cdk-explorer/lib/web/server.ts
```diff
@@ -65,16 +89,29 @@ export async function startWebServer(options: WebServerOptions = {}): Promise<We
...
+    onError: options.onWatcherError ?? ((err) =>
+      process.stderr.write(`assembly watcher error: ${err instanceof Error ? err.message : String(err)}\n`)),
+  });
 
   let stopped = false;
   return {
-    url: `http://${host}:${port}`,
-    stop: () => {
-      if (stopped) return Promise.resolve();
+    url: `http://127.0.0.1:${port}`,
```
rix0rrr: Use `localhost` instead. On some operating systems it uses a faster network path (that skips a few layers), and it can't be wrong even if IPv4 is turned off.

rix0rrr: Also -- why hardcode this in a lot of places? Why not pass our preferred value in the `host` as an argument, and keep it deduplicated and easy to change?

packages/@aws-cdk/cdk-explorer/lib/web/events.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export class SseBroadcaster {
+  private readonly clients = new Set<Response>();
+
+  /**
+   * Express handler for `GET /api/events`. Opens a long-lived SSE stream and
+   * registers the client, removing it when the request closes or the socket
+   * errors so a vanished client is never written to.
+   */
+  public readonly handle = (req: Request, res: Response): void => {
```
rix0rrr: Why not just function syntax?

```ts
public handle(req: Request, res: Response): void {
  // ...
}
```

I think I have this comment in a lot of places, where you use function expression syntax instead of function declaration syntax. I personally find function declaration syntax a lot easier to grok, and I'd prefer we use that as much as possible.

-----

To be frank I'm not even sure whether this binds the `this` that we think it does. I guess it has to because otherwise TypeScript would complain?

Hmm, thinking about how this must work, I think I see why `this` gets bound correctly and you also don't need the `.bind()` if you write it like this. 

Still, when I look at the usage site I don't know any of that, so it still looks sus.

rix0rrr (commented):
> The event carries no payload. The server holds no assembly state, so the client just re-reads /api/tree and /api/policy-validation on receipt. 

Nice! Exactly what I would have done

packages/@aws-cdk/cdk-explorer/lib/web/server.ts
megha-narayanan: will fix to be localhost instead. I was trying to clean up options that we didn't need/ never used, but I can absolutely make this an argument as localhost to make it easy to change.

packages/@aws-cdk/cdk-explorer/lib/web/server.ts
megha-narayanan: fixed, handle is now a normal method and I bind it at the call site: app.get('/api/events', events.handle.bind(events)). I didn't see a linter rule but I do agree anyway.

packages/@aws-cdk/cdk-explorer/lib/web/events.ts
megha-narayanan: Switched to function-declaration syntax. Now that it's a real method I bind it explicitly at the call site (see the /api/events thread).

rix0rrr (approved):

PR #1704, opened 2026-07-04
feat(toolkit-lib): Use COLUMNS env var for tables

faridnsh:
When stdout is not a TTY, process.stdout.columns is undefined and diff tables render at their unbounded natural width. CI log viewers such as CodeBuild then wrap the output and make security diffs difficult to read.

This falls back to a positive integer from the COLUMNS environment variable, allowing the user to set it, while preserving process.stdout.columns precedence. Invalid or unset values retain the existing unbounded behavior.

### Checklist

---


mrgrain (approved):

iankhou:
Thank you for your contribution!

PR #1708, opened 2026-07-07
feat(toolkit-lib): `withListeners` API for IoHost

sai-ray:
The CLI's `CliIoHost` has a listener mechanism that's private and welded to that one class. Programmatic toolkit-lib users have no way to observe or reshape what flows through their IoHost short of writing a whole custom host.

This PR extracts that engine into a private `ListenerRegistry` and adds a public `withListeners(host)` wrapper, so listeners can be attached to any `IIoHost`:

```ts
import { Toolkit, NonInteractiveIoHost, withListeners, byCode, IoRequest, ConfirmationRequest, StackDetailsPayload } from '@aws-cdk/toolkit-lib';

const host = withListeners(new NonInteractiveIoHost());

let count = 0;
const warnings: string[] = [];

// Observe. `byCode<T>` types `msg.data`, and registering returns a disposer.
const dispose = host.on(byCode<StackDetailsPayload>('CDK_TOOLKIT_I2901'), (msg) => { count += msg.data.stacks.length; });
dispose();

// Any predicate is a matcher, so you can match a whole level.
host.on((msg) => msg.level === 'warn', (msg) => { warnings.push(msg.message); });

// Restyle a message, or drop it, without the host knowing.
using _fmt = host.rewrite(byCode<StackDetailsPayload>('CDK_TOOLKIT_I2901'), (msg) => `${msg.data.stacks.length} stacks`, { level: 'debug' });
using _quiet = host.on(byCode('CDK_TOOLKIT_I5031'), () => ({ preventDefault: true }));

// Answer a request so the host is never asked to prompt.
using _yes = host.respond(byCode<IoRequest<ConfirmationRequest, boolean>>('CDK_TOOLKIT_I7010'), true);

const toolkit = new Toolkit({ ioHost: host });
```

A listener can observe a message (`on`/`once`), rewrite its text and level (`rewrite`/`rewriteOnce`), suppress it, or answer a request (`respond`/`respondOnce`). Each registration returns a disposer that works both as a call and under `using`. Documented in the toolkit-lib README under `#### Attaching listeners to an IoHost`.

Nine additions users interact with:

- `withListeners(host)`, which returns the host you passed in, typed `T & IoEmitter`. `IoEmitter` is the six registration methods it adds.
- `MessageMatcher`, the one way to select messages, and `byCode<T>(...codes)` to build one from one or more message codes.
- `MessageListenerResult` and `MessageListenerResultOrPromise` for what a listener may return (`message`, `level`, `action`, `preventDefault`, `respond`), and `DisposeListener` for what registering gives back.
- `RespondOptions` and `RewriteOptions`, the options bags for `respond`/`respondOnce` and `rewrite`/`rewriteOnce`, replacing positional booleans and levels.

Listeners are keyed on a `MessageMatcher`, which is any `(msg: IoMessage<unknown>) => boolean`. A matcher that carries a payload type is just a type guard, so `byCode<StackDetailsPayload>('CDK_TOOLKIT_I2901')` types `msg.data` with nothing named in between. This PR also makes the message makers callable, so `IO.CDK_TOOLKIT_I2901` is itself a matcher, and because a maker narrows all the way to `IoRequest<T, U>`, `respond`'s value is checked against the request's response type. That is why `IMessageMatcher`, `IRequestMatcher`, `MessagePredicate` and `IoHostWithListeners` are deleted rather than published, and why `ListenerRegistry` takes exactly one signature per method, with the typed signatures declared once in `IoEmitter`.

`withListeners` is a Proxy, which is what keeps the inner host's own methods, getters, setters and full type intact, and wrapping is idempotent so there is never a second registry double-handling a message. Four of the proxy's rules are load-bearing rather than cosmetic, around binding `this` for hosts with `#private` fields, letting an own property shadow the additions, keeping the additions off `Object.keys`, and capturing the inner `notify` at wrap time so a later override cannot recurse. Each has a test in `listeners.test.ts` naming the failure it prevents.

`CliIoHost` no longer owns a registry, it is wrapped like anything else. `CliIoHost.instance()` hands out `attachListeners(new CliIoHost(props))`, and `export interface CliIoHost extends IoEmitter {}` merges the added methods into the class type, so every call site in `cdk-toolkit.ts` is textually unchanged while `cli-io-host.ts` loses 505 lines and gains 107. Three things fall out of that re-layering. The `corkReplaying` flag is gone, because the cork buffer now sits below the listener layer instead of replaying back through `notify`, which also fixes a latent double-count where telemetry saw a corked message twice. Listeners lose their privileged tier, so `removeAllListeners`, `addInternal` and `removeUserListeners` are deleted and stack-activity routing registers with plain `on`. And telemetry ordering is guaranteed rather than incidental, since it registers first and a later `preventDefault` can no longer stop it counting a dropped message.

One existing seam needed plumbing rather than porting. `CliIoHost` already had `observeMessages`, which reports the emitted message, its effective form, and whether a listener dropped it, all three for one message, and the snapshot recorder in `test/_helpers/io-recorder.ts` is built on it. It could compute that itself while the listeners ran inline. Now that they run a layer down, it cannot, so the private `attachListeners` takes an optional `ListenerVerdictHook` and the CLI forwards it into the same fan-out. `ObservableIoHost`, `observeMessages` and `IoMessageObservation` all survive, the last as an alias for `ListenerVerdict`, and `io-recorder.ts` is untouched. The hook is private because it only makes sense for a host at the bottom of the stack that wants the whole stream, and it is fixed at wrap time rather than added later, which is why passing it to an already-wrapped host throws.

Two things change in observable behavior, both intended:

- `preventDefault` on a *request* with no answer now throws. A request's declared default is often approval, so resolving with it would approve on the user's behalf. Pair `preventDefault` with `respond`, or use `respond`, which sets both. No CLI path reaches this, since every `@suppressMessages` target is an `info` or `trace` notification.
- A `once` listener is claimed before it is awaited, so concurrent emissions (e.g. parallel stacks) can no longer double-fire it.

Flagging two changes. Message makers gained `.is` recently in #1679 and this PR drops it in favour of calling the maker, which is what makes the matcher interfaces unnecessary, at the cost of overloading `IO.X` as a factory via `.msg()` and a predicate when called. And `removeAllListeners` is deleted, which after #1887 had only test call sites.

### Checklist

---


packages/@aws-cdk/toolkit-lib/lib/api/io/private/message-maker.ts
mrgrain, line 1: Nope, these are private on purpose. Do not export them.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * Either a message/request *maker* — the listener fires for messages with that
+ * maker's `code` — or a custom *predicate* over the message. A maker's `.is`
+ * type guard (e.g. `IO.CDK_TOOLKIT_I7010.is`) is a convenient predicate, but any
+ * `(msg) => boolean` works (e.g. to match a family of codes, or on the message
+ * level). Use `matchAny` to combine several selectors.
+ */
+export type MessageSelector<T> =
+  | IoMessageMaker<T>
+  | IoRequestMaker<T, any>
```
mrgrain: These are not public types, nor should they be. We can accept `string` here (a single code) or a new `MessageMatcher` class (via `IMessageMatcher`). We can make the MessageMakers implement that interface so we can keep passing them in.

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/io/private/message-maker.ts
sai-ray: Done. MessageInfo, CodeInfo, ActionLessMessage, ActionLessRequest are all un-exported again, and IO and the maker types are off the public barrel too. Nothing maker internal is public now

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
sai-ray: Went with the string option. The public selector is now `IoMessageCode` (a single code) or a `(msg) => boolean` predicate, so it no longer references the maker types. I tried the `IMessageMatcher` route first but dropped it. To keep the typed maker path it wanted, IO would have to be public, and exporting IO pulls the maker internal types back onto the public surface through the same forgotten export cascade with api extractor. With IO private there's also no way for an external caller to obtain an `IMessageMatcher`. `CliIoHost` keeps passing makers internally through the private registry unchanged.

mrgrain: yes there is, the CLI is a privileged caller that can use private exports from toolkit-lib.

In either case I don't understand why it's not possible. Can you share the `IMessageMatcher` interface you attempted?

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
```diff
@@ -0,0 +1,243 @@
...
+   * Returns a function that removes the formatter again.
+   *
+   * @example
+   * ```ts
+   * const dispose = host.rewrite('CDK_TOOLKIT_I2901', (msg) =>
+   *   serializeStructure((msg.data as StackDetailsPayload).stacks, true));
+   * ```
+   */
+  rewrite(
+    code: IoMessageCode,
```
mrgrain: why no predicate?

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
```diff
@@ -0,0 +1,243 @@
...
+   * @param suppressQuestion - whether to also suppress surfacing the question
+   *   text. Defaults to `true` (answer silently). Pass `false` to still surface
+   *   the question while answering it.
+   *
+   * @example
+   * ```ts
+   * const dispose = host.respond('CDK_TOOLKIT_I7010', true);
+   * ```
+   */
+  respond(code: IoMessageCode, value: unknown, suppressQuestion?: boolean): () => void;
```
mrgrain: I know is copied from what we currently have, but for a public API we need to think more careful about the design. If we ever want to add an other option this gets messy. We can preemptively put `suppressQuestion` in a property bag (options).

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
```diff
@@ -0,0 +1,243 @@
...
+ * ```ts
+ * const base = new NonInteractiveIoHost();     // or your own host
+ * const host = withListeners(base);
+ * host.on('CDK_TOOLKIT_I2901', (m) => { count += (m.data as StackDetailsPayload).stacks.length; });
+ * const toolkit = new Toolkit({ ioHost: host });
+ * ```
+ */
+export function withListeners(host: IIoHost): IoHostWithListeners {
+  return new ListeningIoHost(host);
+}
```
mrgrain, line 207: Not sure I love this, but I guess it doesn't hurt either. 🤷🏻

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * The result of `withListeners`. It is still an `IIoHost`, so it drops straight
+ * into the `ioHost` slot of a new `Toolkit`, and it exposes methods to observe,
+ * reshape, or answer individual messages without subclassing a host.
+ *
+ * Listeners are keyed on a message `code` (e.g. `'CDK_TOOLKIT_I2901'`). The
+ * codes are listed in the message registry:
+ * https://docs.aws.amazon.com/cdk/api/toolkit-lib/message-registry/
+ */
+export interface IoHostWithListeners extends IIoHost {
```
mrgrain, line 46: do we need to extend the `IIoHost` interface? Or is `IoEmitter` a separate independent interface?

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * @example
+ * ```ts
+ * const base = new NonInteractiveIoHost();     // or your own host
+ * const host = withListeners(base);
+ * host.on('CDK_TOOLKIT_I2901', (m) => { count += (m.data as StackDetailsPayload).stacks.length; });
+ * const toolkit = new Toolkit({ ioHost: host });
+ * ```
+ */
+export function withListeners(host: IIoHost): IoHostWithListeners {
```
mrgrain, line 205: We have this, we can at least make this a generic type so that the inner host keeps its higher fidelity.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
```diff
@@ -0,0 +1,340 @@
...
+   */
+  public on<T>(
+    selector: IoMessageMaker<T> | IoRequestMaker<T, any> | ((msg: IoMessage<any>) => msg is IoMessage<T>),
+    listener: (msg: IoMessage<T>) => MessageListenerResultOrPromise,
+  ): () => void;
+  public on(
+    predicate: (msg: IoMessage<any>) => boolean,
+    listener: (msg: IoMessage<unknown>) => MessageListenerResultOrPromise,
+  ): () => void;
+  public on(selector: MessageSelector<any>, listener: MessageListenerFn): () => void {
```
mrgrain, line 150: since this is a private class, it feels strange to have so many different signatures. But I guess thats because we couldn't make `IIoMessageMatcher` yet.

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
sai-ray: added. missed it earlier.

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
sai-ray: I held off `IoEmitter` because I wasn't sure of the benefit over the wrapper for our cases. The one place it would help is reusing one emitter across many hosts. If that is a expected case, then this would be small addition.

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
sai-ray: Updated. It's a `RespondOptions` object now, so we can add more options later without changing the signature.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
sai-ray: yes, and now the selector went from two maker arms plus a predicate down to a matcher or a predicate.

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
sai-ray: I dug into this and I'm not sure I found the right answer, so I'd value your read.

The generic form that keeps the host's full type is `withListeners<T>(host: T): T & IoHostWithListeners`. My worry is that the wrapper only implements notify, requestResponse, and the listener methods, so I think that type would promise the host's other members while they'd actually be undefined at runtime.

I tried two ways to make it honest and neither quite felt right:

- A Proxy forwarding unknown members to the inner host. Reads seem fine, and a set trap would probably cover writes, but when I wrapped a host that already has a registry (like `CliIoHost`) I ended up with two registries both firing on notify, and I couldn't find a trap that resolved it cleanly.
- Adding the listener methods straight onto the host and handing it back. That one is genuinely the host, but it changes the object that was passed in, and since the CLI host is a shared singleton I think that would affect everyone.

So for now I've gone back to returning `IoHostWithListeners`. It still extends `IIoHost`, which is all the toolkit consumes, and the caller still holds their own typed reference to the host they passed in, so it seems like they don't really lose it. If there's an approach you had in mind that avoids these I'd really appreciate the guidance. I couldn't reason one that holds up.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
sai-ray: Sorry, my earlier statement was off. I got there by trying to give public users a typed `msg.data` without a cast, and I talked myself into thinking that needed IO to be public. It doesn't. A public matcher interface is fine on its own, and only exporting IO itself would leak the maker types. Here's the interface, it's on the branch:

```ts
export interface IMessageMatcher<T> {
  is(msg: IoMessage<unknown>): msg is IoMessage<T>;
}
export interface IRequestMatcher<T, U> extends IMessageMatcher<T> {
  is(msg: IoMessage<unknown>): msg is IoRequest<T, U>;
}
```

The makers implement it so internally the CLI keeps passing them and gets typed data and the maker types stay private.

But the typing goal I was chasing doesn't actually resolve. Internally the CLI can use the matcher because it has the makers. A public user has no matcher to pass, since IO and the makers are private. To get one they'd either need us to export IO, or hand-write the matcher:

```ts
host.on({ is: (m): m is IoMessage<StackDetailsPayload> => m.code === 'CDK_TOOLKIT_I2901' }, (m) => m.data.stacks);
```

But that `m is IoMessage<StackDetailsPayload>` doesn't give much over just casting on the code path:

```ts
host.on('CDK_TOOLKIT_I2901', (m) => (m.data as StackDetailsPayload).stacks);
```

So the typed matcher only seems to pay off for the CLI. For public users it's a similar cast. Given that, does it earn a place on the public surface, or should the matcher be internal and leave the public API as `IoMessageCode | MessagePredicate`?

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
mrgrain: Exactly, I think a proxy is the right solution here!

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
```diff
@@ -0,0 +1,317 @@
...
+ *
+ * The result of `withListeners`. It is still an `IIoHost`, so it drops straight
+ * into the `ioHost` slot of a new `Toolkit`, and it exposes methods to observe,
+ * reshape, or answer individual messages without subclassing a host.
+ *
+ * Listeners are keyed on a message `code` (e.g. `'CDK_TOOLKIT_I2901'`). The
+ * codes are listed in the message registry:
+ * https://docs.aws.amazon.com/cdk/api/toolkit-lib/message-registry/
+ */
+export interface IoHostWithListeners extends IIoHost {
```
mrgrain: Technically this needs to be `IIoHostWithListeners` but like I've said, I don't see why we need a combined interface. `IoEmitter` might still be the better name even if we extend.

packages/@aws-cdk/toolkit-lib/lib/api/io/io-message.ts
```diff
@@ -82,3 +82,21 @@ export interface IoRequest<T, U> extends IoMessage<T> {
...
+}
+
+/**
+ * A request matcher.
+ *
+ * Carries the response type `U` so an answer can be typed.
+ */
+export interface IRequestMatcher<T, U> extends IMessageMatcher<T> {
+  is(msg: IoMessage<unknown>): msg is IoRequest<T, U>;
+}
```
mrgrain: Do these Matchers need to be generic? What functionality are we gaining?

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
```diff
@@ -0,0 +1,317 @@
...
+   * });
+   * ```
+   */
+  on<T>(
+    matcher: IMessageMatcher<T>,
+    listener: (msg: IoMessage<T>) => MessageListenerResultOrPromise,
+  ): () => void;
+  on(
+    selector: IoMessageCode | MessagePredicate,
+    listener: (msg: IoMessage<unknown>) => MessageListenerResultOrPromise,
```
mrgrain: So we allow 3 different types here. Can we limit this? `MessagePredicate` and `IMessageMatcher` seem to be the same thing. Even `IoMessageCode` we can get rid of if we provide a code matcher.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -331,9 +234,10 @@ export class CliIoHost implements IIoHost, ObservableIoHost {
   private corkedCounter = 0;
   private readonly corkedLoggingBuffer: IoMessage<unknown>[] = [];
 
-  // Message listeners in registration order. Each carries a matcher (by code,
-  // or a custom predicate). See `on`/`once`/`rewrite`/`respond`.
-  private readonly messageListeners: MessageListener[] = [];
+  // The shared listener engine. Registration and message transformation live
+  // here; this host does its own I/O (writing, prompting, telemetry, observers)
+  // around `registry.apply`. See `on`/`once`/`rewrite`/`respond`.
+  private readonly registry = new ListenerRegistry();
```
mrgrain, line 241: What is this registry? If we need it, this tells you that our wrapping pattern doesn't work as designed.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
mrgrain, line 1: Please check which types need exporting from here. Also maybe some are duplicated with public types, not sure.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
mrgrain: But we now have `IIoMessageMatcher!`

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
```diff
@@ -0,0 +1,342 @@
...
+   */
+  public addInternal(matches: (msg: IoMessage<unknown>) => boolean, fn: MessageListenerFn): () => void {
+    return this.add({ once: false, internal: true, fn, matches });
+  }
+
+  /**
+   * Remove every listener registered via `on`/`once`/`rewrite`/`respond`,
+   * keeping the host's internal listeners so the host keeps working afterwards.
+   */
+  public removeUserListeners(): void {
```
mrgrain: To be fair, I'd rather get rid of the internal concept completely...

mrgrain (changes requested):
Also some changes have been made on main that solve dispose much nicer. We should include them here.

packages/@aws-cdk/toolkit-lib/lib/api/io/io-message.ts
sai-ray: The generic only existed to carry the payload type and the makers do that themselves nowthat they are callable so deleted both. `io-message.ts` has no changes in this PR any more.

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
sai-ray: Down to one.

    export type MessageMatcher = (msg: IoMessage<unknown>) => boolean;

`byCode` replaced the `IoMessageCode` arm as per suggestion. Payload typing survives without a second arm because a narrowing matcher is just a type guard, and a type guard is assignable to `MessageMatcher`.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
sai-ray: this comment reframed the whole PR for me. `CliIoHost` is wrapped like any other host now and a declaration merge keeps every existing call site compiling.

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
sai-ray: Dropped the combined interface. `IoEmitter` holds the six registration methods and does not extend `IIoHost`,

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
sai-ray: It went the other way, the interface is deleted. Makers are callable now, so a maker is the type guard directly and there is nothing left for the interface to do. One signature per method on the registry.

packages/@aws-cdk/toolkit-lib/lib/api/io/private/listener-registry.ts
sai-ray: Went through them, and it is down to two exports, matchAny and ListenerRegistry.

packages/@aws-cdk/toolkit-lib/lib/api/io/listeners.ts
sai-ray: Update, it exists now and it is separate. If it extended `IIoHost` the intersection would restate what the host already provides and flatten the caller's own type.

PR #1714, opened 2026-07-08
fix(integ-runner): existing snapshot dir interferes with real deployment

rix0rrr:
In a recent change to integ-runner, we started reusing the synth output directory upon reading so we don't have to re-synth for every single file we need to read from the synth directory.

However, we accidentally left the directory too long, where it started interfering with a synth that actually need to be deployed.

Now clean up that directory in between checking the snapshot and doing the actual deployable synth.

In order to fix this, this had to turn into a giant refactoring of the `integ-runner`. Big changes:

- Turn off the update workflow by default.
- 

---


sai-ray (approved):

rix0rrr:
Nevermind, this doesn't fix it :(

PR #1715, opened 2026-07-08
chore: extract assembly read-lock factory into core

megha-narayanan:
Matching LSP side of #1706. Moves the assembly read-lock acquirer out of
lib/lsp into lib/core/assembly-lock.ts so the LSP and web server build the
read lock from one shared factory. Pure refactor, LSP behavior is unchanged,
and server.ts re-exports AssemblyLock so existing importers keep resolving it.

Fixes #

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
@@ -35,6 +35,7 @@ import { synthFailureDiagnostics } from './synth-diagnostics';
 import { sourceTargetAtTemplateOffset } from './template-locator';
 import { WATCH_EXCLUDE_DEFAULTS } from '../../../toolkit-lib/lib/actions/watch/private/helpers';
 import { createIgnoreMatcher } from '../../../toolkit-lib/lib/util/glob-matcher';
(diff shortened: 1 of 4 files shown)


packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -57,13 +58,8 @@ import type { SynthRunResult } from '../core/synth-runner';
...
 
-/**
- * A held read lock on the cloud assembly directory; `release()` unlocks it.
- * In production this wraps the Toolkit's `fromAssemblyDirectory().produce()` readable.
- */
-export interface AssemblyLock {
-  release(): Promise<void>;
-}
+// Re-exported from core so existing importers keep resolving `AssemblyLock` here.
+export type { AssemblyLock };
```
ShadowCat567: This looks a little weird to me, why can't we update the import in other places to point towards core?

megha-narayanan: yes oops this is just for the test, fixed

ShadowCat567 (approved):

PR #1716, opened 2026-07-08
fix: validate template paths and document the LSP server

megha-narayanan:
- onDefinition now bounds the client-supplied *.template.json path to the project's cdk.out (realpath on both sides) before reading, matching the file-read discipline used elsewhere in the package. Prevents reading arbitrary files via a crafted textDocument/definition request.
- Replace the placeholder README with real content: purpose, how to launch (cdk lsp / startLspServer), features, a security note that auto-synth and 'Synth now' run the cdk.json app command with your credentials on save (enable only on trusted projects), and editor-integration notes.
- Remove the unused express / @types/express dependencies; the web interface is not part of this branch.

Fixes #

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/README.md
+++ b/packages/@aws-cdk/cdk-explorer/README.md
@@ -1,10 +1,11 @@
 # @aws-cdk/cdk-explorer
 
 A Language Server (LSP) for AWS CDK apps. It runs your CDK app, reads the
-synthesized cloud assembly, and surfaces that back in your editor: code lenses
-on the constructs you author, hover details from the generated CloudFormation,
-go-to-definition from a template to the construct that created it, and
-diagnostics from policy validation.
+synthesized cloud assembly from your project's `cdk.out` directory, and
+surfaces that back in your editor: code lenses on the constructs you author,
+hover details from the generated CloudFormation, go-to-definition from a
+template to the construct that created it, and diagnostics from policy
+validation.
 
 Used by the `cdk lsp` command in the AWS CDK CLI. It is for CDK developers who
 want in-editor feedback without leaving their source files.
@@ -30,7 +31,12 @@ stdin/stdout. You can also start it programmatically:
   generated template.
 - Go to definition: from a position in a synthesized `*.template.json` back to the
   construct source that produced it.
-- Diagnostics: policy validation report findings appear as squiggles in source.
+- Diagnostics: findings from the policy validation report appear as squiggles in
+  your source. The rule description and any suggested fix ride in the diagnostic
+  message, so your editor shows them when you hover the squiggle and in its
+  problems list. The report identifies a violating construct rather than a source
+  line, so a squiggle sits on the line that creates the construct, not on the
+  specific property at fault.
 
 Source-linked features (code lenses, hover, go to definition) currently work for
 TypeScript and Python.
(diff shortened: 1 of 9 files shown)


packages/@aws-cdk/cdk-explorer/README.md
```diff
@@ -1 +1,90 @@
...
+
+    cdk lsp
+
+Your editor's LSP client launches that command and talks to it over
+stdin/stdout. You can also start it programmatically:
+
+    import { startLspServer } from '@aws-cdk/cdk-explorer';
+    startLspServer();
+
+## Features
```
ShadowCat567, line 25: This section seems a bit weird to me, why are we advertising codelens as the primary feature here, can you maybe reorganize this so we have a list of the features? - mostly so we are not projecting one feature as being more important than the others

packages/@aws-cdk/cdk-explorer/README.md
```diff
@@ -1 +1,90 @@
...
+on the constructs you author, hover details from the generated CloudFormation,
+go-to-definition from a template to the construct that created it, and
+diagnostics from policy validation.
+
+Used by the `cdk lsp` command in the AWS CDK CLI. It is for CDK developers who
+want in-editor feedback without leaving their source files.
+
+## Status
+
+This release ships the LSP server only. The web interface is not part of this
```
ShadowCat567: We don't need to mention that the web interface is not here yet, we don't need to mention it in the README, once it arrives, we can add it

megha-narayanan: okay, will do. can just add that as a note in the ticket then, i know they asked about whether the web explorer was part of this yet

packages/@aws-cdk/cdk-explorer/README.md
megha-narayanan: sounds good, will just include the synth triggering and the resource codelenses both in a flat list.

packages/@aws-cdk/cdk-explorer/README.md
```diff
@@ -1 +1,76 @@
...
+## Features
+
+- Code lenses on source lines that create resources: `Creates <Type>` opens that
+  resource in the synthesized template (a picker when a line maps to several),
+  plus `Synth now` / `Enable auto-synth` / `Disable auto-synth` in the file header.
+- Hover: a construct's resolved CloudFormation properties and a link to the
+  generated template.
+- Go to definition: from a position in a synthesized `*.template.json` back to the
+  construct source that produced it.
+- Diagnostics: policy validation report findings appear as squiggles in source.
```
ShadowCat567: Can you clarify a bit more how this works, since we don't just show the squiggles, I think there is also a hover that explains how to fix it. Also don't the squiggles appear on the same line as the construct, not the actual source of the problem (since we don't have that info through the validator)

packages/@aws-cdk/cdk-explorer/README.md
ShadowCat567, line 1: Somewhere in here I think we want to mention that we pull this information from a user's `cdk.out` directory

packages/@aws-cdk/cdk-explorer/README.md
megha-narayanan: yep, I'll expand this. I'll make both explicit, lmk if the new wording looks good

packages/@aws-cdk/cdk-explorer/README.md
megha-narayanan: adding to the intro paragraph

ShadowCat567 (approved):

PR #1719, opened 2026-07-08
feat(cli): typescript init templates use tsc + tsx and swc/jest, on typescript 7

iankhou:
## Summary

TypeScript **7.0.2** was published on 2026-07-08. `ts-node@10` reads the compiler's `ts.sys` API, which TypeScript 7 no longer exposes, and ts-node has no compatible release. `ts-jest@29` likewise caps its peer dependency at `typescript <7`. Two consequences:

1. Any generated CDK project whose owner manually upgrades `typescript` past `~5.9` crashes on `cdk synth`:
   ```
   TypeError: Cannot read properties of undefined (reading 'fileExists')
       at readConfig (.../ts-node/dist/configuration.js:91:33)
   ```
2. The `init-typescript-app` integ test enumerates TypeScript versions live from npm, so `7.0` entered the matrix immediately and the `integ_init-templates` jobs fail on every PR.

This PR moves the TypeScript init templates to TypeScript 7 and replaces the compiler-API-dependent tooling:

- **App execution**: `npx tsc && npx tsx bin/<name>.ts` — `tsc` type-checks (`noEmit`), then [tsx](https://tsx.is/) (esbuild-based) runs the app. Follows @mrgrain's review suggestion, with plain `npx` chaining instead of `npx -c` for portability across npm versions and Windows.
- **Tests**: jest with [`@swc/jest`](https://swc.rs/docs/usage/jest) (transpile-only, no type checking). Faster than `ts-jest`, which we also can't use with TypeScript 7 due to `ts-jest` depending on `ts-node`. Type errors in `test/` are still caught by `npm run build`, since the tsconfig includes the test directory.

Neither tsx nor swc drives the TypeScript compiler API, so the templates no longer break when the compiler's internals change.

## Template changes

Made changes to typescript templates as follows:

For `app` + `sample-app` typescript variants
- Upgrade typescript to v7
- `cdk.template.json`: use app command `npx tsc && npx tsx bin/<name>.ts`
- `package.json`: drop `ts-node` and `ts-jest`; add `tsx`, `@swc/core`, `@swc/jest`
- `tsconfig.json`: add `noEmit`, `isolatedModules` (esbuild transpiles file-by-file), and an explicit `types: ["jest", "node"]` list — TS 6+ no longer auto-includes all of `node_modules/@types`
- `jest.config.js`: transform `@swc/jest`
- `README`: `build`/`watch` descriptions now say "type-check" since `tsc` no longer emits

For `lib` typescript variant

- Similar to above changes, but omit `noEmit`, since there is no `app` command.

## Test changes

- The cli-integ version-matrix test installs `tsx` alongside the matrix's `typescript` version

## Verified locally with a built CLI

- Rendered all three templates into scratch projects with `typescript@7.0.2`: `npm install` resolves, `npm run build` passes under the Go-based compiler, `npm test` passes through `@swc/jest`, and the app command runs end to end
- A deliberate type error fails `npx tsc` with exit 1, blocking synth
- The template tsconfigs type-check under TS 5.5, 5.9, 6.0, and 7.0 (the integ test matrix spans versions younger than 2 years)

## User-facing notes

- Existing projects are not affected — this only changes what `cdk init` generates for new projects.
- New projects start on TypeScript 7. `tsc` still type-checks every synth, but tsx (esbuild) executes the app, so synth and watch iterations are faster than ts-node's compiler-API path.
- Jest no longer type-checks test files at test-time (swc is transpile-only); `npm run build` covers that.

---

### DIFF
--- a/packages/aws-cdk/lib/init-templates/app/typescript/cdk.template.json
+++ b/packages/aws-cdk/lib/init-templates/app/typescript/cdk.template.json
@@ -1,5 +1,5 @@
 {
-  "app": "npx tsc && node bin/%name%.js",
+  "app": "npx tsc && npx tsx bin/%name%.ts",
   "watch": {
     "include": ["**"],
     "exclude": [
--- a/packages/aws-cdk/lib/init-templates/app/typescript/cdk.template.json
+++ b/packages/aws-cdk/lib/init-templates/app/typescript/cdk.template.json
@@ -1,5 +1,5 @@
 {
-  "app": "npx ts-node --prefer-ts-exts bin/%name%.ts",
+  "app": "npx tsc && node bin/%name%.js",
   "watch": {
     "include": ["**"],
     "exclude": [
--- a/packages/@aws-cdk-testing/cli-integ/tests/init-typescript-app/init-typescript-app.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/init-typescript-app/init-typescript-app.integtest.ts
@@ -44,10 +44,9 @@ TYPESCRIPT_VERSIONS.forEach(tsVersion => {
     // still want to test with older versions as well.
     await removeDevDependencies(context);
 
+    // The generated app compiles and runs through `tsc && node` — the TypeScript
+    // compiler itself is the only toolchain needed for `cdk synth`.
     await shell.shell(['npm', 'install', '--save-dev', `typescript@${tsVersion}`]);
-
-    // After we've removed devDependencies we need to re-install ts-node because it's necessary for `cdk synth`
-    await shell.shell(['npm', 'install', '--save-dev', 'ts-node@^10']);
 
     await shell.shell(['npm', 'install']); // Older versions of npm require this to be a separate step from the one above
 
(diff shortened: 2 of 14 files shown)


rix0rrr:
Wait, isn’t the problem that ts-node doesn’t support TS7?

Then why are we/they pulling it in?

packages/aws-cdk/lib/init-templates/app/typescript/cdk.template.json
```diff
@@ -1,5 +1,5 @@
 {
-  "app": "npx ts-node --prefer-ts-exts bin/%name%.ts",
+  "app": "npx tsc && node bin/%name%.js",
```
mrgrain: I was thinking this: 
```suggestion
  "app": "npx -c \"tsc && tsx bin/%name%.js\"",
```

mrgrain (changes requested):

iankhou:
> Wait, isn’t the problem that ts-node doesn’t support TS7?
> 
> Then why are we/they pulling it in?

It's automatically pulled in from NPM when we run our cli-integ test matrix, per the following logic:

https://github.com/aws/aws-cdk-cli/blob/a7a40e09097b7910ce090e9c4daf67bd41644ca8/packages/%40aws-cdk-testing/cli-integ/lib/npm.ts#L52

Anything newer than 2 years old is pulled in.

packages/aws-cdk/lib/init-templates/app/typescript/cdk.template.json
iankhou: Switched to tsc && tsx. Per our conversation, tsx is faster than ts-node anyway, and tsc is faster than it used to be, and now feasible here.

iankhou:
Integ tests are passing on all tested version of typescript (released within 2 years).

rix0rrr (approved):

mrgrain (approved):

Issue #1709, opened 2026-07-07
toolkit-lib: make stack event polling interval configurable


svozza:
### Describe the feature

`StackActivityMonitor` already accepts `pollingInterval` (default `2_000` ms), but none of the three construction sites (deploy and destroy monitors in `lib/api/deployments/deploy-stack.js`, rollback monitor in `lib/api/deployments/deployments.js`) thread it through, and no public options interface exposes it. We'd like a supported way for programmatic consumers to set it.

### Use Case

We're the Powertools for AWS Lambda (TypeScript) team. Our e2e fleet drives `toolkit-lib` programmatically: a GitHub Actions matrix of 36 jobs (9 packages × 2 Node versions × 2 architectures) starting within seconds of each other, each deploying several stacks concurrently. At 2 s polling per active stack operation, ~50–100 concurrent operations put 25–50 `DescribeStackEvents` requests/sec on the account-level CloudFormation read API. CloudTrail for a representative run window shows a sustained ~8–10 calls/sec, all from the toolkit, with `ThrottlingException`s interleaved every 1–2 s at peak. The result is jobs failing randomly with:

```
CDK_TOOLKIT_E5500 'Error occurred while monitoring stack: Throttling: Rate exceeded'
CDK_TOOLKIT_E7900 '<stack>: destroy failed Throttling: Rate exceeded'
```

The `ConfiguredRetryStrategy(7, ...)` in the toolkit's SDK config can't reliably outlast persistent saturation, so the destroy path exhausts retries and throws. For CI, 2 s event granularity buys nothing — we'd happily poll every 10–15 s, cutting our call volume ~5×.

We're unblocked today via a monkeypatch on `StackActivityMonitor.prototype.start` that rewrites `pollingInterval`: https://github.com/aws-powertools/powertools-lambda-typescript/blob/fix/e2e-cfn-throttling-polling-patch/packages/testing/src/patchCdkStackEventPolling.ts. It works, but it depends on toolkit-lib internals and will break silently on a refactor — we'd rather delete it in favor of a supported option.

### Proposed Solution

Thread `pollingInterval` from a public option through `Deployments` into the three `new StackActivityMonitor({...})` call sites. Candidates, in our order of preference:

1. `ToolkitOptions` (e.g. `stackMonitoring: { pollingInterval }`) — set once, applies consistently to deploy, destroy, and rollback.
2. Per-action on `DeployOptions`/`DestroyOptions`.
[...]

PR #1724, opened 2026-07-10
feat(toolkit-lib): configurable stack stabilization polling interval on deploy, destroy and prepareStack

### LINKED ISSUE #1723: toolkit-lib: make stack stabilization (DescribeStacks) polling interval configurable
### Describe the feature

Follow-on to #1709/#1710, which made the `StackActivityMonitor` event-polling interval (`DescribeStackEvents`) configurable. There's a second hardcoded polling loop those didn't touch: the stabilization waiter. `stabilizeStack()` in `lib/api/deployments/cfn-api.ts` — via the exported `waitForStackDeploy()`/`waitForStackDelete()` — calls the module-private `waitFor(valueProvider, timeout = 5000)`, polling `DescribeStacks` every **5 s per in-flight stack** to detect a terminal state. The interval is hardcoded with no option threading to it (as of 1.33.0: `waitFor` at `cfn-api.js:100`, `stabilizeStack` at `cfn-api.js:440`).

We'd like the same treatment as #1710: a supported way to raise this interval.

### Use Case

Same programmatic e2e fleet as #1709. After #1710 shipped and we slowed the event monitor to 10 s, CI kept losing unrelated matrix jobs to `Throttling: Rate exceeded` on deploys — CloudTrail shows `DescribeStacks` from this stabilization loop became the dominant CFN read-API caller once the event loop was slowed. For CI, 5 s completion granularity buys nothing; 10–15 s is fine.

### Proposed Solution

Add an optional interval param to `stabilizeStack` (default `5000`, so it's backward compatible), forward it from `waitForStackDeploy`/`waitForStackDelete`, and pass it at the call sites. The plumbing #1710 added is already in scope there: `stackEventPollingInterval` sits right beside the `waitForStackDeploy`/`waitForStackDelete` calls in `deploy-stack.ts` (deploy, destroy, and the delete-before-recreate path). Reusing that one option so it governs both the event and stabilization loops is our preference. If you'd rather keep the two loops independently tunable, a sibling `stackStabilizationPollingInterval` option works too — happy to implement whichever you prefer.

Two call sites to confirm during implementation: `waitForStackDelete` is also invoked from `Deployments` (changeset cleanup), and the rollback path calls `stabilizeStack` directly — the latter is the same rollback monitor #1710 explicitly deferred, so we're happy to either include it here or leave it consistent with that deferral.

### Acknowledgements

- [x] I may be able to implement this feature request

### CDK version used

`@aws-cdk/toolkit-lib` 1.33.0

GitHub Actions ubuntu-latest runners, Node.js 22/24

svozza:
Fixes #1723

### Motivation

Follow-on to #1709/#1710, which made the `StackActivityMonitor` event-polling interval (`DescribeStackEvents`) configurable via `stackEventPollingInterval`. There's a second hardcoded polling loop that wasn't touched: the stabilization waiter. `stabilizeStack()` in `lib/api/deployments/cfn-api.ts` — via the exported `waitForStackDeploy()`/`waitForStackDelete()` — calls the module-private `waitFor(valueProvider, timeout = 5000)`, polling `DescribeStacks` every 5s per in-flight stack to detect a terminal state, with no option threading to it.

Programmatic consumers running large CI matrices who already slowed the event monitor via `stackEventPollingInterval` saw `DescribeStacks` from this stabilization loop become the dominant CloudFormation read-API caller instead, still triggering throttling.

### Design decision: one parameter, not two

`stackEventPollingInterval` now governs both the event-polling and stabilization-polling intervals, rather than adding a new sibling parameter (e.g. `stackStabilizationPollingInterval`). Going with reuse over a second parameter to avoid growing the public options surface on `DeployOptions`/`DestroyOptions` for a distinction most consumers won't need to control separately — both loops exist to reduce `DescribeStack*` call volume against the same CloudFormation rate limit. Happy to split it into a separate parameter in a follow-up if maintainers would rather keep the two tunable independently.

### Changes

- Added an optional `stabilizationPollingInterval`/`pollingInterval` parameter to `waitForStackDelete()`, `waitForStackDeploy()`, and `stabilizeStack()` in `cfn-api.ts`. Default unchanged (5000ms).
- Threaded the existing `stackEventPollingInterval` option through to that new parameter — it already sits right beside the `waitForStackDeploy`/`waitForStackDelete` calls in `deploy-stack.ts` (deploy, destroy, and the delete-before-recreate path).
- Threaded through:
  - `deployStack()`/`destroyStack()` in `deploy-stack.ts` (all three call sites: delete-before-recreate, post-deploy stabilization, post-destroy stabilization)
  - `Deployments.cleanupChangeSet()` (the change-set cleanup path, called from `prepareStack()` and from `Toolkit.deploy()` when a user declines the deploy confirmation prompt)
- Updated doc comments on `stackEventPollingInterval` at each layer (`DeployOptions`, `DestroyOptions`, `DeployStackOptions` in both `deploy-stack.ts` and `deployments.ts`) to note it also governs stabilization polling.
- No public API surface changes — `stackEventPollingInterval` already existed; only its scope of effect changed, and doc comments were updated to reflect that.

### Not included

Per the issue, two call sites needed confirming during implementation:
- `waitForStackDelete` from `Deployments.cleanupChangeSet` — **included** (see above).
- The rollback path (`Deployments.rollbackStack`), which calls `stabilizeStack` directly — **deferred**. This is the same rollback monitor that #1710 explicitly deferred for event polling, so leaving it out here keeps this PR consistent with that scope. Happy to add `stackEventPollingInterval` support to `RollbackStackOptions` in a follow-up if maintainers want it included.

### Testing

- New `cfn-api-stabilization-polling-interval.test.ts`: directly exercises `stabilizeStack`, `waitForStackDeploy`, and `waitForStackDelete` with fake timers, asserting `DescribeStacks` polling cadence at both the 5s default and a custom interval.
- Extended `deploy-stack-polling-interval.test.ts`: asserts `deployStack()`/`destroyStack()` forward `stackEventPollingInterval` to `waitForStackDeploy`/`waitForStackDelete` as the stabilization interval, alongside the existing `StackActivityMonitor` assertions.
- Extended `cloudformation-deployments.test.ts`: asserts `prepareStack()` forwards `stackEventPollingInterval` through `cleanupChangeSet()` to `waitForStackDelete`.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/actions/deploy/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/actions/deploy/index.ts
@@ -212,10 +212,7 @@ export interface BaseDeployOptions {
   readonly express?: boolean;
 
   /**
-   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation.
-   *
-   * Also governs the interval used to poll CloudFormation for the stack to stabilize (reach a terminal
-   * state) after the deployment has been issued.
+   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring stack operations and waiting for stack stabilization.
    *
    * Increase this value to reduce the number of `DescribeStackEvents`/`DescribeStacks` calls,
    * e.g. when many concurrent stack operations are hitting CloudFormation API rate limits.
--- a/packages/@aws-cdk/toolkit-lib/lib/actions/deploy/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/actions/deploy/index.ts
@@ -212,9 +212,12 @@ export interface BaseDeployOptions {
   readonly express?: boolean;
 
   /**
-   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation
+   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation.
    *
-   * Increase this value to reduce the number of `DescribeStackEvents` calls,
+   * Also governs the interval used to poll CloudFormation for the stack to stabilize (reach a terminal
+   * state) after the deployment has been issued.
+   *
+   * Increase this value to reduce the number of `DescribeStackEvents`/`DescribeStacks` calls,
    * e.g. when many concurrent stack operations are hitting CloudFormation API rate limits.
    *
    * @default 2000
--- a/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
@@ -28,9 +28,12 @@ export interface DestroyOptions {
   readonly express?: boolean;
 
   /**
-   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation
+   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation.
    *
-   * Increase this value to reduce the number of `DescribeStackEvents` calls,
+   * Also governs the interval used to poll CloudFormation for the stack to stabilize (reach a terminal
+   * state) after the destroy has been issued.
+   *
+   * Increase this value to reduce the number of `DescribeStackEvents`/`DescribeStacks` calls,
    * e.g. when many concurrent stack operations are hitting CloudFormation API rate limits.
    *
    * @default 2000
--- a/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/actions/destroy/index.ts
@@ -28,10 +28,7 @@ export interface DestroyOptions {
   readonly express?: boolean;
 
   /**
-   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation.
-   *
-   * Also governs the interval used to poll CloudFormation for the stack to stabilize (reach a terminal
-   * state) after the destroy has been issued.
+   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring stack operations and waiting for stack stabilization.
    *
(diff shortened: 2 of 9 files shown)


packages/@aws-cdk/toolkit-lib/lib/actions/deploy/index.ts
```diff
@@ -212,9 +212,12 @@ export interface BaseDeployOptions {
   readonly express?: boolean;
 
   /**
-   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation
+   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring a stack operation.
    *
-   * Increase this value to reduce the number of `DescribeStackEvents` calls,
+   * Also governs the interval used to poll CloudFormation for the stack to stabilize (reach a terminal
+   * state) after the deployment has been issued.
```
mrgrain: Let's use this everywhere. The "Also..." is a typical AI addition that irrelevantly keeps some sort of history. We define this option controls all polling. That's enough to describe it.

```sug gestion
   * Time in milliseconds to wait between polling CloudFormation for stack events while monitoring stack operations and waiting for stack stabilization.
```

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/actions/deploy/index.ts
svozza: Applied everywhere — folded into a single sentence, no more "Also governs..." addendum:

`Time in milliseconds to wait between polling CloudFormation for stack events while monitoring stack operations and waiting for stack stabilization.`

Fixed in 58394ae5.

svozza: Fixed!

mrgrain (approved):

PR #1729, opened 2026-07-13
chore: prevent future SDK duplication

rix0rrr:
We were hit by pretty heavy aws-sdk-js-v3 package duplication (we had 5 copies of some of the core packages in there at some point).

Add a script that ensures that our build fails if we have more than one copy of every package.

---


projenrc/check-sdk-duplication.task.ts
```diff
@@ -0,0 +1,122 @@
...
+  process.stderr.write(`\n[sdk-duplication-check] ❌ ${violations.length} package(s) exceed the maximum of ${MAX_ALLOWED_DUPLICATES} resolved versions:\n\n`);
+
+  for (const { packageName, versions } of violations) {
+    process.stderr.write(`  ${packageName}: ${versions.length} versions\n`);
+    for (const v of versions) {
+      process.stderr.write(`    - ${v}\n`);
+    }
+  }
+
+  console.error('To fix: run \'yarn dedupe --pattern "@aws-sdk/*" --pattern "@smithy/*"\' and verify the result.\n');
```
ShadowCat567: According to the `yarn dedupe` docs, we don't need the `--pattern` flag to dedupe sets of versions: https://yarnpkg.com/cli/dedupe
```
yarn dedupe @aws-sdk/* @smithy/*
```

ShadowCat567 (approved):

PR #1730, opened 2026-07-14
feat: synth freshness

megha-narayanan:
Batches LSP auto-synth-on-save into a single trailing re-synth, like `cdk watch`.
A synth requested while one is running sets a queued flag instead of being
dropped; the running synth drains it by running exactly one more synth once it
finishes. Overlapping saves collapse into that one follow-up, so once saves stop
the assembly always catches up to the last save.

One design decision worth thinking about: 
No retry cap on external lock conflicts. The latch guarantees the LSP
  never contends with its own synth, so a `cdk.out` lock conflict is always an
  external writer (a terminal `cdk synth`/`cdk watch`). We retry indefinitely
  with backoff and full jitter rather than drop the synth. Tradeoff: if an
  external writer never releases, auto-synth stays blocked for the session. This
  matches `cdk watch`.

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
@@ -59,6 +59,23 @@ import type { SynthRunResult } from '../core/synth-runner';
 const REFRESH_LOCK_RETRIES = 10;
 const REFRESH_LOCK_RETRY_MS = 50;
 
+/**
+ * Synth concurrency latch, mirroring `cdk watch` (see toolkit.ts). 'open' = idle,
+ * 'synthing' = a synth is running, 'queued' = a save arrived mid-synth and a
+ * follow-up is owed. No 'pre-ready' (unlike watch there is no ready gate; the LSP
+ * synths on demand).
+ */
+type SynthLatch = 'open' | 'synthing' | 'queued';
+
+/**
+ * When an external writer (a terminal `cdk synth`/`cdk watch`) holds the cdk.out
+ * write lock, the synth retries with exponential backoff + full jitter rather
+ * than giving up, so the LSP always eventually synths the latest save once the
+ * writer releases. Delay is capped; attempts are not.
+ */
+const SYNTH_LOCK_RETRY_BASE_MS = 100;
+const SYNTH_LOCK_RETRY_MAX_MS = 2_000;
+
 export interface LspHandlerOptions {
   /** Override readAssembly for tests. Defaults to reading <applicationDir>/cdk.out. */
   readonly readAssembly?: (assemblyDir: string) => Promise<AssemblyReadResult>;
@@ -177,7 +194,10 @@ export function createLspHandlers(options: LspHandlerOptions): LspHandlers {
 
   let applicationDir: string | undefined;
   let shutdownRequested = false;
-  let synthInFlight = false;
+  // Concurrency latch, mirroring `cdk watch` (toolkit.ts): a synth requested
+  // while one runs is batched into a single follow-up, so synths run one at a
+  // time and the assembly catches up to the latest save.
+  let synthLatch: SynthLatch = 'open';
   // URIs (source files, or cdk.json) currently showing synth-failure diagnostics.
   let synthFailureUris = new Set<string>();
   let autoSynthEnabled = false; // off by default; user enables via the CodeLens toggle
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -276,20 +296,47 @@ export function createLspHandlers(options: LspHandlerOptions): LspHandlers {
...
-  // the same LSP instance: a second call while the first is running is dropped
-  // (not queued) and returns lock-conflict immediately. The Toolkit's RWLock
-  // would reject it anyway, but this short-circuits before any setup work.
+  // One synth, retrying the write lock with exponential backoff + full jitter.
+  // The latch guarantees we never contend with our own synth, so a lock-conflict
+  // here is an external writer (a terminal `cdk synth`/`cdk watch`) holding
+  // cdk.out; retry until it releases rather than dropping the synth.
+  async function runSynthWithRetry(): Promise<SynthRunResult> {
+    let delayMs = SYNTH_LOCK_RETRY_BASE_MS;
+    for (;;) {
```
ShadowCat567: should this be a while loop instead of an infinite for loop?

rix0rrr: I was going to say it would be a `while (true)`, but upon further reflection any old deadline is probably better than no deadline at all.

How about:

```ts
const deadline = Date.now() + 3_600_000; // 1 hour
while (Date.now() < deadline) {
  // ...
}
```

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -276,20 +296,47 @@ export function createLspHandlers(options: LspHandlerOptions): LspHandlers {
...
+      const result = await (synthRunner ? synthRunner(currentProjectDir()) : Promise.resolve({ status: 'error', message: 'No synth runner configured' } as const));
+      if (result.status !== 'lock-conflict') {
+        publishSynthDiagnostics(result);
+        return result;
+      }
+      await new Promise<void>((resolve) => setTimeout(resolve, Math.random() * delayMs));
```
rix0rrr: Super nit: I usually prefer sticking the `setTimeout` bit into an `async function sleep(ms: number)` helper, so it reads super obvious at the call site.

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
```diff
@@ -276,20 +296,47 @@ export function createLspHandlers(options: LspHandlerOptions): LspHandlers {
...
     try {
-      const result = await (synthRunner ? synthRunner(currentProjectDir()) : Promise.resolve({ status: 'error', message: 'No synth runner configured' } as const));
-      publishSynthDiagnostics(result);
+      // Drain the queued slot with a loop (not recursion) so the stack stays
+      // flat. There is no await between reading the latch as not-'queued' and the
+      // finally resetting it to 'open', so under the single-threaded event loop
+      // any request that set 'queued' did so during an await and is observed
+      // here -- no wakeup is lost.
+      let result = await runSynthWithRetry();
+      while ((synthLatch as SynthLatch) === 'queued') {
```
rix0rrr, line 340: Por qué the cast?

rix0rrr (approved):
Provisionally approved!

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
megha-narayanan: Yep, will do!

packages/@aws-cdk/cdk-explorer/lib/lsp/server.ts
megha-narayanan: typescript yells at me because it doesn't realize that runSynthWithRetry can change synthLatch. I get "This comparison appears to be unintentional because the types '"synthing"' and '"queued"' have no overlap." otherwise

ShadowCat567: I think you can set `synthLatch` to `SynthLatch.synthing` (similar for the other situations) to get it to recognize it is a `SynthLatch` object, since right how `synthLatch` is of type `string` which is why Typescript is not happy

ShadowCat567 (approved):

PR #1731, opened 2026-07-15
fix: keep source-tracing helpers out of toolkit-lib public API

megha-narayanan:
Follow-up to review feedback on #1681. `findCreationStackTrace` / `findMutationStackTraces` were re-exported from toolkit-lib's public `api/source-tracing` index, adding new public API surface.

Instead, cdk-explorer imports `findCreationStackTrace` directly from toolkit-lib's private module via a relative path, the same mechanism already used in `lsp/server.ts`. Keeps the helpers out of toolkit-lib's public API.

Verification: cdk-explorer + toolkit-lib compile clean, cdk-explorer core jest suites pass.

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
@@ -2,9 +2,8 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { buildConstructTreeAsync, CloudAssembly, MANIFEST_FILE, type ConstructTreeNode } from '@aws-cdk/cloud-assembly-api';
 import { VALIDATION_REPORT_FILE, type PolicyValidationReportJson } from '@aws-cdk/cloud-assembly-schema';
-/* eslint-disable import/no-relative-packages */
+import { findCreationStackTrace } from '../api-private';
 import { SourceMapResolver, isWithinRoot, type SourceLocation } from './source-resolver';
-import { findCreationStackTrace } from '../../../toolkit-lib/lib/api/source-tracing/private/stack-source-tracing';
 
 /**
  * A construct from the cloud assembly, decorated with the user source location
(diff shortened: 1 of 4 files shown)


ShadowCat567 (approved):

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
```diff
@@ -2,8 +2,9 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { buildConstructTreeAsync, CloudAssembly, MANIFEST_FILE, type ConstructTreeNode } from '@aws-cdk/cloud-assembly-api';
 import { VALIDATION_REPORT_FILE, type PolicyValidationReportJson } from '@aws-cdk/cloud-assembly-schema';
-import { findCreationStackTrace } from '@aws-cdk/toolkit-lib';
+/* eslint-disable import/no-relative-packages */
```
mrgrain: Please let's not disable the rule here. There should be an existing single file in CLI that lists all private imports from toolkit-lib.

https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/api/index.ts
https://github.com/aws/aws-cdk-cli/blob/main/packages/aws-cdk/lib/api-private.ts(<-- this one probably)

mrgrain (changes requested):

packages/@aws-cdk/cdk-explorer/lib/core/assembly-reader.ts
megha-narayanan: Got it, made a file like this for the explorer package.

megha-narayanan: To clarify, the reason I can't just use the file you mentioned directly is because it isn't re-exported from aws-cdk's index, so importing it cross-package would itself be a relative-package import, triggering the exact rule. Anyway, dependency direction forbids that.

rix0rrr (approved):

PR #1732, opened 2026-07-16
feat: explorer UI improvements

megha-narayanan:
UI improvements to the CDK Web Explorer's violations panel, construct tree, file picker, and template pane diagnostics.

**Violations Panel**
- Human-readable violation titles: Violation headers show the description (e.g. "Require point-in-time recovery for a DynamoDB table") as the primary title instead of the raw rule name, with the rule name shown right-aligned in monospace as a secondary identifier
- Search bar: Added a search input (top-right in header) that filters violations across all text fields (description, rule name, construct path, suggested fix)
- Resource filter on double-click: Double-clicking a construct in the tree or a squiggle in the source pane filters the violations panel to show only violations for that resource (and its children)
- Dismissable filter pill: When filtered, the header shows "Violations for [ResourceName ×]" with a clear button pill/chip
- Right-aligned rule names: Monospace rule name pushed to far-right of each violation row (matching construct tree type alignment) 
- Two-row header layout: Severity + description on row 1, count/source subtitle on row 2
- Tighter expanded content: Reduced gap between violation header and expanded body, indented expanded content under the header
- Clickable construct paths: Occurrence paths styled as blue underlined links with single-click navigation (was double-click)

Screenshot 2026-07-16 at 12 23 46 PM

**Template Pane**
- Squiggle diagnostics: Resources with violations now show wavy underline squiggles in the template pane, which skip leading whitespace: Underlines start at the first non-space character
- YAML/JSON format toggle moved to header: Toggle is now in the upper-right of the template pane header (always visible regardless of scroll), only appears when a template is loaded. On switch, takes you to the same PLACE in the other format, not the same line number 

Screenshot 2026-07-16 at 1 01 35 PM

**Construct Tree**
- Removed opacity on inherited colors: Tree labels with inherited severity colors now display at full opacity
- Default cursor on non-clickable rows: Non-expandable rows show default cursor instead of pointer
Screenshot 2026-07-16 at 12 24 38 PM

**File Picker**
- Folder icon button: Replaced the grey "Open" button with a folder icon
- Dropdown overlay: File picker overlays the code pane instead of replacing it (code stays mounted underneath)
- No animation replay: Opening/closing the file picker no longer re-triggers the navigation highlight animation
- Click-outside to close: Clicking anywhere outside the picker dismisses it

Screenshot 2026-07-16 at 1 03 54 PM

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/frontend/App.tsx
+++ b/packages/@aws-cdk/cdk-explorer/frontend/App.tsx
@@ -6,7 +6,7 @@ import SpaceBetween from '@cloudscape-design/components/space-between';
 import Spinner from '@cloudscape-design/components/spinner';
 import * as React from 'react';
 import { buildSourceAnchorIndex, findConstructAtLine } from '../lib/web/source-nav';
-import { api, type DirEntry, type TemplateResponse, type TreeResponse, type ViolationsResponse } from './api';
+import { api, type TemplateResponse, type TreeResponse, type ViolationsResponse } from './api';
 import { CodeViewer, type Diagnostic } from './components/CodeViewer';
 import { ConstructTree } from './components/ConstructTree';
 import type { Language } from './syntax';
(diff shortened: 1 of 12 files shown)


rix0rrr (approved):
Good stuff! 

I'll ship it, but I don't think we need a full file picker with folder browsing and everything. From the stack traces in the diagnostics we know exactly which files we might be interested in showing. So why not just allow choosing between those files?

megha-narayanan:
> Good stuff!
> 
> I'll ship it, but I don't think we need a full file picker with folder browsing and everything. From the stack traces in the diagnostics we know exactly which files we might be interested in showing. So why not just allow choosing between those files?

Implemented that, here's what those look like now: 
Screenshot 2026-07-16 at 3 43 36 PM
Screenshot 2026-07-16 at 3 45 55 PM

PR #1746, opened 2026-07-21
feat(cdk-explorer): add cdk/getConstructTree LSP route

megha-narayanan:
Exposes the source-decorated construct tree over the language server so a client can overlay source links onto its own tree without re-implementing resolution. The server already builds this tree for hover, definition, and CodeLens; this adds a `cdk/getConstructTree` request that serializes it.

Each entry is flattened and keyed by construct path (as in tree.json), carrying the resolved source location, the template file, and the character offset of the resource's block in that template.

- No new resolution. The handler serializes the tree the server already caches (`cachedIndex`), so source locations and template files are exactly what hover and definition use. There is no second resolver to drift.
- Flat entries, not a nested tree. The consumer keeps its own tree and only needs a path to location lookup, so the response is one entry per construct keyed by path.
- templateOffset resolved per request. Each resource's block offset comes from cloud-assembly-api's template ranges, parsing each template at most once per request, so clients get a jump-to-resource position with no template-parsing dependency of their own.
- no-assembly before synth. Until the app is synthesized there is no tree to serve, so the request returns a no-assembly status rather than erroring.

Fixes #

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/cdk-explorer/lib/lsp/construct-tree-request.ts
+++ b/packages/@aws-cdk/cdk-explorer/lib/lsp/construct-tree-request.ts
@@ -1,0 +1,48 @@
+import { RequestType0 } from 'vscode-languageserver';
+import type { SourceLocation } from '../core/source-resolver';
+
+/**
+ * Custom LSP request that returns the source-resolved construct tree, flattened
+ * to one entry per construct keyed by its tree path.
+ *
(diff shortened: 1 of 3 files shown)


rix0rrr (approved):

ShadowCat567:
@mergify queue

PR #1747, opened 2026-07-22
fix: route metadata through toolkit-lib

mrgrain:
The `cdk metadata` command now delegates synthesis and single-stack selection to the existing toolkit-lib `Toolkit.list()` action instead of maintaining a separate legacy selection path. This keeps the CLI thin while preserving the command's YAML and `--json` output.

`CliIoHost` listeners can now override an effective message action, allowing the structured list result to remain tagged as `metadata`; matching and telemetry continue to use the originally emitted action. Focused IO snapshots protect both output formats and listener tests cover emitted-versus-effective action behavior.

Validation: `yarn compile`, `yarn eslint`, and focused Jest coverage for the metadata command, `CliIoHost`, and `IoHostRecorder` (90 tests).

### Checklist
- [ ] Integration tests added/updated (not required; no AWS interaction changed)
  - Release notes for the new version: N/A

---


packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -226,8 +226,33 @@ export class CdkToolkit {
...
+    try {
+      await this.toolkit.list(this.props.cloudExecutable, {
+        stacks: {
+          patterns: [stackName],
+          strategy: StackSelectionStrategy.PATTERN_MUST_MATCH_SINGLE,
+          expand: ExpandStackSelection.NONE,
+        },
+      });
+    } finally {
+      this.ioHost.removeAllListeners();
```
iankhou, line 254: This is for error-contingent cleanup of the listeners we create with `.once` right? Why not use the function that removes the listener for each one? Like:

```ts
let disposers = [
    this.ioHost.once(...),
    this.ioHost.once(...),
    this.ioHost.once(...)
];

...

try {
    ...
} finally {
    disposers.forEach(cleanupFunc => cleanupFunc());
}

```

iankhou (changes requested):

packages/aws-cdk/lib/cli/cdk-toolkit.ts
mrgrain: Could do, but `this.ioHost.removeAllListeners();` does the same and is already used elsewhere. I think in general we need a better (more generic) way of handling these. Can I do this as a followup?

iankhou: It does the same thing now, but would wipe other listeners too, if we introduce them in the future, and might result in an unexpected/difficult to debug situation. But yep, OK as a follow-up.

iankhou (approved):

packages/aws-cdk/lib/cli/cdk-toolkit.ts
mrgrain: Yeah, I see. But then the CLI only ever runs one command... I think about it.

PR #1748, opened 2026-07-22
feat: integrate cdk explorer branch

megha-narayanan:
Merges the feat/cdk-explorer branch into main. The change is additive and introduces no behavior change to existing CLI commands.

Builds on the language server merged in #1681 and adds the web explorer: a new cdk explore command that serves a local, read-only web UI over a synthesized cloud assembly.

- New `cdk explore` command in `packages/aws-cdk`. It starts a localhost web server over the cloud assembly in cdk.out, takes a single `--port` option (default 4200), and does not synthesize.
- New web layer in `@aws-cdk/cdk-explorer` (lib/web, lib/core, and the frontend SPA): an Express backend serving a small JSON API plus an embedded single-page app, and chokidar watchers for the assembly and the source tree.
- Uses `@aws-cdk/cloud-assembly-api` template-range resolution for the template viewer, and reads the synth-lock path exported from `@aws-cdk/toolkit-lib` to detect when source files are stale relative to the last synth.

Capabilities (folds in #1653, #1698, #1706, #1732, #1738):
- Construct tree annotated with per-node policy-validation severity, and inherited-severity coloring on ancestor nodes.
- Three-way navigation: double-click a construct to open its source line and its synthesized template resource, and jump from a template resource back to the source that created it.
- Source and template viewers with Prism syntax highlighting, a YAML/JSON toggle on the template, and wavy underlines on lines that carry violations.
- Violations panel sorted by severity, with text search and a construct-path filter.
- Live refresh over Server-Sent Events: the tree and violations update when cdk.out changes, including an external cdk synth or cdk watch, and a banner flags a source file that has been edited since the last synth.
- Security posture: the server binds localhost only. File and template reads are confined to the project directory, with path-escape and symlink attempts rejected, size-capped, and taken under the assembly read lock so a request never observes a mid-synth assembly. The explorer never runs the app or spawns a synth.
- The web explorer is a pure reader over cdk.out. Synth-on-save stays with the language server.

Screenshot 2026-07-22 at 10 46 34 AM

### Checklist

---


packages/@aws-cdk/cdk-explorer/tsconfig.frontend.json
mrgrain, line 1: Works better if this file lives in `frontend/tsconfig.json`

mrgrain (changes requested):
Any chance we can replace with a smaller, more modern alternative? I suspect we only need a small subset of the features express offers.

https://e18e.dev/docs/replacements/express.html#replacements-for-express

packages/@aws-cdk/cdk-explorer/lib/web/local-only.ts
```diff
@@ -0,0 +1,220 @@
...
+export const TOKEN_QUERY_PARAM = 'token';
+
+/**
+ * A fresh session token. 256 bits from the CSPRNG, base64url so it survives a URL
+ * and a `Set-Cookie` without escaping. Held in memory for the life of one
+ * `cdk explore`: nothing writes it to disk, so a bookmarked URL stops working once
+ * the server restarts — which is the point. A token that outlived the session
+ * would be a credential sitting in the user's home directory.
+ */
+export function newSessionToken(): string {
```
iankhou, line 116: Nice!

PR #1758, opened 2026-07-23
refactor: migrate all subprocess call sites to the shared subprocess module

iankhou:
## Summary

Follow-up to #1763, which adds the `@aws-cdk/private-tools/lib/subprocess` module.

This PR migrates every `child_process` call site in the shipped packages onto the shared module, so that no code path **both escapes and executes**:

| Call site | Before | After |
|---|---|---|
| `init/os.ts`, `init/init.ts` | `spawn(argv, {shell: true})` + broken hand-rolled escaping | `run()` — no shell, `inherit-stderr` keeps npm/pip progress bars |
| `telemetry/library-version.ts` | `exec()` string | `run()` argv |
| toolkit-lib `ec2-detection.ts` | `execSync` with `2>nul` | `runSync()`, stderr discarded via stdio |
| toolkit-lib `cloud-assembly/exec.ts` (synth `app`) | `spawn(str, {shell: true})` + split2 | `runUserCommandLine()`, line-buffered, `collect: false` |
| `commands/docs.ts` (`--browser`) | `exec()` | `runUserCommandLine()` |
| cdk-assets-lib `shell.ts` | own spawn + broken escaping copy | delegates to `run()`; spawn failures rethrow the OS error so docker's ENOENT → 'please install docker' guidance is preserved |
| toolkit-lib `environment.ts` | `quoteSpaces()` (space-only quoting) | allowlist-based `quoteShellPart()` |

The deleted `windowsEscape` helpers carried an escaping bug (`shellMeta.has(x)` instead of `has(c)`) that made Windows escaping a silent no-op — this PR removes the pattern rather than fixing it. `grep -rn runUserCommandLine` now returns exactly the two sinks where a shell is the documented contract (synth `app`, `--browser`) — the complete trust-model surface.

Drops `split2` from toolkit-lib (replaced by the module's line buffering).

Not in scope (follow-ups): `cdk-build-tools`/`integ-runner`/`cli-integ` migration, the eslint `no-restricted-imports` ban on `child_process`, and a Windows CI leg for the module.

### Checklist
- [x] Manually verified: init (git/npm/metachar paths), synth (shell features, exit codes, 200k-line streaming, UTF-8), docs --browser, cdk-assets shell paths (hostile args, docker ENOENT, EPIPE)

---


packages/@aws-cdk/toolkit-lib/lib/api/aws-auth/ec2-detection.ts
```diff
@@ -68,10 +68,12 @@ function detectEc2Linux(): boolean {
...
   // On Windows EC2 instances the board asset tag is an instance ID, readable
-  // from the registry without elevated privileges.
-  const tag = execSync(
-    'reg query "HKLM\\SYSTEM\\HardwareConfig\\Current" /v BaseBoardAssetTag 2>nul',
-    { encoding: 'utf-8', timeout: 500 },
+  // from the registry without elevated privileges. stderr is discarded (the
+  // key does not exist off EC2); a non-zero exit throws, which detectEc2()
+  // treats as "assume EC2".
+  const tag = runSync(
+    ['reg', 'query', 'HKLM\\SYSTEM\\HardwareConfig\\Current', '/v', 'BaseBoardAssetTag'],
```
iankhou, line 75: We don't include `2>nul` here as it's an instruction to the shell, not an argument. We preserve the behavior in https://github.com/aws/aws-cdk-cli/pull/1758/changes#diff-e2cd2b44a5e07b43d817ab35dc1c72ae4150b4b45fe640c75d5f1ef7f200c7f8R192, where the third 'ignore' in the array sends error output to the `null` device.

iankhou:
> Total lines changed 1086 is greater than 1000. Please consider breaking this PR down.

New change is smaller.

ShadowCat567:
> The deleted windowsEscape helpers carried an escaping bug (shellMeta.has(x) instead of has(c)) that made Windows escaping a silent no-op — this PR removes the pattern rather than fixing it.

Does this mean we still can't run tests on windows? Are customers affected by this at all or is this just testing infra?

packages/@aws-cdk/cdk-assets-lib/lib/private/shell.ts
```diff
@@ -1,79 +1,80 @@
...
+  readonly shellEventType: ShellEventType;
 }
 
-function handleShellOutput(
-  chunk: Buffer | string,
-  options: ShellOptions,
-  shellEventType: ShellEventType,
-): void {
+function handleShellOutput(props: HandleShellOutputProps): void {
+  const { chunk, options, shellEventType } = props;
```
ShadowCat567: if we're just going to expand the options back out again, what's the point of making `HandleShellOutputProps`?

packages/@aws-cdk/cdk-assets-lib/lib/private/shell.ts
```diff
@@ -104,65 +105,3 @@ class ProcessFailed extends Error {
...
-    return chars.some((c) => str.indexOf(c) !== -1);
-  };
-}
-
-/**
- * Escape a shell argument for POSIX shells
- *
- * Wrapping in single quotes and escaping single quotes inside will do it for us.
- */
-function posixEscape(x: string) {
```
ShadowCat567, line 146: You made note of removing the `windowsEscape`, is `posixEscape` no longer needed because of the refactor?

packages/@aws-cdk/cdk-assets-lib/lib/private/shell.ts
iankhou: Good point, will remove the property bag here

iankhou:
> > The deleted windowsEscape helpers carried an escaping bug (shellMeta.has(x) instead of has(c)) that made Windows escaping a silent no-op — this PR removes the pattern rather than fixing it.
> 
> Does this mean we still can't run tests on windows? Are customers affected by this at all or is this just testing infra?

It's not really related to being able to test on Windows. The original escaping bug allowed special characters like `,`, `|`, `^`, when it should not have, since it passed whole strings instead of validating individual characters. The approach that I used removes the entire helper, removing the bug as well.

We still don't have CI in Windows in this PR, but I can add it as a follow-up. I'm not certain why we don't already have it.

packages/@aws-cdk/cdk-assets-lib/lib/private/shell.ts
iankhou: It is also unneeded, since we use `renderForDisplay` in this change to subsume it. `renderForDisplay` uses similar logic, except it uses an allowlist for characters instead of a denylist (previous logic), which is safer.

Both `windowsEscape` and `posixEscape` are made obsolete from this PR, but `windowsEscape` was also buggy, which is why I mentioned it. The "logic" for both is still used in `renderForDisplay`.

iankhou:
> > The deleted windowsEscape helpers carried an escaping bug (shellMeta.has(x) instead of has(c)) that made Windows escaping a silent no-op — this PR removes the pattern rather than fixing it.
> 
> Does this mean we still can't run tests on windows? Are customers affected by this at all or is this just testing infra?

In terms of customer impact:
- We fix a bug in `cdk init` on Windows that would have misparsed input containing special characters like `|`, and `^`, which weren't in the deny list
- We put in quotes everything that doesn't fit the allowlist in [`DISPLAY_SAFE`](https://github.com/aws/aws-cdk-cli/blob/3eea30399e6385c32987fcfd54648db7fe0e1b8f/packages/%40aws-cdk/private-tools/lib/subprocess/index.ts#L309), so more stuff will be quoted, but it's just stuff displayed to the user

ShadowCat567 (approved):

PR #1761, opened 2026-07-24
feat(cli): add --watch option to cdk validate

iankhou:
https://github.com/aws/aws-cdk-cli/issues/1762

Continuously observe project files and re-run validation whenever changes are detected. Does not deploy.

Reuses settings from `cdk.json` that currently control behavior for `cdk watch`. It's possible that we may want to author a separate config for validate's watch and deploy's watch.

### Checklist

---


packages/aws-cdk/lib/cli/cdk-toolkit.ts
```diff
@@ -655,8 +682,54 @@ export class CdkToolkit {
...
+    });
+  }
+
+  /**
+   * Continuously observe project files, and invoke a command every time changes are detected.
+   *
+   * The files to observe are configured with the "watch" key of `cdk.json`.
+   * Used by both 'cdk deploy --watch' and 'cdk validate --watch'.
+   */
+  private async watchProjectFiles(props: {
```
iankhou, line 705: Share the logic from `cdk watch` a.k.a. `cdk deploy --watch` with `cdk validate --watch` using `watchProjectFiles`, which takes in a cdk command agnostically.

mrgrain (changes requested):
Hmm, so this will once again bring the CLI and Toolkit-lib further a part. I don't like this. E.g. this will make it now super complex for me to move over CLI `watch` to use toolkit-lib `watch`.

Can you think of a better way of handling this while also future proofing? As a general guidance, we want _all new features_ to go through `toolkit-lib`.

iankhou:
Closing for a different approach that uses `toolkit-lib` for file-watching. Reuse deploy-watch logic and refactor it out. https://github.com/aws/aws-cdk-cli/pull/1766.

PR #1769, opened 2026-07-28
fix(cli): version check trusts stale registry mirrors and recommends non-upgrades

mrgrain:
Fixes #

A user reported that the CLI version check recommended upgrading to 3.0.0, a version that was accidentally published and has since been deprecated, while npmjs.com already pointed `latest` back at 2.x. The check shelled out to `npm view`, which honors the local npm configuration, so a stale corporate registry mirror or npm cache could feed it bogus metadata. Because the check only tested that latest differs from the current version, any bogus value was presented as an upgrade recommendation.

The check now queries `registry.npmjs.org` directly over HTTPS, following the same pattern as the notices data source, so stale mirrors and local npm caches can no longer influence the result. It reuses the CLI's existing proxy agent to keep working behind corporate proxies, and no longer depends on an `npm` binary being on the PATH. It is also faster, since the previous implementation spawned two npm subprocesses inside a 3-second timeout.

As defense in depth, an upgrade is now only recommended when the reported latest version is strictly newer than the running one. The deprecation warning for the running version is kept independent of that comparison, so users on a deprecated version (like the accidental 3.0.0) are still told to move off it even though that means a downgrade.

The running version may legitimately not exist on the registry (developer builds), in which case the check proceeds without deprecation info instead of failing.

### Checklist

---

### DIFF
--- a/packages/aws-cdk/lib/cli/cli.ts
+++ b/packages/aws-cdk/lib/cli/cli.ts
@@ -244,7 +244,7 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
 
     // Do PSAs here
     if (shouldDisplayVersionMessage()) {
-      await displayVersionMessage(ioHelper);
+      await displayVersionMessage(ioHelper, { agent: proxyAgent });
     }
 
     await refreshNotices;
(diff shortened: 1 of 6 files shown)


iankhou:
Was the change also tested manually?

mrgrain:
> Was the change also tested manually?

yes

iankhou (approved):

PR #1779, opened 2026-07-29
feat(cli): send telemetry from a detached subprocess to unblock CLI exit

sanjanaravikumar-az:
This PR removes the telemetry request from the CLI's critical path. The CLI posts telemetry at the end of every invocation. That POST is 3–4 round trips to an HTTPS endpoint, and at ~100ms+ latency it adds meaningful time to
every `cdk` command — for a fire-and-forget send nobody is waiting on. This moves the POST into a detached child process:  the CLI writes the payload and exits immediately, and the child delivers the telemetry on its own and outlives the parent.

#### How it works
- A dedicated bundled entry point (`sender-bundle.js`) is spawned detached and unref'd. It reuses the real `proxy-agent`, so proxy / PAC / SOCKS / `NO_PROXY` behaviour matches the rest of the CLI.
- The parent hands the batch off via a file. The proxy URL and the CA bundle **path** (not the certificate bytes) are passed as plain data, since an `Agent` can't cross a process boundary. Forwarding the path also keeps payload size independent of CA-bundle size.
- Delivery is fire-and-forget: the CLI never learns whether the POST succeeded.


packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-cli-telemetry-disable-sends-no-data.integtest.ts
```diff
@@ -5,8 +5,8 @@ integTest(
   withDefaultFixture(async (fixture) => {
     const output = await fixture.cdk(['cli-telemetry', '--disable'], { verboseLevel: 3 });
 
-    // Check the trace that telemetry was not executed successfully
-    expect(output).not.toContain('Telemetry Sent Successfully');
+    // Check the trace that telemetry was never handed to a sender
+    expect(output).not.toContain('Telemetry dispatched');
```
iliapolo, line 9: But now our tests won't ensure telemetry is actually being sent - asserting on a dispatch is not enough. We need a way for the sender to communicate back to the test that the telemetry endpoint responded with 200.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-does-not-block-exit.integtest.ts
```diff
@@ -0,0 +1,55 @@
...
+      });
+      const blackHoleMs = Date.now() - blackHoleStart;
+
+      const overhead = blackHoleMs - disabledMs;
+      fixture.log(`synth with telemetry disabled: ${disabledMs}ms, pointed at a black hole: ${blackHoleMs}ms (overhead ${overhead}ms)`);
+
+      // The detached sender is what hangs on the black hole, not us. The headroom is generous
+      // because CI machines are noisy; what this rules out is the CLI blocking on the request
+      // timeout, which shows up as whole seconds.
+      expect(overhead).toBeLessThan(2000);
```
iliapolo, line 47: Add a note that this must assert on a number lower than our request timeout:

https://github.com/aws/aws-cdk-cli/blob/5a8df75269eebe8e2c639945c39b09898b6433e9/packages/aws-cdk/lib/cli/telemetry/sender.ts#L45

packages/aws-cdk/lib/cli/telemetry/sender.ts
```diff
@@ -0,0 +1,692 @@
...
+    return await postViaProxy(url, proxy, payload, cfg.ca, timeoutMs);
+  } catch (e: any) {
+    return { sent: false, via: 'skipped', reason: `${e?.name ?? 'Error'}: ${e?.message}` };
+  }
+}
+
+/**
+ * Resolve the proxy to use for `endpoint` from proxy environment variables.
+ *
+ * Faithfully re-implements `proxy-from-env@1`, which is what `proxy-agent` falls back to in the
```
iliapolo, line 299: This introduces massive amounts of code - how do we know it works?

An alternative would be to create a new CLI entry point that depends on the standard `proxy-agent` package and configure it for bundling as well. 

https://github.com/aws/aws-cdk-cli/blob/750d650aa0abe5d65cc362aabfbd558d7996395c/.projenrc.ts#L1450-L1452

We already have two entrypoints in `cdk-assets` for example:

https://github.com/aws/aws-cdk-cli/blob/750d650aa0abe5d65cc362aabfbd558d7996395c/.projenrc.ts#L866-L869 

We should be mindful about how much will it balloon the tarball, but lets try it.

packages/aws-cdk/lib/cli/telemetry/sink/endpoint-sink.ts
```diff
@@ -88,67 +140,87 @@ export class EndpointTelemetrySink implements ITelemetrySink {
...
+      proxyUrl: this.proxyUrl,
+      ca: this.caCert,
+    });
 
-      await this.ioHelper.defaults.trace(`Telemetry Unsuccessful: POST ${url.hostname}${url.pathname}: ${res.statusCode}:${res.statusMessage}`);
+    const payloadBytes = Buffer.byteLength(payload);
+    if (payloadBytes > MAX_DISPATCH_PAYLOAD_BYTES) {
+      // Writing this much to the child's stdin would block our own exit. Drop the batch; it is
+      // not going to get smaller on a retry.
+      await this.ioHelper.defaults.trace(`Telemetry dropped: payload of ${payloadBytes} bytes exceeds ${MAX_DISPATCH_PAYLOAD_BYTES}`);
```
iliapolo, line 169: Previously when the payload was large we would have seen evidence for that on the server logs (because the request was still being sent) - now we won't even try so we might be blind to a sudden increase in large payloads. 

Instead of dropping, rewrite the payload to indicate that the original payload was too large - and send that one. We can then add a metric on the server we can track to see if this happens a lot.

iliapolo (changes requested):

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-cli-telemetry-disable-sends-no-data.integtest.ts
rix0rrr: What we probably need is to stand up an HTTP server, set that as the telemetry endpoint, then assert on what gets sent to that endpoint.

Or in this case, we need to assert that after X seconds, we still didn't get any data POSTed to that endpoint.

And that holds for all tests, it will be a better one than asserting on the log line.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-does-not-block-exit.integtest.ts
```diff
@@ -0,0 +1,55 @@
...
+integTest(
+  'cdk synth does not wait for the telemetry endpoint',
+  withDefaultFixture(async (fixture) => {
+    const sockets: net.Socket[] = [];
+    const blackHole = net.createServer((socket) => {
+      // Accept and hold. Never respond, never close.
+      sockets.push(socket);
+    });
+    await new Promise<void>((ok) => blackHole.listen(0, '127.0.0.1', ok));
+    const port = (blackHole.address() as AddressInfo).port;
```
rix0rrr: Put all of this in a helper, not in the test itself. That way, we can reuse a server component for other integ tests.

For fun and games, make it implement `I[Async]Disposable` and use `using` (although a `try/finally` on a "normal" object with a `dispose` method will do in a pinch)

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-goes-through-a-proxy.integtest.ts
```diff
@@ -0,0 +1,95 @@
+import * as https from 'node:https';
+import type { AddressInfo } from 'node:net';
+import * as mockttp from 'mockttp';
+import { integTest, withDefaultFixture } from '../../lib';
+import { startProxyServer } from '../../lib/proxy';
+
+/**
+ * Telemetry has to keep working for users behind a corporate proxy.
+ *
+ * This matters more than it looks. The POST is made by a detached child process that has no access
```
rix0rrr: > This matters more than it looks

???????????

Please get the clanker-speech out of my sight.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-goes-through-a-proxy.integtest.ts
```diff
@@ -0,0 +1,95 @@
...
+ * deliberately out of scope (the proxy will not trust the local server's self-signed certificate,
+ * which does not matter -- the proxy records the decrypted request either way).
+ */
+integTest(
+  'telemetry is delivered through a configured proxy',
+  withDefaultFixture(async (fixture) => {
+    // Stand-in for the telemetry endpoint. Never actually serves a response to the proxy; it only
+    // needs to occupy a port so the CONNECT target is real.
+    const { key, cert } = await mockttp.generateCACertificate();
+    const endpointServer = https.createServer({ key, cert }, (_req, res) => {
```
rix0rrr, line 30: Same -- get the details of this server out of the test. Create a disposable class or object to represent the running server.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-goes-through-a-proxy.integtest.ts
```diff
@@ -0,0 +1,95 @@
...
+        ],
+        modEnv: {
+          CDK_HOME: fixture.integTestDir,
+          TELEMETRY_ENDPOINT: telemetryEndpoint,
+        },
+        verboseLevel: 3, // trace
+      });
+
+      // The parent reports the hand-off, not the delivery.
+      expect(output).toContain('Telemetry dispatched');
```
rix0rrr: Do we care about what it logs? I don't.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-goes-through-a-proxy.integtest.ts
```diff
@@ -0,0 +1,95 @@
...
+      await proxyServer.stop();
+      await new Promise<void>((ok) => endpointServer.close(() => ok()));
+    }
+  }),
+);
+
+/**
+ * Poll `fn` until it returns something truthy, or give up after `timeoutMs`.
+ */
+async function waitFor<A>(fn: () => Promise<A | undefined>, timeoutMs: number): Promise<A | undefined> {
```
rix0rrr: Surely this helper function exists somewhere already? Find it or put this in a shared place.

packages/aws-cdk/bin/cdk
```diff
@@ -1,4 +1,18 @@
 #!/usr/bin/env node
+// Publish our own location so the CLI can respawn us as a detached telemetry sender.
+// This is the only place that knows it reliably; process.argv[1] may be a .bin symlink,
+// the `cdk` alias package's wrapper, or an embedding script.
+process.env.CDK_CLI_BIN_PATH = __filename;
```
rix0rrr: Yeah but even if it is a symlink, we are able to launch it (because we are apparently launching it!).

What's the problem?

packages/aws-cdk/bin/cdk
```diff
@@ -1,4 +1,18 @@
...
+
+// That detached sender is this same script with a flag. Dispatch before requiring the CLI,
+// whose bundle costs ~600ms to load and which the sender does not need.
+if (process.env.CDK_TELEMETRY_SENDER === '1') {
+  require("../lib/cli/telemetry/sender").main();
```
rix0rrr: Maybe call that file `sender-bundle.js` as well, just for clarity.

packages/aws-cdk/lib/cli/io-host/cli-io-host.ts
```diff
@@ -407,8 +430,10 @@ export class CliIoHost implements IIoHost, ObservableIoHost {
       try {
         sinks.push(new EndpointTelemetrySink({
```
rix0rrr: Why not a new sink type? Why not a `SubprocessTelemetrySink` ?

Seems more accurate.

packages/aws-cdk/bin/cdk
rix0rrr: Also, is this CLI now communicating with itself via environment variables? 

This could have been:

```ts
require("../lib/cli/telemetry/sender").main(__filename);
```

packages/aws-cdk/lib/cli/telemetry/cli-bin-path.ts
```diff
@@ -0,0 +1,46 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { cliRootDir } from '../root-dir';
+
+/**
+ * Environment variable through which `bin/cdk` records its own location.
+ *
+ * `bin/cdk` is the only place that knows this reliably, so it publishes `__filename` here.
+ */
+export const CLI_BIN_PATH_ENV = 'CDK_CLI_BIN_PATH';
```
rix0rrr: We don't need this.

packages/aws-cdk/lib/cli/telemetry/cli-bin-path.ts
```diff
@@ -0,0 +1,46 @@
...
+ * - installed via the `cdk` alias package, it resolves to that package's wrapper, not ours;
+ * - used programmatically (`require('aws-cdk').cli()`), it is the caller's own script -- respawning
+ *   it would re-run somebody else's program.
+ *
+ * So we prefer the path `bin/cdk` published about itself, and fall back to walking up from this
+ * module to the package root (which works both from `lib/` in source and from the bundle).
+ *
+ * Returns undefined if no candidate exists on disk, in which case telemetry is skipped.
+ */
+export function cliBinPath(env: NodeJS.ProcessEnv = process.env): string | undefined {
```
rix0rrr: If we don't use the environment variable we don't need this guesswork. Just get rid of all of this.

packages/aws-cdk/lib/cli/telemetry/sender.ts
```diff
@@ -0,0 +1,692 @@
...
+import * as tls from 'node:tls';
+
+/**
+ * The detached telemetry sender.
+ *
+ * This module is executed in a short-lived, detached child process (see `bin/cdk`, which
+ * dispatches here when `CDK_TELEMETRY_SENDER=1`). Its only job is to POST a telemetry payload
+ * that it receives on stdin, and then exit.
+ *
+ * IMPORTANT: this file must only import Node built-ins.
```
rix0rrr: Not important at all. We are perfectly allowed to do targeted imports of small-ish utility files. Keeping the additional bundle small and the load time low is an admirable objective, but:

- Not to the point of foolishness
- Code duplication is the enemy of future change
- Even if it takes 600ms to load, who cares. There's nobody waiting for it anyway, by design!

For example, `ProxyAgentProvider` seems important. And if we had made a new class for the `SubprocessSink` we could reuse the `EndpointSink` here.

Redo this please.

packages/aws-cdk/lib/cli/telemetry/sink/endpoint-sink.ts
```diff
@@ -1,14 +1,31 @@
...
+/**
+ * Largest payload we are willing to hand to the detached sender.
+ *
+ * The payload is written to the child's stdin. Once it exceeds the OS pipe buffer plus libuv's
+ * own buffering, `write()` no longer completes eagerly and the parent ends up waiting for the
+ * child to drain -- which is exactly the blocking behaviour the detached sender exists to remove.
+ * Measured on Linux/Node 20, the parent still exits in ~37ms at 200KB but stalls for seconds at
+ * 400KB, so 64KB leaves a wide margin. Realistic batches are 3-10KB.
+ */
+const MAX_DISPATCH_PAYLOAD_BYTES = 65_536;
```
rix0rrr: I don't like this, at all.

First of all:

> Realistic batches are 3-10KB.

Is that true? It feels not true to me. Where did you get that number?

In any case, if buffering and draining on `stdin` is a problem (which I get because we are now forced to wait for the load time of the subprocess), then there is an *extremely* simple solution for that:

Write to a file instead, and have the subprocess read the file! 

The filesystem will easily absorb files of 100k-1M within milliseconds, hardly any waiting necessary.

packages/aws-cdk/lib/cli/telemetry/sink/endpoint-sink.ts
```diff
@@ -25,23 +42,56 @@ export interface EndpointTelemetrySinkProps {
...
+   * @default - only the system trust store
    */
-  readonly agent?: Agent;
+  readonly caCert?: string;
 }
 
 /**
  * The telemetry client that hits an external endpoint.
+ *
+ * The HTTP POST itself does not happen in this process. Events are handed to a detached child
```
rix0rrr: As I said already, please don't change the functionality of this class. Add a new class instead.

This component does what it does, and it does that usefully and well. You just happen to need a different component. Don't turn your screwdriver into a hammer. Just buy a new tool.

packages/aws-cdk/lib/cli/telemetry/sink/endpoint-sink.ts
```diff
@@ -88,67 +140,87 @@ export class EndpointTelemetrySink implements ITelemetrySink {
...
+          CDK_TELEMETRY_SENDER: '1',
+        },
+      });
+
+      // The child is on its own from here; a spawn failure must not surface anywhere. These fire
+      // after the CLI may already have exited, so they cannot go through the IoHost -- see
+      // `debugTrace`.
+      child.on('error', (e: Error) => {
+        debugTrace(`failed to spawn sender: ${e.message}`);
+      });
```
rix0rrr: Okay. We logged something. And then what, are we going to return an error from this function?

(Answer: no, we would fall through and `return true`)

Turn an error into a properly reported error/exception and handle/log it in one place.

packages/aws-cdk/test/cli/telemetry/resolve-proxy-parity.test.ts
```diff
@@ -0,0 +1,83 @@
+/**
+ * Differential parity test: our built-ins-only proxy resolution vs. the real thing.
+ *
+ * The CLI itself resolves proxies with `proxy-agent`, which delegates to `proxy-from-env` whenever
+ * the user did not pass `--proxy`. The detached telemetry sender cannot use `proxy-agent` (it has
+ * no dependencies available), so `resolveProxy` re-implements that logic. This test pins the
+ * re-implementation to the original by running both over the same table of environments.
```
rix0rrr: These should be lies. Why would we not have dependencies available?

packages/aws-cdk/test/cli/telemetry/sender.test.ts
```diff
@@ -0,0 +1,582 @@
...
+  close(): Promise<void>;
+}
+
+interface ConnectProxyOptions {
+  readonly requireAuth?: string;
+  readonly delayConnectResponseMs?: number;
+  readonly appendAfterConnectResponse?: string;
+}
+
+async function startConnectProxy(options: ConnectProxyOptions = {}): Promise<Proxy> {
```
rix0rrr, line 82: I'm pretty sure I saw a `startProxyServer` function somewhere already.

packages/aws-cdk/test/cli/telemetry/sender.test.ts
```diff
@@ -0,0 +1,582 @@
...
+
+jest.setTimeout(30_000);
+
+interface Endpoint {
+  readonly url: string;
+  readonly received: Array<{ body: string; headers: http.IncomingHttpHeaders }>;
+  close(): Promise<void>;
+}
+
+async function startEndpoint(ca: TestCa, options: { statusCode?: number; urlHost?: string } = {}): Promise<Endpoint> {
```
rix0rrr, line 45: Hey this could be reused in those other tests where I complained you did inline servers.

packages/aws-cdk/test/cli/telemetry/sender.test.ts
```diff
@@ -0,0 +1,582 @@
...
+    url: `http://127.0.0.1:${port}`,
+    connects,
+    authHeaders,
+    close: () => new Promise<void>((ok) => server.close(() => ok())),
+  };
+}
+
+const BODY = { events: [{ identifiers: { sessionId: 'test-session' } }] };
+
+describe('sender', () => {
```
rix0rrr, line 262: What all are we testing here, and why?

rix0rrr (changes requested):
There is still a lot of extraneous detail in the PR body that you could get rid of. 

A reader of this PR cares about what we do to the CLI and why. They don't need to care about all the things we tried to do to the server.

The bottom line is:

Posting to an HTTPS server can take 3-4 RTTs, and with latencies of potentially ~100ms (or more) that becomes significant.

No more explanation necessary than that.

rix0rrr:
It would be helpful to re-review if you could go through our comments and either respond to them if you have questions or disagree, or close the conversation if you feel you've addressed the request. Thanks! 🙏

sanjanaravikumar-az:
I reworked most of the design, and made some changes:

**Sink**
- Added a new `SubprocessTelemetrySink` instead of mutating the existing endpoint sink. The endpoint sink was left untouched, then deleted once nothing referenced it. 

**Proxy**
- The sender is now a second bundled entry point (`sender-bundle.js`) so it can just use the real proxy-agent like cdk-assets and integ-runner already use. 
- Bundle is ~487KB gz. ~1MB of that is the PAC WASM, which is already in `index.js`. Keeping it: PAC/SOCKS support is the entire reason corporate-proxy users need this.

**No env-var handshake**
- removed `CDK_CLI_BIN_PATH` and `cli-bin-path.ts`, and `bin/cdk` is reverted to its original 3 lines. The sender is resolved by path and spawned as `sender-bundle.js`.

**Test servers**
- One shared disposable helper (`dispose`/`using`), used by both the unit tests and the integ tests. No inline servers left.

**Payload hand-off**
- File hand-off instead of the stdin size cap. The payload goes to a file and carries the CA *path*, not the cert bytes, so batch size no longer depends on bundle size. There's a regression test with a >128KB CA asserting the payload stays under 4KB.

**Error handling**
- One place now: `dispatch()` throws, `flush()` logs once. 
- node reports spawn failures asynchronously, so a failed spawn was reading as success. Now `child.pid === undefined` throws and surfaces as a single `Dropped N event(s)`.
- Dropped `previousSendFailed` 

**Tests assert delivery, not logs**
- mockttp endpoint everywhere. Positive tests assert the batch actually arrives; negative tests assert nothing arrives after a quiet period. The block-exit test asserts both that a connection was made and that the CLI had already exited.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-cli-telemetry-disable-sends-no-data.integtest.ts
sanjanaravikumar-az: yes, these now talk to a real endpoint (mockttp, via a shared disposable helper) and assert on what actually shows up. it waits out a quiet period and asserts nothing was POSTed; the positive tests assert the batch lands. took out the log-line asserts.

packages/aws-cdk/lib/cli/telemetry/sender.ts
sanjanaravikumar-az: now there's a second bundled entry point (sender-bundle.js) using the real ProxyAgentProvider. It costs ~487KB gz, ~1MB of which is the PAC engine already duplicated in index.js. And testing is done with the proxy / SOCKS5 / TLS tests.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-does-not-block-exit.integtest.ts
sanjanaravikumar-az: server lives in one shared disposable helper (dispose/using). nothing inline in the test.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-goes-through-a-proxy.integtest.ts
sanjanaravikumar-az: yes changed

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-goes-through-a-proxy.integtest.ts
sanjanaravikumar-az: got it, now its the same shared disposable helper and test just spins it up.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-telemetry-goes-through-a-proxy.integtest.ts
sanjanaravikumar-az: changed it so now it only asserts on what reaches the endpoint now.

packages/aws-cdk/bin/cdk
sanjanaravikumar-az: makes sense, changing it back

packages/aws-cdk/lib/cli/telemetry/cli-bin-path.ts
sanjanaravikumar-az: no env-var handshake anymore - removed both CDK_CLI_BIN_PATH and cli-bin-path.ts. But it can't be an in-process require, since the child has to outlive the CLI. It's spawned as sender-bundle.js by path.

packages/aws-cdk/lib/cli/telemetry/sender.ts
sanjanaravikumar-az: done: builtins-only is out, sender-bundle.js uses the real ProxyAgentProvider, and the duplicated proxy code is deleted. also split SubprocessTelemetrySink.

packages/aws-cdk/lib/cli/telemetry/sink/endpoint-sink.ts
sanjanaravikumar-az: the 3-10KB was measured, but i suppose it doesn't matter now since the payload carries the CA path instead of bytes. there's a >128KB CA test asserting the payload stays under 4KB.

packages/aws-cdk/lib/cli/telemetry/sink/endpoint-sink.ts
sanjanaravikumar-az: fair enough - added SubprocessTelemetrySink

packages/aws-cdk/test/cli/telemetry/resolve-proxy-parity.test.ts
sanjanaravikumar-az: the published package ships zero runtime deps, so a plain file can't require('proxy-agent'). that's why the workaround existed but the bundled entry point fixed it, so this isnt necessary anymore

packages/aws-cdk/test/cli/telemetry/sender.test.ts
sanjanaravikumar-az: yes, it's the only one now, all the inline proxy/CONNECT servers are contained in it

packages/aws-cdk/test/cli/telemetry/sender.test.ts
sanjanaravikumar-az: reused. the helper covers the HTTPS endpoint, the CONNECT proxy and the SOCKS5 server, and unit + integ tests

packages/aws-cdk/test/cli/telemetry/sender.test.ts
sanjanaravikumar-az: its supposed to be proxy parity with the CLI: real CONNECT proxy, real SOCKS5, 407, NO_PROXY, TLS identity pinning (including an IP literal through a proxy, both directions), plus the >128KB CA size regression.

sanjanaravikumar-az:
> It would be helpful to re-review if you could go through our comments and either respond to them if you have questions or disagree, or close the conversation if you feel you've addressed the request. Thanks! 🙏

yes so sorry for the late replies, redid the design now with your suggestions!

packages/aws-cdk/test/cli/telemetry/sender.test.ts
rix0rrr: Are we testing our test utility?

sanjanaravikumar-az: mm no, the thing we're testing is the real client: https-proxy-agent/socks-proxy-agent from npm, picked by our ProxyAgentProvider. only way a bug is on both sides is if npm's socks client shares it.
the socks tests also assert connects[0] is the endpoint host, plus a socks5:// fail-closed case.

packages/@aws-cdk-testing/cli-integ/tests/telemetry-integ-tests/cdk-cli-telemetry-disable-sends-no-data.integtest.ts
rix0rrr: And yet here and below I still see assertions on `expect(output)`. Let's get rid of those.

Here is how to write tests (ideally):

# One conceptual behavior is asserted per test

Any assertion is either:

- The behavior we are testing, which we should only be testing once; OR
- A proxy for the behavior we are testing (for example, sending to the endpoint is hard to test so we test for a log line instead)

In this case, "sending to the endpoint", or "asserting on the data that gets sent" is NOT hard to test, so we don't need to test for a proxy.

Is "this log line appears in the output" the behavior we are testing? If we *did* want to write a test to say "the log contains X when telemetry is sent", then we would have written a single test to assert exactly that. So there is no reason for this to appear in 10 different tests.

This is either copy/paste detritus, or tests are asserting too much at once. In either case, we should get rid of those asserts.

I know you didn't make this mess, but you did touch it and we always leave the camping grounds cleaner than we found them 😉 .

packages/aws-cdk/lib/cli/telemetry/sender-bundle.ts
```diff
@@ -0,0 +1,86 @@
...
+const hardKill = setTimeout(() => process.exit(0), HARD_KILL_MS);
+hardKill.unref();
+
+// Always exit 0: nobody reads this status, and a non-zero exit would make a failed delivery look
+// like a crashed CLI.
+const done = () => {
+  clearTimeout(hardKill);
+  process.exit(0);
+};
+void main().then(done, done);
```
rix0rrr: There is a `thenable.finally()`.

packages/aws-cdk/lib/cli/telemetry/sender-bundle.ts
```diff
@@ -0,0 +1,86 @@
...
+  let cfg: TelemetrySenderConfig;
+  try {
+    cfg = JSON.parse(raw) as TelemetrySenderConfig;
+  } catch (e: any) {
+    trace(`Malformed payload: ${e?.message}`);
+    return;
+  }
+
+  try {
+    const statusCode = await sendTelemetry(cfg);
```
rix0rrr, line 60: I would have expected this process to use `EndpointTelemetrySink`. Why doesn't it?

packages/aws-cdk/lib/toolkit-error.ts
```diff
@@ -0,0 +1,12 @@
...
+// Re-exported from its defining module rather than from the `@aws-cdk/toolkit-lib` barrel, so the deep
+// path is stated once instead of copied into every file that needs it.
+//
+// Required for `sender.ts`, `post-telemetry.ts` and `proxy-agent.ts`: those are the detached telemetry
+// sender's bundle graph, and the barrel would drag the whole toolkit (~11MB) into it for the sake of
+// one error class. The other telemetry files import it for consistency rather than necessity.
+//
+// `lib/api-private.ts` is not a substitute even though it re-exports from the same package: it also
+// exports `deployStack`, `cfnApi`, the change-set describer and the activity printer, so importing it
+// would pull the entire deployment path into the sender's graph -- the opposite of the point.
```
rix0rrr: Still call this something like `api-private-error.ts` or something, to make clear that it mimics `api-private.ts` but is smaller.

Also, please get rid of the sass.

rix0rrr (commented):
Almost there!

packages/aws-cdk/lib/cli/telemetry/sender-bundle.ts
sanjanaravikumar-az: the sink doesn't do the POST, `post-telemetry.ts` does, and the sink just wrapped it. So this sender calls the same code. Also, importing the sink also pulls the toolkit barrel and aws sdk into the child (which is like ~8.6MB), and it won't bundle without a generated `cli-version.json` - and the child won't use it because it's all parent side stuff: an IoHost, a 30s batch interval, and a 500ms connectivity precheck.

packages/aws-cdk/lib/cli/telemetry/sender-bundle.ts
sanjanaravikumar-az: updated!

rix0rrr (approved):

iliapolo (approved):

PR #1781, opened 2026-07-29
feat(cli-integ): also run integ test suites on Windows - DO NOT MERGE

iankhou:
# Do not merge this. This is a POC.

[17 minute workflow run](https://github.com/aws/aws-cdk-cli/actions/runs/30676732584).

## Description
Adds windowsTestRunsOn to the integ workflow. Every integ suite (currently running on Ubuntu) will now also run on GitHub's `windows-latest`.

Windows jobs download the built artifacts to run the suites. The artifacts are built on Linux runners.

Made modifications to the testing harness, since a lot of code was linux-specific and did not run properly on Windows. 

For Windows tests, we excluded integration tests that use Linux docker images ([`DOCKER_TESTS_SKIPPED_ON_WINDOWS`](https://github.com/aws/aws-cdk-cli/pull/1781/changes#diff-337e402a94a0f7713f6f4b431ace2760e4e9941d727c9eca655b76124be9e50fR22)), which don't work on `windows-latest`'s kernel. We also use Dev Drive, a Windows 11 feature that helps to optimize file I/O operations that would otherwise take a longer time on Windows (because of NTFS and synchronous Windows Defender scanning).

We removed the dependency on `pm2` (long install time) to run Verdaccio, and run it directly, speeding up Windows jobs by 54s on average. This speeds up Linux runs too (avg 5s per job).

What I've been doing to test changes:

Run the `integ` workflow using changes from this branch. The auto-run integ on each commit doesn't test Windows, since the repo uses the integ.yml from main, not this branch. I recommend you cancel the auto run to keep from wasting Atmosphere run time.

```shell
gh workflow run integ.yml --repo aws/aws-cdk-cli --ref iankhou-windows-integ
```

and then manually approve the integ workflow.

### In-flight work
- `windows-latest` only has 4x CPU cores, which makes it too small to run all of our integ tests on Windows efficiently (lot of wait time, see integ test report); we currently have 4, 8, 16, and 32 core linux machines: see https://github.com/aws/aws-cdk-cli/settings/actions/runners, we can't get more machines this way, but will need to follow steps here to get machines on CodeBuild: https://docs.aws.amazon.com/codebuild/latest/userguide/action-runner.html - we cannot accept a significant degradation in integ-test run on PRs
- Integ tests excluded via `DOCKER_TESTS_SKIPPED_ON_WINDOWS` use Linux-based images, which we cannot run or build on `windows-latest` without configuring a Linux kernel, which in our case, would require virtualization on the hosted runner, and also takes more RAM - in theory, this is possible, but I haven't investigated it fully, it is a big lift and probably will not make the initial windows-integ-test change
- (Lower priority) Additional optimizations like caching the Verdaccio install (takes longer on Windows), caching the cli-integ install, skipping setup-node in cases where the container already has it (windows-latest ships with Node 22)

### Checklist

---

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -867,7 +867,7 @@ jobs:
           RELEASE_TAG: latest
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           INTEG_LOGS: logs
-        run: npx run-suite --shard="${{ matrix.shard }}/24" --use-cli-release=${{ steps.versions.outputs.cli_version }} --framework-version=${{ steps.versions.outputs.lib_version }} ${{ matrix.suite }}
+        run: npx run-suite --shard="${{ matrix.shard }}/12" --use-cli-release=${{ steps.versions.outputs.cli_version }} --framework-version=${{ steps.versions.outputs.lib_version }} ${{ matrix.suite }}
       - name: Set workflow summary
         id: set_workflow_summary
         if: always()
@@ -918,18 +918,6 @@ jobs:
           - 10
           - 11
           - 12
-          - 13
-          - 14
-          - 15
-          - 16
-          - 17
-          - 18
-          - 19
-          - 20
-          - 21
-          - 22
-          - 23
-          - 24
   integ_toolkit-lib_windows:
     needs: prepare
     runs-on: windows-latest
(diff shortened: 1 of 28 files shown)


dgandhi62:
Question - can I merge this? :)

iankhou:
@dgandhi62 could you do the following?
- verify that we still need the 300_000 ms timeout for integ tests, and remove or reduce where able
- split this into two PRs: 1. code changes that make the integ tests able to run on windows and 2. workflow logic that actually runs the windows tests, and schedule

PR #1785, opened 2026-07-30
fix: change-set diff can miss changes to JSON-typed properties

### LINKED ISSUE #1780: (diff): change set method does not detect changes to JSON-typed properties (AWS::QBusiness::DataSource Configuration)
### Describe the bug

Since CDK CLI version 2.1129.0, `cdk diff` using the default change set method (`--method=auto` or `--method=change-set`) does not detect changes to the `Configuration` property of `AWS::QBusiness::DataSource` resources. The diff reports "There were no differences" even though the property has been modified in the template.

Using `--method=template` correctly detects the change in all versions.

The `Configuration` property is a JSON-typed (opaque blob) property. The issue seems specific to how the change set method handles this type of property.

## Bisect Results

| CLI Version | Detects diff (change set method)? |
|-------------|-----------------------------------|
| 2.1127.0    | ✅ Yes |
| 2.1128.0    | ✅ Yes |
| 2.1129.0    | ❌ No |
| 2.1130.0    | ❌ No |
| 2.1131.0    | ❌ No |
| 2.1132.0    | ❌ No |
| 2.1133.0    | ❌ No |

### Regression Issue

2.1128.0

### Expected Behavior

`cdk diff` with `--method=auto` or `--method=change-set` should detect changes to JSON-typed properties like `Configuration` on `AWS::QBusiness::DataSource`.

### Current Behavior

`cdk diff` with the change set method reports "There were no differences" when the `Configuration` property has been modified. The change is only visible with `--method=template`.

### Reproduction Steps

1. Deploy a stack containing an `AWS::QBusiness::DataSource` with a `Configuration` property (e.g., a SharePoint data source with `inclusionFileNamePatterns`)
2. Modify the `Configuration` property in the CDK code (e.g., change a regex pattern value)
3. Run `cdk diff --method=change-set` → reports "There were no differences"
4. Run `cdk diff --method=template` → correctly detects the change
5. Run `cdk deploy` → successfully applies the change

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.1133.0

### Framework Version

_No response_

### Node.js Version

24.x

### OS

Windows / Ubuntu (reproduced on both)

### Language

TypeScript

### Language Version

TypeScript (5.6.2)

### Other information

PS C:\Users\user\project> npm install -g aws-cdk@2.1133.0

changed 1 package in 2s

PS C:\Users\user\project> cdk diff -c accountName=my-account --no-color "my-stack-with-qbusiness-datasource"

Bundling asset my-stack-with-qbusiness-datasource/log-delivery-function/Code/Stage...

  cdk\cdk.out\bundling-temp-xxxxx\index.js  4.1kb

Done in 16ms

start: Building my-stack-with-qbusiness-datasource Template
success: Built my-stack-with-qbusiness-datasource Template
start: Publishing my-stack-with-qbusiness-datasource Template (123456789012-eu-west-1-xxxxxxxx)
success: Published my-stack-with-qbusiness-datasource Template (123456789012-eu-west-1-xxxxxxxx)

Hold on while we create a read-only change set to get a diff with accurate replacement information (use --method=template to use a less accurate but faster template-only diff)

Stack my-stack-with-qbusiness-datasource
There were no differences
Omitted 1 changes because they are likely mangled non-ASCII characters. Use --strict to print them.
✨  Number of stacks with differences: 0

PS C:\Users\user\project> npm install -g aws-cdk@2.1132.0

changed 1 package in 2s

PS C:\Users\user\project> cdk diff -c accountName=my-account --no-color "my-stack-with-qbusiness-datasource"

Bundling asset my-stack-with-qbusiness-datasource/log-delivery-function/Code/Stage...

  cdk\cdk.out\bundling-temp-xxxxx\index.js  4.1kb

Done in 11ms

start: Building my-stack-with-qbusiness-datasource Template
success: Built my-stack-with-qbusiness-datasource Template
start: Publishing my-stack-with-qbusiness-datasource Template (123456789012-eu-west-1-xxxxxxxx)
success: Published my-stack-with-qbusiness-datasource Template (123456789012-eu-west-1-xxxxxxxx)

Hold on while we create a read-only change set to get a diff with accurate replacement information (use --method=template to use a less accurate but faster template-only diff)

Stack my-stack-with-qbusiness-datasource
There were no differences
Omitted 1 changes because they are likely mangled non-ASCII characters. Use --strict to print them.
✨  Number of stacks with differences: 0

PS C:\Users\user\project> npm install -g aws-cdk@2.1128.0

changed 1 package in 2s

PS C:\Users\user\project> cdk diff -c accountName=my-account --no-color "my-stack-with-qbusiness-datasource"

Bundling asset my-stack-with-qbusiness-datasource/log-delivery-function/Code/Stage...

  cdk\cdk.out\bundling-temp-xxxxx\index.js  4.1kb

Done in 16ms

start: Building my-stack-with-qbusiness-datasource Template
success: Built my-stack-with-qbusiness-datasource Template
start: Publishing my-stack-with-qbusiness-datasource Template (123456789012-eu-west-1-xxxxxxxx)
success: Published my-stack-with-qbusiness-datasource Template (123456789012-eu-west-1-xxxxxxxx)

Hold on while we create a read-only change set to get a diff with accurate replacement information (use --method=template to use a less accurate but faster template-only diff)

Stack my-stack-with-qbusiness-datasource
Resources
[~] AWS::QBusiness::DataSource my-sharepoint-data-source mySharepointDataSourceLogicalId
 └─ [~] Configuration
     └─ [~] .additionalProperties:
         └─ [~] .inclusionFileNamePatterns:
             └─ @@ -1,4 +1,4 @@
                [ ] [
                [-]   ".*PREFIX-.*\\.docx$",
                [-]   ".*PREFIX-.*\\.doc$"
                [+]   "PREFIX-NEW-.*\\.docx$",
                [+]   "PREFIX-NEW-.*\\.doc$"
                [ ] ]
Omitted 1 changes because they are likely mangled non-ASCII characters. Use --strict to print them.
✨  Number of stacks with differences: 1

mrgrain:
Fixes #1780

Since 2.1129.0, `cdk diff` can report "There were no differences" for a change you actually made. It happens when:

1. your stack has a resource with a JSON-typed property, such as `Configuration` on `AWS::QBusiness::DataSource`
2. you change a value inside that JSON
3. you run `cdk diff` with the default `--method=auto`, or with `--method=change-set`

The change is invisible in the diff, but `cdk deploy` applies it, and `--method=template` shows it. Anything else changing on the same resource is hidden too.

The cause is on the CloudFormation side. To tell you whether a change replaces a resource, we ask CloudFormation to describe the change set including property values. For a JSON blob it cannot validate, CloudFormation drops the whole resource from that response and only mentions it in a warning. We read the empty response as "nothing changed here" and threw away the correct answer we already had.

We now notice that warning and ask CloudFormation a second time, without property values, then combine both answers. You get the replacement information where it is available, and the change is no longer lost where it is not.

Diffs of stacks with many changes were affected by the same blind spot: only the first page of changes was ever read, so changes beyond it were silently dropped. All pages are now read.

## Behaviour changes worth mentioning

**`deploy` now asks CloudFormation for property values too.** `describeCurrentState` always sends `IncludePropertyValues: true`, and the deploy path reaches it through `waitAndThrowOnProblem`. Previously only `diff` requested property values; `deploy` described change sets without them. Worth knowing because it widens where the parameter is used, beyond the command this fix is about.

**`--method=execute-change-set` errors changed class and wording.** A change set that is not usable previously raised `ToolkitError('ChangeSetNotReady')` with "is not ready for execution (status: …)". It now raises `DeploymentError`, and the message depends on why: "is still being created" while it is still being created, "cannot be executed (`STATUS`)" for a change set that is empty or gone, and the full resource-level diagnosis for one that failed to create. The last of those is the reason for the change.

## Technical notes

The trigger is a `[WARN] --include-property-values option can return incomplete ChangeSet data` prefix in `StatusReason`; `Status` stays `CREATE_COMPLETE`, so this is the only signal available. It appears only when data was actually omitted, so the second `DescribeChangeSet` call is made only when needed.

Change set describing moved into a `ChangeSetDescriber` class in the new `api/change-sets` module, replacing free functions and scattered `describeChangeSet` calls. Its methods name what the caller gets: `describeCurrentState` (any state), `waitForSettled` (done being created), `waitForReport` (plus diagnosis), `waitAndThrowOnProblem`, and `describeForExecution` (usable for execution). The last one exists because an empty or deleted change set is a normal no-op for `deploy` but cannot be executed, so `diagnoseChangeSet` gained a `requireExecutable` option and a `change-set-not-ready` problem source instead of callers re-checking status themselves.

`StackDiagnosis` values are wrapped in an internal `Diagnosis` class so callers can write `diagnosis.throwOnError()`. The public `StackDiagnosis` union and `DiagnosedStack.result` are unchanged, but `StackProblemSource` gains the `change-set-not-ready` variant, which is worth a look as that type is public.

Covered by unit tests per merge scenario, including the exact response CloudFormation returns for #1780, and a `toolkit.diff` reproduction that fails without the fix.

There is no integration test, because this cannot practically be tested in CI. As detailed in https://github.com/aws/aws-cdk-cli/issues/1780#issuecomment-5130161886, only three resource types in the entire registry declare a property with an empty schema and so trigger the behaviour: `AWS::QBusiness::DataSource.Configuration`, `AWS::ControlTower::LandingZone.Manifest`, and `AWS::QuickSight::Flow.FlowDefinition`. The latter two need a Control Tower landing zone or a QuickSight subscription, and Amazon Q Business closes to new customers on 31 July 2026, so a fixture built on it would stop working in the clean accounts CI uses. A working QBusiness reproduction was used to verify the fix by hand instead.

### Checklist

---


packages/@aws-cdk-testing/cli-integ/lib/regions.ts
```diff
@@ -66,3 +73,15 @@ export const BEDROCK_AGENT_REGIONS = allBut([
   'mx-central-1',
   'us-west-1',
 ]);
+
+/**
+ * Regions that support AWS::QBusiness resources
+ *
+ * Listed rather than excluded, because Amazon Q Business exists in only a handful of regions.
+ */
+export const QBUSINESS_REGIONS = only([
```
iliapolo: Looks like leftover from an attempt to write an integ test.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -528,39 +530,48 @@ class FullCloudFormationDeployment {
   /**
    * Check rollback/replacement constraints and execute the change set if all checks pass.
    */
-  private async checkAndExecuteChangeSet(changeSetDescription: DescribeChangeSetCommandOutput): Promise<DeployStackResult> {
-    if (changeSetDescription.Status !== 'CREATE_COMPLETE') {
-      const status = changeSetDescription.Status ?? 'UNKNOWN';
-      const reason = changeSetDescription.StatusReason ? `: ${changeSetDescription.StatusReason}` : '';
-      throw new ToolkitError(
```
iliapolo, line 535: PR body states:

> A change set that is not usable previously raised ToolkitError('ChangeSetNotReady') with "is not ready for execution (status: …)". It now raises DeploymentError

I don't actually see where a `DeploymentError` would be thrown though. What i'm seeing is that `await this.cfn.executeChangeSet` will be called and throw a raw SDK error - doesn't seem expected.

iliapolo (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
mrgrain: `this.cfn.executeChangeSet` happened before. I just collapsed the single use private method  [executeChangeSet](https://github.com/aws/aws-cdk-cli/pull/1785/changes/BASE..f7e657f887e2a6efe70b399e32d2365c87ccc23f#diff-878c988044bbe730ec228c192facb2c64ba74b8e50476a1063f288c8f568f97dL589) into `checkAndExecuteChangeSet`.

The `DeploymentError` is now thrown at call sites of `checkAndExecuteChangeSet`. In order to get a `ChangeSetReport` this is checked.

iliapolo: Where is `DeploymentError` thrown? Prior to this if a changeset was in `CREATE_PENDING` this function would immediately throw a `ToolkitError`, now I see it would reach `await this.cfn.executeChangeSet`:

https://github.com/aws/aws-cdk-cli/blob/08b0d250e297e0d23e758a1914b4e69d080aa911/packages/%40aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts#L554

No?

iliapolo: Oh you mean inside `describeForExecution` - got it 👍

iliapolo (approved):

PR #1789, opened 2026-07-31
feat(cli-integ): bootstrap test environments with a random qualifier

mrgrain:
Fixes #

The integ tests bootstrap every environment with the fixed default qualifier, so the bootstrap bucket always gets the same deterministic name. With recycled test environments this regularly makes the bucket name unavailable — S3's eventual consistency lets CloudFormation believe the previous bucket still exists and the bootstrap fails. We currently paper over this with a retry loop.

This change bootstraps each test fixture with its own random qualifier and toolkit stack name, making bucket names unique so the collision can't occur (the retry stays as a safety net). As a bonus, every test now exercises the custom bootstrap qualifier feature instead of always running on the default.

Apps automatically receive the qualifier: `TestFixture` keeps an app context dict (seeded with the qualifier) that the app fixtures write into the app's `cdk.json`. This also covers invocations bypassing `fixture.cdk()`, and command-line `--context` still wins. Tests can add their own context via `addAppContext()` or the nestable `withAppContext()` block. Tests that hardcoded default-qualifier resource names now derive them from the fixture.

### Checklist

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
@@ -56,7 +56,17 @@ export function withSpecificCdkApp(
       stackNamePrefix,
       context.output,
       context.aws,
-      context.randomString);
+      context.randomString,
+    );
+    if (context.disableBootstrap) {
+      // Tests that disable the default bootstrap manage their own bootstrap
+      // stack (name, qualifier, template) and pass the matching context and
+      // arguments explicitly. Don't preconfigure a qualifier for them; e.g.
+      // the legacy bootstrap refuses to run with any qualifier set.
+      await fixture.removeAppContext('@aws-cdk/core:bootstrapQualifier');
+    } else {
+      await fixture.writeAppContext();
+    }
     await fixture.ecrPublicLogin();
 
     let success = true;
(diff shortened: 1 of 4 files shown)


packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
```diff
@@ -662,9 +746,14 @@ export class TestFixture extends ShellHelper {
...
   }
 
+  /**
+   * Look up the ECR repository name of this fixture's bootstrap stack
+   *
+   * Expects the environment to have been bootstrapped already with
+   * `toolkitStackName: fixture.bootstrapStackName` (or via `ensureBootstrapped`,
+   * which uses that stack name by default).
+   */
   public async bootstrapRepoName(): Promise<string> {
```
mrgrain, line 782: Removed the `ensureBootstrapped()` call that used to be here. It was vestigial: it bootstrapped the default `CDKToolkit` stack, which never matched the `fixture.bootstrapStackName` lookup below — the assertion only ever passed because all callers (the gc ECR tests) explicitly bootstrap that stack first. With the new per-fixture bootstrap it would have become harmful, re-bootstrapping the caller's stack with `--bootstrap-kms-key-id AWS_MANAGED_KEY` and triggering a CloudFormation update (bucket encryption change) mid-test.

iliapolo (approved):

PR #1791, opened 2026-07-31
fix(toolkit-lib): refactor ignores a custom toolkit stack name

### LINKED ISSUE #1790: refactor: does not respect custom toolkit stack name when uploading large templates
### Describe the feature

`cdk refactor` should respect a custom toolkit stack name (from `--toolkit-stack-name` or the `toolkitStackName` setting in `cdk.json`) when it looks up the bootstrap stack.

### Use Case

When refactoring stacks whose templates exceed the 50KiB limit, the refactor action must upload templates to the bootstrap staging bucket. To find that bucket, the large-template path in `@aws-cdk/toolkit-lib` (`lib/api/refactoring/stack-definitions.ts`) constructs `new EnvironmentResourcesRegistry()` without a toolkit stack name, so `ToolkitInfo.lookup()` always falls back to the default `CDKToolkit` name. The CLI side compounds this: `cli.refactor(...)` in `packages/aws-cdk/lib/cli/cli.ts` does not forward the resolved `toolkitStackName` (unlike deploy/import, which thread it through `Deployments`).

As a result, any environment bootstrapped under a custom toolkit stack name fails large-template refactors with:

```
Refactor failed: Template too large to refactor ("cdk bootstrap" is required)
```

even though the environment is correctly bootstrapped.

We hit this in the CLI integration tests while moving them to randomly-qualified, custom-named bootstrap stacks (#1789), but it affects any customer using `--toolkit-stack-name` / `toolkitStackName`.

### Proposed Solution

Thread the toolkit stack name through the refactor action:

- Add it to the refactor options in `@aws-cdk/toolkit-lib` and pass it to `EnvironmentResourcesRegistry` in `stack-definitions.ts`.
- Forward the already-resolved `toolkitStackName` in the CLI's `case 'refactor'` handler.

### Other Information

_No response_

### Acknowledgements

- [x] I may be able to implement this feature request

### CDK CLI Version

2.1134.0

CI (GitHub Actions, cli-integ tests)

---

*This issue was filed by an AI agent (Kiro) on behalf of @mrgrain.*

mrgrain:
Fixes #1790

`cdk refactor` failed with `Template too large to refactor ("cdk bootstrap" is required)` in environments bootstrapped under a custom toolkit stack name, even though they were bootstrapped correctly. Refactoring a template larger than 50KiB has to upload it to the bootstrap staging bucket, and that lookup always searched for `CDKToolkit`, ignoring both `--toolkit-stack-name` and `toolkitStackName` in `cdk.json`. The same blind spot silently disabled the bootstrap version check, so a bootstrap stack too old to refactor was only caught later, inside CloudFormation.

`cdk refactor` now also accepts `--toolkit-stack-name`, which was rejected here while `deploy`, `import` and `bootstrap` all accept it.

Both of these now work as you'd expect: point `cdk refactor` at a custom-named bootstrap stack, by flag or config, and large templates are staged in the right bucket.

### Checklist

---


packages/aws-cdk/lib/cli/user-input.ts
```diff
@@ -1848,6 +1848,13 @@ export interface RefactorOptions {
    */
   readonly force?: boolean;
 
+  /**
+   * The name of the existing CDK toolkit stack (used to find the staging bucket for templates that are too large to be sent inline)
+   *
+   * @default - undefined
+   */
+  readonly toolkitStackName?: string;
+
```
iankhou, line 1857: Isn't the default `CDKToolkit`?

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -799,7 +799,8 @@ class CloudControlHotswapStack extends cdk.Stack {
       instruction: process.env.DYNAMIC_CC_PROPERTY_VALUE
         ? `You help query the table ${table.tableName}. ${process.env.DYNAMIC_CC_PROPERTY_VALUE}. ${process.env.DYNAMIC_CC_PROPERTY_VALUE_2 ?? 'original'}`
         : `You help query the table ${table.tableName}. original. original`,
-      foundationModel: 'anthropic.claude-instant-v1',
+      // Amazon models don't require a per-account access grant
+      foundationModel: 'amazon.nova-micro-v1:0',
```
iankhou: How does this relate to the refactor change?

mrgrain: It doesn't but the integ tests are currently broken: Both hotswap tests die deploying the `cc-hotswap` stack with: `Access denied for operation 'AWS::Bedrock::Agent'`
  
Turns out the test app's Bedrock agent uses `anthropic.claude-instant-v1` — that model has been retired and doesn't even show up in `list-foundation-models` anymore. Bedrock reports a gone model as `AccessDenied`, so it looks like a permissions problem when it isn't ([model lifecycle docs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html)).
 
Fix: switched the agent to `amazon.nova-micro-v1:0` — broadly available, no access grant needed, and the test never invokes the model anyway.

packages/aws-cdk/lib/cli/user-input.ts
mrgrain: Well it depends. This file is auto-generated from the CLI config. In the sense of the config and what values are passed around in the code, the default is `undefined.` From the perspective of the CLI I don't think we want to give this a default value; empty is fine.

Note this is not a user facing file and how the existing `toolkitStackName` also have the same default. If we were to change it, definitely not in this PR.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
mrgrain: But it doesn't actually fix it.

mrgrain: This is the actual reason: https://docs.aws.amazon.com/bedrock/latest/userguide/agents-classic-maintenance-mode.html

packages/aws-cdk/lib/cli/user-input.ts
iankhou: I see, that makes sense since we set the default later in the chain. But might be worth bringing the default "left" in the future.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
iankhou: Thanks for the context!

iankhou: Integ test run is still failing with:

```
  cdktest-0hiui07qsdl7-cc-hotswap | 2/7 | 1:58:11 PM | CREATE_FAILED           | AWS::Bedrock::Agent               | Agent Resource handler returned message: "Access denied for operation 'AWS::Bedrock::Agent'." (RequestToken: 72ff3305-abfe-0301-bf72-0d5295a364d9, HandlerErrorCode: AccessDenied)

```

mrgrain: resolve in https://github.com/aws/aws-cdk-cli/pull/1793 instead

iankhou (approved):

PR #1799, opened 2026-08-04
chore: improve logging in tests

ShadowCat567:
Updates logging in tests. 

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/corking.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/corking.ts
@@ -13,15 +13,10 @@ const SECRETS = new Set<string>();
 
 /**
  * Register values to be replaced with a placeholder in all buffered output
- *
- * Short values are ignored: they are unlikely to be real secrets and are likely to occur
- * as substrings of legitimate output, where redacting them would corrupt the log.
  */
-export function registerSecrets(...values: Array<string | undefined>) {
+export function registerSecrets(...values: Array<string>) {
   for (const value of values) {
-    if (value && value.length >= 8) {
-      SECRETS.add(value);
-    }
+    SECRETS.add(value);
   }
 }
 
--- a/packages/@aws-cdk-testing/cli-integ/lib/corking.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/corking.ts
@@ -3,6 +3,39 @@
  */
 import * as stream from 'stream';
 
+/**
+ * Values that must be scrubbed from any buffered test output
+ *
+ * Registered values are process-global because `MemoryStream`s are created per test,
+ * while credentials are established once per worker process.
+ */
+const SECRETS = new Set<string>();
+
+/**
+ * Register values to be replaced with a placeholder in all buffered output
+ *
+ * Short values are ignored: they are unlikely to be real secrets and are likely to occur
+ * as substrings of legitimate output, where redacting them would corrupt the log.
+ */
+export function registerSecrets(...values: Array<string | undefined>) {
+  for (const value of values) {
+    if (value && value.length >= 8) {
+      SECRETS.add(value);
+    }
+  }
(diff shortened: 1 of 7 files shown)


packages/@aws-cdk-testing/cli-integ/lib/corking.ts
```diff
@@ -3,6 +3,39 @@
...
+ *
+ * Registered values are process-global because `MemoryStream`s are created per test,
+ * while credentials are established once per worker process.
+ */
+const SECRETS = new Set<string>();
+
+/**
+ * Register values to be replaced with a placeholder in all buffered output
+ *
+ * Short values are ignored: they are unlikely to be real secrets and are likely to occur
```
iliapolo: But the registration happens explicitly, I think we can trust the caller here. 

Also in general silently ignoring creates unexpected behavior, if anything - i'd throw an error here because it means that either the caller is wrong, or there __is__ actually a short secret.

packages/@aws-cdk-testing/cli-integ/lib/corking.ts
```diff
@@ -3,6 +3,39 @@
...
+ * as substrings of legitimate output, where redacting them would corrupt the log.
+ */
+export function registerSecrets(...values: Array<string | undefined>) {
```
iliapolo: Why allow for `undefined`?

```suggestion
export function registerSecrets(...values: Array<string>) {
```

iliapolo (dismissed):

packages/@aws-cdk-testing/cli-integ/lib/corking.ts
ShadowCat567: This would be the case for when we are pulling secrets from environment variable and they are possibly undefined. However, since we control the environment that should not be the case...

mrgrain (changes requested):
~~I think in addition we can also update `renderEnv` in `tests/cli-integ-tests/proxy/proxy.integtest.ts` to redact the values of the credentials env vars explicitly (no matter their value).~~

Never mind, already happening.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/proxy/proxy.integtest.ts
```diff
@@ -138,7 +140,10 @@ function isolatedDockerCommands(proxyPort: number, caBundlePath: string) {
 }
 
 function renderEnv(env: Record<string, string | undefined>) {
-  return Object.entries(env).filter(([_, v]) => v).map(([k, v]) => `${k}='${v}'`).join(' ');
+  return Object.entries(env)
+    .filter(([k, v]) => v && !CREDENTIAL_ENV_VARS.includes(k))
+    .map(([k, v]) => `${k}='${v}'`)
+    .join(' ');
```
mrgrain, line 146: Probably better to replace this with `<REDACTED>` then removing it?

mrgrain (changes requested):

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/proxy/proxy.integtest.ts
ShadowCat567: Why? We don't seem to be using the credentials the come in through this route

ShadowCat567: I attempted to replace with redacted here, but that causes our proxy test to fail because it seems like we try to get env variables from here first and if we can't find them here, we fall back to looking at the rest of what was provided to us (and there are other env vars that are not the `CREDENTIAL_ENV_VARS` that pass through here). So it is actually better to remove them entirely than to redact them.

mrgrain (approved):

PR #1800, opened 2026-08-06
feat(cli): add "errors-only" progress mode that only prints errors

9pace:
Adds a third \`--progress\` mode: \`errors-only\`. It prints nothing during deployment except failures.

**Motivation:**
The existing modes waste LLM agent tokens for no benefit. `bar` never reaches agents anyway (degrades to `events` on non-TTY), so agents get the full per-event firehose today. An agent can't act on `CREATE_IN_PROGRESS` lines — it only needs *succeeded → outputs* or *failed → why*, and `errors-only` preserves both.

**Diff between events mode and errors-only mode**
image

**Measured** :
- `events`: 2,793 bytes / 36 lines (~700 tokens)
- `errors-only`: 1,169 bytes / 23 lines (~290 tokens)

**Design decisions:**
- `errors-only` survives the non-TTY degradations and `--concurrency` (only `bar` falls back to `events`)
- Named `errors-only` rather than `quiet` to avoid the existing `--quiet` flag on synth/diff, and to avoid the log-level connotation of a bare `errors` (webpack's `errors-only` stats preset and AWS CLI's `--only-show-errors` set the precedent)
- Defaulting to `errors-only` when an AI agent is detected will follow in a separate PR

### Checklist

---


mrgrain (changes requested):
Cool idea, but can we can it something else than quiet? We already have --quiet flag and this is going to be confusing.

packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -115,6 +116,12 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
       },
     });
 
+  // Progress updates are wasted tokens for AI agents
+  if (guessAgent() && configuration.settings.get(['progress']) === undefined) {
+    ioHost.stackProgress = StackActivityProgress.ERRORS_ONLY;
+    await ioHost.defaults.debug('AI agent detected, defaulting to --progress "errors-only"');
+  }
```
mrgrain: this is the key change in this PR, isn't it? I'd prefer to split it into two PRs (1/ add the flag 2/ make it the default for agents)

packages/@aws-cdk/toolkit-lib/lib/private/activity-printer/errors-only.ts
```diff
@@ -0,0 +1,61 @@
...
+  public start(state: { stack: CloudFormationStackArtifact }) {
+    super.start(state);
+    this.printedFailures = 0;
+  }
+
+  public stop() {
+    super.stop();
+
+    if (this.failures.some((f) => this.isProvisionalFailure(f))) {
+      this.stream.write(chalk.yellow('\n ⚠️  Some resources failed to delete but were skipped. These resources may still exist and could incur charges. Clean them up manually.\n'));
```
mrgrain: can we centralize this message so we don't have two copies of it?

mrgrain (changes requested):
Almost!

packages/aws-cdk/lib/cli/cli.ts
9pace: removed the default for agents from this pr

packages/@aws-cdk/toolkit-lib/lib/private/activity-printer/errors-only.ts
9pace: moved up into ActivityPrinterBase.skippedDeletesWarning()

mrgrain (approved):

PR #1803, opened 2026-08-08
fix(toolkit-lib): stale DescribeStacks read fails fresh stack creates, masked as NoStack

### LINKED ISSUE #1802: (toolkit-lib): stale DescribeStacks read fails fresh stack creates, and the error is masked as NoStack
## Describe the bug

Under CloudFormation read pressure, a fresh stack create via `Toolkit.deploy()` intermittently fails with an error that says nothing useful:

```
NoStack: ❌  <StackName> failed: NoStack: CloudFormationStack object does not hold a stack
```

Two bugs compound.

**1. A stale `DescribeStacks` read is treated as terminal.** For a fresh CREATE via changeset, `ExecuteChangeSet` moves the stack `REVIEW_IN_PROGRESS` → `CREATE_IN_PROGRESS`. Because `DescribeStacks` is eventually consistent, a later poll can still see `REVIEW_IN_PROGRESS`. `StackStatus.isInProgress` excludes review, so that read falls into the `isReviewInProgress` carve-out in `stabilizeStack` (`cfn-api.ts:426`), which returns the stack as *stable*. `waitForStackDeploy` then throws `StackDeployFailed` because `REVIEW_IN_PROGRESS` isn't a success status. The carve-out is right at the start of a wait (abandoned changeset), but wrong mid-wait — a real stack can't go `CREATE_IN_PROGRESS` → `REVIEW_IN_PROGRESS`, so it can only be a stale read.

**2. The diagnosis path then masks that error.** In `monitorDeployment` (`deploy-stack.ts:718`), the catch block passes `finalState.wrapped` to `diagnoseFromErrorCollection`. `finalState` is the *pre-deploy* lookup, which for a new stack holds no `Stack`, so the `wrapped` getter throws `NoStack`. That happens while evaluating an argument inside the `catch`, so it replaces the original `DeploymentError` entirely.

Bug 2 isn't specific to bug 1 — it masks **any** deployment error on a fresh create, including a plain `ROLLBACK_COMPLETE` from a resource failure.

Observed on `@aws-cdk/toolkit-lib` 1.35.0; both paths unchanged on `main` at `65517408`.

## Expected Behavior

A stale read shouldn't fail a deployment that CloudFormation goes on to complete. And a fresh-create failure should surface its real cause — `NoStack` is an internal invariant violation that shouldn't be reachable from a normal failed deploy.

## Current Behavior

The deployment succeeds in CloudFormation, but `deploy` throws `NoStack` roughly one stabilization interval after `ExecuteChangeSet`. CloudTrail across three failing stacks shows every `DescribeStacks` returning 200 with no `errorCode` (a genuinely absent stack returns `ValidationError`), under heavy `ThrottlingException` on CFN reads — consistent with a stale replica read rather than a missing stack.

This hit the [Powertools for AWS Lambda (TypeScript)](https://github.com/aws-powertools/powertools-lambda-typescript) e2e CI: ~40 parallel jobs deploying small stacks into one account/region, ~15% of matrix cells failing per run ([baseline run](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/31155239957), 6/40 failed). Tracked downstream as [aws-powertools/powertools-lambda-typescript#5537](https://github.com/aws-powertools/powertools-lambda-typescript/issues/5537).

## Reproduction Steps

Both bugs reproduce as unit tests against `main`, using the existing `FakeCloudFormation` / `mockCloudFormationClient` / `advanceTime` helpers:

- **Bug 1**: `DescribeStacks` returning `CREATE_IN_PROGRESS` → `REVIEW_IN_PROGRESS` → `CREATE_COMPLETE` should deploy successfully. Currently throws `DeploymentError: ... failed to deploy: REVIEW_IN_PROGRESS` at `cfn-api.ts:399`.
- **Bug 2**: a fresh create driven to `ROLLBACK_COMPLETE` (via the fake's `alwaysFailResources`) should surface `StackCreationFailed`. Currently throws `NoStack` from `stack-helpers.ts:75`, via `deploy-stack.ts:718`.

We also confirmed the masking directly: making `wrapped` non-throwing reveals the real error underneath — `DeploymentError` / `StackCreationFailed` / "The stack named X failed creation ...: ROLLBACK_COMPLETE".

## Possible Solution

We're working on a fix and will open a PR.

Sketch: for bug 2, don't evaluate `finalState.wrapped` when the pre-deploy stack doesn't exist — guard on `finalState.exists` or re-look-up the stack for diagnosis, preserving the existing rethrow-when-`noProblem` behaviour. For bug 1, treat `REVIEW_IN_PROGRESS` seen after a real in-progress state as a stale read and keep polling, bounded so the abandoned-changeset escape hatch survives (`stabilizeStack` is stateless across polls today).

We validated both as a load-time monkey-patch in the Powertools repo — retrying `waitForStackDeploy` on the spurious error, and making `wrapped` non-throwing. The 40-cell matrix went [40/40 green](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/31224476369) vs 6/40 failing on baseline, and in a later run 13 genuine fresh-create failures all surfaced their real `DeploymentError`s through the path that previously produced `NoStack`.

## Additional Information/Context

Related: #1709 and #1723 (closed) made the polling intervals configurable — same read-pressure environment. #1777 (open) is the closest sibling: also `monitorDeployment`, also read pressure turning a successful deploy into a reported failure, via the unguarded `monitor.stop()`.

## CDK CLI Version

N/A — used programmatically via `@aws-cdk/toolkit-lib` `^1.35.0`

## Framework Version

aws-cdk-lib `^2.261.0`

## Node.js Version

24

## OS

Linux (GitHub Actions `ubuntu-latest`)

## Language

TypeScript

## Language Version

TypeScript `^6.0.3`

svozza:
Fixes #1802

Two independent bugs that combine to fail fresh stack creates under CloudFormation read pressure with `NoStack: CloudFormationStack object does not hold a stack`. One commit each.

### 1. A stale `DescribeStacks` read failed a running deployment

`DescribeStacks` is eventually consistent, so a poll issued after `ExecuteChangeSet` can still report the pre-execution `REVIEW_IN_PROGRESS` status. `StackStatus.isInProgress` excludes review, so that read fell into the `isReviewInProgress` carve-out in `stabilizeStack`, which returned the stack as *stable*; `waitForStackDeploy` then rejected the deployment with `StackDeployFailed` while CloudFormation went on to reach `CREATE_COMPLETE`.

A stack cannot transition from an in-progress state back to `REVIEW_IN_PROGRESS`, so that status is a stale read whenever it is reported for a stack whose execution we know has been issued. `stabilizeStack` now keeps polling in that case.

Two details worth calling out:

- **Identified by stack id, not by a flag.** Polling by name can observe a *different* stack that a concurrent operation created under the same name (`stack A: UPDATE_IN_PROGRESS` → deleted → `stack B` same name: `REVIEW_IN_PROGRESS`). That review status is genuine, and treating it as stale would wait forever, since `waitFor` has no timeout. Comparing stack ids keeps the two cases apart.
- **`monitorDeployment` passes the executing stack id in.** Nothing guarantees the *first* `DescribeStacks` after execution observes the new status, so recognising a stale read cannot depend on having seen the operation in progress first.

Stale reads are tolerated in bounded number (`STALE_REVIEW_READ_TOLERANCE`), so a stack genuinely left in `REVIEW_IN_PROGRESS` still terminates the wait and reaches the pre-execution behaviour the carve-out was written for — nothing moves an unexecuted ChangeSet on its own.

### 2. `NoStack` masked the real error on any failed create

`monitorDeployment` passed `finalState.wrapped` to the diagnoser from inside its catch block. `finalState` is still the pre-deploy lookup there, which holds no stack when creating one from scratch, so the getter threw `NoStack` — and because that happened while evaluating an *argument*, it replaced the deployment error being reported.

This affected **every** failed deployment of a new stack, not only those caused by bug 1: a resource failure that rolled the stack back reported `NoStack` instead of naming the resource.

It now describes the stack that was actually deployed. The pre-deploy lookup is the wrong input even when it does hold a stack, since it describes a state the deployment has since left and the diagnoser reads the status off it. Diagnosing is best-effort, so a failed lookup leaves the original deployment error to propagate rather than replacing it with an `ErrorDiagnosisFailed` that says less.

### Testing

Unit tests only; no new AWS resource types or cross-service interactions, so no integ test.

- `cfn-api-stabilization.test.ts` (new) — stale read mid-wait, stale read on the first poll, a different stack id treated as genuine, persistent review terminating rather than hanging, and the abandoned-ChangeSet escape hatch still failing fast.
- `deploy-stack-error-surfacing.test.ts` (new) — failing create via change-set and direct, with and without rollback, plus a failing update of an existing stack to guard against a fix that only works when the stack is missing.

Every new test was confirmed to fail against the unfixed code and pass after. Full `toolkit-lib` suite passes (1896 tests).

`deploy-stack-polling-interval.test.ts` needed one update: it asserts `waitForStackDeploy`'s exact argument list, which the new parameter changes.

### How this was found

The e2e CI for [Powertools for AWS Lambda (TypeScript)](https://github.com/aws-powertools/powertools-lambda-typescript) — ~40 parallel jobs deploying small stacks into one account/region — failed ~15% of matrix cells per run. CloudTrail across three failing stacks showed every `DescribeStacks` returning 200 with no `errorCode` (a genuinely absent stack returns `ValidationError`) under heavy `ThrottlingException` on CFN reads, consistent with a stale replica read.

Both fixes were validated there as a [load-time monkey-patch](https://github.com/aws-powertools/powertools-lambda-typescript/blob/77b3e8f8/packages/testing/src/patchToolkitLib.ts) before being written properly here — [`patchWaitForStackDeploy`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/77b3e8f8/packages/testing/src/patchToolkitLib.ts#L77-L113) retried stabilization on the spurious error (bug 1), and [`patchWrappedGetter`](https://github.com/aws-powertools/powertools-lambda-typescript/blob/77b3e8f8/packages/testing/src/patchToolkitLib.ts#L116-L123) made `wrapped` non-throwing (bug 2).

Results: the matrix went [40/40 green](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/31224476369) against [6/40 failing on baseline](https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/31155239957), and in a later run 13 genuine fresh-create failures all surfaced their real `DeploymentError`s through the path that previously produced `NoStack`.

The patch is a workaround rather than a model for this PR — it retries at the `waitForStackDeploy` boundary rather than fixing the stale-read classification inside `stabilizeStack`, and making `wrapped` return `{}` hides a real invariant instead of not violating it. What it does establish is that the two behaviours being changed here are the ones responsible for the failures.

### Checklist

---


packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -382,8 +388,9 @@ export async function waitForStackDeploy(
   ioHelper: IoHelper,
   stackName: string,
   stabilizationPollingInterval?: number,
+  executingStackId?: string,
```
svozza, line 391: Flagging a design choice for maintainer input.

This adds a fifth positional parameter, so `waitForStackDeploy` now reads `(cfn, ioHelper, stackName, stabilizationPollingInterval, executingStackId)`, and `stabilizeStack` below takes the same shape. Two optional `undefined`-able tail parameters of different types are easy to transpose at a call site.

An options object would read better and scale if more parameters get added. I kept positional parameters to keep the diff minimal, and because each function has a single caller today — but I'm happy to switch in this PR if you'd prefer. Neither function is exported from `lib/index.ts`, so there's no API Extractor impact either way.

Happy to go whichever way you'd rather.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -404,15 +411,25 @@ export async function waitForStackDeploy(
...
+ *
+ * @param executingStackId - the id of a stack whose ChangeSet execution has already been issued. `REVIEW_IN_PROGRESS`
+ * reported for that stack is a stale read rather than a stable status, because execution has moved it on already.
  */
 export async function stabilizeStack(
   cfn: ICloudFormationClient,
   ioHelper: IoHelper,
   stackNameOrArn: string,
   pollingInterval?: number,
+  executingStackId?: string,
```
mrgrain, line 423: This is not needed, `stackNameOrArn` can already take a Stack ARN as input and Stack Id is just the Stack ARN.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -421,9 +438,20 @@ export async function stabilizeStack(
     }
     const status = stack.stackStatus;
     if (status.isInProgress) {
+      inProgressStackId = stack.stackId;
```
mrgrain, line 441: This seems good as a general idea, but will need a code change:

- If the originally provided `stackNameOrArn` was NOT an ARN, then we "upgrade" the value to the ARN from the first check.
- If it already was an ARN, we keep using it.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -421,9 +438,20 @@ export async function stabilizeStack(
...
+      staleReviewReads = 0;
       await ioHelper.defaults.debug(format('Stack %s has an ongoing operation in progress and is not stable (%s)', stackDisplayName, status));
       return undefined;
     } else if (status.isReviewInProgress) {
+      // A stack cannot go from an in-progress state back to REVIEW_IN_PROGRESS, so for the same stack id this is a
+      // stale read from an eventually consistent DescribeStacks replica. Tolerate a bounded number of them so that
+      // a stack genuinely left in review - which nothing will move on its own - still terminates the wait.
+      if (stack.stackId === inProgressStackId && staleReviewReads < STALE_REVIEW_READ_TOLERANCE) {
+        staleReviewReads++;
+        await ioHelper.defaults.debug(format('Stack %s reported REVIEW_IN_PROGRESS after being in progress; treating as a stale read (%s)', stackDisplayName, status));
```
mrgrain, line 451: Can you phrase this a bit more concrete on what the next action is (which I guess is we try again a few more times)

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
mrgrain: See other comment, it's not needed because `stabilizeStack` already takes a `stackNameOrArn`. We can make the same change to `waitForStackDeploy`. The main todo then is to make sure that we use the display name for logs and errors.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -24,6 +24,12 @@ import type { IoHelper } from '../io/private';
 import type { ResourcesToImport } from '../resource-import';
 import { StackArtifactSourceTracer } from '../source-tracing/private/stack-source-tracing';
 
+/**
+ * How many consecutive REVIEW_IN_PROGRESS reads to attribute to DescribeStacks eventual consistency
+ * before accepting the status as genuine.
+ */
+const STALE_REVIEW_READ_TOLERANCE = 5;
```
mrgrain, line 31: q: what is the upper limit for wall clock time impact here? How does this compare to CFNs published eventual consistency guarantees?

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -705,7 +705,7 @@ class FullCloudFormationDeployment {
 
     let finalState = this.cloudFormationStack;
     try {
-      const successStack = await waitForStackDeploy(this.cfn, this.ioHelper, this.stackName, this.options.stackEventPollingInterval);
+      const successStack = await waitForStackDeploy(this.cfn, this.ioHelper, this.stackName, this.options.stackEventPollingInterval, stackArn);
```
mrgrain, line 708: Probably just this...

```suggestion
      const successStack = await waitForStackDeploy(this.cfn, this.ioHelper, stackArn, this.options.stackEventPollingInterval);
```

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -715,10 +715,16 @@ class FullCloudFormationDeployment {
...
       // If this is a deployment error, route the diagnosis and error reporting through the central code for that
       if (ToolkitError.isDeploymentError(e)) {
-        const diagnosis = await this.diagnoser.diagnoseFromErrorCollection(monitor.errors, finalState.wrapped, true, {
-          rollbackEnabled: this.options.rollback !== false,
-        });
-        diagnosis.throwOnError();
+        // `finalState` is still the pre-deploy lookup, which holds no stack at all when we were creating one from
+        // scratch, and describes a state the deployment has since left when it does. Describe the stack we actually
+        // deployed instead. Diagnosing is best-effort: if it cannot tell us anything, report the deployment error.
+        const deployedState = await CloudFormationStack.lookup(this.cfn, stackArn).catch(() => undefined);
```
mrgrain, line 721: but this can also fail... what then?

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -715,10 +715,16 @@ class FullCloudFormationDeployment {
...
+        if (deployedState?.exists) {
+          const diagnosis = await this.diagnoser.diagnoseFromErrorCollection(monitor.errors, deployedState.wrapped, true, {
+            rollbackEnabled: this.options.rollback !== false,
+          });
+          diagnosis.throwOnError();
+        }
```
mrgrain, line 727: this whole code block seems very convoluted to me and the new if is making it even worse. can you take a step back and look at what this is aiming to do? What errors can be throw and when and how do we react to it?

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
svozza: Addressed in #1845. If the describe fails, `diagnoseDeploymentFailure` returns without throwing, so the original `DeploymentError` propagates — diagnosis can only ever replace it with something that says more. There's a test that fails the describe with a throttling error and asserts the real `ROLLBACK_COMPLETE` error survives.

The failure is also logged at debug now, matching `stack-diagnoser.ts:163` — previously a throttled or access-denied diagnosis was invisible even under `-v`.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
svozza: Fair — took a step back on this in #1845.

The invariant the block was groping at: diagnosis is an *enrichment* of the deployment error, so it may only replace that error with something that says more. Enumerated what can be thrown:

| Path | Old | New |
|---|---|---|
| fresh create, no stack in `finalState` | `wrapped` threw `NoStack`, replacing the real error | describes the deployed stack instead |
| describe fails while diagnosing | unreachable — threw before getting there | returns, original error propagates |
| diagnosis finds no cause | original rethrown (correct) | unchanged |

That's now one helper holding to the invariant, and the catch block is three lines.

Two things fell out of it. `finalState`'s initialiser turned out to be provably dead once the catch stopped reading it, so the mutable local that meant two different things at two points in the method no longer starts out holding a pre-deploy value — which is what made the bug representable. And with the helper describing the stack directly rather than via `CloudFormationStack.lookup(...).wrapped`, the `wrapped` getter had no callers left anywhere in the repo, so it's gone.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
svozza: Agreed, and verified — a stack ARN *is* the stack id, and `CloudFormationStack.lookup` passes it straight to `DescribeStacks` as `StackName`, so passing the ARN gives identity-pinning for free. A replacement stack created under the same name simply isn't found by the old ARN. Dropping `executingStackId` and the id comparison entirely. Moving to the follow-up PR with the rest of the stale-read fix.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
svozza: Will do — `stackNameOrArn` on `waitForStackDeploy` too. Noting the display-name point: the two `DeploymentError` messages interpolate the raw `stackName`, so they'd leak a full ARN; will run them through `stackNameFromArn()` as `stabilizeStack` already does. Follow-up PR.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
svozza: Taking this suggestion. Depends on the two above. Follow-up PR.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
svozza: Makes sense — that covers the callers passing a bare name (`toolkit-info.ts:57`, `deployments.ts:622`), so they get identity-pinning after their first successful read. Follow-up PR.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
svozza: Will do. Follow-up PR.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
svozza: Honest answer on the wall clock: there is no published CloudFormation bound to compare against, so I can'''t derive the tolerance from one.

What I could establish. AWS annotates read consistency on exactly five CloudFormation APIs, all StackSets — `DescribeStackSet`, `DescribeStackSetOperation` and `ListStackSets` as strongly consistent, `ListStackSetOperations` and `ListStackSetOperationResults` as eventually consistent. No stack-level read carries an annotation, `DescribeStacks` included, so its consistency model is simply undocumented — that cuts both ways and I'''m not going to read an eventual-consistency guarantee into the silence. `ExecuteChangeSet` says only "After the call successfully completes, CloudFormation starts updating the stack. Use the DescribeStacks action to view the status of the update" — it points you at `DescribeStacks` without saying when the new status becomes visible there. The SLA covers 500/503 only, so a 200 carrying stale data isn'''t in scope. EC2 and IAM have dedicated eventual-consistency pages; CloudFormation has none, and isn'''t in IAM'''s cross-reference list. The closest AWS puts a number on it anywhere is the general glossary: "Consistency is usually reached within seconds" — hedged, and not CFN-specific.

So the tolerance is an empirical margin, not a derived one, and I'''ll say that in the code comment rather than implying otherwise.

Worst case in wall clock, which is fully determined by our own code: the counter resets on any in-progress read, so it bounds *consecutive* stale reads — 5 reads x the stabilization interval per stale burst, so 10s at the 2s default and 50s at the 10s interval that prompted #1709/#1723. It only spends that in the genuinely-abandoned-changeset case, where the wait previously exited immediately; on the stale-read case it saves a failed deployment.

One thing that argues the bound matters less than it looks: pinning to the stack ARN per your other comments removes the known way a persistent `REVIEW_IN_PROGRESS` could be attributed to our own operation, so the counter becomes a backstop rather than the mechanism. Happy to drop it if you'''d rather not carry an arbitrary constant.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
svozza: Answered by mrgrain'''s review — `stabilizeStack` already takes a `stackNameOrArn`, and a stack ARN is the stack id, so the extra parameter isn'''t needed at all. It'''s gone, along with the id-comparison logic it existed for.

svozza:
Splitting this per @mrgrain's suggestion, error-masking fix first.

**#1845** — `NoStack` masking the real error. Independent of the stale read: it affected *every* failed deployment of a new stack, so a plain resource failure that rolled the stack back reported `NoStack` rather than naming the resource. Rewritten rather than patched, per the review; addresses the two `monitorDeployment` comments here.

**Stale `REVIEW_IN_PROGRESS` read** — follow-up PR, incorporating the six comments on this one. The main change is that `executingStackId` turned out to be unnecessary: `stabilizeStack` already accepts a `stackNameOrArn` and a stack ARN is the stack id, so passing the ARN gives identity-pinning for free. That deletes a parameter, a comparison, and a test.

I've replied on each thread above pointing to where it's handled. Will close this once #1845 is in, unless you'd rather I retarget it for the second half.

Thanks for the review, the ARN point in particular made that fix materially smaller.

svozza:
Both halves are now filed, so closing this in favour of them:

- **#1845** (merged) — `NoStack` masking the real error on any failed deployment of a new stack.
- **#1848** — the stale `REVIEW_IN_PROGRESS` read, incorporating the six comments on this PR.

Two things worth flagging from your review, @mrgrain:

The ARN point was right and made the fix materially smaller — `executingStackId` is gone entirely, along with the id-comparison logic it existed for, since `stabilizeStack` already takes a `stackNameOrArn` and a stack ARN *is* the stack id.

On the wall-clock question: rather than keep guessing at the retry constant, I ran the case against CloudFormation directly — a CREATE ChangeSet whose only resource fails immediately, executed, polling `DescribeStacks` at 0.5s. `REVIEW_IN_PROGRESS` appears exactly once, as the stack initial state, and is never re-entered, including on the fail-before-any-resource path I had assumed could return to it. So the retry budget now only covers the genuinely ambiguous leading-read case, and once an in-progress status has been seen a review read is known to be stale and needs no budget at all. Full event log is in #1848.

Thanks for the review — it caught two unbounded-polling bugs and a false-success on `cdk rollback` that I would otherwise have shipped.

PR #1805, opened 2026-08-09
fix(toolkit-lib): finalizing deployment after refactor intermittently fails while stacks are still in UPDATE_IN_PROGRESS

### LINKED ISSUE #1804: (refactor): finalizing deployment fails because stacks are still in UPDATE_IN_PROGRESS after the refactor completes
### Describe the bug

`cdk refactor --unstable=refactor` executes the stack refactor itself successfully, but the finalizing deployment that runs right after it ("Deploying updated stacks to finalize refactor...") frequently fails (intermittently; 2 out of 3 attempts in my environment) with:

```
❌  Refactor failed: ❌  MyStack2 failed: ValidationError: Stack:arn:aws:cloudformation:us-east-1:123456789012:stack/MyStack2/xxxx is in UPDATE_IN_PROGRESS state and can not be updated.
```

The refactor is applied correctly (resources are moved to their new locations); only the finalizing deployment fails.

The cause is a race condition between the CloudFormation stack refactor API and the finalizing deployment:

- CloudFormation marks the stack refactor's `ExecutionStatus` as `EXECUTE_COMPLETE` while the affected stacks themselves are still in `UPDATE_IN_PROGRESS` for a few more seconds (observed: 0.5-3 seconds between the last `IMPORT_COMPLETE` stack event and the stack becoming stable).
- The CLI only waits for the refactor status: `RefactoringContext.execute()` ends with `waitUntilStackRefactorExecuteComplete`, which only polls `DescribeStackRefactor`'s `ExecutionStatus` (https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/api/refactoring/context.ts#L102-L104).
- `Toolkit._refactor()` then immediately calls `this.deploy(...)` without any further waiting (https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts#L1674-L1678).

Since the assembly is already synthesized and the assets are already published at that point, the deployment reaches `CreateChangeSet` within ~1-2 seconds, while the stack may still be in `UPDATE_IN_PROGRESS`, and CloudFormation rejects it with the `ValidationError` above. Whether it fails depends on where the waiter's 6-second polling of `DescribeStackRefactor` lands relative to this window: when the lag is 2-3 seconds the deployment frequently loses the race (2 out of 3 attempts in my environment), and when the lag is sub-second it can succeed, which makes the failure intermittent but frequent.

I reproduced this with CLI 2.1129.0, but the code path is unchanged on current `main` and in the latest release (2.1135.1), so the latest version is affected as well.

### Regression Issue

_No response_

### Expected Behavior

After the stack refactor reaches `EXECUTE_COMPLETE`, the CLI should wait for the affected stacks to stabilize (leave `UPDATE_IN_PROGRESS`) before starting the finalizing deployment, so that the whole `cdk refactor` command completes successfully.

### Current Behavior

The refactor itself succeeds, and then the finalizing deployment fails:

```
Refactoring...
✅  Stack refactor complete
Deploying updated stacks to finalize refactor...
...
❌  Refactor failed: ❌  MyStack2 failed: ValidationError: Stack:arn:aws:cloudformation:us-east-1:123456789012:stack/MyStack2/xxxx is in UPDATE_IN_PROGRESS state and can not be updated.
```

The command exits with the error even though the resource moves were applied successfully.

### Reproduction Steps

1. Deploy a stack with a couple of resources, e.g.:

```ts
export class MyStack extends Stack {
  constructor(scope: Construct, id: string, props?: StackProps) {
    super(scope, id, props);

    new Queue(this, 'Queue');
    new Topic(this, 'Topic');
  }
}
```

2. Move the resources into a different construct scope (a refactor within the same stack), e.g.:

```ts
class Messaging extends Construct {
  constructor(scope: Construct, id: string) {
    super(scope, id);

    new Queue(this, 'Queue');
    new Topic(this, 'Topic');
  }
}

export class MyStack extends Stack {
  constructor(scope: Construct, id: string, props?: StackProps) {
    super(scope, id, props);

    new Messaging(this, 'Messaging');
  }
}
```

3. Run `npx cdk refactor --unstable=refactor` and confirm the prompt.

The refactor completes, then the finalizing deployment fails with the `ValidationError` above. In my environment this reproduces 2 out of 3 times. It is timing-dependent (a race condition), so it may occasionally succeed.

### Possible Solution

In `RefactoringContext.execute()`, after `waitUntilStackRefactorExecuteComplete`, wait for all stacks involved in the mappings (source and destination stacks) to stabilize before returning. The existing `stabilizeStack` helper in `packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts` (which polls `DescribeStacks` until the stack leaves `*_IN_PROGRESS`) can be reused for this. Placing the wait inside `execute()` (rather than in `Toolkit._refactor()`) also protects any other caller of `execute()`.

I will submit a PR for this.

_No response_

### CDK CLI Version

2.1129.0

### Framework Version

2.261.0

### Node.js Version

24.19.0

### OS

macOS 15

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

go-to-k:
Fixes #1804

### Problem

CloudFormation marks a stack refactor's `ExecutionStatus` as `EXECUTE_COMPLETE` while the affected stacks may still be in `UPDATE_IN_PROGRESS` for a few more seconds (observed: 0.5-3 seconds). `RefactoringContext.execute()` only waits for the refactor status (`waitUntilStackRefactorExecuteComplete`), and `Toolkit._refactor()` then immediately starts the finalizing deployment. Since synthesis is done and the assets are already published at that point, the deployment reaches `CreateChangeSet` within ~1-2 seconds and frequently fails with:

```
ValidationError: Stack:arn:... is in UPDATE_IN_PROGRESS state and can not be updated.
```

Whether it fails depends on where the waiter's 6-second polling of `DescribeStackRefactor` lands relative to this window, so the failure is intermittent but frequent (2 out of 3 attempts in the reporter's environment).

### Fix

After `waitUntilStackRefactorExecuteComplete`, `RefactoringContext.execute()` now waits for every stack involved in the mappings (source and destination) to stabilize, reusing the existing `stabilizeStack` helper from `api/deployments/cfn-api`. The wait is placed inside `execute()` rather than in `Toolkit._refactor()` so that any caller of `execute()` is protected.

### Testing

- New unit tests in `test/api/refactoring/context-execute.test.ts` (fake timers, same style as `cfn-api-stabilization-polling-interval.test.ts`): `execute()` does not resolve while the stack is `UPDATE_IN_PROGRESS` and resolves only after it reaches `UPDATE_COMPLETE`; in a cross-stack refactor both source and destination stacks are waited on.
- Verified against a real AWS environment: with this fix, in one of the runs the CLI observed the stack still in `UPDATE_IN_PROGRESS (Refactor ... initiated)` right after the refactor waiter returned, waited for it to stabilize, and the finalizing deployment then succeeded:

```
[03:17:33] Waiting for stack RefactorReproStack to finish creating or updating...
[03:17:33] Stack RefactorReproStack has an ongoing operation in progress and is not stable (UPDATE_IN_PROGRESS (Refactor 7689cad2-... initiated))
✅  Stack refactor complete
Deploying updated stacks to finalize refactor...
✅  RefactorReproStack (no changes)
```

This is exactly the timing where the current CLI fails.

---


packages/@aws-cdk/toolkit-lib/lib/api/refactoring/context.ts
```diff
@@ -102,6 +103,14 @@ export class RefactoringContext {
     await cfn.waitUntilStackRefactorExecuteComplete({
       StackRefactorId: refactor.StackRefactorId,
     });
+
+    // The refactor reaches EXECUTE_COMPLETE while the affected stacks may still
+    // be in UPDATE_IN_PROGRESS for a few more seconds. Wait for them to
+    // stabilize, so that callers can immediately start another stack operation.
+    const stackNames = [...new Set(mappings.flatMap((m) => [m.source.stack.stackName, m.destination.stack.stackName]))];
```
mrgrain: can we use the stack arn here?

mrgrain (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/refactoring/context.ts
go-to-k: yes, changed it.

mrgrain (approved):

mrgrain:
@go-to-k this fails on merge. can you update your branch and check the tests?

mrgrain (changes requested):
@go-to-k this fails on merge. can you update your branch and check the tests?

go-to-k:
@mrgrain Fixed it!

mrgrain (approved):

go-to-k:
Could you please re-run the failed integ job? `aws-cdk-lib:jar:2.266.0` was not yet on Maven Central at the time of the run (it is now).

PR #1818, opened 2026-08-11
fix(toolkit-lib): every change-set deployment announces "waiting in review for manual execution (--no-execute)" although the change set is executed

### LINKED ISSUE #1815: (deploy): every change-set deployment prints "Changeset ... created and waiting in review for manual execution (--no-execute)" although the change set is executed
### Describe the bug

Since #1273, every `cdk deploy` that uses the default `change-set` method and has actual changes prints:

```
MyStack: creating CloudFormation changeset...
Changeset arn:aws:cloudformation:...:changeSet/cdk-deploy-change-set/... created and waiting in review for manual execution (--no-execute)
...
✅  MyStack
```

The message claims the change set is waiting for manual execution, but the change set **is** executed right after (the deployment completes and `describe-change-set` shows `ExecutionStatus: EXECUTE_COMPLETE`). `--no-execute` was never requested. Display-only issue; the deployment itself works.

Cause:

- Since #1273, the deploy action always creates the change set upfront to get an accurate diff for the approval prompt: it calls `Deployments.prepareStack()`, which internally forces `execute: false`, and then executes the prepared change set in a second phase (https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts — "When using change-set method, always create the change set upfront").
- `deploy-stack.ts` prints the "waiting in review for manual execution (--no-execute)" message whenever `execute` is `false` (https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts#L486-L488). It cannot distinguish the internal first phase of an executing deployment from a user-requested `--no-execute`, so the message is printed for every change-set deployment with changes.

This also surfaces in flows that deploy internally, e.g. the finalizing deployment after `cdk import` ("Finish with a cdk deploy now?"), which is how I originally noticed it.

#1273 first shipped in `aws-cdk` 2.1118.1. I reproduced the message with a plain `cdk deploy` (a simple stack update, no flags besides `--require-approval never`) on 2.1135.0.

### Regression Issue

2.1118.0

### Expected Behavior

The "waiting in review for manual execution (--no-execute)" message is only printed when the created change set is actually left unexecuted, i.e. when the user asked for it (`--method=prepare-change-set` / `--no-execute` variants, including `cdk import --no-execute`).

### Current Behavior

The message is printed on every change-set deployment that has changes, and then the change set is executed anyway.

### Reproduction Steps

1. Deploy any stack.
2. Change something in the stack and run `cdk deploy` again (default change-set method).
3. Observe the `Changeset ... created and waiting in review for manual execution (--no-execute)` line, followed by the deployment executing and succeeding.

### Possible Solution

Only the caller knows whether `execute: false` is user-requested or the internal first phase of an executing deployment. Move the announcement out of `deploy-stack.ts` to the places where that intent is known: `Deployments.prepareStack()` (which already receives this distinction via `cleanupOnNoOp`) and the resource importer (for `cdk import --no-execute`).

I will submit a PR for this.

_No response_

### CDK CLI Version

2.1135.0

### Framework Version

2.261.0

### Node.js Version

24.19.0

### OS

macOS 15

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

go-to-k:
Fixes #1815

### Problem

Since #1273, the deploy action always creates the change set upfront (to get an accurate diff for the approval prompt) by calling `Deployments.prepareStack()`, which internally forces `execute: false`, and then executes the prepared change set in a second phase. `deploy-stack.ts` prints

```
Changeset ... created and waiting in review for manual execution (--no-execute)
```

whenever `execute` is `false`, so every change-set deployment with actual changes now prints this message and then executes the change set anyway. The message is misleading: `--no-execute` was never requested, and the change set does not wait for anything. This also surfaces in flows that deploy internally, e.g. the finalizing deployment after `cdk import`.

### Fix

`deploy-stack.ts` cannot know whether `execute: false` is user-requested or the internal first phase of an executing deployment, so the caller now tells it, via a single intent-revealing option: `willExecuteChangeSet` (default `false`: the change set is the user's final artifact). Both behaviors that depend on this intent are derived from the one flag:

- `deploy-stack.ts` announces a created-but-not-executed change set as waiting for manual execution only when `willExecuteChangeSet` is not set (this is the fix);
- `Deployments.prepareStack()` cleans up an empty change set only when `willExecuteChangeSet` is set (this replaces the former `cleanupOnNoOp` option, which encoded the same fact under a behavior-specific name).

The deploy flows (the toolkit deploy action and the CLI's deploy flow, which share `prepareStack`) pass `willExecuteChangeSet: isExecutingChangeSetDeployment(deploymentMethod)`; `prepareStack` passes the flag through to `deployStack` unchanged. The default preserves today's behavior for every other caller, including `cdk import --no-execute`, which keeps its announcement without any change to the importer.

The option lives on the internal `DeployStackOptions` interfaces (`api/deployments` is not part of the toolkit-lib public API), so there is no public API change.

### Alternative considered

An alternative implementation was initially committed (e2c6083): removing the announcement from `deploy-stack.ts` entirely and emitting it from the callers that know the user's intent (`prepareStack` and the resource importer). Its appeal is that it structurally removes `deploy-stack`'s ability to assert a user intent it cannot know.

It was replaced with the current approach because:

- the announcement stays at a single emission site, at the moment the change set is created;
- the importer is left untouched (`cdk import --no-execute` keeps its announcement via the default, instead of re-implementing it);
- the default preserves today's behavior for every caller of `deployStack` with `execute: false`.

Both directions are kept in the branch history.

### Testing

- New unit tests in `test/api/deployments/deploy-stack.test.ts`: with `execute: false`, the change set is announced as awaiting manual execution by default, and not announced with `willExecuteChangeSet: true`.
- New unit tests in `test/api/deployments/cloudformation-deployments.test.ts`: `prepareStack` passes `willExecuteChangeSet` through to `deployStack` unchanged (set for the internal prepare of an executing deployment, unset for a user-requested `--no-execute` prepare).
- Verified against a real AWS environment:
  - CLI 2.1135.0 (released): a plain `cdk deploy` with changes prints the misleading message and then executes the change set.
  - This branch: the same deploys (both a stack creation and an update with changes) print no such message and execute normally; `cdk deploy --method=prepare-change-set` still prints the message and leaves the change set unexecuted (stack `UPDATE_COMPLETE` and unchanged, change set `CREATE_COMPLETE`/`AVAILABLE`).

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts
@@ -1,5 +1,4 @@
 import { randomUUID } from 'node:crypto';
-import { format } from 'node:util';
 import * as cdk_assets from '@aws-cdk/cdk-assets-lib';
 import type * as cxapi from '@aws-cdk/cloud-assembly-api';
 import chalk from 'chalk';
@@ -97,6 +96,19 @@ export interface DeployStackOptions {
    * @default - Change set with default options
    */
   readonly deploymentMethod?: DeploymentMethod;
+
+  /**
+   * Whether to announce a change set that is created but deliberately not
+   * executed (change-set method with `execute: false`) as waiting for manual
+   * execution.
+   *
+   * Set to false when the change set is the internal first phase of an
+   * executing deployment (two-phase deploy), where the announcement would be
+   * misleading: the change set is about to be executed.
+   *
+   * @default true
+   */
+  readonly announceNoExecuteChangeSet?: boolean;
 
   /**
    * Force deployment, even if the deployed template is identical to the one we are about to deploy.
@@ -417,6 +429,7 @@ export class Deployments {
       envResources: env.resources,
       tags: options.tags,
       deploymentMethod: options.deploymentMethod,
+      announceNoExecuteChangeSet: options.announceNoExecuteChangeSet,
       forceDeployment: options.forceDeployment,
       parameters: options.parameters,
       usePreviousParameters: options.usePreviousParameters,
@@ -450,6 +463,12 @@ export class Deployments {
     const result = await this.deployStack({
       ...options,
       deploymentMethod: { ...options.deploymentMethod, execute: false },
+      // Announce the created change set only when the user explicitly asked
+      // for --no-execute (the change set is the final result). When this
+      // prepare is the internal first phase of an executing deployment
+      // (cleanupOnNoOp), the change set is about to be executed and the
+      // announcement would be misleading.
+      announceNoExecuteChangeSet: !options.cleanupOnNoOp,
     });
 
     // With execute: false, the only possible result type is did-deploy-stack
@@ -464,14 +483,6 @@ export class Deployments {
     if (result.noOp && options.cleanupOnNoOp) {
       const changeSetName = options.deploymentMethod.changeSetName ?? DEFAULT_DEPLOY_CHANGE_SET_NAME;
       await this.cleanupChangeSet(options.stack, changeSetName, options.stackEventPollingInterval);
-    }
-
-    // Announce the created change set only when the user explicitly asked for
-    // --no-execute (the change set is the final result). When the caller forced
-    // execute: false internally (two-phase deploy), the change set is about to
-    // be executed and the announcement would be misleading.
-    if (!result.noOp && !options.cleanupOnNoOp) {
-      await announceChangeSetAwaitingExecution(this.ioHelper, result.changeSet?.ChangeSetId);
     }
 
     return result;
@@ -819,15 +830,4 @@ class ParallelSafeAssetProgress extends BasePublishProgressListener {
 
 function suffixWithErrors(msg: string, errors?: string[]) {
   return errors && errors.length > 0 ? `${msg}: ${errors.join(', ')}` : msg;
-}
-
-/**
- * Announce a change set that was deliberately created without being executed
- * (--no-execute) and is now waiting for manual execution.
- */
-export async function announceChangeSetAwaitingExecution(ioHelper: IoHelper, changeSetId: string | undefined): Promise<void> {
-  await ioHelper.defaults.info(format(
-    'Changeset %s created and waiting in review for manual execution (--no-execute)',
-    changeSetId,
-  ));
 }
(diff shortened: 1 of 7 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts
```diff
@@ -449,6 +463,12 @@ export class Deployments {
     const result = await this.deployStack({
       ...options,
       deploymentMethod: { ...options.deploymentMethod, execute: false },
+      // Announce the created change set only when the user explicitly asked
+      // for --no-execute (the change set is the final result). When this
+      // prepare is the internal first phase of an executing deployment
+      // (cleanupOnNoOp), the change set is about to be executed and the
+      // announcement would be misleading.
+      announceNoExecuteChangeSet: !options.cleanupOnNoOp,
```
9pace: If caller passed `announceNoExecuteChangeSet` here it would be ignored. Effectively the `options.announceNoExecuteChangeSet` is unused.

9pace (changes requested):
Thanks for the contribution! You have correctly identified an issue here. 

on `announceNoExecuteChangeSet`:

- `PrepareStackOptions` extends `Omit<DeployStackOptions, 'deploymentMethod'>`, so it inherits `announceNoExecuteChangeSet`, but `prepareStack()` overwrites it after the spread with `!options.cleanupOnNoOp`. A caller-supplied value is ignored.

- `cleanupOnNoOp` and `announceNoExecuteChangeSet` are two booleans derived from one underlying fact: whether the change set is the user's final artifact (`--no-execute`) or the first phase of a deployment that will execute it. The proposed change (Syncing them with an inversion at one call site) invites drift, and the announce flag is named for the message it controls rather than the intent behind it.

Suggestion: Rename `cleanupOnNoOp` to something intent-revealing (e.g. `willExecuteChangeSet`) and derive both behaviors (empty-change-set cleanup and announcement suppression) from it.

go-to-k:
Thanks for the review! Applied in d4e3f5fb: renamed `cleanupOnNoOp` to `willExecuteChangeSet` and derived both behaviors from that single flag.

9pace (commented):
Thanks for the quick turnaround, this is the shape I was hoping for. Looks good.

9pace (approved):

PR #1819, opened 2026-08-11
fix(cli): deployments skipped even though changes exist when previous deployment was with hotswap

ShadowCat567:
When making multiple hotswap deployments in a row, hotswap should compare between the template currently being deployed and the last hotswap template. This happens properly most of the time, however when a change is made that would cause the template being hotswapped to be the same as the template in CloudFormation, `canSkipDeploy` was not checking the cached hotswap template and did not register that a change could be deployed via hotswap. 
Updates the behavior of `canSkipDeploy` so it checks the hotswap cache if we are deploying with hotswap and does not skip the hotswap deployment if the cached hotswap template and the template currently being deployed differ. 
This also applies when performing a Cloudformation deployment after performing a hotswap deployment.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
@@ -40,7 +40,7 @@ import { changeSetHasNoChanges } from '../diagnosing/stack-diagnoser';
 import type { EnvironmentResources, StringWithoutPlaceholders } from '../environment';
 import { HotswapPropertyOverrides, ICON, createHotswapPropertyOverrides } from '../hotswap/common';
 import { tryHotswapDeployment } from '../hotswap/hotswap-deployments';
-import { invalidateHotswapTemplateCache } from '../hotswap/hotswap-template-cache';
+import { invalidateHotswapTemplateCache, readHotswapTemplateCache } from '../hotswap/hotswap-template-cache';
 import type { IoHelper } from '../io/private';
 import type { ResourcesToImport } from '../resource-import';
 import { StackActivityMonitor } from '../stack-events';
@@ -926,6 +926,18 @@ async function canSkipDeploy(
   if (cloudFormationStack.stackStatus.isFailure) {
     await ioHelper.defaults.debug(`${deployName}: stack is in a failure state`);
     return false;
+  }
+
+  if (deployStackOptions.deploymentMethod?.method === 'hotswap') {
+    const hotswapCache = await readHotswapTemplateCache(
+      deployStackOptions.stack.assembly.directory,
+      deployStackOptions.stack.stackName,
+      deployStackOptions.stack.template,
+    );
+    if (hotswapCache && JSON.stringify(deployStackOptions.stack.template) !== JSON.stringify(hotswapCache.deployedRootTemplate)) {
+      await ioHelper.defaults.debug(`${deployName}: template has changed in relation to last successful hotswap deployment`);
+      return false;
+    }
   }
 
   // We can skip deploy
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -927,6 +928,18 @@ async function canSkipDeploy(
     return false;
   }
 
+  if (deployStackOptions.deploymentMethod?.method === 'hotswap') {
+    const hotswapCache = await readHotswapTemplateCache(
+      deployStackOptions.stack.assembly.directory,
+      deployStackOptions.stack.stackName,
+      deployStackOptions.stack.template,
+    );
+    if (hotswapCache && JSON.stringify(deployStackOptions.stack.template) !== JSON.stringify(hotswapCache.deployedRootTemplate)) {
```
iliapolo: Use [`diffTemplate`](https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/cloudformation-diff/lib/diff-template.ts#L92) to avoid any ordering problems (and possible CFN quirks that don't represent a real difference).

iliapolo (changes requested):

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
ShadowCat567: Updated and added a test for when there is no change between hotswap templates

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
```diff
@@ -927,6 +929,18 @@ async function canSkipDeploy(
     return false;
   }
 
+  if (deployStackOptions.deploymentMethod?.method === 'hotswap') {
```
iliapolo: Gating on `hotswap` will cause:

1. Make some changes.
2. Deploy with hotswap.
3. Revert the changes.
4. Deploy regularly.

(4) will be skipped because it won't enter this block and the deployed template will be the same as the currently generated one. We therefore shouldn't gate.

packages/@aws-cdk/toolkit-lib/test/api/deployments/deploy-stack.test.ts
```diff
@@ -676,6 +677,52 @@ test('deploy is not skipped if notificationArns are different', async () => {
   expect(mockCloudFormationClient).toHaveReceivedCommand(CreateChangeSetCommand);
 });
 
+test('deploy is not skipped if we are deploying with hotswap and there is a change from the last hotswap deployment', async () => {
```
iliapolo:
```suggestion
test('deploy is not skipped if there is a change from the last hotswap deployment', async () => {
```

packages/@aws-cdk/toolkit-lib/test/api/deployments/deploy-stack.test.ts
```diff
@@ -676,6 +677,52 @@ test('deploy is not skipped if notificationArns are different', async () => {
...
+  givenTemplateIs(defaultTargetTemplate());
+  (readHotswapTemplateCache as jest.Mock).mockResolvedValue({
+    deployedRootTemplate: { Resources: { Different: { Type: 'Test::Resource::Type' } } },
+    nestedStacks: {},
+  });
+
+  // WHEN
+  await testDeployStack({
+    ...standardDeployStackArguments(),
+    deploymentMethod: { method: 'hotswap' },
```
iliapolo: No needed now.

iliapolo (changes requested):

iliapolo (approved):

PR #1821, opened 2026-08-12
fix(toolkit-lib): clean up the diff change set when early validation fails

### LINKED ISSUE #1767: (cli): changeset fails to delete when pre-deployment validation fails
### Describe the bug

1. cdk diff a new stack with a bucket, but the bucket already exists
2. changeset created, fails with Resource name conflict (expected)
3. no cli output at all - have to go to the console to view or go into verbose mode
4. changeset doesn't get deleted, stack stays in "REVIEW IN PROGRESS" state
5. cdk diff again, doesn't delete the existing changeset and can't do a proper diff (has to use template) 
6. the only way to delete the changeset is to go into the console and delete

### Regression Issue

_No response_

### Expected Behavior

cdk diff deletes the changeset and stack doesn't permanently stay in REVIEW sgtate

### Current Behavior

first diff runs with: 
```[20:50:42] Changeset cdk-diff-change-set on stack mystackis still creating
[20:50:47] DeploymentError: Early validation failed for change set cdk-diff-change-set:
mystack/bucket/Resource  (AWS::S3::Bucket mybucket)
  Resource of type 'AWS::S3::Bucket' with identifier 'mybucket' already exists. (at
  /Resources/mybucket)
Source Location: ...aws-cdk-lib, jsii runtime, node internals...
                 (no user code in 10 frames, use --stack-trace-limit to capture more)
Could not create a change set, will base the diff on template differences (run again with -v to see the reason)```

second diff:
```
[20:55:12] Attempting to create ChangeSet with name cdk-diff-change-set for stack mystack
[20:55:12] AlreadyExistsException: ChangeSet cdk-diff-change-set cannot be created due to a mismatch with existing attribute ClientToken
Could not create a change set, will base the diff on template differences (run again with -v to see the reason)
```


### Reproduction Steps

1. `cdk diff mystack` with a bucket in the construct with the name that already exists on a bucket somewhere 
2. cfn validation fails and stays in review state
3. `cdk diff mystack` again

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.1133.0

### Framework Version

_No response_

### Node.js Version

26.5.0

### OS

macos tahoe 26.5.2

### Language

Python

### Language Version

_No response_

### Other information

would also like to add that i don't see any flag that disables this pre-deployment check

### PR DESCRIPTION
**Pull request by lemon0333**
Fixes #1767

When a `cdk diff` change set fails early validation (e.g. a resource that already exists), `createChangeSetAndCleanup` in `toolkit-lib`'s `cfn-api.ts` threw from `waitAndThrowOnProblem` before the change-set and empty-stack cleanup could run. That orphaned the change set and left a new stack stuck in `REVIEW_IN_PROGRESS`, which then blocked subsequent change-set creation and forced the user to delete it manually in the console.

This change runs the cleanup on the failure path too — deleting the change set and, for a brand new stack, the empty review stack — as a best-effort step that does not mask the original validation error. Successful diffs are unchanged.

Verified with a new unit test in `diff.test.ts` (change set fails validation → `DeleteChangeSet` and `DeleteStack` are called); the full `diff.test.ts` suite passes (22/22).

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
@@ -227,27 +227,9 @@ async function createChangeSetAndCleanup(
   const changeSetId = changeSet.Id ?? options.changeSetName;
   const stackId = changeSet.StackId ?? options.stack.stackName;
 
-  // Remove the change set (and, for a brand new stack, the empty stack that a
-  // CREATE change set leaves in REVIEW_IN_PROGRESS). This has to run whether the
-  // change set succeeds or fails validation: otherwise a change set that fails
-  // early validation is orphaned and leaves the stack stuck in REVIEW_IN_PROGRESS,
-  // which then blocks subsequent change set creation.
-  const cleanup = async () => {
-    await cleanupOldChangeset(options.cfn, ioHelper, changeSetId, stackId);
-
-    if (!options.exists) {
-      await ioHelper.defaults.debug(format('Deleting empty stack created by diff changeset: %s', stackId));
-      await options.cfn.deleteStack({
-        StackName: stackId,
-        ClientRequestToken: randomUUID(),
-      });
-    }
-  };
-
-  let createdChangeSet: ChangeSetReport;
   try {
     // Fetching all pages if we'll execute, so we can have the correct change count when monitoring.
-    createdChangeSet = await new ChangeSetDescriber({
+    return await new ChangeSetDescriber({
       cfn: options.cfn,
       ioHelper,
       stackNameOrArn: stackId,
@@ -260,22 +242,24 @@ async function createChangeSetAndCleanup(
     // otherwise consumes this error at debug level before falling back to a
     // template-only diff, so without this the reason is invisible unless -v is set.
     await ioHelper.defaults.warn(format('Change set %s failed: %s', changeSetId, e));
-
-    // Best-effort cleanup so a failed change set doesn't leak (which would leave a
-    // new stack stuck in REVIEW_IN_PROGRESS). Deletion can legitimately fail -- e.g.
-    // missing cloudformation:DeleteChangeSet/DeleteStack permissions -- so we log
-    // that but must not let it mask the original failure surfaced above.
-    try {
-      await cleanup();
-    } catch (cleanupError) {
-      await ioHelper.defaults.warn(format('Failed to clean up change set %s after a creation error: %s', changeSetId, cleanupError));
-    }
     throw e;
+  } finally {
+    // Always clean up the change set (and, for a brand new stack, the empty stack a
+    // CREATE change set leaves in REVIEW_IN_PROGRESS) whether or not creation
+    // succeeded. Otherwise a change set that fails early validation is orphaned and
+    // leaves the stack stuck in REVIEW_IN_PROGRESS, which blocks subsequent change
+    // set creation. The failure reason has already been surfaced above, so a cleanup
+    // error can surface on its own.
+    await cleanupOldChangeset(options.cfn, ioHelper, changeSetId, stackId);
+
+    if (!options.exists) {
+      await ioHelper.defaults.debug(format('Deleting empty stack created by diff changeset: %s', stackId));
+      await options.cfn.deleteStack({
+        StackName: stackId,
+        ClientRequestToken: randomUUID(),
+      });
+    }
   }
-
-  await cleanup();
-
-  return createdChangeSet;
 }
 
 /**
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts [hunk 1]
```diff
@@ -223,33 +223,51 @@ async function createChangeSetAndCleanup(
...
+      cfn: options.cfn,
+      ioHelper,
+      stackNameOrArn: stackId,
+      changeSetNameOrArn: changeSetId,
+    }).waitAndThrowOnProblem({
+      diagnoser: options.diagnoser,
     });
+  } catch (e) {
+    // Best-effort cleanup so a failed change set doesn't leak; don't let a
+    // cleanup failure mask the original creation/validation error.
```
ShadowCat567: Can you log the error we caught here? Currently the error that gets thrown will still not be visible to the user because it gets consumed here: https://github.com/aws/aws-cdk-cli/blob/536ad69012bdc5d2fa3fdfe6418e7036a3909082/packages/%40aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts#L74-L92 (the actual error is only revealed if you use the `--debug` flag)

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -223,33 +223,51 @@ async function createChangeSetAndCleanup(
...
+    try {
+      await cleanup();
+    } catch (cleanupError) {
+      await ioHelper.defaults.debug(format('Failed to clean up change set after a creation error: %s', cleanupError));
+    }
```
ShadowCat567, line 261: having this try/catch may be overly defensive, I am not necessarily against it but why are we expecting that deletion might fail?

ShadowCat567 (commented):
Hello @lemon0333! Thanks for the PR :D it looks good overall! I just have a few comments that can help improve it

lemon0333:
Thanks for the review @ShadowCat567! Addressed both:

1. **Visibility** — the caught error is now `warn`-logged (`Change set <id> failed: <reason>`) so the reason is visible without `-v`, since `createDiffChangeSet` otherwise only consumes it at `debug` before falling back to the template diff.

2. **The try/catch around cleanup** — it's there so that a cleanup failure doesn't *mask* the original validation error (the useful one we want to rethrow). Deletion can legitimately fail — e.g. a role missing `cloudformation:DeleteChangeSet`/`DeleteStack` — and in that case we now `warn`-log the cleanup failure but still rethrow the original error. I added a comment making that rationale explicit; happy to simplify further if you'd prefer.

Also added a test assertion that the failure reason is surfaced. Full `diff.test.ts` passes (22/22).

packages/@aws-cdk/toolkit-lib/lib/api/deployments/cfn-api.ts
```diff
@@ -223,33 +223,58 @@ async function createChangeSetAndCleanup(
...
 
-  // If the stack didn't exist before, creating a CREATE changeset will have
-  // put it in REVIEW_IN_PROGRESS state. Delete the empty stack to clean up.
-  if (!options.exists) {
-    await ioHelper.defaults.debug(format('Deleting empty stack created by diff changeset: %s', changeSet.StackId ?? options.stack.stackName));
-    await options.cfn.deleteStack({
-      StackName: changeSet.StackId ?? options.stack.stackName,
-      ClientRequestToken: randomUUID(),
+  let createdChangeSet: ChangeSetReport;
+  try {
```
ShadowCat567, line 230: I spent some more time thinking about this, I think the cleanest way to do this might be using a `try/catch/finally`:
```
try {
  return await new ChangeSetDescriber({...});
} catch (e) {
  await io.Helper.defaults.warn(...); // print the error
  throw e;
} finally { // ensures we always clean up whether we ran into an error or not
  // do clean up, surface errors if any appear
}
```
this means we are not calling the `cleanup` function in 2 places (and we probably don't need it to begin with). Let me know what you think!

ShadowCat567 (commented):
Hello @lemon0333! Put some more thought into the `try/catches`, let me know what you think of my proposal

lemon0333:
Nice, that's cleaner — done in the latest commit. Moved the cleanup into a `finally` so it runs once whether or not the change set succeeds, kept the `warn` in `catch` (which runs before `finally`, so the failure reason is surfaced even if cleanup itself errors), and dropped the extra defensive try/catch. `diff.test.ts` still passes (22/22). Thanks for the suggestion! 🙏

ShadowCat567 (approved):
Nice job @lemon0333!

Issue #1823, opened 2026-08-12
toolkit-lib: cdk refactor deep-clones every resource to drop one metadata key (~25% of the digest phase)


Adityaj0:
### Describe the bug

`cdk refactor` spends roughly a quarter of its digest phase deep-cloning resources in order to delete a single metadata key.

`stripConstructPath` removes `Metadata['aws:cdk:path']` before a resource is hashed, and does it by round-tripping the entire resource through `JSON.stringify` + `JSON.parse`:

[`toolkit-lib/lib/api/refactoring/digest.ts#L141-L149`](https://github.com/aws/aws-cdk-cli/blob/main/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/digest.ts#L141-L149)

```ts
function stripConstructPath(resource: any): any {
  if (resource?.Metadata?.['aws:cdk:path'] == null) {
    return resource;
  }

  const copy = JSON.parse(JSON.stringify(resource));
  delete copy.Metadata['aws:cdk:path'];
  return copy;
}
```

The cost scales with the size of the resource, not with the thing being removed. Every CDK-generated resource carries `aws:cdk:path`, so the clone happens for essentially every resource — including the large ones (inline policy documents, state machine definitions, inline Lambda code), where a full serialize-and-parse is most expensive.

The clone is also unnecessary. The result is only read: it is handed straight to `stripReferences`, which builds its own new tree anyway. Nothing downstream mutates it, and the only key being removed is one level down.

A `RefactoringContext` calls `computeResourceDigests` four times for one environment — the deployed and the local stacks, each in both graph directions — so each resource is cloned four times per refactor.

### Expected Behavior

Removing one metadata key should cost the same regardless of how large the resource is. A shallow copy of the resource and of its `Metadata` map achieves exactly the same result, still leaves the caller's template untouched, and produces byte-identical digests.

### Current Behavior

Synthetic CDK-shaped templates (nested policy documents, tags, `Ref`/`Fn::GetAtt` cross-references, `DependsOn`, and `aws:cdk:path` on every resource), timing the four `computeResourceDigests` calls that one `RefactoringContext` performs. Cold node process per measurement, min of 7 runs:

| template shape | JSON size per side | `main` | with a shallow copy | improvement |
|---|---|---|---|---|
| 1 stack × 200 resources | 0.6 MiB | 50.0 ms | 38.1 ms | **23.8%** |
[...]

PR #1827, opened 2026-08-12
refactor(toolkit-lib): compute each resource's property hash once per refactor

### LINKED ISSUE #1824: toolkit-lib: cdk refactor hashes every resource's properties four times (~35% of the digest phase)
### Describe the bug

`cdk refactor` hashes every resource's properties four times per environment, though the result is the same every time.

One `RefactoringContext` computes digests in four passes:

[`toolkit-lib/lib/api/refactoring/context.ts#L47-L48`](https://github.com/aws/aws-cdk-cli/blob/main/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/context.ts#L47-L48)

```ts
const moves = resourceMoves(props.deployedStacks, props.localStacks, 'direct', props.ignoreModifications);
const additionalOverrides = structuralOverrides(props.deployedStacks, props.localStacks);
```

`resourceMoves` digests both the deployed and the local stacks, and `structuralOverrides` calls it again for the `'opposite'` graph direction — four `computeResourceDigests` calls over the same two sets of resources.

A digest is built from two inputs:

[`digest.ts#L59-L65`](https://github.com/aws/aws-cdk-cli/blob/main/packages/@aws-cdk/toolkit-lib/lib/api/refactoring/digest.ts#L59-L65)

```ts
const depDigests = Array.from(graph.outNeighbors(id)).map((d) => result[d]);
const propertiesHash = hashObject(stripReferences(stripConstructPath(resource), exports));
const toHash = resource.Type + propertiesHash + depDigests.join('');
```

Only `depDigests` depends on the graph direction. `propertiesHash` — which is where nearly all the CPU goes, since it walks and hashes the resource's entire property tree — is a pure function of the resource and the stacks' exports, and is byte-identical in the `'direct'` and `'opposite'` passes. It is nevertheless recomputed from scratch in each one, so every resource is stripped, copied and SHA-256'd twice per stack set.

### Expected Behavior

The direction-independent half of the digest should be computed once per resource and reused by the passes that need it. The digests themselves must not change — only how often they are computed.

### Current Behavior

Synthetic CDK-shaped templates (nested policy documents, tags, `Ref`/`Fn::GetAtt` cross-references, `DependsOn`, `aws:cdk:path` on every resource), timing exactly the four `computeResourceDigests` calls a `RefactoringContext` performs. Cold node process per measurement, min of 7 runs:

| template shape | JSON size per side | `main` | sharing property hashes | improvement |
|---|---|---|---|---|
| 1 stack × 200 resources | 0.6 MiB | 48.6 ms | 32.6 ms | **32.9%** |
| 5 stacks × 200 resources | 3.0 MiB | 238.2 ms | 156.4 ms | **34.4%** |
| 10 stacks × 400 resources | 22.2 MiB | 1542.1 ms | 961.6 ms | **37.6%** |

The measurement is a single cold execution, so the cache starts empty exactly as it does in a real CLI invocation: two of the four passes populate it and two read from it. (Timing repeated iterations in one process inflates the number to ~80%, because the cache is then already warm — that figure would not be honest.)

### Reproduction Steps

```ts
import { computeResourceDigests } from '@aws-cdk/toolkit-lib/lib/api/refactoring/digest';

// deployed and local: CloudFormationStack[] built from CDK-shaped templates
const t0 = process.hrtime.bigint();
computeResourceDigests(deployed, 'direct');
computeResourceDigests(local, 'direct');
computeResourceDigests(deployed, 'opposite');   // re-hashes every deployed resource
computeResourceDigests(local, 'opposite');      // re-hashes every local resource
console.log(Number(process.hrtime.bigint() - t0) / 1e6, 'ms');
```

A CPU profile shows `hashObject` and `stripReferences` accounting for the bulk of all four passes, with the third and fourth repeating the work of the first and second.

### Possible Solution

Give `computeResourceDigests` an optional cache of property hashes keyed by the resource object, and let `RefactoringContext` own one for the duration of the operation:

```ts
export type PropertyHashCache = Map<CloudFormationResource, string>;

export function computeResourceDigests(
  stacks: CloudFormationStack[],
  direction: GraphDirection = 'direct',
  propertyHashes: PropertyHashCache = new Map(),
): Record<string, string> { ... }
```

```ts
// in computeDigestsInTopologicalOrder
let propertiesHash = propertyHashes.get(resource);
if (propertiesHash == null) {
  propertiesHash = hashObject(stripReferences(stripConstructPath(resource), exports));
  propertyHashes.set(resource, propertiesHash);
}
```

with the cache created in the `RefactoringContext` constructor and threaded through `resourceMoves` / `structuralOverrides` / `resourceDigests`.

Scoping the cache to one operation rather than holding it in module state matters: keys are the resource objects themselves, so a longer-lived cache could outlive the templates it was built from and go stale if a caller mutated one. Defaulting the parameter to a fresh `Map` keeps every existing caller's behaviour unchanged.

Related but separate: `stripConstructPath` deep-clones each resource via `JSON.parse(JSON.stringify(...))`, which is another ~20-25% of the same code path. Filed separately so the two can be reviewed independently; they compose.

I have this change with a regression test asserting that a shared cache produces the same digests as an unshared one in both directions; all 55 tests under `test/api/refactoring/` pass.

### CDK CLI Version

main (`536ad69`)

### Node.js Version

v24.1.0

### OS

macOS 15 (Darwin 25.2.0)

### Language

TypeScript

Adityaj0:
Fixes #1824

### Reason for this change

A `RefactoringContext` computes digests four times for one environment: `resourceMoves` handles the deployed and the local stacks in the `'direct'` direction, then `structuralOverrides` repeats both in the `'opposite'` direction.

Only the dependency half of a digest depends on the direction. The other input — `hashObject(stripReferences(stripConstructPath(resource), exports))`, which is where nearly all the CPU goes because it walks and hashes the resource's whole property tree — is a pure function of the resource and the stacks' exports, and is byte-identical in every pass. It is recomputed from scratch in each of them.

### Description of changes

Added an optional `PropertyHashCache` (a `Map` keyed by the resource object) to `computeResourceDigests`, consulted before hashing and populated after. `RefactoringContext` creates one in its constructor and threads it through `resourceMoves`, `structuralOverrides` and `resourceDigests`, so all four passes share it.

Two deliberate choices:

- **The cache is scoped to one operation, not module state.** Keys are the resource objects themselves, so a longer-lived cache could outlive the templates it was built from and go stale if a caller mutated one. Owning it in the constructor bounds its lifetime to the refactor.
- **The parameter defaults to a fresh `Map`.** Every existing caller of `computeResourceDigests` — including the tests — behaves exactly as before without changes.

Digests are unchanged; only how often they are computed changes.

- New regression test asserting that digests computed with a cache shared across both graph directions are identical to digests computed without one, for both directions.
- All 55 tests under `test/api/refactoring/` pass.
- Benchmark — synthetic CDK-shaped templates (nested policy documents, tags, `Ref`/`Fn::GetAtt` cross-references, `DependsOn`, `aws:cdk:path` on every resource), timing exactly the four `computeResourceDigests` calls a `RefactoringContext` performs. Cold node process per measurement, min of 7 runs:

  | template shape | JSON size per side | before | after | improvement |
  |---|---|---|---|---|
  | 1 stack × 200 resources | 0.6 MiB | 48.6 ms | 32.6 ms | 32.9% |
  | 5 stacks × 200 resources | 3.0 MiB | 238.2 ms | 156.4 ms | 34.4% |
  | 10 stacks × 400 resources | 22.2 MiB | 1542.1 ms | 961.6 ms | 37.6% |

  Each measurement is a single cold execution, so the cache starts empty exactly as in a real CLI invocation: two passes fill it, two read from it. Timing repeated iterations inside one process would report ~80%, but only because the cache is already warm by then — that number would not reflect what a user sees.

Note: #1823 covers the deep clone in `stripConstructPath`, a separate ~20-25% in the same code path. The two changes are independent and compose.

### Checklist

---


mrgrain (changes requested):
@Adityaj0 thanks for the fix, can you resolve the merge conflicts?

otaviomacedo (approved):

Adityaj0:
@mrgrain done.

mrgrain (approved):

PR #1831, opened 2026-08-14
fix(toolkit-lib): hotswap silently drops empty-string Lambda Description

### LINKED ISSUE #1830: (hotswap): clearing a Lambda's Description via hotswap silently does nothing
### Describe the bug

`cdk deploy --hotswap` (or `--hotswap-fallback`) can silently fail to apply a Lambda `Description` change when the new value is an empty string, while reporting the deployment as successful (or as a no-op) — the Lambda's actual `Description` in AWS is never updated.

### Root cause

`evaluateLambdaFunctionProps` in `packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts`:

```ts
const configurations = description || environment ? { description, environment } : undefined;
return code || configurations ? { code, configurations } : undefined;
```

`description` is `string | undefined`, populated only when `Description` was actually part of the classified diff (see the `case 'Description':` branch above). An empty string `""` is a legitimate, meaningful new value — but `description || environment` treats it identically to "Description didn't change" (`undefined`), since `""` is falsy in JS.

If the only classified hotswappable change is `Description` (no `Code`/`Environment` change), `configurations` becomes `undefined`, `evaluateLambdaFunctionProps` returns `undefined`, and the caller (`isHotswappableLambdaFunctionChange`, line ~51) takes the "nothing to do here" branch and returns without ever calling `UpdateFunctionConfiguration`. Because the resource change is dropped entirely (not classified as non-hotswappable, not classified as hotswappable-with-a-no-op — just absent), `hotswappableChanges` in `tryHotswapDeployment` can end up empty, which is reported as `noOp: true` — i.e. CDK tells the user there was nothing to hotswap, when in fact there was a real, un-applied template change.

### Reproduction

1. Deploy a stack with a Lambda function that has `Description: "some text"`.
2. Change the template to `Description: ""` (no other Code/Environment changes).
3. Run `cdk deploy --hotswap`.

**Expected:** the Lambda's `Description` in AWS is cleared to `""`.
**Actual:** `cdk deploy --hotswap` reports success (or "no changes to hotswap"), but the Lambda's `Description` in AWS is left unchanged at `"some text"` — silent infrastructure drift between the template and the deployed resource, with no error or warning.

### Fix

PR incoming — checks `!== undefined` instead of relying on truthiness, matching how the values are actually populated (only set when that property was part of the diff).

### CDK CLI Version

Reproduced by reading current `main` branch source; not version-specific — the logic has been truthy-checked since this code path was introduced.

N/A — logic bug, not environment-specific

Adityaj0:
fixes #1830

### Reason for this change

`evaluateLambdaFunctionProps` (`packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts`) decided whether a hotswappable Lambda config change existed with:

```ts
const configurations = description || environment ? { description, environment } : undefined;
return code || configurations ? { code, configurations } : undefined;
```

`description` is `string | undefined`, set only when `Description` was actually part of the classified property diff. An empty string `""` is a real, meaningful new value (e.g. a user clearing a Lambda's description) — but `||` can't distinguish it from "not set," since `""` is falsy in JS.

When the *only* hotswappable change on a Lambda is `Description -> ""` (no `Code`/`Environment` change), `configurations` becomes `undefined`, so the function returns `undefined`. The caller (`isHotswappableLambdaFunctionChange`) reads that as "nothing to do here" and drops the resource change entirely — it isn't even reported as non-hotswappable. If that was the deployment's only change, `tryHotswapDeployment` ends up with an empty `hotswappableChanges` list, which is reported to the user as `noOp: true` (nothing to hotswap) or a clean success — while the Lambda's `Description` in AWS silently never gets updated. This is infrastructure drift with no error, warning, or fallback to a full CloudFormation deployment.

### Description of changes

Switches both truthiness checks to `!== undefined`, matching how `code`/`description`/`environment` are actually populated in the loop above (only ever assigned when that property appeared in the diff). This mirrors the existing correct pattern already used a few lines below for applying the update (`lambdaCodeChange.configurations.description !== undefined`).

Added a regression test, `'calls the updateLambdaConfiguration() API when a Lambda function Description is changed to an empty string'`, to `test/api/hotswap/lambda-functions-hotswap-deployments.test.ts` (both `FALL_BACK` and `HOTSWAP_ONLY` modes, matching the file's existing `describe.each` pattern). Verified it fails on the pre-fix code (the expected `UpdateFunctionConfigurationCommand` is never sent) and passes with the fix.

Full hotswap suite: `test/api/hotswap/` — 285 passed, no regressions.

### Checklist
- [ ] Integration tests added/updated (not applicable — logic-only change to a helper function already covered by extensive unit tests; no new resource type or cross-service interaction)

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
@@ -234,9 +234,8 @@ async function evaluateLambdaFunctionProps(
     }
   }
 
-  const configurations =
-    description !== undefined || environment !== undefined ? { description, environment } : undefined;
-  return code !== undefined || configurations !== undefined ? { code, configurations } : undefined;
+  const configurations = description !== undefined || environment ? { description, environment } : undefined;
+  return code || configurations ? { code, configurations } : undefined;
 }
 
 interface LambdaFunctionCode {
--- a/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
@@ -234,8 +234,9 @@ async function evaluateLambdaFunctionProps(
     }
   }
 
-  const configurations = description || environment ? { description, environment } : undefined;
-  return code || configurations ? { code, configurations } : undefined;
+  const configurations =
+    description !== undefined || environment !== undefined ? { description, environment } : undefined;
+  return code !== undefined || configurations !== undefined ? { code, configurations } : undefined;
 }
 
 interface LambdaFunctionCode {
--- a/packages/@aws-cdk/toolkit-lib/test/api/hotswap/lambda-functions-hotswap-deployments.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/hotswap/lambda-functions-hotswap-deployments.test.ts
@@ -797,6 +797,65 @@ describe.each([HotswapMode.FALL_BACK, HotswapMode.HOTSWAP_ONLY])('%p mode', (hot
   );
 
   test(
+    'calls the updateLambdaConfiguration() API when a Lambda function Description is changed to an empty string',
+    async () => {
+      // GIVEN
+      setup.setCurrentCfnStackTemplate({
+        Resources: {
+          Func: {
+            Type: 'AWS::Lambda::Function',
+            Properties: {
+              Code: {
+                S3Bucket: 's3-bucket',
+                S3Key: 's3-key',
+              },
+              FunctionName: 'my-function',
+              Description: 'Old Description',
+            },
+            Metadata: {
+              'aws:asset:path': 'asset-path',
+            },
+          },
+        },
+      });
+      const cdkStackArtifact = setup.cdkStackArtifactOf({
+        template: {
+          Resources: {
+            Func: {
+              Type: 'AWS::Lambda::Function',
+              Properties: {
+                Code: {
+                  S3Bucket: 's3-bucket',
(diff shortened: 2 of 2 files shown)


packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
```diff
@@ -234,8 +234,9 @@ async function evaluateLambdaFunctionProps(
     }
   }
 
-  const configurations = description || environment ? { description, environment } : undefined;
-  return code || configurations ? { code, configurations } : undefined;
+  const configurations =
+    description !== undefined || environment !== undefined ? { description, environment } : undefined;
+  return code !== undefined || configurations !== undefined ? { code, configurations } : undefined;
```
ShadowCat567: I understand why you are applying this change to `description` but `environment`, `code`, and `configurations`, are objects, `{}`/empty object is truthy for them (https://masteringjs.io/tutorials/fundamentals/empty-object-truthy) so I don't think they need the undefined check.

ShadowCat567 (commented):
Looks good overall! Minor comment about using `!== undefined` for the objects

packages/@aws-cdk/toolkit-lib/lib/api/hotswap/lambda-functions.ts
Adityaj0: Good catch, you're right — pushed a fix that reverts `code`/`environment`/`configurations` back to their original truthy checks and keeps `!== undefined` only on `description`, the one place it was actually load-bearing. Full hotswap suite (285 tests) still green.

ShadowCat567 (approved):

PR #1847, opened 2026-08-18
fix(cli): guide Stage users to a working pattern in stack-selection error

lemon0333:
Fixes #1451

When an app has more than one stack and no selector is given, `cdk deploy` prints:

> Since this app includes more than a single stack, specify which stacks to use (wildcards are supported) or specify `--all`

When the stacks live inside a **Stage**, this message is unhelpful: their ids are namespaced like `StageName/StackName`, and neither a bare stack name nor `--all` is the obvious way to target them, so users are left guessing (see #1451).

This change detects stacks that are nested inside a Stage (i.e. stacks that are not top-level stacks) and additionally points the user at a wildcard pattern that selects them, e.g.:

> Some of these stacks are nested inside a Stage. To select the stacks in a Stage, use a pattern that matches their full path, e.g. `'StageName/*'`

The hint is only appended when staged stacks are actually present, so flat apps are unaffected. The Stage detection is structural (stacks present recursively but not at the top level) rather than string-based, so a top-level stack whose display name happens to contain `/` is not misclassified. The same message is shared by the CLI stack-selection path (`cloud-assembly.ts`) and the `toolkit-lib` `selectStacksV2` path, so both are updated for consistency.

Unit tests were added on both selector/error paths (Stage guidance present when nested, absent for flat apps, and stage patterns de-duplicated across multiple stages).

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
@@ -163,8 +163,7 @@ export class StackAssembly implements IReadableCloudAssembly {
       case StackSelectionStrategy.ONLY_SINGLE:
         if (topLevelStacks.length !== 1) {
           // @todo text should probably be handled in io host
-          throw new ToolkitError('MultipleStacksWithoutSelector', 'Since this app includes more than a single stack, specify which stacks to use (wildcards are supported) or specify `--all`\n' +
-          `Stacks: ${allStacks.map(x => x.hierarchicalId).join(' · ')}`);
+          throw new ToolkitError('MultipleStacksWithoutSelector', multipleStacksWithoutSelectorMessage(topLevelStacks, allStacks));
         }
         return { stacks: new StackCollection(this, topLevelStacks) };
       default:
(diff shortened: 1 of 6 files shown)


mrgrain (changes requested):
Nice! Can you please add test cases to the CLI packages that logs io snpashots? See https://github.com/aws/aws-cdk-cli/tree/main/packages/aws-cdk/test/commands and https://github.com/aws/aws-cdk-cli/tree/main/packages/aws-cdk/test/commands/__io_snapshots__/deploy

lemon0333:
Thanks for the review, @mrgrain! I've added command-level tests that go through the io-snapshot harness.

In `packages/aws-cdk/test/commands/deploy.test.ts` (`multi-stack selection` block) there are now two tests that drive `cdk deploy` with no selector:

- **Stage-nested app** (no top-level stacks, stacks live under `MyStage`) — asserts the error guides the user to the Stage wildcard pattern, e.g. `'MyStage/*'`.
- **Flat multi-stack app** — asserts the error does *not* mention Stages.

Each records an entry under `packages/aws-cdk/test/commands/__io_snapshots__/deploy/` via the same `IoHostRecorder`/`matchSnapshot()` afterEach the other deploy tests use.

One thing worth flagging: the guidance itself is raised as a thrown `ToolkitError` (the deploy path aborts at stack selection, and the message is rendered at the CLI top level by `prettyPrintError`), so it never crosses the `IoHost` `notify`/`requestResponse` boundary the recorder observes. The committed snapshots therefore capture the IO stream up to the point deploy aborts (synthesis, then stop) — protecting that no stray user-facing IO is emitted on this path — while the guidance text is asserted directly on the rejected error. This mirrors the existing `deploy failures` test, which also asserts on the thrown error's message rather than an IO line.

If you'd prefer the guidance to also be emitted through the `IoHost` (so the message text lands in the snapshot as well), I'm happy to add that — just let me know, since it's a small behavioral change to how the error is surfaced.

mrgrain:
> One thing worth flagging: the guidance itself is raised as a thrown `ToolkitError` (the deploy path aborts at stack selection, and the message is rendered at the CLI top level by `prettyPrintError`), so it never crosses the `IoHost` `notify`/`requestResponse` boundary the recorder observes.

That's fine, we are aware of it! Thanks for making the change. Once conflicts are resolved this should be good to merge! Thanks again.

mrgrain (approved):

PR #1774, opened 2026-07-29
chore: revert "refactor: migrate all subprocess call sites to the shared subprocess module"

mrgrain:
Reverts aws/aws-cdk-cli#1758

We decided this is too risky right now since we don't have windows integration tests as part of the release flow. We will prioritize the Windows integ tests first.

---

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #1849, opened 2026-08-18
refactor: migrate all subprocess call sites to the shared subprocess module

iankhou:
This re-submits changes from #1758, which was merged and then reverted in #1774 because we weren't able to run Windows integ tests on it yet.

Tested in workflow: https://github.com/aws/aws-cdk-cli/actions/runs/32169009705, which ran the changes through the integ tests on both Linux and Windows using throwaway branch https://github.com/aws/aws-cdk-cli/tree/refs/heads/iankhou-windows-x-shell-interactions.

## Summary

Follow-up to #1763, which adds the `@aws-cdk/private-tools/lib/subprocess` module.

This PR migrates every `child_process` call site in the shipped packages onto the shared module, so that no code path **both escapes and executes**:

| Call site | Before | After |
|---|---|---|
| `init/os.ts`, `init/init.ts` | `spawn(argv, {shell: true})` + broken hand-rolled escaping | `run()` — no shell, `inherit-stderr` keeps npm/pip progress bars |
| `telemetry/library-version.ts` | `exec()` string | `run()` argv |
| toolkit-lib `ec2-detection.ts` | `execSync` with `2>nul` | `runSync()`, stderr discarded via stdio |
| toolkit-lib `cloud-assembly/exec.ts` (synth `app`) | `spawn(str, {shell: true})` + split2 | `runUserCommandLine()`, line-buffered, `collect: false` |
| `commands/docs.ts` (`--browser`) | `exec()` | `runUserCommandLine()` |
| cdk-assets-lib `shell.ts` | own spawn + broken escaping copy | delegates to `run()`; spawn failures rethrow the OS error so docker's ENOENT → 'please install docker' guidance is preserved |
| toolkit-lib `environment.ts` | `quoteSpaces()` (space-only quoting) | allowlist-based `quoteShellPart()` |

The deleted `windowsEscape` helpers carried an escaping bug (`shellMeta.has(x)` instead of `has(c)`) that made Windows escaping a silent no-op — this PR removes that pattern. `grep -rn runUserCommandLine` now returns exactly the two sinks where a shell is the documented contract (synth `app`, `--browser`) — the complete trust-model surface.

Drops `split2` from toolkit-lib (replaced by the module's line buffering).

Follow-ups include `cdk-build-tools`/`integ-runner`/`cli-integ` migrations.

### Checklist
- [x] Manually verified: init (git/npm/metachar paths), synth (shell features, exit codes, 200k-line streaming, UTF-8), docs --browser, cdk-assets shell paths (hostile args, docker ENOENT, EPIPE)

---


ShadowCat567 (commented):
This looks like it should be fine, but do we want to wait for the Windows tests PR to be ready as well?

iankhou:
> This looks like it should be fine, but do we want to wait for the Windows tests PR to be ready as well?

@ShadowCat567 I ran the Windows workflow against this branch to validate that it will not cause a regression, so I think this is safe to merge.

You're right that work on Windows tests is not complete, but they're now more a matter of how often we run them, since we know they work already.

ShadowCat567 (approved):

PR #1854, opened 2026-08-19
chore(cli-integ): make the integ test harness work on Windows

dgandhi62:
This pr is created for maintaining separation of concerns from here - https://github.com/aws/aws-cdk-cli/pull/1781

Here, we make modifications to the testing harness, since a lot of the older code was linux-specific and did not run properly on Windows. The code to enable the windows tests will be a follow up. 

---

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/integ-test.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/integ-test.ts
@@ -246,14 +246,13 @@ export async function atomicWrite(fileName: string, contents: string) {
   const tmp = `${fileName}.${process.pid}`;
   await fs.promises.writeFile(tmp, contents);
 
-  const RENAME_RETRYABLE = ['EPERM', 'EACCES'];
   const maxAttempts = 10;
   for (let attempt = 1; ; attempt++) {
     try {
       await fs.promises.rename(tmp, fileName);
       return;
     } catch (e: any) {
-      if (!RENAME_RETRYABLE.includes(e.code) || attempt >= maxAttempts) {
+      if (!['EPERM', 'EACCES'].includes(e.code) || attempt >= maxAttempts) {
         // Final failure: don't leave the temp file behind as litter.
         await fs.promises.rm(tmp, { force: true }).catch(() => undefined);
         throw e;
--- a/packages/@aws-cdk-testing/cli-integ/lib/integ-test.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/integ-test.ts
@@ -229,12 +229,38 @@ function slugify(x: string) {
   return x.replace(/[^a-zA-Z0-9_,]+/g, '-');
 }
 
-async function atomicWrite(fileName: string, contents: string) {
+/**
+ * Write a file by writing to a temp file and renaming it into place.
+ *
+ * On POSIX the final rename atomically replaces any existing destination, and
+ * concurrent writers of the same target harmlessly clobber each other. On
+ * Windows, replacing a destination that another process currently has open (or
+ * that is in a "delete pending" state from a concurrent replace) fails with
+ * EPERM/EACCES. Multiple test workers rewrite shared log files (notably
+ * `0-header.md`) at once, so ride out that transient window by retrying the
+ * rename a handful of times before giving up.
+ */
+export async function atomicWrite(fileName: string, contents: string) {
   await fs.promises.mkdir(path.dirname(fileName), { recursive: true });
 
   const tmp = `${fileName}.${process.pid}`;
   await fs.promises.writeFile(tmp, contents);
-  await fs.promises.rename(tmp, fileName);
+
+  const RENAME_RETRYABLE = ['EPERM', 'EACCES'];
+  const maxAttempts = 10;
+  for (let attempt = 1; ; attempt++) {
+    try {
+      await fs.promises.rename(tmp, fileName);
+      return;
+    } catch (e: any) {
+      if (!RENAME_RETRYABLE.includes(e.code) || attempt >= maxAttempts) {
+        // Final failure: don't leave the temp file behind as litter.
+        await fs.promises.rm(tmp, { force: true }).catch(() => undefined);
+        throw e;
+      }
+      await new Promise(ok => setTimeout(ok, Math.floor(Math.random() * 20) + 5));
+    }
+  }
 }
 
 function readSkipFile(filePath?: string): string[] {
--- a/packages/@aws-cdk-testing/cli-integ/lib/shell.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/shell.ts
@@ -286,7 +286,7 @@ export function rimraf(fsPath: string): boolean {
 
     // Remove links without recursing into their target: a directory may
     // link to shared content that other tests are still using (e.g. the
-    // shared 'node_modules' on Windows).
+    // shared 'node_modules').
     if (stat.isSymbolicLink()) {
       try {
         fs.unlinkSync(fsPath);
--- a/packages/@aws-cdk-testing/cli-integ/lib/shell.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/shell.ts
@@ -3,6 +3,7 @@ import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import type { TestContext } from './integ-test';
+import { isWindows } from './platform';
 import { Process } from './process';
 import type { TemporaryDirectoryContext } from './with-temporary-directory';
 
@@ -284,15 +285,18 @@ export function rimraf(fsPath: string): boolean {
     let success = true;
     const stat = fs.lstatSync(fsPath);
 
-    // Remove links without recursing into their target: a directory may
-    // link to shared content that other tests are still using (e.g. the
-    // shared 'node_modules').
+    // This test's private directory contains a 'node_modules' symlink into a
+    // machine-wide shared install that other running tests also link to. Delete
+    // the link itself and stop — do NOT recurse through it, or we'd delete the
+    // shared install's contents out from under those other tests.
     if (stat.isSymbolicLink()) {
-      try {
-        fs.unlinkSync(fsPath);
-      } catch {
-        // On Windows, directory links (junctions) must be removed with rmdir
+      // On POSIX, unlink removes a symlink whatever its target type. On
+      // Windows, a link to a directory (or a junction) must be removed with
+      // rmdir, while a link to a file must be removed with unlink.
+      if (isWindows() && isDirectoryLink(fsPath)) {
         fs.rmdirSync(fsPath);
+      } else {
+        fs.unlinkSync(fsPath);
       }
       return true;
     }
@@ -322,6 +326,18 @@ export function rimraf(fsPath: string): boolean {
 
     throw e;
   }
+}
+
+/**
+ * Whether a symlink resolves to a directory.
+ *
+ * `statSync` follows the link, so a directory target means a directory link.
+ * A dangling link (target already removed) returns undefined; treat it as a
+ * directory, since the only links we create are directory links (the shared
+ * 'node_modules' junction) and those still need `rmdir` on Windows.
+ */
+function isDirectoryLink(linkPath: string): boolean {
+  return fs.statSync(linkPath, { throwIfNoEntry: false })?.isDirectory() ?? true;
 }
 
 export function addToShellPath(x: string) {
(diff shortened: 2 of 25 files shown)


packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
```diff
@@ -1045,6 +1065,70 @@ export async function installNpmPackages(fixture: TestFixture, packages: Record<
...
   }, undefined, 2), { encoding: 'utf-8' });
 
+  if (process.platform === 'win32') {
+    // Installing aws-cdk-lib means writing out tens of thousands of small
+    // files, which is very slow on Windows (minutes instead of seconds),
+    // and every concurrent jest worker doing so at once makes it slower
+    // still. Install every distinct package set only once per machine and
+    // junction it into the test directory.
+    const sharedNodeModules = await sharedPackageSetInstall(fixture, packages);
+    fs.symlinkSync(sharedNodeModules, path.join(fixture.integTestDir, 'node_modules'), 'junction');
```
iliapolo: Interesting. Lets make use of it on linux as well! We've seen network errors due to multiple `npm install` processes and I remember also seeing some slow installation in our CodeBuild canary runs.

iliapolo (changes requested):

packages/@aws-cdk-testing/cli-integ/tests/init-csharp/init-csharp.integtest.ts
```diff
@@ -10,6 +10,6 @@ import { integTest, withTemporaryDirectory, ShellHelper, withPackages } from '..
     await shell.shell(['cdk', 'init', '--lib-version', context.library.requestedVersion(), '-l', 'csharp', template]);
     await context.library.initializeDotnetPackages(context.integTestDir);
     await shell.shell(['cdk', 'synth']);
-  })));
+  })), 300_000);
```
iankhou: For each instance where we adjust test timeouts, please verify in the workflow run whether it needs this much time.

dgandhi62: I can reduce the timeouts in some to make it closer to the test times. Depending on the urgency of this pr, we can have a separate one to make efficiency improvements in the tests themselves

dgandhi62: For the ask of updating timeouts based on test times, that is done. If the ask is to identify test improvements to reduce times, that might have to be a separate pr

iankhou: No ask for the latter. Reducing the timeouts is OK.

packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
iankhou: This was one of the biggest changes for bringing down Windows integ test times, since `npm install` is so much less efficient. Concur on running it on Linux. @dgandhi62 thanks for the changes, can you show us a workflow run using the latest commit to validate it works as expected (tests pass and shorter test times)?

dgandhi62: Running it with the latest code now. Will link it here once done

dgandhi62: https://github.com/aws/aws-cdk-cli/actions/runs/32524341885

iankhou (approved):

iankhou (approved):

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
```diff
@@ -282,7 +282,22 @@ export class ShellHelper {
...
-    const isDir = fs.lstatSync(fsPath).isDirectory();
+    const stat = fs.lstatSync(fsPath);
+
+    // Remove links without recursing into their target: a directory may
+    // link to shared content that other tests are still using (e.g. the
+    // shared 'node_modules').
+    if (stat.isSymbolicLink()) {
+      try {
+        fs.unlinkSync(fsPath);
+      } catch {
```
iliapolo: Why rely on exceptions for handling win/linux? we already have a bunch of code testing for windows explicitly, lets use that (better yet - create a util `isWindows()` function). Then the code is self-describing, no comments needed.

packages/@aws-cdk-testing/cli-integ/lib/integ-test.ts
```diff
@@ -229,12 +229,38 @@ function slugify(x: string) {
...
+ * rename a handful of times before giving up.
+ */
+export async function atomicWrite(fileName: string, contents: string) {
   await fs.promises.mkdir(path.dirname(fileName), { recursive: true });
 
   const tmp = `${fileName}.${process.pid}`;
   await fs.promises.writeFile(tmp, contents);
-  await fs.promises.rename(tmp, fileName);
+
+  const RENAME_RETRYABLE = ['EPERM', 'EACCES'];
```
iliapolo: UPPER_CASE in the middle a function hurts my eyes. Please lowercase or move to a module constant. Actually there is no need to the `const` at all - you are just using it once.

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
```diff
@@ -282,7 +282,22 @@ export class ShellHelper {
...
+        // On Windows, directory links (junctions) must be removed with rmdir
+        fs.rmdirSync(fsPath);
```
iliapolo: How do you know the link is a directory though?

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
```diff
@@ -282,7 +282,22 @@ export class ShellHelper {
 export function rimraf(fsPath: string): boolean {
   try {
     let success = true;
-    const isDir = fs.lstatSync(fsPath).isDirectory();
+    const stat = fs.lstatSync(fsPath);
+
+    // Remove links without recursing into their target: a directory may
+    // link to shared content that other tests are still using (e.g. the
```
iliapolo: I don't understand...

1. Why is this being called on a shared directory as tests are running?
2. If another process is relying on the symlink, why is it ok for us to remove it?

iliapolo (changes requested):

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
dgandhi62: `rimraf` is called on the private directory of each test, not on the shared install directory.. That private directory contains a node_modules symlink pointing at the shared install. The comment was not clear enough - I've changed it

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
dgandhi62: The only symlink on windows is the node_modules directory link. But you're right. I updated it to check whether its a directory or not first

packages/@aws-cdk-testing/cli-integ/lib/integ-test.ts
dgandhi62: inlined it

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
dgandhi62: Updated this!

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
```diff
@@ -282,7 +283,25 @@ export class ShellHelper {
...
+
+    // This test's private directory contains a 'node_modules' symlink into a
+    // machine-wide shared install that other running tests also link to. Delete
+    // the link itself and stop — do NOT recurse through it, or we'd delete the
+    // shared install's contents out from under those other tests.
+    if (stat.isSymbolicLink()) {
+      // On POSIX, unlink removes a symlink whatever its target type. On
+      // Windows, a link to a directory (or a junction) must be removed with
+      // rmdir, while a link to a file must be removed with unlink.
+      if (isWindows() && isDirectoryLink(fsPath)) {
```
iliapolo: Why not:

```suggestion
      if (isWindows() && stats.isDirectory()) {
```

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
```diff
@@ -309,14 +328,26 @@ export function rimraf(fsPath: string): boolean {
...
+/**
+ * Whether a symlink resolves to a directory.
+ *
+ * `statSync` follows the link, so a directory target means a directory link.
+ * A dangling link (target already removed) returns undefined; treat it as a
+ * directory, since the only links we create are directory links (the shared
+ * 'node_modules' junction) and those still need `rmdir` on Windows.
+ */
+function isDirectoryLink(linkPath: string): boolean {
+  return fs.statSync(linkPath, { throwIfNoEntry: false })?.isDirectory() ?? true;
```
iliapolo: This function is not needed. But in general - is there a good reason why we would call this on a path that doesn't exist? 

Agents will always prefer not to throw - make sure you evaluate that decision every time, because this hides very subtle bugs.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/watch/watch-helpers.ts
```diff
@@ -33,12 +35,32 @@ export async function waitForCondition(condition: () => boolean): Promise<void>
   expect(condition()).toBe(true);
 }
 
+/**
+ * Spawn a long-running `cdk watch` process.
+ *
+ * On Windows the CLI is an npm .cmd shim, which `spawn` can only start
+ * through a shell ('spawn cdk ENOENT' otherwise).
+ */
+export function spawnWatch(args: string[], options: SpawnOptions): ChildProcess {
```
iliapolo, line 44: Why is this called `spawnWatch` but actually just runs `cdk`? Also I'm sure we already have this function somewhere that can be reused.

iliapolo (changes requested):

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
dgandhi62: We used `lstatsync` earlier, which describes the symlink, and not the target. isDirectory needs to operate on the target.

Ref - https://www.geeksforgeeks.org/node-js/node-js-fs-lstatsync-method/
https://www.geeksforgeeks.org/node-js/node-js-stats-isdirectory-method-from-fs-stats-class/

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
dgandhi62: I didn't understand - when would the path not exist?

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/watch/watch-helpers.ts
dgandhi62: I'm not sure why (I inherited this code from the original pr). The function is here - https://github.com/aws/aws-cdk-cli/blob/dcf9d18512c6fbe4d115eb510380e22ec327c188/packages/%40aws-cdk-testing/cli-integ/lib/process.ts#L74

My guess is that we need a way to kill the process at the end specifically for watch, and returning the child process gives us that. The return type of the older function is completely non-destructive.

I think this might explain the naming too

dgandhi62: If you look at the integ tests, we've just abstracted away common code. I don't think this is a big change per say

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
iliapolo: Discussed offline, we are already not recursing if its a symlink to a directory, so we just need to handle the unliking correctly:

```ts
      // On POSIX, unlink removes a symlink whatever its target type. On
      // Windows, a link to a directory (or a junction) must be removed with
      // rmdir, while a link to a file must be removed with unlink.
      if (stat.isSymbolicLink() && isWindows() && isDirectoryLink(fsPath)) {
        fs.rmdirSync(fsPath);
      } else {
        fs.unlinkSync(fsPath);
      }

```

packages/@aws-cdk-testing/cli-integ/lib/shell.ts
iliapolo: Exactly - it should always exist, which is why we **should** throw

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/watch/watch-helpers.ts
iliapolo: I was just confused by the name because it doesn't actually spawn `cdk watch`. Its fine.

iliapolo (approved):

PR #1861, opened 2026-08-20
refactor(toolkit-lib): batch asset-active lookups in cdk gc with a multi-pattern search

### LINKED ISSUE #1860: perf(toolkit-lib): cdk gc's active-asset lookup is O(assets x stacks x template size)
### Describe the bug

`cdk gc` (garbage collection of orphaned S3/ECR bootstrap assets) determines whether an asset is still referenced by any stack via `ActiveAssetCache.contains()` in [`packages/@aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts`](https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/toolkit-lib/lib/api/garbage-collection/stack-refresh.ts):

```ts
public contains(asset: string): boolean {
  for (const stack of this.stacks) {
    if (stack.includes(asset)) {
      return true;
    }
  }
  return false;
}
```

This is called once **per asset** from `garbageCollectEcr`/`garbageCollectS3` in `garbage-collector.ts`, each call doing a full linear scan over every remembered stack template (raw template JSON text), searching for the asset's hash as a substring.

For an account with `S` stacks (each up to hundreds of KB) and `A` orphaned assets accumulated over time (easily tens of thousands in a long-lived account), this makes the asset-lookup phase of `cdk gc` cost roughly `O(A x S x avg template size)` — every asset re-scans every stack template from scratch.

### Expected Behavior

`cdk gc`'s active-asset lookup should scale roughly with total template size plus the number of assets being checked, not their product — it shouldn't get quadratically slower as an account accumulates more stacks and more garbage over time.

### Observed Behavior

In a synthetic benchmark modeling a large, long-lived account (500 stacks x ~20KB templates, 50,000 orphaned assets processed in batches of 1000, matching `cdk gc`'s real batch size), the asset-lookup phase alone took ~13.3s.

### What's the environment?

- aws-cdk-cli main branch
- N/A (algorithmic, reproducible via unit benchmark, not environment-specific)

### Other

I'll follow up with a PR that replaces the per-asset linear scan with a single multi-pattern search (Aho-Corasick) built once per batch, cutting the above benchmark to ~5.3s (~2.5x) with results verified identical to the current implementation (no false negatives/positives — this matters a lot here since a false negative would delete an asset that's still in use).

Adityaj0:
fixes #1860

### Reason for this change

`cdk gc`'s `ActiveAssetCache.contains()` checked whether an asset is still referenced by any stack via a linear per-stack `String.includes()` scan, called once **per asset**:

```ts
public contains(asset: string): boolean {
  for (const stack of this.stacks) {
    if (stack.includes(asset)) {
      return true;
    }
  }
  return false;
}
```

`garbageCollectEcr`/`garbageCollectS3` call this once per asset in every batch (batch size 1000), so for an account with `S` stacks and `A` accumulated orphaned assets, the lookup phase alone costs `O(A x S x avg template size)` — every single asset re-scans every stack template from scratch. In a long-lived account with hundreds of stacks and tens of thousands of orphaned S3 objects / ECR images, this dominates `cdk gc` runtime.

### Description of changes

- Added `ActiveAssetCache.containsAny(assets: string[])`, which builds a single [Aho-Corasick](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm) multi-pattern search index over a batch of candidate asset identifiers and scans each remembered stack template **exactly once**, regardless of how many candidates are in the batch. This brings the cost down to `O(stacks x avg template size + sum of candidate lengths)` per batch.
- `garbageCollectEcr`/`garbageCollectS3` now call `containsAny()` once per batch (of up to 1000 assets/tags) instead of calling `contains()` once per asset.
- Kept `contains(asset)` as a thin wrapper around `containsAny([asset])` for any other/future single-item callers.
- The multi-pattern search is exact substring matching — same semantics as before, so results are byte-for-byte identical to the old implementation. This matters a lot here: a false negative would mean deleting an asset that's actually still referenced by a live stack.

### Testing

- Added `test/api/garbage-collection/stack-refresh.test.ts`, including:
  - basic `contains`/`containsAny` behavior across multiple stacks
  - overlapping/substring patterns (one candidate being a substring of another)
  - a template-boundary false-positive regression test (concatenating templates without a separator could otherwise create a spurious match spanning two templates)
  - a randomized-workload test that cross-checks `containsAny()` against the naive per-candidate `contains()` for exact agreement (no false negatives, no false positives)
- Existing `garbage-collection.test.ts` suite passes unchanged (40/40 tests).

### Metrics

Synthetic benchmark modeling a large, long-lived account (500 stacks x ~20KB templates, 50,000 orphaned assets processed in batches of 1000, matching `cdk gc`'s real batch size):

| | Before | After | Speedup |
|---|---|---|---|
| Asset-lookup phase | ~13.3s | ~5.3s | ~2.5x |

Results were identical between old and new implementations across all benchmarked scales (3,000 / 20,000 / 50,000 assets) — verified via direct set comparison, not just count.

### Checklist

- [ ] Integration tests added/updated (not applicable — algorithmic change, no new AWS resource types or cross-service interactions)

----

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/test/api/garbage-collection/stack-refresh.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/garbage-collection/stack-refresh.test.ts
@@ -81,10 +81,16 @@ describe(ActiveAssetCache, () => {
       function randHash(rng: () => number) {
         return Array.from({ length: 12 }, () => Math.floor(rng() * 16).toString(16)).join('');
       }
-      // Simple deterministic PRNG so failures are reproducible.
+      // Simple deterministic PRNG so failures are reproducible. Math.imul keeps
+      // the multiply within a 32-bit-safe integer (no bitwise operators, no
+      // precision loss from JS doubles on the full product).
       let seed = 42;
       const rng = () => {
-        seed = (seed * 1103515245 + 12345) & 0x7fffffff;
+        seed = Math.imul(seed, 1103515245) + 12345;
+        seed = seed % 0x7fffffff;
+        if (seed < 0) {
+          seed += 0x7fffffff;
+        }
         return seed / 0x7fffffff;
       };
 
--- a/packages/@aws-cdk/toolkit-lib/test/api/garbage-collection/stack-refresh.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/garbage-collection/stack-refresh.test.ts
@@ -1,0 +1,109 @@
+import { ActiveAssetCache } from '../../../lib/api/garbage-collection/stack-refresh';
+
+describe(ActiveAssetCache, () => {
+  test('contains returns false when no stacks are remembered', () => {
+    const cache = new ActiveAssetCache();
+    expect(cache.contains('some-asset-hash')).toBe(false);
+  });
+
+  test('contains finds an asset referenced anywhere within a remembered template', () => {
+    const cache = new ActiveAssetCache();
+    cache.rememberStack('{"Resources":{"Bucket":{"Properties":{"Key":"prefix-abc123def-suffix.zip"}}}}');
+    expect(cache.contains('abc123def')).toBe(true);
+    expect(cache.contains('not-present')).toBe(false);
+  });
+
+  test('contains searches across all remembered stacks, not just the first', () => {
+    const cache = new ActiveAssetCache();
+    cache.rememberStack('{"Resources":{"A":"nothing-relevant-here"}}');
+    cache.rememberStack('{"Resources":{"B":"has-the-hash-xyz789-in-it"}}');
+    cache.rememberStack('{"Resources":{"C":"also-irrelevant"}}');
+    expect(cache.contains('xyz789')).toBe(true);
+  });
+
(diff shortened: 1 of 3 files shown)


packages/@aws-cdk/toolkit-lib/test/api/garbage-collection/stack-refresh.test.ts
```diff
@@ -0,0 +1,115 @@
...
+        return seed / 0x7fffffff;
+      };
+
+      const cache = new ActiveAssetCache();
+      const embeddedHashes: string[] = [];
+      for (let i = 0; i < 20; i++) {
+        const h = randHash(rng);
+        embeddedHashes.push(h);
+        cache.rememberStack(`{"Resources":{"R${i}":{"Key":"prefix-${h}-suffix"}}}`);
+      }
```
mrgrain, line 103: Yikes, let's see if we can simplify this test case. This seems like a lot of workarounds to test the actual feature. Why is this needed?

mrgrain (changes requested):

PR #1867, opened 2026-08-20
feat: add new github actions workflow for windows integ tests

dgandhi62:
Adds Windows instances of the five integ suites (`cli-integ-tests`, `toolkit-lib`, `telemetry`, `init-templates`, `tool-integrations`) and wires up when they run. Windows tests take about ~24 minutes to run on the default core by github. Therefore, we don't run them on every pr. Instead, we run them:

- Nightly at 06:00 UTC, unattended. A failure files a tracking issue labelled `potential-regression`.
- On demand when a PR carries the `pr/test-windows` label. This failure then blocks the PR like any other integ failure. When adding the label, a new run needs to be triggered through any event like a push for it to take effect.

Linux is excluded from the nightly since it already runs on every PR

A schedule-only `integ_windows_report_failure` job files a GitHub issue on failure. Label-triggered PR runs file nothing. The job also must keep the default `GITHUB_TOKEN`. Issues created with it don't trigger other workflows, which is what stops `issue-regression-labeler` from stripping `potential-regression` off an issue whose body has no regression checkbox.

### Prerequisites

Requires the `pr/test-windows` label to exist in the repository so maintainers can opt a PR into the Windows suites. The nightly needs no label (it runs on a cron schedule). The failure-report issue uses the existing `potential-regression` label, already managed by `issue-regression-labeler`.

### Checklist

---

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -4,8 +4,16 @@ name: integ
 on:
   pull_request_target:
     branches: []
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
   merge_group: {}
   workflow_dispatch: {}
+  schedule:
+    - cron: 0 6 * * *
 jobs:
   determine_env:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


dgandhi62:
For reference, passing run - https://github.com/aws/aws-cdk-cli/actions/runs/33178564563

rix0rrr (approved):

PR #1875, opened 2026-08-22
fix(toolkit-lib): hotswap template cache ignores target environment

### LINKED ISSUE #1874: hotswap template cache is not keyed by target environment, can reuse stale state across accounts/regions
### Describe the bug

The on-disk hotswap cache written by \`hotswap-template-cache.ts\` (\`<assemblyDir>/.hotswap-cache/<stackName>.json\`) is keyed only by the cloud assembly directory and the stack name:

```ts
function cachePath(assemblyDir: string, stackName: string): string {
  return path.join(assemblyDir, CACHE_DIR, `${stackName}.json`);
}
```

The cached payload stores the last-deployed root template and, for nested stacks, the physical stack names — both of which are meaningful only for the specific AWS account/region they were captured against. Nothing in the cache key or the cached data records which environment produced it.

\`hotswapDeployment()\` in \`hotswap-deployments.ts\` resolves the target environment via \`sdkProvider.resolveEnvironment(stack.environment)\` and then unconditionally trusts whatever is in the cache file for that stack name:

```ts
const resolvedEnv = await sdkProvider.resolveEnvironment(stack.environment);
...
const hotswapCache = await readHotswapTemplateCache(stack.assembly.directory, stack.stackName, stack.template);
const currentTemplate = hotswapCache ?? await loadCurrentTemplateWithNestedStacks(stack, sdk);
```

The only place the cache is invalidated is after a full CloudFormation deployment (\`invalidateHotswapTemplateCache\`, called from \`deploy-stack.ts\`). Switching the *target environment* between hotswap-only sessions never invalidates it.

### Impact / repro

1. A CDK app resolves its environment from CLI credentials at deploy time (a common pattern for per-developer sandbox accounts, or simply switching \`AWS_PROFILE\`/\`CDK_DEFAULT_ACCOUNT\` without re-synthesizing).
2. Run \`cdk watch\` (hotswap) against Account A for stack \`MyStack\`. On success, \`writeHotswapTemplateCache\` persists Account A's deployed template and physical resource names to \`cdk.out/.hotswap-cache/MyStack.json\`.
3. Without an intervening full (non-hotswap) deploy, switch credentials and run \`cdk watch\` again for the same stack name against Account B.
4. \`readHotswapTemplateCache\` returns Account A's stale cache. The hotswap diff is computed against Account A's deployed state instead of Account B's, and any hotswap operations that rely on the cached physical resource names target Account A's resource identifiers while using an SDK client authenticated for Account B.

This can cause a hotswap to silently skip changes that are actually needed, or to attempt an operation against a physical resource name that doesn't exist (or, worse, happens to collide with an unrelated resource) in the new account/region.

### Expected behavior

The hotswap cache should never be usable across two different target environments — a cache entry written for one account/region should not be readable when hotswapping the same stack name against a different account/region.

### Where

- \`packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-template-cache.ts\`
- \`packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts\`
- \`packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts\`

A fix (folding the resolved environment into the cache key) plus a regression test is up in a PR shortly.

Adityaj0:
Closes #1874

### Reason for this change

The on-disk hotswap cache (`.hotswap-cache/<stackName>.json`, in `hotswap-template-cache.ts`) was keyed only by cloud assembly directory + stack name. Its payload — the last-deployed root template and, for nested stacks, physical resource names — is only valid for the specific account/region it was captured against, but nothing in the key or the data records which environment produced it.

`hotswapDeployment()` resolves the target environment and then unconditionally trusts the cache for that stack name, regardless of whether it was written for a different account/region. The only invalidation path is a full CloudFormation deployment for the same stack name — switching *target environments* between hotswap-only sessions (e.g. `cdk watch` runs against different accounts via `AWS_PROFILE`, common for environments resolved from CLI credentials at deploy time) never invalidates it.

Concretely: hotswap against Account A, then without an intervening full deploy, switch credentials and hotswap the same stack name against Account B. The stale Account-A cache is returned, so the diff is computed against the wrong deployed state and any hotswap operations that rely on cached physical resource names can target the wrong account's resources.

### Description of changes

Fold the resolved environment (`account/region`) into the hotswap cache key (`cachePath`), threading it through `readHotswapTemplateCache`, `writeHotswapTemplateCache`, and `invalidateHotswapTemplateCache`, and updating all three call sites (`hotswap-deployments.ts`, `deploy-stack.ts`) to pass the stack's resolved environment. A cache entry can now only ever be read back for the environment it was written for; a different environment for the same stack name/assembly directory is treated as a cache miss (falls back to loading the template from CloudFormation), and invalidating one environment's cache no longer touches another's.

Added a regression test (`cache from one environment is never returned for a different environment`) to `hotswap-template-cache.test.ts` that writes a cache entry for environment A and asserts reading it back for environment B returns `undefined`, reading it back for environment A still returns the cached state, and invalidating environment B's cache doesn't remove environment A's. Verified this test fails against the pre-fix code (the stale cache is returned across environments) and passes after the fix. Updated the existing tests in the same file to pass an environment argument. Ran the full `hotswap` and `deployments` unit test suites (494 tests) — all pass. Confirmed `tsc --noEmit` has no new type errors from this change.

### Checklist

- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk-cli/blob/main/CONTRIBUTING.md) and the [DESIGN GUIDELINES](https://github.com/aws/aws-cdk-cli/blob/main/docs/DESIGN_GUIDELINES.md)

----

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts
@@ -439,7 +439,12 @@ class FullCloudFormationDeployment {
     const deploymentMethod = this.deploymentMethod ?? { method: 'change-set' };
 
     // if there is a hotswap cache, clear it when a full Cloudformation of any kind happens
-    await invalidateHotswapTemplateCache(this.stackArtifact.assembly.directory, this.stackArtifact.stackName);
+    const deploymentEnv = this.options.resolvedEnvironment;
+    await invalidateHotswapTemplateCache(
+      this.stackArtifact.assembly.directory,
+      this.stackArtifact.stackName,
+      `${deploymentEnv.account}/${deploymentEnv.region}`,
+    );
 
     if (deploymentMethod.method === 'direct' && this.options.resourcesToImport) {
       throw new ToolkitError('ImportRequiresChangeSet', 'Importing resources requires a changeset deployment');
@@ -967,10 +972,12 @@ async function canSkipDeploy(
   }
 
   // treat template in the hotswap cache as the source of truth
+  const hotswapCacheEnv = deployStackOptions.resolvedEnvironment;
   const hotswapCache = await readHotswapTemplateCache(
     deployStackOptions.stack.assembly.directory,
     deployStackOptions.stack.stackName,
     deployStackOptions.stack.template,
+    `${hotswapCacheEnv.account}/${hotswapCacheEnv.region}`,
   );
   if (hotswapCache && diffTemplate(hotswapCache.deployedRootTemplate, deployStackOptions.stack.template).differenceCount > 0) {
     await ioHelper.defaults.debug(`${deployName}: template has changed in relation to last successful hotswap deployment`);
--- a/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/hotswap/hotswap-deployments.ts
@@ -174,7 +174,8 @@ async function hotswapDeployment(
 
   // Check for a cached template from a previous hotswap deployment.
   // Use if available, represents the current state of the resources involved in hotswap.
-  const hotswapCache = await readHotswapTemplateCache(stack.assembly.directory, stack.stackName, stack.template);
+  const environmentKey = `${resolvedEnv.account}/${resolvedEnv.region}`;
+  const hotswapCache = await readHotswapTemplateCache(stack.assembly.directory, stack.stackName, stack.template, environmentKey);
   const currentTemplate = hotswapCache ?? await loadCurrentTemplateWithNestedStacks(stack, sdk);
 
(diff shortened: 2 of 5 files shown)


ShadowCat567 (approved):

ShadowCat567:
@Adityaj0 Can you update the integration tests for hotswap? This is the file you will want to update: https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk-testing/cli-integ/tests/cli-integ-tests/hotswap/cdk-hotswap-deployment-caches-template.integtest.ts

ShadowCat567 (changes requested):
@Adityaj0 can you update the hotswap cache integration tests? See my previous comment for the file you should be updating

ShadowCat567 (approved):

PR #1881, opened 2026-08-22
feat(toolkit-lib): support CloudFormation rollback triggers

### LINKED ISSUE #5170: Support CloudFormation rollback triggers
CloudFormation supports specifying 0-5 rollback triggers as CloudWatch Metric Alarms, which when triggered will automatically cause the stack update to be cancelled. Also a monitoring time of 0-180 minutes can be specified, which means a pause time CloudFormation will wait for any of the alarms to be triggered, or a rollback to be manually triggered, before cleaning up any resources.

There should be a way to use these with AWS CDK.

### Use Case

Rollback triggers have obvious uses to make stack updates more reliable.

### Proposed Solution

Similar to `--notification-arns` currently in `deploy` command, add `--rollback-trigger-alarm-arns` option to be able to list 1-5 CloudWatch Alarms that automatically trigger a rollback. Also add `--monitoring-time-minutes` option which can be used to add 0-180 minutes of pause time after a stack update before the cleanup phase starts. Both options can be specified independently, as they are useful on their own.

### Other

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-rollback-triggers.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_RollbackConfiguration.html

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

mattmb-aws:
Expose CloudFormation rollback triggers (rollback alarms + monitoring time) so a stack create/update is automatically rolled back when a CloudWatch alarm breaches.

- toolkit-lib: new public `rollbackConfiguration` deploy option (RollbackConfiguration/RollbackTrigger/RollbackTriggerType). It is validated (<=5 triggers, valid alarm ARNs, 0-180 min) and threaded through to the Create/Update/CreateChangeSet calls, with change detection so an unchanged config does not force a deployment.
- cli: new `cdk deploy --rollback-trigger-alarm-arns` and `--monitoring-time-minutes` flags. CLI ARNs default to metric alarms; composite alarms are reachable via the programmatic API.

Follows the same undefined/[]/[...] management semantics as notificationArns.

Fixes aws/aws-cdk#5170

Fixes #

### Checklist

---


packages/aws-cdk/lib/cli/cli.ts
```diff
@@ -437,12 +437,22 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           throw new ToolkitError('ConflictingExecuteAndMethod', 'Can not supply both --[no-]execute and --method at the same time');
         }
 
+        // Only manage rollback configuration when the user opts in via either flag. Every alarm ARN
+        // is treated as a metric alarm; composite alarms are only reachable through the programmatic API.
+        const rollbackConfiguration = (args.rollbackTriggerAlarmArns !== undefined || args.monitoringTimeMinutes !== undefined)
```
momo-bee: 🔴 **--monitoring-time-minutes alone silently clears a stack's existing rollback triggers.**

This condition treats `--monitoring-time-minutes` as sufficient, on its own, to opt into CDK managing rollback configuration. When only that flag is passed, `rollbackTriggerAlarmArns` is undefined, so `triggers: (undefined ?? []).map(...)` becomes `triggers: []`. Per the documented semantics in `toolkit.ts` and `deploy/index.ts` ('`{ triggers: [] }`: CDK manages it and clears any existing triggers'), and per `canSkipDeploy` in `deploy-stack.ts` (which only skips the rollback check when `rollbackConfiguration === undefined`), this object is not undefined, so it is compared and forwarded — and an existing set of alarms on the stack is cleared. The CLI help text for `--monitoring-time-minutes` itself claims 'Requires --rollback-trigger-alarm-arns to have any effect', which is the opposite of what the code does.

```
const rollbackConfiguration = (args.rollbackTriggerAlarmArns !== undefined || args.monitoringTimeMinutes !== undefined)
  ? {
    triggers: (args.rollbackTriggerAlarmArns ?? []).map((arn: string) => ({ arn })),
    monitoringTimeInMinutes: args.monitoringTimeMinutes,
  }
  : undefined;
```

**Why it matters:** cause: user runs `cdk deploy --monitoring-time-minutes 30` on a stack that already has rollback alarms configured (via a previous CDK deploy or out-of-band) intending only to extend the monitoring window -> mechanism: the CLI sends `RollbackConfiguration: { RollbackTriggers: [], MonitoringTimeInMinutes: 30 }` to CloudFormation, which is a valid distinct configuration from the existing one -> consequence: CloudFormation silently removes all rollback alarms from the stack on the next deploy, with no warning, contradicting both the CLI's own documented behavior and the user's evident intent, and removing an operational safety mechanism that will not be missed until the next incident it should have caught.

**Suggestion:** Only build a management-triggering config when `rollbackTriggerAlarmArns` is explicitly provided (i.e. gate on `args.rollbackTriggerAlarmArns !== undefined` alone), and either ignore a lone `--monitoring-time-minutes` or throw a `ToolkitError` telling the user it requires `--rollback-trigger-alarm-arns` — matching the CLI help text's own claim.

_[code-review-sage]_

packages/aws-cdk/test/commands/deploy.test.ts
```diff
@@ -429,6 +429,42 @@ describe('deploy parameters forwarded to CloudFormation', () => {
     );
   });
```
momo-bee: 🟡 **No test exercises --monitoring-time-minutes without --rollback-trigger-alarm-arns.**

The new CLI tests (`cli-arguments.test.ts`, `deploy.test.ts`) cover ARNs-provided and no-flags-provided cases, but never the monitoring-time-only case — exactly the path with the silent-clear defect above.

```
test('rollback triggers are converted and forwarded to deployStack', async () => { ... }
test('an invalid rollback trigger arn is rejected', async () => { ... }
```

**Why it matters:** cause: the gap in the argument-combination test matrix -> mechanism: the monitoring-time-only path was never exercised end-to-end -> consequence: the defect above shipped undetected and a future regression on this specific combination would also go unnoticed.

**Suggestion:** Add a test asserting the behavior for `cdk deploy --monitoring-time-minutes 30` with no `--rollback-trigger-alarm-arns`, once the fix above defines the intended behavior (ignore, or throw).

_[code-review-sage]_

packages/aws-cdk/lib/cli/parse-command-line-arguments.ts
```diff
@@ -495,6 +495,17 @@ export function parseCommandLineArguments(args: Array<string>): any {
...
+        .option('rollback-trigger-alarm-arns', {
+          type: 'array',
+          desc: 'ARNs of CloudWatch alarms that CloudFormation monitors during the deployment. If any alarm goes to ALARM state the deployment is rolled back. A maximum of 5 can be specified. Alarms are treated as metric alarms (AWS::CloudWatch::Alarm).',
+          nargs: 1,
+          requiresArg: true,
+        })
+        .option('monitoring-time-minutes', {
+          default: undefined,
+          type: 'number',
+          desc: 'The number of minutes CloudFormation continues to monitor the rollback trigger alarms after the deployment completes, before cleaning up old resources. Must be between 0 and 180. Requires --rollback-trigger-alarm-arns to have any effect.',
```
momo-bee: 🟡 **Help text claims a flag dependency the implementation does not enforce.**

This description states `--monitoring-time-minutes` 'Requires --rollback-trigger-alarm-arns to have any effect', but `cli.ts` builds a non-undefined rollback configuration whenever either flag is set, so the flag does have an effect alone (see the red finding). This is the same defect surfacing as a docs/code fidelity gap.

```
desc: 'The number of minutes CloudFormation continues to monitor the rollback trigger alarms after the deployment completes, before cleaning up old resources. Must be between 0 and 180. Requires --rollback-trigger-alarm-arns to have any effect.',
```

**Why it matters:** cause: help text and code diverge on the same behavior -> mechanism: a user reading `--help` is told the flag is a no-op without the other, and acts accordingly -> consequence: they are misled into believing the flag is safe to use alone, which is precisely the scenario that clears their triggers.

**Suggestion:** Once the CLI gating logic is fixed to genuinely require `--rollback-trigger-alarm-arns` for `--monitoring-time-minutes` to take effect, this text becomes accurate and needs no further change; if a standalone-with-warning approach is chosen instead, update the text accordingly.

_[code-review-sage]_

packages/@aws-cdk/toolkit-lib/lib/util/cloudformation.ts
```diff
@@ -7,6 +10,70 @@ export function validateSnsTopicArn(arn: string): boolean {
...
+ */
+export function toCloudFormationRollbackConfiguration(config?: RollbackConfiguration): CfnRollbackConfiguration | undefined {
+  if (config === undefined) {
+    return undefined;
+  }
+
+  const triggers = config.triggers ?? [];
+  if (triggers.length > MAX_ROLLBACK_TRIGGERS) {
+    throw new ToolkitError(
+      'InvalidRollbackConfiguration',
```
momo-bee: 🟡 **Monitoring time is not validated as an integer, allowing fractional minutes CloudFormation may reject or truncate.**

The range check accepts any finite number in [0, 180], including non-integers (e.g. 12.5) and NaN is falsy on both comparisons so `NaN < 0` and `NaN > 180` are both false, letting `NaN` through silently.

```
const monitoringTime = config.monitoringTimeInMinutes;
if (monitoringTime !== undefined && (monitoringTime < 0 || monitoringTime > MAX_MONITORING_TIME_IN_MINUTES)) {
  throw new ToolkitError(...)
}
```

**Why it matters:** cause: no integer/finite check on `monitoringTimeInMinutes` -> mechanism: a fractional or NaN value passes local validation and is forwarded in the `CreateChangeSetCommand`/`UpdateStackCommand` payload -> consequence: the failure moves from a clear, immediate `ToolkitError` at the CLI boundary to a less clear CloudFormation API rejection (or, for NaN, a request with `MonitoringTimeInMinutes: NaN` that may serialize unpredictably), making the error harder to diagnose for the user.

**Suggestion:** Add `!Number.isInteger(monitoringTime)` (or `Number.isFinite`) to the guard, e.g. `if (monitoringTime !== undefined && (!Number.isInteger(monitoringTime) || monitoringTime < 0 || monitoringTime > MAX_MONITORING_TIME_IN_MINUTES))`.

_[code-review-sage]_

mrgrain:
## Design verdict: PASS
- **Risk / criticality:** medium design risk, critical review depth (blast radius: LARGE — 524 LOC added, touches `packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/stack-helpers.ts`, import fan-out 12)
- **Problem:** CDK deploy has no native way to configure CloudFormation rollback triggers (alarm-based auto-rollback on stack create/update), forcing users to manage them out-of-band.
- **Why it matters:** Users who want automatic rollback safety on deploy — a widely-used CFN feature, tracked upstream as `aws/aws-cdk#5170` — currently can't express it through CDK.
- **Solution assessment:** Mirrors the existing `notificationArns` pattern end-to-end (interface → validation → threading through toolkit-lib and CLI → change-detection in `canSkipDeploy`) — low architectural risk, easy to reason about. The one place the mirroring breaks down is the CLI's opt-in condition combined with the `undefined`-handling asymmetry between `notificationArns` and `rollbackConfiguration` — see the 🔴 below.

## 🔴 Must-fix (1)

### `--monitoring-time-minutes` alone silently clears a stack's existing rollback triggers
**File:** `packages/aws-cdk/lib/cli/cli.ts:442`

```ts
const rollbackConfiguration = (args.rollbackTriggerAlarmArns !== undefined || args.monitoringTimeMinutes !== undefined)
  ? {
    triggers: (args.rollbackTriggerAlarmArns ?? []).map((arn: string) => ({ arn })),
    monitoringTimeInMinutes: args.monitoringTimeMinutes,
  }
  : undefined;
```

Passing `--monitoring-time-minutes` without `--rollback-trigger-alarm-arns` builds `{ triggers: [] }` — which, per the code's own documented semantics ("`{ triggers: [] }`: CDK manages it and clears any existing triggers"), is a real management directive, not a no-op. `canSkipDeploy` only skips the rollback check when `rollbackConfiguration === undefined`, so this gets forwarded to CloudFormation and wipes any alarms already on the stack.

This directly contradicts the flag's own `--help` text: *"Requires --rollback-trigger-alarm-arns to have any effect."*

**Consequence:** a user runs `cdk deploy --monitoring-time-minutes 30` on a stack that already has rollback alarms, intending only to extend the monitoring window — and CloudFormation silently removes all rollback alarms on the next deploy, with no warning, removing an operational safety mechanism that won't be missed until the next incident it should have caught.

**Suggestion:** gate rollback-management purely on `args.rollbackTriggerAlarmArns !== undefined`; either ignore a lone `--monitoring-time-minutes` or throw a `ToolkitError` telling the user it requires `--rollback-trigger-alarm-arns` — matching what the help text already claims.

## 🟡 Should-fix (3)

### No test covers `--monitoring-time-minutes` without `--rollback-trigger-alarm-arns`
**File:** `packages/aws-cdk/test/commands/deploy.test.ts:431`
The new CLI tests cover "ARNs provided" and "no flags" but never the monitoring-time-only combination — exactly the path with the defect above. Add a test once the intended behavior (ignore, or throw) is decided. Add IO Snapshots for these as well.

### Help text claims a flag dependency the implementation doesn't enforce
**File:** `packages/aws-cdk/lib/cli/parse-command-line-arguments.ts:507`
Same root cause as the 🔴 above, surfacing as a docs/code fidelity gap — the `--help` text is currently *aspirational*, not descriptive. Resolves itself once the CLI gating is fixed.

### Monitoring time isn't validated as an integer
**File:** `packages/@aws-cdk/toolkit-lib/lib/util/cloudformation.ts:46`
```ts
const monitoringTime = config.monitoringTimeInMinutes;
if (monitoringTime !== undefined && (monitoringTime < 0 || monitoringTime > MAX_MONITORING_TIME_IN_MINUTES)) {
  throw new ToolkitError(...)
}
```
Accepts fractional values (e.g. `12.5`) and lets `NaN` through silently (both comparisons are false for `NaN`). Moves the failure from a clear `ToolkitError` at the CLI boundary to a murkier CloudFormation API rejection. Suggest adding `!Number.isInteger(monitoringTime)` to the guard.

## Ship summary
**Not ready** — 1 must-fix: `--monitoring-time-minutes` alone silently clears a stack's existing rollback triggers, contradicting its own documented behavior. Design is otherwise sound (follows the established `notificationArns` pattern, proportionate scope, real linked feature request `aws/aws-cdk#5170`).

---
_Created by an AI Agent on behalf of @mrgrain_

mrgrain (changes requested):
see comment

PR #1885, opened 2026-08-23
feat: show the target environment in cdk diff stack headers

badmintoncryer:
Relates to #286

`cdk diff` printed only the stack name, so nothing in the output showed which account and region the diff was computed against. An env-agnostic stack diffed under the wrong profile looked identical to a correct one.

Each stack header now shows the resolved environment:

```console
Stack MyStackName (aws://123456789012/us-east-1)
```

- The value is the *resolved* environment, so env-agnostic stacks show their real target instead of `unknown-account`/`unknown-region`.
- `resolveEnvironment()` was already awaited on both paths to feed `ResourceMigrator`, so this adds no extra STS call.
- Only the root stack is annotated; nested stacks always share their parent's environment.
- `cdk diff --template` is unchanged, since that path never contacts AWS.

Shown by default rather than behind a flag. Happy to gate it if you'd prefer.

Not `Fixes`, because the issue also asks for per-resource region attribution, which a CloudFormation template doesn't carry. This covers the stack-level half.

### Checklist
- [x] This change contains a major version upgrade for a dependency and I confirm all breaking changes are addressed

---


packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
```diff
@@ -232,6 +243,7 @@ export class DiffFormatter {
       changeSet: this.templateInfo.changeSet,
       mappings: this.mappings,
       logicalIdMap: buildLogicalToPathMap(this.templateInfo.newTemplate).toPath,
+      environmentSuffix: this.environmentSuffix,
```
mrgrain: let's pass the environmentString around, not the formatted suffix. or maybe we can just use `this.templateInfo.environment` anyway

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
```diff
@@ -256,7 +270,7 @@ export class DiffFormatter {
     try {
       // must output the stack name if there are differences, even if quiet
       if (stackName && (!options.quiet || !diff.isEmpty)) {
-        stream.write(format(`Stack ${chalk.bold(stackName)}\n`));
+        stream.write(format(`Stack ${chalk.bold(stackName)}${environmentSuffix ?? ''}\n`));
```
mrgrain: this is where we should call format on the environment string

packages/@aws-cdk/toolkit-lib/lib/api/diff/diff-formatter.ts
```diff
@@ -344,15 +359,17 @@ export class DiffFormatter {
...
-    const { oldTemplate, newTemplate, stackName, nestedStacks, changeSet, logicalIdMap } = params;
+    const { oldTemplate, newTemplate, stackName, nestedStacks, changeSet, logicalIdMap, environmentSuffix } = params;
 
     const diff = this.computeDiff(stackName, oldTemplate, newTemplate, changeSet, this.mappings);
     const permissionChangeType = permissionTypeFromDiff(diff);
 
     const stream = new StringWriteStream();
     if (!options.quiet || permissionChangeType !== PermissionChangeType.NONE) {
-      stream.write(format(`Stack ${chalk.bold(stackName)}\n`));
+      stream.write(format(`Stack ${chalk.bold(stackName)}${environmentSuffix ?? ''}\n`));
```
mrgrain: this is where we should call format on the environment string

mrgrain (changes requested):

badmintoncryer:
Thank you for your review @mrgrain !

I've resolved your comments.

mrgrain (approved):

PR #1896, opened 2026-08-25
refactor(cli): consolidate stack selection on toolkit-lib selectors

mrgrain:
### Fixes

Stack selection with `PATTERN_MUST_MATCH` reported a misleading "Stack selection is ambiguous" error when nothing matched; it now reports "No stacks match the name(s) ...", the message the CLI has always shown. Selection with `failOnEmpty: false` from an app without any stacks now returns an empty selection for every strategy, as documented, instead of throwing confusing errors for some of them.

CLI behavior is otherwise unchanged: same selection results, same messages, protected by the existing test suites.

### Maintainer improvements

The CLI no longer carries its own legacy stack selection implementation next to toolkit-lib's. Commands now describe their selection as a single toolkit-lib `StackSelector` built in `cli.ts`, and `CdkToolkit` passes it straight into `selectStacksV2()` — removing the legacy `CloudAssembly`/`BaseStackAssembly` classes, the duplicated enums and validation checks, and the layer-by-layer translation between the two models. Common selectors are shared between production code and tests, and are unit tested.

---


packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-selectors.ts
```diff
@@ -1,6 +1,76 @@
...
+
+/**
+ * Select all top-level stacks
+ */
+export function selectAllTopLevel(expand: ExpandStackSelection = ExpandStackSelection.UPSTREAM): StackSelector {
+  return {
+    strategy: StackSelectionStrategy.MAIN_ASSEMBLY,
+    expand,
+  };
+}
```
iankhou, line 65: In the old code [we ignored `expand` when allTopLevel was false (for diff it always was) and patterns.length was not > 0](https://github.com/aws/aws-cdk-cli/pull/1896/changes#diff-460061298c994a2e2fa25bd1c7dfd960089ddfa848b8d3f1f9521a3d72d37529L106-L112). With this change, we collapse that condition tree and thus pass `expand` for that "third" case since we go through [this condition](https://github.com/aws/aws-cdk-cli/pull/1896/changes#diff-2526cd99ab2386c4ccf4aa857f11eb29ba833ce3d8a24386afeeb066b55fa0a2R1391) via `selectTopLevelOrMatchingStacks`. So this is a behavior change that we also don't currently test for.

iankhou: The new behavior seems to be more "correct," though. But worth a note in the description for the behavior change.

iankhou (changes requested):

iankhou (approved):

PR #1898, opened 2026-08-25
feat(integ-runner): detect bootstrap errors and retry tests in valid regions

mrgrain:
The integ-runner now detects when a test failed only because its AWS environment is not bootstrapped. Previously a single non-bootstrapped region would fail every test scheduled into it with a cryptic deployment error and keep receiving more tests, ruining the whole run.

When a bootstrap error is detected, the integ-runner now:

- stops scheduling tests into the affected environment
- retries the affected tests in the remaining environments
- fails with a clear message if no environments are left
- prints a summary with the fix for each skipped environment, e.g. `cdk bootstrap --profile dev aws://123456789012/eu-west-1`

toolkit-lib now reports bootstrap problems with a new `BootstrapError` carrying the affected account and region. 

Re-work of #1115 

### Checklist

---


packages/@aws-cdk/integ-runner/lib/workers/environment-pool.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * Unique key for a profile+region combination
+ *
+ * The account is deliberately not part of the key: it is not known upfront
+ * and the profile+region pair already identifies the credentials used.
+ */
+function key(env: TestEnvironment): string {
+  return JSON.stringify([env.profile, env.region]);
```
mrgrain: there should be a better way to create a key...

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts
```diff
@@ -790,7 +790,12 @@ export class Deployments {
     try {
       await envResources.validateVersion(requiresBootstrapStackVersion, bootstrapStackVersionSsmParameter);
     } catch (e: any) {
-      throw new ToolkitError('BootstrapVersionValidation', `${stackName}: ${formatErrorMessage(e)}`);
+      if (ToolkitError.isBootstrapError(e)) {
+        // Preserve the error type (and the environment it carries), so that
+        // callers can still detect a bootstrap problem programmatically.
+        throw new BootstrapError(e.name, `${stackName}: ${formatErrorMessage(e)}`, e.environment, e);
+      }
```
mrgrain: why not just re throw the original?

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
```diff
@@ -1086,6 +1086,11 @@ export class Toolkit extends CloudAssemblySourceBuilder {
         // "bold(stackname) failed: ResourceNotReady: <error>"
         const code = ToolkitError.isToolkitError(e) ? e.name : 'DeployStackFailed';
         const newMessage = [`❌  ${chalk.bold(stack.stackName)} failed:`, ...(e.name ? [`${e.name}:`] : []), e.message].join(' ');
+        if (ToolkitError.isBootstrapError(e)) {
+          // Preserve the error type (and the environment it carries), so that
+          // callers can still detect a bootstrap problem programmatically.
+          throw new BootstrapError(code, newMessage, e.environment, e);
+        }
```
mrgrain: why not re-throw the original?

packages/@aws-cdk/integ-runner/lib/workers/environment-pool.ts
mrgrain: Fixed — now `${profile ?? ""}/${region}`, since regions never contain a slash.

I went with `JSON.stringify` initially only to avoid reasoning about separators at all, but that is over-engineering for two known-shape fields, and it makes the key unreadable in a debugger.

One thing worth being explicit about: the account is deliberately *not* part of the key. It is unknown when the pool is built (we only learn it from a failure), so including it would make `isAvailable({profile, region})` and `isAvailable({profile, region, account})` different environments and the removal would not stick. There is a test covering that.

packages/@aws-cdk/toolkit-lib/lib/api/deployments/deployments.ts
mrgrain: Changed to a plain `throw e`, but flagging the tradeoff rather than pretending it is free: the only thing this wrapper does is prepend the stack name, so rethrowing drops it for bootstrap errors.

I think that is the right call and not just simpler — a bootstrap problem is environment-scoped, not stack-scoped. Naming one arbitrary stack out of the batch that happened to hit it first is misleading, since every stack in that environment is equally affected. To make sure nothing is actually lost, I added the environment to the messages themselves (`... found 5 in aws://123456789012/us-east-1`), which is the information a user needs and the stack name never gave them.

If you would rather keep the stack name, the alternative is `BootstrapError.withStackName(e, stackName)` doing the re-wrap while preserving type, environment, and cause. Say the word and I will switch — but I would not add that constructor for its own sake.

packages/@aws-cdk/toolkit-lib/lib/toolkit/toolkit.ts
mrgrain: Changed to a plain `throw e`. Same reasoning as the `Deployments` one, but here it is a visible CLI behaviour change that deserves a second opinion.

Bootstrap errors now bypass the `❌  <stackname> failed: ...` banner and surface their own message directly. I think that reads better — the banner frames a bootstrap problem as a deployment failure of one particular stack, when nothing was ever deployed and the fix is environment-wide. The integration test that pins that exact string is about `ResourceNotReady`, so it is unaffected.

The cost is one inconsistency: a bootstrap failure is the only deploy-time error that does not get the banner. Happy to wrap it back in if you would rather keep the output uniform.

rix0rrr (approved):

PR #1909, opened 2026-08-28
fix(toolkit-lib): stack selection ignores all but one exclusion pattern

### LINKED ISSUE #1908: (cli): only one stack exclusion pattern takes effect
### Describe the bug

Stack patterns are matched with picomatch, which reads a leading `!` as a negation, so a single `!Stack` already selects everything but that stack. Each pattern is matched on its own and the results are unioned, so a second exclusion undoes the first: `!A` contributes every stack except A, `!B` contributes every stack except B, and the union is every stack in the app.

Mixing an exclusion with a regular pattern goes wrong the same way. `StackA !Canary` returns everything except Canary, instead of just StackA.

### Expected Behavior

`cdk ls '!Prod/Canary' '!Prod/StackB'` lists neither Canary nor StackB.

### Current Behavior

Both are listed, along with everything else.

### Reproduction Steps

An app with four stacks: `DevStack`, `Prod/StackA`, `Prod/StackB`, `Prod/Canary`.

```console
$ cdk ls '!Prod/Canary'
DevStack
Prod/StackA
Prod/StackB

$ cdk ls '!Prod/Canary' '!Prod/StackB'
DevStack
Prod/StackA
Prod/StackB
Prod/Canary

$ cdk ls 'Prod/StackA' '!Prod/Canary'
DevStack
Prod/StackA
Prod/StackB
```

The same applies to `deploy`, `diff`, `synth` and `destroy`, which all take the same patterns.

### Possible Solution

Match the non-excluding patterns first, then subtract everything the excluding ones match, rather than unioning both. `!(...)` needs to stay a regular extglob.

Related to #758, which asks for a way to run against everything but a named stack.

### CDK CLI Version

2.1139.0

### Framework Version

_No response_

### Node.js Version

v20.19.2

### OS

macOS

### Language

TypeScript

badmintoncryer:
Fixes #1908
Relates to #758

Stack patterns are matched with picomatch, which reads a leading `!` as a negation, so `cdk deploy '!Stack'` already works today. Each pattern was matched on its own and the results unioned, so a second exclusion undid the first: `!A` contributes everything but A, `!B` contributes everything but B, and the union is every stack in the app. `StackA !Canary` went wrong the same way, returning everything except Canary instead of just StackA.

The non-excluding patterns are now matched first, and everything the excluding ones match is subtracted from that. Exclusions on their own start from the whole assembly, so `!Stack` keeps meaning "everything but Stack".

```console
$ cdk ls '!Prod/Canary' '!Prod/StackB'    # before: all four stacks
DevStack
Prod/StackA
```

- Patterns without a `!` take exactly the path they took before, and no existing test needed changing.
- The new selection is always a subset of the old one, so nothing gets deployed or destroyed that wasn't before. Checked over every combination of up to three patterns against a six-stack app: 1463 combinations, no case where a stack was added.
- The one behaviour people could be relying on is the union that made multiple exclusions a no-op. That reads as a bug rather than something to keep, but say the word if you'd rather it warn for a release first.
- `!(...)` is extglob syntax that already means "anything but", so it stays a regular pattern.
- An empty pattern list still selects nothing, leaving `PATTERN_MUST_MATCH` to reject it.
- `suggestionsForPatterns` asked whether a pattern matched the resulting selection, which is always true for an exclusion, so a typo like `!Cnary` was silently ignored. Exclusions are now checked against the assembly instead.
- Excluding a stack that a selected stack depends on does not drop it from the deployment; expansion still pulls it back in, as `--exclusively` exists for. Covered by a test and mentioned in the README.

Only `toolkit-lib` changes. The CLI passes its positional arguments straight through as `patterns`, so `deploy`, `diff`, `synth`, `destroy` and `ls` all pick this up.

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
@@ -205,11 +205,16 @@ export class StackAssembly implements IReadableCloudAssembly {
   private suggestionsForPatterns(patterns: string[], matched: StackCollection): Record<string, string[]> {
     const suggestions: Record<string, string[]> = {};
     for (const pattern of patterns) {
-      if (matched.stackArtifacts.some((stack) => picomatch(stack.hierarchicalId, pattern))) {
+      // An exclusion never matches the selection it produced, so judge it by what it removes.
+      const exclusion = isExclusionPattern(pattern);
+      const effective = exclusion ? pattern.slice(1) : pattern;
+      const haystack = exclusion ? this.allStacks : matched.stackArtifacts;
+
+      if (haystack.some((stack) => picomatch(stack.hierarchicalId, effective))) {
         continue;
       }
       suggestions[pattern] = this.allStacks
-        .filter((stack) => picomatch(stack.hierarchicalId.toLowerCase(), pattern.toLowerCase()))
+        .filter((stack) => picomatch(stack.hierarchicalId.toLowerCase(), effective.toLowerCase()))
         .map((stack) => stack.hierarchicalId);
     }
     return suggestions;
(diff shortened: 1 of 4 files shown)


mrgrain (changes requested):
I don't think we are keen to maintain a custom extension to our glob matching library.

What's not okay with writing this?

```ts
cdk ls '!(Prod/Canary|Prod/StackB)'
```

mrgrain:
Although I agree that our current way of matching is strange. 🤔 

```
    const matchingPattern = (pattern: string) => (stack: cxapi.CloudFormationStackArtifact) => picomatch(stack.hierarchicalId, pattern);
    const matchedStacks = flatten(patterns.map(pattern => stacks.filter(matchingPattern(pattern))));
```

I am also concerned about breaking users.

mrgrain:
I guess today we are treating spaces as `OR` and you are suggesting a targeted `AND ` behavior. But I am not convinced that moving away from a clear behaviourally rule is helpful.

badmintoncryer:
@mrgrain 

You're right, `!(A|B)` covers it. I went in trying to write `!A !B`, and when that didn't work I took it for a problem. I just hadn't thought of the other form. 
And fair point on the rule, keeping spaces as a plain `OR` is worth more than the shorthand I was after.

Closing this. Thanks for taking a look!

mrgrain:
Happy to take a PR for better documentation!

mrgrain:
```
// micromatch-style semantics: OR over positives, AND over negations
function createMatcher(patterns, options) {
  const positives = [];
  const negatives = [];

  for (const pattern of patterns) {
    // let picomatch's own parser decide whether the pattern is negated;
    // scan() strips the '!' prefix(es) and handles edge cases like '!(a|b)' and '!!foo'
    const { negated, input, prefix } = pm.scan(pattern);
    if (negated) {
      negatives.push(pm(input.slice(prefix.length), options));
    } else {
      positives.push(pm(pattern, options));
    }
  }

  return (item) => {
    const included = positives.length === 0 || positives.some((m) => m(item));
    const excluded = negatives.some((m) => m(item));
    return included && !excluded;
  };
}
```

PR #1920, opened 2026-09-01
feat: exclude stacks from the selection with `!` patterns

### LINKED ISSUE #758: aws-cdk: Add --except flag in CDK CLI
### Describe the feature

cdk diff, cdk synth, cdk deploy all have option either to deploy all or deploy specific stack.

There must be an option where I do a cdk diff --except abc-stack where it does a diff for all the stacks except abc-stack.

### Use Case

When you create cdk during ci/cd pipeline, there might be situations where you will first create 4 stacks and 1 stack will be dependent on the output of those 4 stacks created which we might reference resources.

Example, in my case - 
1. I create VPC, Elastic Beanstalk.
2. This elastic beanstalk creates an internal ALB.
3. I need a privatelink connectivity to this internal ALB so I create an NLB in another stack.
4. To reference the ALB created, we need to get information like VPC of the ALB etc to create NLB in. 
5. We use fromLookup with Beanstalk Env name/tag to do that (since Beanstalk does not provide output for Load Balancer properties).
6. We also add ALB to NLBs Target Group which Requires ALB Construct.

In this case, if I do a diff during my first run, it will fail since my ALB with Beanstalk has not been created to populate the Context using fromLookup.

### Proposed Solution

cdk diff --except abc-stack 
cdk deploy --except abc-stack 
cdk synth --except abc-stack 

So either I create 2 stages where I do cdk diff --except nlb-stack and cdk deploy --except nlb-stack in 1st stage and cdk diff nlb-stack and cdk deploy nlb-stack in 2nd stage in my ci cd pipeline.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.99.1

MacOS

### LINKED ISSUE #1908: (cli): only one stack exclusion pattern takes effect

badmintoncryer:
Fixes #1908
Resolves #758

Implements @mrgrain's snippet from #1909. (That PR could not be reopened after the force push, so this is a new one.)

`picomatch.parse().negated` splits the patterns into positives and negations, and the selection is OR over the positives, AND over the negations. Every pattern is compiled exactly as picomatch defines it - a negated matcher accepts the stacks that survive it - so nothing is ever stripped or rewritten, and `!(A|B)` stays an extglob.

```console
$ cdk ls '!Prod/Canary' '!Prod/StackB'   # before: all 4 stacks
DevStack
Prod/StackA
```

Two changes from the snippet:

- `scan().negated` reports `true` for `!!Stack` where picomatch matches it as a positive, and slicing the `!` prefix off turns `'!'` into an empty pattern that picomatch refuses to compile. `parse().negated` is the flag the matcher itself uses, and compiling the raw pattern sidesteps the rest.
- `positives.length === 0` meaning "everything" would make `patterns: []` select every stack, which `select behavior: none` in `cloud-assembly.test.ts` pins down. Now it needs at least one negation.

Also:

- Patterns without `!` are unchanged and no existing test needed changes. The new selection is always a subset of the old one, so the change only ever narrows: anyone leaning on the union bug, e.g. `'Prod/*' '!Prod/Canary'`, will stop deploying `Canary`. That is the point of the fix, but say the word if you'd rather have it warn for a release first.
- The `does not exist` warning compared each raw pattern against the selection it produced. With subtraction in play that turns false (`'Stack1' '!Stack1'` would warn that `Stack1` does not exist), so patterns are now checked against the assembly, negations with the `!` off - which also gets `destroy '!stack1'` its casing suggestion.
- Excluding a dependency of a selected stack does not drop it, `--exclusively` still does. Test + README.

toolkit-lib only, so `deploy` / `diff` / `synth` / `destroy` / `ls` all get this.

### Checklist

---


packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
```diff
@@ -267,6 +266,43 @@ export class StackAssembly implements IReadableCloudAssembly {
...
+ * `!`s cancels out, as it does in picomatch itself.
+ */
+function parsePattern(pattern: string) {
+  const { prefix, input } = picomatch.scan(pattern);
+
+  return {
+    source: pattern,
+    excludes: prefix.replace(/[^!]/g, '').length % 2 === 1,
+    glob: input.slice(prefix.length),
+  };
```
mrgrain: Right. We really really don't want this kind of code. It's a code smell and seems to indicate that we are trying to shoehorn something into picomatch that doesn't work. Thanks for catching the edge cases though.

Sounds like there's a bug in picomatch with regards to scan and negate? Should we try and fix this upstream?

packages/aws-cdk/README.md
```diff
@@ -293,6 +293,18 @@ In order to deploy them, you can list the stacks you want to deploy. If your app
...
+
+```console
+$ # every stack except NlbStack
+$ cdk deploy '!NlbStack'
+
+$ # every stack under Prod, except the canary
+$ cdk deploy 'PipelineStack/Prod/**' '!PipelineStack/Prod/Canary'
+```
+
+The selection is the union of the other patterns, minus everything the exclusions match; `!(...)` is extglob syntax, not an exclusion. When you only pass exclusions, they apply to every stack in the app. Note that `--all` cannot be combined with patterns, so use `**` when you want to spell out the starting point. Stacks that a selected stack depends on are still added to the deployment even if you excluded them; pass `--exclusively` (`-e`) to keep them out.
```
mrgrain: "`!(...)` is extglob syntax, not an exclusion." this is probably correct, but not sure what it is supposed to mean 😅 

"When you only pass exclusions, they apply to every stack in the app." similar here.

But don't worry about it. I'll adjust this before me merge.

mrgrain (changes requested):
Hmm. Not as easy as I hoped.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
badmintoncryer: Thanks. I've realized that Picomatch provides a `parse()` function that can correctly parse both `!A`(`negated`=`true`) and `!!A`(`negated`=`false`). I've updated to use this function, and I'll check it later. Please wait a while.

packages/aws-cdk/README.md
badmintoncryer: Thank you so much...

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
badmintoncryer: I've checked it over and this looks right to me now. Could you take another look?

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
```diff
@@ -267,6 +271,30 @@ export class StackAssembly implements IReadableCloudAssembly {
...
+ * Match the stacks a set of patterns selects: OR over the selecting patterns,
+ * AND over the negated ones. Every pattern is compiled exactly as picomatch
+ * defines it - a negated matcher accepts the stacks that survive it - and
+ * `parse().negated` decides which group a pattern belongs to.
+ *
+ * Negations on their own start from every stack, so `!Stack` selects every
+ * stack but that one. No patterns at all still selects nothing.
+ */
+function matcherFor(patterns: string[]): (hierarchicalId: string) => boolean {
+  const matchers = patterns.map(pattern => ({ negates: picomatch.parse(pattern).negated, matches: picomatch(pattern) }));
```
mrgrain: looks like `picomatch(pattern, undefined, true).state` is the same result as `picomatch.parse(pattern)` so we can save a parse here.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
```diff
@@ -267,6 +271,30 @@ export class StackAssembly implements IReadableCloudAssembly {
...
+  const positives = matchers.filter(matcher => !matcher.negates);
+  const negatives = matchers.filter(matcher => matcher.negates);
```
mrgrain: this probably can be a single loop (`for ... each` or `Array.reduce`).

mrgrain (changes requested):
some small performance optimizations please, otherwise great!

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
badmintoncryer: I've updated it to use `for ... of` and removed the `{ negates, matches }` wrapper object.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
```diff
@@ -267,6 +271,32 @@ export class StackAssembly implements IReadableCloudAssembly {
...
+ * `parse().negated` decides which group a pattern belongs to.
+ *
+ * Negations on their own start from every stack, so `!Stack` selects every
+ * stack but that one. No patterns at all still selects nothing.
+ */
+function matcherFor(patterns: string[]): (hierarchicalId: string) => boolean {
+  const positives: picomatch.Matcher[] = [];
+  const negatives: picomatch.Matcher[] = [];
+  for (const pattern of patterns) {
+    (picomatch.parse(pattern).negated ? negatives : positives).push(picomatch(pattern));
```
mrgrain:
```suggestion
    const parsed = picomatch(pattern, undefined, true);
    (parsed.state.negated ? negatives : positives).push(parsed);
```

I think this saves a parse?

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/private/stack-assembly.ts
mrgrain: @badmintoncryer I think this is still missing?

badmintoncryer: I'm sorry. I've misunderstood that `save` means `remain`. I've resolved it!

packages/aws-cdk/README.md
```diff
@@ -293,6 +293,18 @@ In order to deploy them, you can list the stacks you want to deploy. If your app
...
+
```
mrgrain: To deploy every stack in the app, use the `--all` flag or the wildcard `*`.
If your app has stacks inside stages, keep in mind how far each wildcard reaches:

- `*` matches one level, so it only picks up top-level stacks (same as `--all`)
- `**` matches any number of levels, so it picks up stacks at every depth

You can also combine these patterns.
For example, to deploy everything inside the `Prod` stage of a pipeline:

```console
cdk deploy 'PipelineStack/Prod/**'
```

##### Excluding stacks

Put a `!` in front of a pattern to leave those stacks out.
Wrap the pattern in quotes so your shell leaves the `!` alone.

Every stack except `NlbStack`:

```
cdk deploy '!NlbStack'
```

Everything under `Prod`, except the canary:

```
cdk deploy 'PipelineStack/Prod/**' '!PipelineStack/Prod/Canary'
```

CDK picks the stacks your normal patterns match, then drops the ones your `!` patterns match.
If you only pass `!` patterns, they apply to every stack in the app.
`--all` doesn't work together with patterns, so use `**` instead.
An excluded stack still deploys if a selected stack needs it — add `--exclusively` (`-e`) to skip dependencies.
And `!(...)` is shell syntax, not a CDK exclusion.

##### Deploying stacks in parallel

badmintoncryer: Thank you very much. I've updated README.

packages/@aws-cdk/toolkit-lib/lib/api/cloud-assembly/stack-selector.ts
```diff
@@ -78,6 +78,14 @@ export interface StackSelector {
...
    * A list of patterns to match the stack hierarchical ids
    * Only used with `PATTERN_*` selection strategies.
+   *
+   * A pattern starting with `!` excludes the stacks it matches. The selection is
+   * the union of the other patterns, minus everything the excluding ones match;
+   * exclusions on their own start from every stack. `!(...)` is extglob syntax
+   * and is matched as a regular pattern.
+   *
+   * - `['!Prod/Canary']` selects every stack except `Prod/Canary`
+   * - `['Prod/**', '!Prod/Canary']` selects every stack under `Prod` but that one
```
mrgrain:
```suggestion
   * A list of patterns to match the stack hierarchical ids.
   * Only used with `PATTERN_*` selection strategies.
   *
   * Patterns are globs. `*` and `**` match different levels of stacks nested
   * in stages. A `!` in front removes those stacks from the selection.
   *
   * - `*` matches only top-level stacks, any stacks in stages are ignored
   * - `**` matches any number of levels, so it picks up all stacks in stages
   * - `MyStage/*` matches all stacks in a particular stage
   * - `!MyStack` matches all stacks except the excluded one
   *
   * Selection first picks all stacks the normal patterns match, then removes
   * any matches of negated patterns (starting with `!`) patterns match.
   * More technically: Positive patterns use OR; negations use AND; the order
   * of patterns is not preserved.
```

mrgrain (approved):

PR #1941, opened 2026-09-03
fix(toolkit-lib): throttled final event poll fails a successful deploy

### LINKED ISSUE #1777: (toolkit-lib): throttled final stack-event poll in monitor.stop() fails an already-successful deployment
## Describe the bug

`FullCloudFormationDeployment.monitorDeployment()` ends with an unguarded `finally`
(`packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts`):

```ts
let finalState = this.cloudFormationStack;
try {
  const successStack = await waitForStackDeploy(this.cfn, this.ioHelper, this.stackName);
  if (!successStack) { throw new DeploymentError(...); }
  finalState = successStack;   // <-- the deployment has SUCCEEDED here
} catch (e: any) {
  ...
  throw e;
} finally {
  await monitor.stop();        // <-- can throw, and a throw here discards the success
}
```

`monitor.stop()` issues one further `DescribeStackEvents` read:

`stop()` → `finalPollToEnd()` → `readNewEvents()` → `StackEventPoller.poll()` → `doPoll()`

On this path `readNewEvents()` has **no** `try`/`catch`, unlike the periodic `tick()`, which
catches, emits `CDK_TOOLKIT_E5500`, and carries on:

```ts
} catch (e) {
  await this.ioHelper.notify(IO.CDK_TOOLKIT_E5500.msg(
    util.format('Error occurred while monitoring stack: %s', e), { error: e },
  ));
}
```

Because a `throw` inside a `finally` discards whatever the `try` was about to return, a
transient `Throttling` on that final read **throws away an already-successful deployment
result and propagates as a stack failure**. The stack is `UPDATE_COMPLETE` in CloudFormation
while the toolkit reports it failed and `cdk deploy` exits non-zero.

The decisive point is that this final poll is purely presentational — its own comment says so:

```ts
// Do a final poll for all events. This is to handle the situation where DescribeStackStatus
// already returned an error, but the monitor hasn't seen all the events yet and we'd end
// up not printing the failure reason to users.
await this.finalPollToEnd(oldMonitorId);
```

It exists to make console output complete. It cannot affect whether the deployment succeeded,
so it should not be able to change the reported outcome.

The same unguarded shape appears at the `destroyStack` monitor in the same file, and
`Deployments.rollbackStack` has a third `monitor.stop()` call site.

## Expected Behavior

A read-only, post-completion, presentation-only poll cannot change a deployment's outcome. If
the final poll fails, the toolkit reports the successful deployment and at most warns that
trailing stack events could not be fetched — the same treatment `tick()` already gives an
identical failure.

## Current Behavior

A stack that CloudFormation reports as `UPDATE_COMPLETE` is reported by the toolkit as failed,
and the process exits non-zero, aborting the remainder of a multi-stack deploy:

```
❌  <stack-name> failed: Throttling: Rate exceeded
```

CloudFormation's own stack-event history showed the stack reaching `UPDATE_COMPLETE` roughly two
and a half minutes *before* the toolkit reported it as failed, and the only error in the run was
a CloudFormation read throttle.

The `DescribeStackEvents` poller was demonstrably being throttled. Below is the non-fatal
`CDK_TOOLKIT_E5500` occurrence from the same run, from a global install of `aws-cdk@2.1133.0`
(stack name omitted; `$metadata` and `Error` were collapsed by Node's inspector):

```
Error occurred while monitoring stack: Throttling: Rate exceeded
    at ProtocolLib.getErrorSchemaOrThrowBaseException (/usr/local/lib/node_modules/aws-cdk/lib/index.js:37185:63)
    at AwsQueryProtocol.handleError (/usr/local/lib/node_modules/aws-cdk/lib/index.js:39041:65)
    at AwsQueryProtocol.deserializeResponse (/usr/local/lib/node_modules/aws-cdk/lib/index.js:39011:22)
    at async /usr/local/lib/node_modules/aws-cdk/lib/index.js:20535:24
    at async /usr/local/lib/node_modules/aws-cdk/lib/index.js:31100:22
    at async /usr/local/lib/node_modules/aws-cdk/lib/index.js:29786:40
    at async /usr/local/lib/node_modules/aws-cdk/lib/index.js:30715:26
    at async _StackEventPoller.doPoll (/usr/local/lib/node_modules/aws-cdk/lib/index.js:90499:26)
    at async _StackEventPoller.poll (/usr/local/lib/node_modules/aws-cdk/lib/index.js:90455:24) {
  '$fault': 'client',
  '$retryable': undefined,
  '$metadata': [Object],
  Type: 'Sender',
  Code: 'Throttling',
  Error: [Object]
}
```

Those offsets resolve against the published `aws-cdk@2.1133.0` bundle, so they are reproducible
from the artifact — `npm pack aws-cdk@2.1133.0` and check:

- `lib/index.js:90455` → `const events = await this.doPoll();`, inside `poll()` (defined at 90454)
- `lib/index.js:90499` → `await this.cfn.describeStackEvents({ StackName: ..., NextToken: ... })`,
  inside `doPoll()` (defined at 90492)

Note that 90499 is the *paginating* call inside `doPoll`'s loop, so a single logical poll can
issue several `DescribeStackEvents` requests, each with its own retry budget.

That occurrence was swallowed by `tick()` as designed, and the deploy continued. The unguarded
`finalPollToEnd` path reaches the identical call from `stop()`, where the same error is fatal
instead.

## Reproduction Steps

The defect is in the control flow and does not need real throttling to demonstrate — inject the
failure on the event poll while the stack reaches a successful terminal state. Sketch in the
style of the existing `toolkit-lib` tests (`test/_helpers/mock-sdk.ts`):

```ts
import {
  DescribeStackEventsCommand, DescribeStacksCommand, StackStatus,
} from '@aws-sdk/client-cloudformation';
import { mockCloudFormationClient, restoreSdkMocksToDefault } from '../_helpers/mock-sdk';

test('a throttled final event poll must not fail an already-successful deployment', async () => {
  restoreSdkMocksToDefault();

  // The stack reaches a successful terminal state, so waitForStackDeploy() resolves.
  mockCloudFormationClient.on(DescribeStacksCommand).resolves({
    Stacks: [{
      StackName: 'my-stack',
      StackStatus: StackStatus.UPDATE_COMPLETE,
      CreationTime: new Date(),
    }],
  });

  // Every DescribeStackEvents call throttles. tick() swallows these by design;
  // the final poll inside monitor.stop() does not.
  mockCloudFormationClient.on(DescribeStackEventsCommand).rejects(
    Object.assign(new Error('Rate exceeded'), {
      name: 'Throttling',
      $metadata: { httpStatusCode: 400 },
    }),
  );

  // Fails: rejects with `Throttling: Rate exceeded` out of the `finally`,
  // even though the deployment succeeded.
  await expect(deployments.deployStack({ stack: testStack({ stackName: 'my-stack' }) }))
    .resolves.toMatchObject({ type: 'did-deploy-stack' });
});
```

Note the CloudFormation client's own retry strategy will spend its budget on the rejecting mock,
so this wants fake timers (`test/_helpers/fake-time.ts`) to stay fast.

Organically, the trigger is any deploy that holds enough stacks in flight to saturate the
account-level CloudFormation read quota:

1. `cdk deploy --all --concurrency N` with `N` above ~20, all stacks in one account and region.
   The monitor polls `DescribeStackEvents` every 2 s per in-flight stack, so `N / 2` requests
   per second exceeds the account quota for that API.
2. Let a stack complete successfully while the account is throttling.
3. If that stack's final poll exhausts its retry budget, the successful stack is reported as
   failed and the whole deploy exits non-zero.

## Possible Solution

Guard inside `finalPollToEnd`, which covers the deploy, destroy, and rollback monitors from one
place and mirrors what `tick()` already does:

```ts
private async finalPollToEnd(monitorId: string) {
  try {
    if (this.readPromise) {
      await this.readPromise;
    }
    await this.readNewEvents(monitorId);
  } catch (e) {
    // A presentation-only final poll must never change the outcome of the operation
    await this.ioHelper.notify(IO.CDK_TOOLKIT_E5500.msg(
      util.format('Error occurred during final stack event poll: %s', e), { error: e },
    ));
  }
}
```

Guarding at the `finally { await monitor.stop(); }` call sites instead would cover deploy and
destroy but miss `Deployments.rollbackStack`.

Happy to open a PR if you'd prefer this placement, or a different one.

## Additional Information/Context

**Why this is easy to hit.** The `DescribeStackEvents` quota is **10 requests/second and is not
adjustable** — `aws service-quotas list-service-quotas --service-code cloudformation` reports
`Adjustable: false` for it (and for `DescribeStacks` and `GetTemplate`). With a 2 s interval per
in-flight stack, more than ~20 concurrent stack operations in one account and region
structurally exceeds that quota for this API alone, leaving SDK retries as the only thing
keeping the deploy alive.

**Operators cannot widen the retry window.** `SDK.cloudFormation()` hardcodes
`retryStrategy: new ConfiguredRetryStrategy(7, cappedExponentialBackoff(1000, 15_000))`, which
overrides `AWS_MAX_ATTEMPTS` and `AWS_RETRY_MODE`. That also means there is no way to opt into
`adaptive` mode, whose client-side rate limiter would suit this workload well.

**The tolerance for sustained throttling narrowed in 2.1122.0.** #1500 changed the
CloudFormation client's retry strategy:

```diff
-      retryStrategy: new ConfiguredRetryStrategy(11, (attempt: number) => 1000 * (2 ** attempt)),
+      retryStrategy: new ConfiguredRetryStrategy(7, cappedExponentialBackoff(1000, 15_000)),
```

That reduces attempts 11 → 7 *and* lowers the per-delay ceiling from `@smithy/util-retry`'s
`MAXIMUM_RETRY_DELAY` (20 s) to an explicit 15 s, shortening how long a single call chain can
ride out a throttling burst from roughly 131 s to roughly 45 s. I verified the boundary against
the published artifacts: `aws-cdk@2.1121.0` carries the old values, `2.1122.0` the new ones.

That change looks entirely reasonable for the problem it solved — the PR notes an uncapped
backoff on this client could make the SDK spend tens of minutes, which is a real bug. Flagging
it only because it means less headroom for the failure described here, and because the
consequence of running out of headroom is currently an incorrectly failed deployment rather
than a warning. The polling interval and the number of API calls per poll are unchanged, so
this is purely a tolerance change, not a request-rate change.

**Relationship to #1709 / #1710.** #1710 made the polling interval configurable, which lowers
the request *rate*. This issue is about the *consequence* when a poll does fail, which #1710
does not change — and the failure described here was observed on a CLI that already contained
#1710.

`stackEventPollingInterval` is also unreachable for `cdk deploy` users. In the published
`aws-cdk@2.1133.0` bundle it appears only as `stackEventPollingInterval: options.…` (four
threading sites), never sourced from `argv`, and the bundle contains no corresponding CLI
option string. So the CLI still polls at the 2 s default with no way to change it, and the
`toolkit-lib` escape hatch does not help anyone invoking `cdk` directly.

**Related.** #879 is the analogous unguarded-monitor report for `cdk watch` log monitoring.

## CDK CLI Version
```
2.1133.0
```
*(version confirmed by resolving the stack-trace offsets above against the published bundle;
defect confirmed by inspection of that bundle's `finalPollToEnd` and of `main`; also present in
2.1129.0)*

## Node.js Version
```
24.x where the failure was observed; v24.18.0 for the source analysis
```

## OS
```
Observed on Linux arm64 (container). Source analysis and version verification on macOS 26.5.2 (Darwin arm64).
```

## Language
TypeScript

---

*Authored by Claude Code (Anthropic's agentic CLI). This is AI-assisted analysis of the public
`aws-cdk-cli` source, prompted by a failure seen in a CI pipeline running `cdk deploy --all` at
high `--concurrency`. The source reading, reasoning, and text above are the model's work; the
reproduction sketch has not been executed. Please review the reasoning on its merits rather
than assuming human verification.*

polothy:
> [!NOTE]
> **This PR was created by AI** (Claude Code, Anthropic's agentic CLI), including the code, tests and this description, working from the analysis in #1777. A human directed the work and reviewed the result, but the reasoning and wording below are the model's. Please review it on its merits rather than assuming human verification of every claim.

fixes #1777

That issue has the full diagnosis, the throttling arithmetic and the reproduction; this description covers only what was implemented.

## The fix

`StackActivityMonitor.finalPollToEnd()` is now guarded, which is the one place all three `monitor.stop()` call sites go through — `FullCloudFormationDeployment.monitorDeployment`, `destroyStack` (both in `deploy-stack.ts`) and `Deployments.rollbackStack`. Guarding at the call sites instead would have missed rollback.

Three changes in `packages/@aws-cdk/toolkit-lib`:

1. **`finalPollToEnd()` — the final read no longer propagates.** A failed `readNewEvents()` is reported as `CDK_TOOLKIT_W5500` and `stop()` returns normally. This read only completes the event log; the operation's outcome came from `DescribeStacks` via `waitForStackDeploy` long before.

2. **`finalPollToEnd()` — the awaited in-flight `readPromise` is caught.** This path was equally unguarded and is a separate failure mode: a poll that is still in flight when `stop()` is called, and rejects. It is swallowed rather than reported, because `tick()` awaits that same promise inside its own `try`/`catch` and has already emitted `CDK_TOOLKIT_E5500` for it. Without this catch, the rejection skips the final read entirely — the very flush `stop()` exists to perform.

3. **`tick()` — `readPromise` is cleared in a `finally`.** It was cleared only on the success path, so after any failed poll the field kept an already-rejected promise, and the next `stop()` re-raised it before doing anything. Now the field only ever holds a read that is genuinely in flight, which makes (2)'s "already reported by `tick()`" an invariant rather than an assumption.

Both (1) and (2) are load-bearing; see the mutation testing below.

## Error path

```
                     doPoll() ── DescribeStackEvents ── SDK retries exhausted (Throttling)
                        │
                        ▼
                    poll() → readNewEvents()
                        │
        ┌───────────────┴────────────────┐
        │                                │
   from tick()                    from stop() → finalPollToEnd()
        │                                │
        │                    ┌───────────┴────────────┐
        │                    │                        │
        │            await readPromise         await readNewEvents()
        │            (in-flight poll)             (the final read)
        │                    │                        │
   ┌────▼─────┐         ┌────▼─────┐            ┌─────▼──────┐
   │ caught   │         │ caught   │            │  caught    │
   │ E5500    │         │ ignored  │            │  W5500     │
   │ (before  │         │ (tick    │            │  (new)     │
   │  & after)│         │  reports)│            │            │
   └──────────┘         └──────────┘            └────────────┘
```

Before this PR, only the left branch was caught. Either branch under `stop()` threw, and because both call sites invoke `stop()` from a `finally`, the throw replaced the value the `try` was about to return:

```
finalPollToEnd() throws
  → monitor.stop() throws
  → finally { await monitor.stop() }      ← discards the successful return value
  → monitorDeployment() rejects
  → deployStack() rejects  →  Deploy action rejects  →  exec() rejects
  → cli() .catch()  →  process.exitCode = 1
```

## Before / after for `cdk` users

Scenario: a stack reaches `UPDATE_COMPLETE`, and the account is throttling `DescribeStackEvents` (10 req/s, non-adjustable) hard enough that a call exhausts its retries.

**Before**

```
 ✅  my-stack-a
 ❌  my-stack-b failed: Throttling: Rate exceeded
```

- The stack is `UPDATE_COMPLETE` in CloudFormation but reported as failed.
- `cdk deploy` exits `1`.
- Remaining stacks in a `--all` run are abandoned.
- `toolkit-lib` callers get a rejected `deployStack()` for a deployment that succeeded.

**After**

```
 ✅  my-stack-a
 ✅  my-stack-b
```

plus one warning on the affected stack:

```
[Warning at my-stack-b] Could not read the final stack events, the event log may be incomplete (Throttling). Run again with -v to see the full error.
```

- Exit code `0`; the rest of the `--all` run continues.
- Programmatic callers get `{ type: 'did-deploy-stack', ... }`.
- The only loss is presentational: trailing stack events for that stack may be missing from the printed log, which is what the warning says.

The full error, stack trace and all, is emitted as a `debug` message rather than printed by default, so `-v` shows:

```
Error occurred during final stack event poll: Throttling: Rate exceeded
    at ProtocolLib.getErrorSchemaOrThrowBaseException (...)
    at ... (11 more frames)
```

`CDK_TOOLKIT_W5500` still carries the untouched error object as its `ErrorPayload`, so programmatic consumers get the whole thing regardless of verbosity.

Unchanged: the periodic polls. `tick()` already swallowed throttles and re-polls 2s later, so missed events catch up there. Also unchanged: any failure from `DescribeStacks`/`waitForStackDeploy`, which is how the outcome is actually determined, still fails the deployment.

## Why `warn` and a new message code

The first draft reused `CDK_TOOLKIT_E5500`, matching what `tick()` emits for the identical failure on the identical API call. Changed to a new `CDK_TOOLKIT_W5500` (warn, `ErrorPayload`) because:

- `error` is the one level `CliIoHost.selectStreamFromLevel` does not redirect to stdout under `--ci`, so a green deploy wrote to stderr — noise for exactly the CI pipelines reporting this issue.
- Reporting `error` on an operation that succeeded misstates the outcome. #1777 asked for "at most warns".
- Sibling best-effort failures in the same directory already warn — `hook-result-details.ts` does so twice for the same "optional detail fetch failed, carry on" shape.

`tick()`'s existing `E5500` is deliberately left alone: it fires while the outcome is still unknown, so `error` remains defensible there. `docs/message-registry.md` was regenerated with `npx projen registry`. No API report change — `IO` lives under `lib/api/io/private`.

The warning names the error and points at `-v` rather than inlining it, following the shape #1950 established for `hook-result-details.ts` in this same directory. A throttle here would otherwise put 14 lines of minified SDK stack trace into a log for a deployment that succeeded.

## Trade-off worth reviewing

When the in-flight read fails, the final read now still runs, where previously the throw short-circuited it. Under sustained throttling that is one additional fully-retried `DescribeStackEvents` per stack — `ConfiguredRetryStrategy(7, cappedExponentialBackoff(1000, 15_000))`, so up to roughly a minute — inside a `finally`.

Kept deliberately: that read is the last chance to print the resource-level failure reasons for a stack that genuinely failed, so skipping it would lose diagnostics exactly when they matter. Cost is bounded per stack and concurrent across stacks, so it is tail latency rather than an additive delay, and the alternative it replaces is a wrongly failed deployment.

## Tests

- `test/api/stack-events/stack-activity-monitor.test.ts` — new `stack monitor, failures while reading events` block: a throttled final poll is reported as `W5500` and the full error as `debug` while `stop()` resolves and still emits `I5503`; a poll that fails while `stop()` is waiting on it does not fail `stop()` and does not skip the final read.
- `test/api/deployments/deploy-stack-event-poll-failures.test.ts` (new) — end-to-end: `describeStackEvents` throttling throughout, `deployStack` still returns `did-deploy-stack` for both `change-set` and `direct`, and `destroyStack` still returns its stack ARN.
- `test/api/deployments/cloudformation-deployments.test.ts` — `rollbackStack` under the same throttling still returns `{ success: true }`, pinning the third call site.

Verified by mutation, since these tests must fail for the right reason:

| Mutation | Result |
|---|---|
| Revert the guard entirely | 5 tests fail |
| Keep the swallow, drop the `W5500` notify | 5 tests fail |
| Keep the warning, drop the `debug` message | 1 test fails |
| Remove only the `readPromise` catch | 1 test fails (the in-flight case) |

`test/api/deployments` + `test/api/stack-events` pass (265 tests). Full `toolkit-lib` suite: 1989 pass, with 3 `bootstrap*` suites failing for a pre-existing local reason unrelated to this change (they need `post-compile` to copy `bootstrap-template.yaml`). `eslint` clean.

## Follow-ups, not in this PR

Both come from #1777 and are separate concerns:

- `cappedExponentialBackoff` is deterministic, so `ConfiguredRetryStrategy` replaces smithy's jittered default and throttled pollers re-collide at the same instants. Adding jitter would reduce how often the retry budget is exhausted in the first place.
- `stackEventPollingInterval` exists only in `toolkit-lib`; there is no `cdk deploy` flag for it, so CLI users cannot lower their request rate.

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
@@ -296,11 +296,13 @@ export class StackActivityMonitor {
     // of letting the failure propagate out of stop().
     try {
       await this.readNewEvents(monitorId);
-    } catch (e) {
+    } catch (e: any) {
+      const errorName = e instanceof Error ? e.name : String(e);
       await this.ioHelper.notify(IO.CDK_TOOLKIT_W5500.msg(
-        util.format('Error occurred during final stack event poll, event log may be incomplete: %s', e),
-        { error: e as any },
+        util.format('Could not read the final stack events, the event log may be incomplete (%s). Run again with -v to see the full error.', errorName),
+        { error: e },
       ));
+      await this.ioHelper.defaults.debug(util.format('Error occurred during final stack event poll: %s', e));
     }
   }
 
--- a/packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
@@ -217,7 +217,6 @@ export class StackActivityMonitor {
     try {
       this.readPromise = this.readNewEvents(this.monitorId);
       await this.readPromise;
-      this.readPromise = undefined;
 
       // We might have been stop()ped while the network call was in progress.
       if (!this.monitorId) {
@@ -228,6 +227,9 @@ export class StackActivityMonitor {
         util.format('Error occurred while monitoring stack: %s', e),
         { error: e as any },
       ));
+    } finally {
+      // Clear on both paths, so `readPromise` only ever holds a read that is still in flight.
+      this.readPromise = undefined;
     }
     this.scheduleNextTick();
   }
@@ -283,11 +285,23 @@ export class StackActivityMonitor {
     // the moment we were sure we weren't going to get any new events anymore
     // so we need to do a new one anyway. Need to wait for this one though
     // because our state is single-threaded.
-    if (this.readPromise) {
+    try {
       await this.readPromise;
+    } catch {
+      // A failure of the in-flight poll has already been reported by tick().
     }
 
-    await this.readNewEvents(monitorId);
+    // Reading events only completes the event log shown to the user; it cannot change
+    // whether the monitored operation succeeded. Warn that the log may be short instead
+    // of letting the failure propagate out of stop().
+    try {
+      await this.readNewEvents(monitorId);
+    } catch (e) {
+      await this.ioHelper.notify(IO.CDK_TOOLKIT_W5500.msg(
+        util.format('Error occurred during final stack event poll, event log may be incomplete: %s', e),
+        { error: e as any },
+      ));
+    }
   }
 
   /**
--- a/packages/@aws-cdk/toolkit-lib/docs/message-registry.md
+++ b/packages/@aws-cdk/toolkit-lib/docs/message-registry.md
@@ -120,6 +120,7 @@ Please let us know by [opening an issue](https://github.com/aws/aws-cdk-cli/issu
 | `CDK_TOOLKIT_W5400` | Hotswap disclosure message | `warn` | n/a |
 | `CDK_TOOLKIT_E5001` | No stacks found | `error` | n/a |
 | `CDK_TOOLKIT_E5500` | Stack Monitoring error | `error` | {@link ErrorPayload} |
+| `CDK_TOOLKIT_W5500` | Stack events could not be read; the reported event log may be incomplete | `warn` | {@link ErrorPayload} |
 | `CDK_TOOLKIT_I6000` | Provides rollback times | `info` | {@link Duration} |
 | `CDK_TOOLKIT_I6100` | Stack rollback progress | `info` | {@link StackRollbackProgress} |
 | `CDK_TOOLKIT_E6001` | No stacks found | `error` | n/a |
--- a/packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
+++ b/packages/@aws-cdk/toolkit-lib/lib/api/io/private/messages.ts
@@ -348,6 +348,11 @@ export const IO = {
     description: 'Stack Monitoring error',
     interface: 'ErrorPayload',
   }),
+  CDK_TOOLKIT_W5500: make.warn<ErrorPayload>({
+    code: 'CDK_TOOLKIT_W5500',
+    description: 'Stack events could not be read; the reported event log may be incomplete',
+    interface: 'ErrorPayload',
+  }),
 
   // 6: Rollback (6xxx)
   CDK_TOOLKIT_I6000: make.info<Duration>({
--- a/packages/@aws-cdk/toolkit-lib/test/api/deployments/cloudformation-deployments.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/deployments/cloudformation-deployments.test.ts
@@ -1102,6 +1102,26 @@ test('rollback stack allows rolling back from UPDATE_FAILED', async () => {
   expect(mockCloudFormationClient).toHaveReceivedCommand(RollbackStackCommand);
 });
 
+test('rollback stack is not failed by a throttled stack event poll', async () => {
+  // GIVEN - reading stack events fails throughout, including the final poll in monitor.stop()
+  givenStacks({
+    '*': { template: {}, stackStatus: 'UPDATE_FAILED' },
+  });
+  mockCloudFormationClient.on(DescribeStackEventsCommand).rejects(
+    Object.assign(new Error('Rate exceeded'), { name: 'Throttling' }),
+  );
+
+  // WHEN
+  const response = await deployments.rollbackStack({
+    stack: testStack({ stackName: 'boop' }),
+    validateBootstrapStackVersion: false,
+  });
+
+  // THEN - the rollback succeeded, and the final poll failure was only reported
+  expect(response).toMatchObject({ success: true });
+  ioHost.expectMessage({ level: 'warn', containing: 'Error occurred during final stack event poll' });
+});
+
 test('rollback stack allows continue rollback from UPDATE_ROLLBACK_FAILED', async () => {
   // GIVEN
   givenStacks({
--- a/packages/@aws-cdk/toolkit-lib/test/api/deployments/cloudformation-deployments.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/deployments/cloudformation-deployments.test.ts
@@ -1119,7 +1119,7 @@ test('rollback stack is not failed by a throttled stack event poll', async () =>
 
   // THEN - the rollback succeeded, and the final poll failure was only reported
   expect(response).toMatchObject({ success: true });
-  ioHost.expectMessage({ level: 'warn', containing: 'Error occurred during final stack event poll' });
+  ioHost.expectMessage({ level: 'warn', containing: 'the event log may be incomplete' });
 });
 
 test('rollback stack allows continue rollback from UPDATE_ROLLBACK_FAILED', async () => {
--- a/packages/@aws-cdk/toolkit-lib/test/api/deployments/deploy-stack-event-poll-failures.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/api/deployments/deploy-stack-event-poll-failures.test.ts
@@ -1,0 +1,117 @@
+import type { CloudFormationStackArtifact } from '@aws-cdk/cloud-assembly-api';
+import { deployStack, destroyStack } from '../../../lib/api/deployments/deploy-stack';
+import type { DeployStackOptions as DeployStackApiOptions } from '../../../lib/api/deployments/deploy-stack';
+import { CloudFormationStackDiagnoser } from '../../../lib/api/diagnosing/stack-diagnoser';
+import { NoBootstrapStackEnvironmentResources } from '../../../lib/api/environment';
+import { StackArtifactSourceTracer } from '../../../lib/api/source-tracing/private/stack-source-tracing';
+import { testStack } from '../../_helpers/assembly';
+import { FakeCloudFormation } from '../../_helpers/fake-aws/fake-cloudformation';
+import { advanceTime } from '../../_helpers/fake-time';
+import {
+  mockCloudFormationClient,
+  mockResolvedEnvironment,
+  MockSdk,
+  MockSdkProvider,
+  restoreSdkMocksToDefault,
+} from '../../_helpers/mock-sdk';
+import { TestIoHost } from '../../_helpers/test-io-host';
+
+const ioHost = new TestIoHost();
+const ioHelper = ioHost.asHelper('deploy');
+
+const FAKE_STACK = testStack({
+  stackName: 'withouterrors',
+  template: {
+    Resources: {
+      MyResource: {
+        Type: 'Test::Resource::Type',
+        Properties: {
+          Bar: 'Bar',
+        },
+      },
+    },
+  },
+});
+
+let sdk: MockSdk;
+let sdkProvider: MockSdkProvider;
+const fakeCfn = new FakeCloudFormation();
+
+beforeEach(() => {
+  fakeCfn.reset();
+  ioHost.clear();
+
+  sdkProvider = new MockSdkProvider();
+  sdk = new MockSdk();
+
+  restoreSdkMocksToDefault();
+  fakeCfn.installUsingAwsMock(mockCloudFormationClient);
+
+  jest.useFakeTimers();
+});
+
+afterEach(() => {
+  jest.useRealTimers();
+  jest.restoreAllMocks();
+});
+
+function standardDeployStackArguments(stack: CloudFormationStackArtifact = FAKE_STACK): DeployStackApiOptions {
+  const resolvedEnvironment = mockResolvedEnvironment();
+  return {
+    stack,
+    sdk,
+    sdkProvider,
+    resolvedEnvironment,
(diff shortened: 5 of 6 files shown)


mrgrain (approved):

packages/@aws-cdk/toolkit-lib/lib/api/stack-events/stack-activity-monitor.ts
```diff
@@ -287,11 +289,23 @@ export class StackActivityMonitor {
...
+    // Reading events only completes the event log shown to the user; it cannot change
+    // whether the monitored operation succeeded. Warn that the log may be short instead
+    // of letting the failure propagate out of stop().
+    try {
+      await this.readNewEvents(monitorId);
+    } catch (e) {
+      await this.ioHelper.notify(IO.CDK_TOOLKIT_W5500.msg(
+        util.format('Error occurred during final stack event poll, event log may be incomplete: %s', e),
+        { error: e as any },
+      ));
```
mrgrain, line 304: we can probably do better here in terms of the message, maybe just use the error name in the default warning and emit the full error as debug message that can be enabled using `-v`. We can tell the users about this flag in the message as well.

mrgrain (changes requested):
This looks good. Some minor feedback on the error and you'll need to rebase to make the build pass. Thank you!

polothy:
@mrgrain thanks for the swift feedback! Updated error reporting/handling (in https://github.com/aws/aws-cdk-cli/pull/1941/changes/b90581a59f3298daa6c8eb156389f8153b2d2489) and the branch was re-based.

mrgrain (approved):

polothy:
Thanks so much for the help @mrgrain !

PR #1945, opened 2026-09-04
fix(cloud-assembly-schema): validation crashes with Invalid URL on Node >= 24.20.0

mrgrain:
Since GitHub runners started rolling out Node.js 24.20.0 (via `node-version: lts/*`), builds have been failing intermittently in `@aws-cdk/cloud-assembly-schema` with `TypeError: Invalid URL` during schema validation. The failure appears intermittent because it depends on which Node version the runner has cached; it will become permanent as the rollout completes.

The root cause is a latent bug in jsonschema@1.5.0 (the latest release). When no `base` option is passed to `validate()`, it stringifies `undefined` into its internal base URI, producing ref URIs like `/undefined#/definitions/X`. It then re-parses these against the opaque base `thismessage::/`, which is invalid per the WHATWG URL spec. Older URL parsers tolerated this; the Ada 4.0.0 parser shipped in Node.js 24.20.0 correctly rejects it, so every manifest validation throws. Our schemas and refs are standard and not at fault.

Since no fixed jsonschema release exists, this applies the upstream fix from tdegrunt/jsonschema#424 via `yarn patch`: it splits the `#` fragment manually instead of round-tripping through `new URL()`. The resolution is scoped to the jsonschema dependency of `@aws-cdk/cloud-assembly-schema` and declared in `.projenrc.ts`, because yarn only reads resolutions from the monorepo root. Note that jsonschema is a bundled dependency, so the patched code ships in the published package. The patch should be removed once upstream releases a fix.

Verified by running the full `@aws-cdk/cloud-assembly-schema` test suite (38 passed) and by exercising manifest loading with a simulated strict URL parser that throws like Node 24.20.0 does.

### Checklist

---

### DIFF
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -392,6 +392,11 @@ repoProject.tasks.tryFind('build')!.spawn(gitSecretsScan);
 
 const repo = configureProject(repoProject);
 
+// Exclude dist from the NX cache to avoid restoring stale release artifacts
+repoProject.tryFindObjectFile('nx.json')!.patch(
+  pj.JsonPatch.remove('/targetDefaults/build/outputs/4'), // {projectRoot}/dist
+);
+
 interface GenericProps {
   private?: boolean;
 }
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -476,6 +476,12 @@ const cloudAssemblySchema = configureProject(
 );
 fixupTestTask(cloudAssemblySchema);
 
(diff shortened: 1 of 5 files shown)


ShadowCat567 (approved):
I do not like this because we have to remember that we did this when `jsonschema` is fixed but we need this fix sooner than `jsonschema` is going to roll it out :/

PR #1952, opened 2026-09-08
fix(toolkit-lib): guard against a stack with no metadata in countAssemblyResults

sharonyajain:
`countAssemblyResults` counts annotation error codes after synth by calling `Object.values(stack.metadata)` for every stack in `assembly.stacksRecursively`, with no null guard. When a synthesized stack has no metadata (`stack.metadata` is `undefined`), `Object.values(undefined)` throws:

```
TypeError: Cannot convert undefined or null to object
    at Object.values (<anonymous>)
    at countAssemblyResults (packages/@aws-cdk/toolkit-lib/lib/toolkit/private/count-assembly-results.ts:11)
    at synthAndMeasure (.../toolkit.ts)
    at Toolkit.synth (.../toolkit.ts)
```

This crashes `Toolkit.synth()` for otherwise-valid apps. It was reported downstream in AWS Amplify Gen 2, where a backend defining auth + a multi-model data schema + storage produces at least one nested stack whose `metadata` is undefined: aws-amplify/amplify-backend#3316.

This change guards the access with `stack.metadata ?? {}` (one line) and adds a regression test covering both the undefined-metadata case and the normal annotation-counting path.

Fixes #

### Checklist

---

### DIFF
--- a/packages/@aws-cdk/toolkit-lib/test/toolkit/count-assembly-results.test.ts
+++ b/packages/@aws-cdk/toolkit-lib/test/toolkit/count-assembly-results.test.ts
@@ -1,0 +1,51 @@
+import { countAssemblyResults } from '../../lib/toolkit/private/count-assembly-results';
+
+// Minimal span that records incCounter calls; only the surface
+// countAssemblyResults touches is implemented.
+function fakeSpan() {
+  const counters: Array<{ name: string; delta?: number }> = [];
+  return {
+    counters,
+    span: {
+      incCounter: (name: string, delta?: number) => {
+        counters.push({ name, delta });
+      },
+    } as any,
+  };
+}
+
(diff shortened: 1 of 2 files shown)


mrgrain (approved):

packages/@aws-cdk/toolkit-lib/test/toolkit/count-assembly-results.test.ts
```diff
@@ -0,0 +1,51 @@
+import { countAssemblyResults } from '../../../lib/toolkit/private/count-assembly-results';
```
mrgrain:
```suggestion
import { countAssemblyResults } from '../../../src/toolkit/private/count-assembly-results';
```

mrgrain (changes requested):

mrgrain (approved):

PR #1967, opened 2026-09-14
fix(toolkit-lib): classify create-only properties as non-hotswappable

### LINKED ISSUE #1965: (hotswap): create-only properties (e.g. SQS QueueName/FifoQueue) are classified hotswappable, so --hotswap hard-fails and --hotswap-fallback never falls back
### Describe the bug

The generic Cloud Control hotswap handler classifies a change to a **create-only** property as *hotswappable*, because it does not consult the resource's `createOnlyProperties` before attempting the update. As a result:

- `cdk deploy --hotswap` **hard-fails** (exit 1) with a CloudControl `NotUpdatableException`, instead of skipping the change.
- `cdk deploy --hotswap-fallback` **does not fall back** to a full deployment — it hits the same hard failure — because fallback only triggers for changes classified *non-hotswappable up front*, and these were (incorrectly) classified *hotswappable*.

The change is impossible to apply via hotswap (create-only properties require resource **replacement**), so the correct behavior would be to classify it as non-hotswappable up front. That would let `--hotswap` skip it cleanly and let `--hotswap-fallback` fall back to a full (replacing) deploy.

This is a CDK-side classification issue, not a CloudControl issue — CloudControl correctly rejects the update, and the resource schema already exposes `createOnlyProperties` (retrievable via `cloudformation describe-type`). The handler simply isn't reading it before issuing the `UpdateResource` PATCH.

### Expected behavior

For a changed property that is in the resource type's `createOnlyProperties`:

- `--hotswap`: classify as **non-hotswappable**, report it (like other non-hotswappable changes), and do not attempt a doomed PATCH.
- `--hotswap-fallback`: **fall back to a full CloudFormation deployment** (which performs the replacement).

### Current behavior

`--hotswap`:
```
✨ hotswapping resources:
   ✨ AWS::SQS::Queue 'https://sqs.us-east-1.amazonaws.com/<acct>/sqs-co-repro-v1'
❌  <Stack> failed to deploy
‣ HotswapFailed: [cloudcontrol] Failed to update <queue-url> (AWS::SQS::Queue)
‣ NotUpdatableException: Invalid patch update: createOnlyProperties [/properties/QueueName] cannot be updated
```
(exit 1)

`--hotswap-fallback`: identical hard failure, no “Falling back to doing a full deployment” message (exit 1).

Plain `cdk deploy` (no hotswap flags) works correctly — CloudFormation replaces the queue.

### Reproduction steps

Minimal stack:

```ts
import { Queue } from 'aws-cdk-lib/aws-sqs';

export class SqsCreateOnlyStack extends cdk.Stack {
    super(scope, id, props);
    new Queue(this, 'Q', {
      queueName: 'sqs-co-repro-v1',   // create-only: /properties/QueueName
      fifo: false,                    // create-only: /properties/FifoQueue
      visibilityTimeout: cdk.Duration.seconds(30),
    });
  }
}
```

1. `cdk deploy` (baseline → `sqs-co-repro-v1`).
2. Change `queueName` to `sqs-co-repro-v2`, then run `cdk deploy --hotswap` → hard fail (`NotUpdatableException [/properties/QueueName]`).
3. Same change, run `cdk deploy --hotswap-fallback` → same hard fail, **no fallback**.
4. (fifo variant) set `fifo: true` (name must end `.fifo`, add `contentBasedDeduplication: true`) → both flags fail with `createOnlyProperties [/properties/FifoQueue, /properties/QueueName]`.
5. Plain `cdk deploy` (no flags) → succeeds, queue replaced.

### Verified results matrix

| Change | Flag | Result |
|--------|------|--------|
| `queueName` v1→v2 | `--hotswap` | ❌ exit 1 — `NotUpdatableException [/properties/QueueName]` |
| `queueName` v1→v2 | `--hotswap-fallback` | ❌ exit 1 — same error, no fallback |
| `fifo` false→true | `--hotswap` | ❌ exit 1 — `[/properties/FifoQueue, /properties/QueueName]` |
| `fifo` false→true | `--hotswap-fallback` | ❌ exit 1 — same error, no fallback |
| `queueName` v1→v2 | plain `cdk deploy` | ✅ success — queue replaced |
| `visibilityTimeout` 30→90 (control) | `--hotswap` | ✅ hotswapped (mutable prop, works as expected) |

### Possible solution

The root cause is at the CCAPI hotswap call site, which passes **all** changed property names to `classifyChanges` as hotswappable candidates, without consulting the resource type's `createOnlyProperties`:

https://github.com/aws/aws-cdk-cli/blob/a73eebf18a917db1fd93864785cc434361b6cf95/packages/%40aws-cdk/toolkit-lib/lib/api/hotswap/cloud-control-resource.ts#L17-L22

`classifyChanges` then simply buckets every passed-in name as hotswappable — it has no notion of create-only properties:

https://github.com/aws/aws-cdk-cli/blob/a73eebf18a917db1fd93864785cc434361b6cf95/packages/%40aws-cdk/toolkit-lib/lib/api/hotswap/common.ts#L164-L176

**Fix:** before calling `classifyChanges`, fetch the resource type's `createOnlyProperties` (already available from the CloudControl/registry schema — e.g. via `cloudformation describe-type`, and the handler already reads the schema for the identifier) and exclude any changed property that appears there from the hotswappable set — i.e. treat create-only property changes as **non-hotswappable up front**.

This makes `--hotswap` skip them (reported like other non-hotswappable changes) and `--hotswap-fallback` fall back to a full CloudFormation deployment (which performs the replacement), instead of both hard-failing on a doomed `UpdateResource` PATCH.

### Additional context

Likely related to the same root cause (CCAPI handler not consulting property mutability up front): a DynamoDB `KeySchema` change — which is **not** in `createOnlyProperties` but is effectively immutable server-side — is reported as `hotswapped!` (exit 0) while being silently ignored (no `NotUpdatableException`), leaving the live resource unchanged. That case is being investigated separately.

### CDK CLI Version

`2.1141.0` (build c07d251)

### Framework Version

aws-cdk-lib `2.269.0`

### Node.js Version

24.x

### OS

Amazon Linux 2023

### Language

TypeScript

dedsec-terminal:
fixes #1965.

The generic Cloud Control hotswap handler passed every changed property to classifyChanges as a hotswap candidate without consulting the resource type's createOnlyProperties. A change like SQS QueueName was therefore PATCHed in place and hard-failed with NotUpdatableException under both --hotswap and --hotswap-fallback, and fallback never triggered because the change had been classified hotswappable.

The handler now reads createOnlyProperties from the registry schema via cloudformation DescribeType before classifying, and treats those properties as non-hotswappable up front through the existing classifyChanges path. --hotswap skips them like any other non-hotswappable change, and --hotswap-fallback falls back to a full CloudFormation deployment. When the schema cannot be retrieved the lookup returns empty and behavior is unchanged. A mixed change still hotswaps its mutable properties; only the create-only ones are held back.

Verified with two new unit tests in cloud-control-hotswap-deployments.test.ts (each running in both hotswap-only and fall-back modes): a create-only-only change sends no UpdateResource call, and a mixed change PATCHes only the mutable property. Both tests fail against the unfixed code. Full cloud-control suite passes (63 tests), tsc compiles, and eslint reports no new findings on the touched files (the no-literal-partition errors elsewhere in the package pre-exist on main).

- [x] Integration tests added/updated (not applicable, classification logic covered by unit tests)


packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
```diff
@@ -377,6 +393,51 @@ export class EvaluateCloudFormationTemplate {
...
+      });
+      if (!response.Schema) {
+        return new Set();
+      }
+      const schema = JSON.parse(response.Schema) as { createOnlyProperties?: unknown };
+      if (!Array.isArray(schema.createOnlyProperties)) {
+        return new Set();
+      }
+      const names = new Set<string>();
+      for (const pointer of schema.createOnlyProperties) {
```
ShadowCat567: Minor nit: can you change the name of `pointer` here and in `topLevelPropertyName` to `property` or something similar so the variable name of a bit more descriptive of what we are dealing with?

ShadowCat567 (commented):
Thanks for the contribution @dedsec-terminal! It looks good overall! One minor comment about variable names

packages/@aws-cdk/toolkit-lib/lib/api/cloudformation/evaluate-cloudformation-template.ts
dedsec-terminal: Done — renamed \pointer\ to \property\ in both the loop and \	opLevelPropertyName\.

ShadowCat567 (approved):
Nice work @dedsec-terminal! This will be very helpful for hotswap with CCAPI!

PR #1968, opened 2026-09-15
chore: upload built CLI tarball to S3 for the Windows integ canary

dgandhi62:
On every push to `main`, build and pack the  CLI and upload the npm tarball to the shared cdk-ops canary S3 bucket, keyed by commit hash, along with a  `cli/latest.json` pointer. This is the producer half of the daily Windows canary: aws-cdk-cli can only build on Linux, so this repo emits the artifact and cdk-ops downloads it to run cli-integ-tests on Windows. The other half is tracked here - https://github.com/cdklabs/cdk-ops/pull/5465

### Checklist

---


rix0rrr (approved):
> aws-cdk-cli can only build on Linux

Fine with this change, BUT, the above sentence could also be something we fix, right?

We don't have to accept the world as a given.

PR #1969, opened 2026-09-18
fix(toolkit-lib): restore the replacement guard for Express Mode deployments

### LINKED ISSUE #1931: (cli): #1745 removed the express-mode replacement guard, so --express deploys replacements with rollback disabled
### Describe the bug

`cdk deploy --express` no longer checks whether the change set contains a replacement before deploying it with rollback disabled. Express shipped with a purpose-built clause for this, and #1745 removed it.

2.1129.0 (the first release with express) through 2.1132.0:

```js
const rollback = this.options.rollback ?? true;
const expressNoRollback = this.options.express && this.options.rollback !== true;
...
if ((!rollback || expressNoRollback) && replacement) {
  return { type: 'replacement-requires-rollback' };
}
```

#1745 deleted `expressNoRollback` and added an early return above the guards:

```js
// Route express mode deployments directly to executeChangeset, express mode stacks cannot use rollback API
if (this.options.express) {
  return this.executeChangeSet(changeSetDescription);
}
```

#1785 later restructured that into `if (!this.options.express) { ...guards... }`, same behavior. As of 2.1138.0 an express deploy reaches `executeChangeSet` with no replacement check.

This matters because express sends `{Mode: "EXPRESS"}` with no `DisableRollback`, so the service disables rollback, and CloudFormation refuses replacement updates in that mode.

The consequence is not a failed deploy you retry. Our first `cdk deploy --express` on a healthy stack failed with `Replacement type updates not supported on stack with disable-rollback` on three `AWS::ECS::TaskDefinition` resources, and the stack could not be recovered: express retries hit the same replacement error, `--rollback` and non-express deploys are rejected with `This stack is currently in a non-terminal [UPDATE_FAILED] state`, and `cdk rollback` is refused because `RollbackStack is not supported for stacks that were last updated using EXPRESS deployment mode and are in a failed state` (the same dead end as #1739). We deleted the stack.

Worth flagging separately: the recovery path #1745 added does not cover this shape. It lives in `checkAndExecuteChangeSet`, which only runs on a successfully created change set. CloudTrail for our `--express --rollback` attempt shows `CreateChangeSet` succeeding, then 13 `DescribeChangeSet` polls over roughly 40s, then no `ExecuteChangeSet`. The change set fails on its own, so that code never runs.

### Regression Issue

Introduced in 2.1133.0 by #1745.

2.1132.0

### Expected Behavior

`cdk deploy --express` without an explicit `--rollback` should return `replacement-requires-rollback` when the change set contains a replacement, and prompt, as it did through 2.1132.0. The `--rollback` help text already documents that express defaults rollback to `false`, which is what `expressNoRollback` encoded.

### Current Behavior

No prompt, and a stack that cannot be recovered:

```
$ cdk deploy --express
...
<stack> failed: The following resource(s) failed to update: [<resources>]
Replacement type updates not supported on stack with disable-rollback

$ cdk deploy --express --rollback
<stack> failed: ValidationError: This stack is currently in a non-terminal [UPDATE_FAILED] state.
To update the stack from this state, please use the disable-rollback parameter with update-stack API.
To rollback to the last known good state, use the rollback-stack API

$ aws cloudformation rollback-stack --stack-name <stack>
An error occurred (ValidationError) when calling the RollbackStack operation:
RollbackStack is not supported for stacks that were last updated using EXPRESS deployment
mode and are in a failed state. To recover this stack, submit an UpdateStack request with
the last known stable template.
```

### Reproduction Steps

1. Deploy a stack containing an `AWS::ECS::TaskDefinition` and get it to `UPDATE_COMPLETE`. Every property change replaces a task definition.
2. Change a property on it.
3. On 2.1132.0, `cdk deploy --express` returns `replacement-requires-rollback` and prompts.
4. On 2.1133.0 or later, the same command deploys without prompting and fails with `Replacement type updates not supported on stack with disable-rollback`.

We observed step 4 only. Our project went straight from 2.1114.1, which predates express, to 2.1138.0, so we never ran a version where express had the guard. Step 3 is from reading `packages/@aws-cdk/toolkit-lib/lib/api/deployments/deploy-stack.ts` at the 2.1132.0 tag, where express is an ordinary change-set deployment that reaches the guard and `expressNoRollback` is true whenever `--rollback` is absent.

### Possible Solution

Restore the express case, which is your own code from 2.1132.0:

```js
const expressNoRollback = this.options.express && this.options.rollback !== true;
...
if ((!rollback || expressNoRollback) && replacement) {
  return { type: 'replacement-requires-rollback' };
}
```

#1745 needed express deploys to avoid the `RollbackStack` API, and the two `failpaused-need-rollback-first` returns are the ones that lead there. `replacement-requires-rollback` does not call `RollbackStack`; it prompts and then redeploys with rollback enabled, so it can stay in the express path.

Region us-east-1. Happy to share account and request IDs privately.

### CDK CLI Version

2.1138.0

### Framework Version

aws-cdk-lib 2.266.0

### Node.js Version

v24.18.0

### OS

macOS 26.6.2

### Language

TypeScript

### Language Version

TypeScript 6.0.2

### Other information

Related: #1739, #1745, #1785.

sanjanaravikumar-az:
Fixes #1931. CloudFormation refuses replacement-type updates while rollback is disabled. Express Mode disables rollback unless `--rollback` is passed, so `cdk deploy --express` submitted replacements that CloudFormation rejected, leaving the stack in UPDATE_FAILED with no rollback available. #1745 removed the express half of the guard condition and relaxed the test covering it; #1785 restructured the remainder into `if (!this.options.express)`.

The condition is now derived once in `rollbackDisabled()` (`express ? rollback !== true : rollback === false`) and consumed by the guard, `deployConfig()` and the failure diagnostic — which previously derived `rollbackEnabled` as `rollback !== false`, so under `--express` it reported rollback as enabled and advised re-running with `--no-rollback`. The guard is a single cross-mode check rather than an express-specific branch, and returns `replacement-requires-rollback` so the existing confirm-and-retry-with-rollback prompt engages. A new `CDK_TOOLKIT_W5903` carries a typed payload with `detectedBy: 'change-set' | 'service-error'`.

When the stack is already in a failed state the guard instead throws `ReplacementRequiresUnwedge` carrying recovery guidance for that state - only `UPDATE_FAILED` is told to replay the previous configuration, since a failed initial create has none - because deploying with rollback enabled cannot update an already-failed stack; since the confirmation defaults to yes, returning a result there meant a non-interactive caller ran a deployment CloudFormation rejects with `ValidationError: non-terminal [UPDATE_FAILED]`.

`--method=direct` has no change set to inspect and is deliberately not refused, because replaying the previous configuration that way is the only exit from an already-stranded stack; the rejection is detected after the fact and the user is routed to `--express --rollback`, with the original CloudFormation error still propagating. That routing is gated on `express`, since a standard-mode `--no-rollback` deployment hits the same CloudFormation rejection and must not be pushed towards a sticky mode that gives up `cdk rollback`.

Verified against CloudFormation on a live ECS stack (Memory 1024 to 2048, a genuine replacing change): the gate fires and submits nothing, and retrying with rollback enabled cannot rescue a stack that is already failed. The direct-path hook firing on CloudFormation's resource-level reason, and the documented unwedge still succeeding with the gate upstream, were verified the same way.

```
This deployment replaces a resource, which CloudFormation does not support while rollback is disabled.
BEFORE: UPDATE_COMPLETE  2026-09-18T01:35:42.509000+00:00
AFTER:  UPDATE_COMPLETE  2026-09-18T01:35:42.509000+00:00

ValidationError: This stack is currently in a non-terminal [UPDATE_FAILED] state.
```

Notes:
- New public API: `lib/index.ts` re-exports `./payloads`, so `ReplacementRequiresRollback` and `ReplacedResource` are supported surface; the matched reason constant stays private, as `deploy-stack.ts` is not in the barrel.
- Non-interactive behaviour change: CI no longer auto-runs the retry against an already-failed express stack, which also closes the path described in #1973 and leaves only the result-type question open there.
- `FakeCloudFormation` changed meaning for all future tests: express implies rollback disabled, `FailReason` produces resource-level failure events, and `DescribeChangeSet` now returns the persisted `DeploymentConfig` as the real API does. No other suite needed changing.
- The monitor-flush reorder is not covered by a test, since under fake timers the errors are usually already populated; it rests on code reading plus the live deploys.
- Detection keys on `PolicyAction` only, unchanged and sufficient for #1931, pinned by a negative test and tracked in #1971.
- An Express change set's rollback policy is fixed by `CreateChangeSet` and `ExecuteChangeSet` cannot override it, so rollback safety is read from the persisted `DeploymentConfig` when executing an existing change set and a conflicting execute request is refused with `ChangeSetRollbackPolicyMismatch` rather than silently executed or recreated.

Related: #1972. Docs in #1970. Time-bound: CloudFormation's server-side fix has a tentative ECD of 2026-11-15, after which the guard, the direct-path hook, the matched reason string and `CDK_TOOLKIT_W5903` should be removed together. Draft pending a scoped re-review.

`#NNNN` here means `aws/aws-cdk-cli`; note `aws/aws-cdk#1931` is an unrelated closed issue.


sanjanaravikumar-az:
# Review notes (companion to PR-DESCRIPTION.md — post as a review comment, not in the description)

## Decisions I want explicitly reviewed

### 1. `Replacement === 'True'` detection was DROPPED from this PR
My first revision carried a second commit widening `findReplacements` to also honour `ResourceChange.Replacement`. It has
been **dropped entirely** and filed as ISSUE_A_PLACEHOLDER instead.

Why: `findReplacements` has only ever inspected `PolicyAction`, since the initial commit `1dbec5ce`. I checked whether
that mattered for #1931 by describing a real, unexecuted change set for the exact reproduction change:

```json
{
  "PolicyAction": "ReplaceAndDelete",
  "Action": "Modify",
  "LogicalResourceId": "TaskDef54694570",
  "ResourceType": "AWS::ECS::TaskDefinition",
  "Replacement": "True",
  "Scope": ["Properties"],
  "Details": [
    {"Target":{"Attribute":"Properties","Name":"ContainerDefinitions","RequiresRecreation":"Always"}},
    {"Target":{"Attribute":"Properties","Name":"Memory","RequiresRecreation":"Always"}}
  ]
}
```

`PolicyAction` **is** present, so the shipped detection would have gated the real #1931 change. The widening was
therefore not required to fix this SEV, and it is a behaviour change that also affects the non-express `--no-rollback`
guard. It does not belong in a SEV fix. I had earlier asserted the opposite in a draft description; that was wrong, and
the JSON above is why.

Dropping it broke **zero** tests, which confirms the commit split was categorised correctly rather than leaving orphaned
assertions behind.

### 2. `Conditional` is deliberately excluded
The same real change set contains `CDKMetadata` with `Replacement: "Conditional"` (its `Analytics` property is
`RequiresRecreation: "Conditionally"`). That appears on essentially every CDK deployment, so treating `Conditional` as a
replacement would gate almost every express deploy. A `Conditional` change that does replace is caught by the
direct-path routing — which is part of why that hook exists.


### 3. `--method=direct` intentionally bypasses the change set gate — please do not "fix" it
Direct mode has no change set, and creating one to inspect would defeat the mode. More importantly, a blanket refusal of
`express + direct + rollback-disabled` would remove the only exit from a stranded stack. The hook is in
`monitorDeployment` rather than `directDeployment` on purpose, so it also covers a change set replacement the gate
cannot see; there is a test for that case (`Conditional` change set + rejection at execute time).

### 4. Text matching is the only trigger available
CloudFormation reports this rejection as a resource-level `ResourceStatusReason` with no `HandlerErrorCode:`/`Error Code:`
prefix, so `extractErrorCode()` yields `<ResourceType>:UnknownError` and carries no signal. The string is isolated in
one exported constant with a fragility comment, the structural `rollbackDisabled()` check runs first, and a miss now
logs a debug line listing the reasons actually seen — so CloudFormation's ~11/15 reword shows up in a log rather than as
a second SEV. **If anyone knows a structured discriminator, it should replace this.**

## Test properties worth knowing

15 tests in `test/api/deployments/deploy-stack-express-replacement.test.ts`, 2 in `test/actions/deploy.test.ts`, plus
the two repaired matrix cases in `deploy-stack.test.ts`.

- Negative assertions are enforced, not implied: the gate tests install `ExecuteChangeSet`/`UpdateStack` handlers that
  **throw** if called.
- **Mutation-checked, and the signatures are disjoint:**
  - Rescoping the guard to non-express (the exact #1745/#1785 defect) fails **7 tests** across 2 suites, including both
    fixture shapes, both REGRESSION cases, the pagination test and the two repaired matrix cases.
  - Removing only the `Replacement === 'True'` arm fails **3 tests**, all of them in the `Replacement=True with no
    policy action` fixture arm.

  So a failure names its cause. In the first revision both causes produced the same 6 failures — that conflation was a
  fair criticism and is fixed by parametrising the fixture over change shapes. (The second mutation is measured against
  the dropped commit; on this PR as it stands the fixture list has a single entry.)

- A test now genuinely reproduces the SEV: the fake was taught the real rule (`DeploymentConfig.Mode === 'EXPRESS'`
  without an explicit `DisableRollback: false` ⇒ rollback disabled), so a failed express update terminates at
  `UPDATE_FAILED` instead of rolling back. Previously every express test took the rollback branch, so no test could
  produce a stranded stack at all.
- `expectMessage` now matches on `code` as well as level/substring, so renaming the code cannot leave positive
  assertions passing while negative ones become vacuous.

## Pre-existing issues found but NOT fixed here

- The gate returns before `executeChangeSet`, so the change set it created is left behind in `AVAILABLE` state
  (observed on a real deploy). That is inherent to this result type and predates the regression — worth a follow-up.
- From an already-failed stack the CLI still offers a retry that cannot succeed — filed as ISSUE_C_PLACEHOLDER.
- The empty-diff `✅ (no changes)` / exit 0 lie — filed as ISSUE_B_PLACEHOLDER.

## Disputed review points

- **"Duplicate `import { IO }` violates repo convention."** It does not. `asset-publishing.ts:20-21` in the same
  directory uses exactly the same `import type { ... }` + `import { IO }` pair, and ESLint (which I did get running,
  see below) reports zero problems on the file. `stack-activity-monitor.ts:11` uses the combined form; both exist in the
  codebase. Happy to change it as a style preference, but it is not a convention violation.
- **"ESLint could not be run."** My earlier claim that no Node ≥20 was available was wrong — I had only checked
  `~/.nvm` and `/usr/local/bin` and missed mise-managed toolchains. Node 20.20.2 was available the whole time. ESLint
  now runs and reports **0 problems** across the package's `lib`, `test` and `build-tools`. It caught exactly one real
  issue in my code (trailing blank lines), which is fixed. I should not have declared that gap unclosable without
  checking properly.

